PR #39, opened 2014-12-30
add test commands

nosami:


src/OmniSharp/Api/TestContext/AspNet5TestCommandProvider.cs
```diff
@@ -0,0 +1,47 @@
+using System.IO;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace OmniSharp
+{
+    public class AspNet5TestCommandProvider : ITestCommandProvider
+    {
+        public string GetTestCommand(TestContext testContext)
+        {
+            if (new FileInfo(testContext.ProjectFile).Extension != ".json")
```
davidfowl: Yes, we should be using the projects that omnisharp is aware of. We need to expose the AspNet5Context here so we can see if the project file is one we're aware of:

https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp/AspNet5/AspNet5Context.cs

src/OmniSharp/Api/TestContext/AspNet5TestCommandProvider.cs
```diff
@@ -0,0 +1,47 @@
...
+            {
+                return null;
+            }
+
+            // Find the test command, if any and use that
+            var node = testContext.SyntaxNode;
```
davidfowl: Why not figure this out outside of the ASP.NET specific implementation? and pass in the ISymbol instead of the SyntaxNode.

nosami:
ISymbol doesn't work if the cursor isn't on a symbol. It could be in an empty space inside a method body for example.

davidfowl:
@nosami can't you move the logic to detect the nearest valid method/class to the controller and just pass the relevant symbol.

nosami:
Ah, yes. I had a bit of a misunderstanding there what you wanted. Will sort it out.

PR #44, opened 2015-01-20
Find references - Small fix and tests

Yantrio:
Fixed duplicate entries for properties (the getter/setter and backing properties were being returned)

Still unsure if the reference should return it'sself or not.

Added tests for most situations I can think of.


nosami:
thanks!

Issue #47, opened 2015-01-21
DesignTimeHostManager only waits 1 sec 


jrieken:
the DTH manager gives the dth host 1 second to start and start listening. If you have a slow machine or a machine under load this won't be enough. we should either wait longer or try a few times in a short interval.

https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp/AspNet5/DesignTimeHostManager.cs#L62

nosami:
Think I'd prefer a smaller interval with retry. The 1 second pause is too long for me :)

jrieken:
:+1: seems like you are spoiled with a monster machine ;-)

PR #49, opened 2015-01-21
#47 - DTHManager tries to connect before calling its callback

jrieken:
Instead of waiting 1sec before letting clients connect to the DTH the DTH manager tries to connect in an interval of 500ms to 4seconds. 

@davidfowl - you might know a better fix for this, esp since it will cause one thread to leak in the DTH (https://github.com/aspnet/XRE/blob/dev/src/Microsoft.Framework.DesignTimeHost/Program.cs#L101)


davidfowl:
:shipit: but lets revisit this.

PR #51, opened 2015-01-25
do not call kvm upgrade in build.cmd

filipw:
`kvm upgrade` is not needed because OmniSharp explicitly needs `1.0.0-beta2`. `kvm upgrade` gets the latest version of KVM which is ignore anyway, moreover it is aliasing it as 'default' forcing VS to pick it up. 
so if you run the `build.cmd` you will not be able to compile OmniSharp in VS until you manually reset the `default back to `1.0.0-beta2` which is a bit odd.

Alternatively, the build script could reset the `default` alias to `1.0.0-beta2`.

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -1,7 +1,6 @@
 @echo off
 
 set "KRE_NUGET_API_URL=https://www.myget.org/F/aspnetvnext/api/v2"
-CALL kvm upgrade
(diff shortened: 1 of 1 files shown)


nosami:
Agreed! ```kvm upgrade``` used to be required up until about a week ago when we upgraded the dependencies.

Thanks!

PR #53, opened 2015-01-26
Split Omnisharp into 2 libraries

davidfowl:
- Omnisharp.Http is the http server and omnisharp http protocol
- Omnisharp is the class library that can be used as an normal
C# API


davidfowl:
bah, I'm not sure this makes much sense anymore but take a look

PR #56, opened 2015-01-28
allow any change to be done in OmnisharpWorkspace

### LINKED ISSUE #55: Trying to apply changes to OmnisharpWorkspace results in exception
When any of the endpoints tries to apply *any* solution changes (add file, rename content, refactor etc etc) to the workspace using

                // Attempt to update the workspace
                if (_workspace.TryApplyChanges(solution))
                {
                    response.Changes = changes;
                }

OmniSharp will not do it - instead it will throw an exception.

See screenshot:
screenshot 2015-01-28 08 21 35

filipw:
fixes #55.

Since `OmniSharpWorkspace` inherits from `Workspace` abstract class, it needs to explicitly specify which changes can be done - and it seems reasonable to assume that all changes specified in the `ApplyChangesKind` should be allowed.

This prevents the exception mentioned in the #55 from happening and the buffer is correctly updated. 

However, the changes are not saved to disk at this moment. Roslyn normally does it using `MSBuildWorkspace` not through the base `Workspace` class. I haven't included this in the PR as I am not sure if any strategy for that has been discussed, but I can extend `OmniSharpWorkspace` to persist to disk if needed.

### DIFF
--- a/tests/OmniSharp.Tests/OmnisharpControllerFacts.cs
+++ b/tests/OmniSharp.Tests/OmnisharpControllerFacts.cs
@@ -26,35 +26,6 @@ namespace OmniSharp.Tests
         }
 
         [Fact]
-        public async Task Rename_UpdatesWorkspace()
-        {
-            const string fileContent = @"using System;
-
-namespace OmniSharp.Models
-{
-    public class CodeFormatResponse
-    {
-        public string Buffer { get; set; }
-    }
-}";
-
-            OmnisharpWorkspace workspace;
-            OmnisharpController controller;
-            DocumentInfo document;
-            CreateSimpleWorkspace(out workspace, out controller, out document, "test.cs", fileContent);
-            var result = await controller.Rename(new Models.RenameRequest
-                        {
-                            Line = 7,
-                            Column = 27,
-                            RenameTo = "foo",
-                            FileName = "test.cs"
-                        });
-            var sourceText = await workspace.CurrentSolution.GetDocument(document.Id).GetTextAsync();
-            Assert.Equal(result.Changes.First().Buffer, sourceText.ToString());
-            Assert.Equal(result.Changes.First().FileName, "test.cs");
-        }
-
-        [Fact]
         public async Task UpdateBuffer_HandlesVoidRequest()
         {
             OmnisharpWorkspace workspace;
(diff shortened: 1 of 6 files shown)


davidfowl:
What does this fix right now?

filipw:
There is no more `NotSupportedException` when trying to apply workspace changes. The workspace instance is correctly updated and the changes are propagated correctly to the buffer.

Additionally, the client gets a proper modified buffer response i.e. from the `/rename` endpoint.
All of the changes live in memory only though.

This is because, as mentioned, `OmniSharpWorkspace` does not persist to disk atm. I suggest the save would done in the `OnDocumentChanged` event but it's a natural next step so wanted to put this out there for discussion - especially as I am not aware of how much you have considered this area thus far.

nosami:
I'm not sure what problem you are seeing. /rename seems to work perfectly
in my testing.

Is this error possibly as a result of using the latest Roslyn code drop?

I went to some lengths with OmniSharp v1 to ensure that changes were never
persisted to disk automatically.

filipw:
@nosami this is why it's not working ATM 
https://github.com/dotnet/roslyn/blob/master/src/Workspaces/Core/Portable/Workspace/Workspace.cs#L822

If you don't override, any attempt to apply changes will error out. I attached a screenshot in the issue #55. This happens for me in the latest VS2015 CTP, by just cloning the repo and building from the source.

Regarding saving to disk - I totally get it. This is exactly why I didn't want to arbitrarily decide about doing it and didn't include it in the PR.

filipw:
following a jabbr discussion -> this exception was introduced by this Roslyn change https://roslyn.codeplex.com/SourceControl/changeset/146532ecfdf1e8657d24965515c0fc1650c3d756

I will add tests for rename controller to the PR too.

nosami:
/rename is completely broken without this PR

tests/OmniSharp.Tests/project.json
```diff
@@ -1,15 +1,19 @@
...
-    "commands": {
-        "test": "Xunit.KRunner"
-    },
-    "frameworks": {
-        "aspnet50": { },
-        "aspnetcore50": { }
-    }
+	"version": "1.0.0-*",
+	"dependencies": {
+		"OmniSharp": "1.0.0-*",
```
davidfowl: Don't use tabs

tests/OmniSharp.Tests/OmnisharpControllerFacts.cs
```diff
@@ -2,6 +2,7 @@
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Text;
 using Xunit;
+using System.Linq;
```
davidfowl: Order using properly, System* goes first

nosami:
I'm gonna pick this up.

PR #57, opened 2015-01-29
when computing the outline/structure of a document consider all frameworks

jrieken:
...ks

### DIFF
--- a/src/OmniSharp/Models/Node.cs
+++ b/src/OmniSharp/Models/Node.cs
@@ -11,6 +11,8 @@ namespace OmniSharp.Models
 
         public string Kind { get; set; }
 
+        public IEnumerable<string> Frameworks { get; set; }
+
         public int CompareTo(Node other)
         {
             if (other.Location.Line < Location.Line)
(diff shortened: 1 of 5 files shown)


jrieken:
Code as shown below should give a merged view when being asked for the its sturcture. The individual element have to know to what framework they belong

```
{
    public class A
    {
    
        public void M() 
        {
            
        }
#if ASPNETCORE50
        public void N() 
        {
            
        }
#else
        
        public void O() 
        {
            
        }
#endif
    }
}
```

nosami:
Been thinking about this. Would you ever need to get a list for more than one framework at a time?

Would it be better to request the structure for the framework that you are interested in?

nosami:
OK... I get it now. Ignore my last comment!

jrieken:
:-) with my last change the Node has a list of Frameworks. That would allow further client side filtering or special rendering depending on the selected framework.

src/OmniSharp/Models/Node.cs
```diff
@@ -1,13 +1,78 @@
+using System;
 using System.Collections.Generic;
 
 namespace OmniSharp.Models
 {
-    public class Node
+    public class Node : IComparable<Node>
```
davidfowl: This class really needs to be renamed. I have no idea what a node is.

jrieken:
:ok_hand: going with a list of project names. Node is the worst name of all times ;-) Will go with ```FileMemberElement```

PR #62, opened 2015-02-03
Made process.Exited event work on mono

davidfowl:


src/OmniSharp/Utilities/ProcessExtensions.cs
```diff
@@ -0,0 +1,61 @@
...
+                        var pair = _processes[i];
+                        if (pair.Item1.HasExited)
+                        {
+                            _processes.RemoveAt(i);
+                            pair.Item2();
+                        }
+                    }
+                }
+
+                // REVIEW: Configurable?
```
nosami: I don't think it needs to be configurable.

nosami:
Looks good

nosami:
Doesn't seem to be working for me though. I launched the server like this

```bash
jason@Jasons-MacBook-Pro ~/src/omnisharp-roslyn (watch-process-mono) $ ps ax | grep emacs
 1545   ??  U      0:02.51 /usr/local/Cellar/emacs/HEAD/Emacs.app/Contents/MacOS/Emacs
 1570 s000  S+     0:00.00 grep --color=auto --exclude-dir=.bzr --exclude-dir=.cvs --exclude-dir=.git --exclude-dir=.hg --exclude-dir=.svn emacs
jason@Jasons-MacBook-Pro ~/src/omnisharp-roslyn (watch-process-mono) $ scripts/Omnisharp --hostPID=1545
```

Closing emacs didn't terminate the server.

nosami:
[INFORMATION:OmniSharp.Startup] Omnisharp server running on port '2000' at location '/Users/jason/src/omnisharp-roslyn' on host -1.

Looks like hostPID isn't being passed through.

davidfowl:
@nosami  for some reason @jrieken make it `--hostPID` instead of `-hostPID`, :smile: . I hit the same issue. Even after that though, I found problems where the pid of the bash script is returned instead of the mono process for DTH. So that doesn't properly die.

jchannon:
I've got sublime calling this directly now so nasty Python or hundreds of
lines of code to launch. Once this is done it will be pretty elegant

nosami:
I used --hostPID though!

davidfowl:
@nosami that works :smile:

jrieken:
@davidfowl the rule is ```--<word>``` and ```-<letter>```

jchannon:
Just ran this and got the below screenshot, so it says its killing the DTH and exits. I know it killed either the server or the DTH but not both as there is 1 mono-sgen process still running

screen shot 2015-02-03 at 20 41 02

davidfowl:
@jchannon that's what I said :smile: DTH does not die, the bash process that launched dth died.

jchannon:
Does this kill both now hence the merge?

davidfowl:
No, the other change needs to happen in DTH and won't affect this.

PR #63, opened 2015-02-05
Codeactions

Yantrio:
Started work with codeactions, currently removed NR6.

original v1 requests should still work at a basic level.

### DIFF
--- a/src/OmniSharp/Services/NRefactoryCodeActionProvider.cs
+++ b/src/OmniSharp/Services/NRefactoryCodeActionProvider.cs
@@ -9,7 +9,7 @@
 
 //namespace OmniSharp.Services
 //{
-//#if ASPNET50
+//#if NR6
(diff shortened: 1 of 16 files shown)


src/OmniSharp/Services/NRefactoryCodeActionProvider.cs
```diff
@@ -0,0 +1,33 @@
+﻿//#if ASPNET50
```
davidfowl: you can leave this an do `#if NR6`

src/OmniSharp/Startup.cs
```diff
@@ -70,6 +70,13 @@ public void ConfigureServices(IServiceCollection services, IApplicationLifetime
             // Add test command providers
             services.AddSingleton<ITestCommandProvider, AspNet5TestCommandProvider>();
 
+#if ASPNET50
```
davidfowl: Nuke the ifdef

PR #64, opened 2015-02-08
ensure the buffer is updated

Yantrio:
This fixes a small issue that I was having in brackets where the buffer was not updated properly, so codeactions changed randomly.


nosami:
Thanks!

PR #65, opened 2015-02-08
if the Request has buffer content and filename, always update buffer

filipw:
instead of remembering to update buffer all over the place, if the action takes a `Request` model and if the request has `FileName` and `Buffer` populated, update the buffer automatically

### DIFF
--- a/tests/OmniSharp.Tests/UpdateBufferFilterFacts.cs
+++ b/tests/OmniSharp.Tests/UpdateBufferFilterFacts.cs
@@ -1,10 +1,9 @@
-﻿using Microsoft.AspNet.Mvc;
-using Microsoft.CodeAnalysis.CSharp;
-using OmniSharp.Filters;
-using System.Collections.Generic;
+﻿using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
+using Microsoft.CodeAnalysis.CSharp;
 using Xunit;
+using OmniSharp.Filters;
 
 namespace OmniSharp.Tests
 {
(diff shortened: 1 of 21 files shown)


tests/OmniSharp.Tests/UpdateBufferFilterFacts.cs
```diff
@@ -0,0 +1,47 @@
+﻿using Microsoft.AspNet.Mvc;
+using Microsoft.CodeAnalysis.CSharp;
+using OmniSharp.Filters;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Xunit;
```
nosami: Order by system first, then 3rd party, then local, then alphabetical

filipw: and I was proud of myself for making sure it was alphabetical cause you mentioned sorting the other time too! Where are the coding guidelines yo! :smile:

filipw:
sorted usings. 

Ultimately, the idea is that as long as you use the main Omnisharp controller and take in some sort of a `Request` the buffer will auto update for you, so stuff like #64 will never happen again.

If you have a different controller, you can either call the buffer update it by hand or use the filter too. In fact, it might even be a good idea to make the filter global and let it always run regardless of controller - but that's up for discussion I guess, so didn't want to make any arbitrary decision myself.

nosami:
I think we'll end up using a global filter. I can't think of any situation where you would pass in a buffer from the editor and not expect it to update the server.

filipw:
ok let me push this change to a global filter, will be even cleaner that way

nosami:
Thanks Filip :)

PR #71, opened 2015-02-11
remove duplicate route registration that was introduced with merge stupi...

jrieken:
With the formatAfterKeystroke/formatRange work I originally deleted ```OmnisharpController.FormatDocument``` but due to a merge error it came back... Now it conflicts.


Yantrio:
Should we separate the formatting of the document and the sections of code into seperate files? similar to how we don't throw all Navigation actions under 1 file. 

Maybe we could clarify how we want to organise the code (file per action vs similar actions in 1 file) ?

nosami:
I'd prefer to have them in separate partial controllers but considering that the code is currently busted, I'm going to merge anyway :)

jrieken:
Thanks

PR #72, opened 2015-02-11
sort text changes descending (by span)

jrieken:
The easiest way to apply a bunch of (non-overlapping) text changes is to apply them in reverse order. This changes makes sure that Omnisharp always sort them that way. Also adding test to verify this behaviour and fixing some formatting issues.


nosami:
nice!

jrieken:
thanks

PR #73, opened 2015-02-12
Find usages in file

jrieken:
Allows to limit the find usages call to the current file. This will allow editors to implement a _mark occurrences_ feature (highlighting all occurrences of the symbol under the caret)


nosami:
Nice. Thanks!

Issue #70, opened 2015-02-10
Don't merge your own Pull Requests


nosami:
Ok. I'm guilty of this as well, but from now on, nobody is to merge their own code into master without getting it signed off by one of the other members of the OmniSharp team.

To be merged in, all code must be formatted with spaces not tabs, with 4 spaces wide and using statements ordered correctly. 

Also, unless there are exceptional circumstances and you can rationalise them, you *must* provide tests that show what your code is doing. This is especially important when we are using reflection against a fast moving API. The tests also should serve as documentation for other team members to use.

Anyone in breach of these conditions will be booted off the team :D

davidfowl:
Yes!!!!

Yantrio:
:+1:

shanselman:
Yep.

jrieken:
:+1: 

We should start a contributors guide including the formatting rules etc

mat-mcloughlin:
Agreed, can I suggest here is a good place to start?

https://github.com/NancyFx/Nancy/blob/master/CONTRIBUTING.md

thecodejunkie:
Merging your own pull-requests is pointless.. you might as well push straight to master.. and why not force push while you are on it? ;)

Code -> Push a branch -> Send a PR -> Someone else reviews it -> Pulls it if it's OK

Workflow to live by

mat-mcloughlin:
Also as a side note this. Is it worth revisiting who is able to merge requests. I'm not wanting to push anybodies nose out but I think there should be a couple of core contributors that can/want to monitor the project and keep it going in the right direction.

Thoughts?

filipw:
reminds me of this

![](http://cdn.meme.am/instances/52307408.jpg)

mat-mcloughlin:
ahahaha

jchannon:
Agree. Nancy minions only got write access once it was unmanageable for Andreas & Steve to handle all the PRs even then we have at least 2 minions review before merging.

I'd say Jason should be primary merger and then anyone else he thinks should be allowed to. Others can obviously say I think it looks ok but for now Jason only merges. 

Thoughts?

mat-mcloughlin:
Can you disable PR ability for project owners?

jchannon:
doubt it

nosami:
I don't mind being the owner, but I'm not particularly fussed either way. I think we should just stick to *someone else* merging for a while and see how it goes.

Issue #31, opened 2014-12-18
Formatting options


nosami:
We need to serialize these http://source.roslyn.codeplex.com/#Microsoft.CodeAnalysis.CSharp.Workspaces/Formatting/CSharpFormattingOptions.cs,dac739b7b8a53984

and http://source.roslyn.codeplex.com/#Microsoft.CodeAnalysis.Workspaces/Formatting/FormattingOptions.cs,2ce68197c6bbf674
into the config file, and load them on solution start into Workspace.Options

We also need a web UI for managing these.

Mpdreamz:
Assigning this to me since I'm taking a stab at plugging editorconfig in. Finally making use of the rewrite to managed code :D https://www.nuget.org/packages/editorconfig

jrieken:
:+1:

PR #74, opened 2015-02-13
Added an .editorconfig to dictate project formatting settings

Mpdreamz:
Every editor under the sun supports (or should!) http://editorconfig.org/ 

Visual studio has an extension: https://visualstudiogallery.msdn.microsoft.com/c8bccfe2-650c-4b42-bc5c-845e21f96328

This will force your editor settings to be changed when working on our project to the settings we dictate. 

Editorconfig will leave your defaults alone, so when you switch back to a private project with no editorconfig file it will keep using your default settings there.

Ties into #70


nosami:
kproj files are indented by 2 spaces. I don't think anyone would ever edit those, but it should probably be added.

khellang:
I think [corefx's .editorconfig](https://github.com/dotnet/corefx/blob/master/.editorconfig) makes more sense, maybe something like:

```
# EditorConfig is awesome:
http://EditorConfig.org

# top-most EditorConfig file
root = true

[*]
insert_final_newline = true
indent_style = space
indent_size = 2
trim_trailing_whitespace = true

[*.cs]
indent_size = 4

[*.{csproj,kproj,nuspec}]
indent_size = 2
```

nosami:
@khellang looks good to me.

Mpdreamz:
:+1: thanks @khellang  will update the PR shortly 

That url should also be quoted though, even if editorconfig files can handle invalid keys (part of its test suite).

Mpdreamz:
Updated, do we care about squashing and rebasing?

nosami:
nah! thanks!

jrieken:
:+1:

jrieken:
@Mpdreamz didn't you also do an API (editor config nuget package) so that we can seed the roslyn formatter with those settings?

Mpdreamz:
I did, I just updated it today so that it includes a proper aspnetcore50 build in the nuget package. Just did a `git checkout -b feature/editorconfig-powered-formatting` to start on #31 :)

Mpdreamz:
It wont actually seed once though, it will lookup **everytime** since thats the editorconfig way, i.e you might very well have different settings for `*.generated.cs` vs `*.cs` files. 

Usually an editor opens a file and seeds that file with editorconfig settings, but we dont track editors opening and closing files on the server right? or do we?

jrieken:
yeah - that's what I meant. we should have a thing (service?) which returns an OptionSet that applies to the current file.

Issue #69, opened 2015-02-09
explore use of stdin/stdout


jrieken:
The goal of this exploration is to use stdin/stdout for the communication between omnisharp-roslyn (server) and editors (client). Potential benefits will be

* better performance due to less message overhead (no http headers)
* clean relationship between editor and omnisharp and request ordering - http is multi client, allows for parallel requests etc
* simplified integration into editors as stdin/stdout is easier to support than http
* two-way communication between server and client, e.g. after a file change the server can push validation messages to the client (no need for polling)

Apart from the obvious adoption cost of editors a concern is that stdout is a shared resource that is also used for logging etc. We should not loose the logging features we have today.

nosami:
@filipw changed the code yesterday to add a Global Filter to do the _workspace.EnsureBufferUpdated

You'll need to do it another way.

nosami:
All logging is done through stderr currently.

davidfowl:
I do have a crazy idea that would make mvc work over stdin/out.

nosami:
![image](https://cloud.githubusercontent.com/assets/667194/6110808/41aa7b44-b07c-11e4-8e87-48ac53f81341.png)

filipw:
clean relationship between editor and omnisharp and request ordering - http is multi client, allows for parallel requests etc

I wonder - has anyone considered building and running a cloud based editor using OmniSharp? as long as workspace is per user, not a singleton, that could work. Someone should change it into a business and challenge https://c9.io :smile: 

or, just from the top of my head, how about a scenario where I have a solution on my desktop, but can also edit it from my mobile device? that would work even today, as long as both devices are on the same network

    simplified integration into editors as stdin/stdout is easier to support than http

isn't HTTP the easiest thing to support? cause it works everywhere and with everything? the barrier of entry for any editor/client is probably the lowest when everything is HTTP based

nosami:
The idea isn't to replace HTTP, but to supplement it for scenarios where
performance is a concern.

On 9 Feb 2015 19:15, "Filip W" <notifications@github.com> wrote:
>
>  clean relationship between editor and omnisharp and request ordering -
http is multi client, allows for parallel requests etc
>
> I wonder - has anyone considered building and running a cloud based
editor using OmniSharp? as long as workspace is per user, not a singleton,
that could work. Someone should change it into a business and challenge
https://c9.io
>
> or, just from the top of my head, how about a scenario where I have a
[...]

PR #86, opened 2015-02-16
Stdio work for #69

jrieken:
This PR adds support for standard in/out (in addition to http) to Omnisharp. Things are in ```Omnisharp.Stdio``` which is a console app reading from ```Console.In``` and writing to ```Console.Out```. Messages are JSON and conform to the v8 debugger protocol envelopes (https://code.google.com/p/chromedevtools/wiki/ChromeDevToolsProtocol). The implementation build on top of the ```Post```-MVC-attributes and reuses the Controllers defined in ```Omnisharp```. Log messages are wrapped into event-messages.

stdio


nosami:
Have you got this code hooked up to the editor? 

Interested in knowing how it compares as far as performance goes. I'd expect it to be faster, but is it noticeable?

filipw:
rather than manually reflecting over types you should use MVC infrastructure - in this case, `ActionDescriptors` (https://github.com/aspnet/Mvc/blob/c936ae80ca6b3df61e54e204444e088f5ced8051/src/Microsoft.AspNet.Mvc.Core/DefaultActionDescriptorsCollectionProvider.cs#L34)

This ensures that in the future, if any MVC components are added or customized the actions are still picked up correctly here

jrieken:
@nosami - didn't compare perf yet. don't expect a big difference. the biggest win of this work is that it will enable 2-way communication without polling and assuming we factor out http into ```Omnisharp.Http``` less dependencies - faster startup, smaller bundle

mat-mcloughlin:
I nominate @Yantrio to try adding it to brackets :)

jrieken:
@filipw - You have a point - I will take a look at how to use it outside a normal MVC app.

nosami:
Like @filipw, I'm not too keen on having the list of controllers here https://github.com/OmniSharp/omnisharp-roslyn/pull/86/files#diff-6dcf536ed4debf80eb70667622c41dd8R19 

Seems like we either need to solve that somehow, or have a test to ensure that all controller actions are reachable via stdio. Otherwise, we're bound to forget to update that list at some point!

Also, I'm not too keen on having 2 Program.cs files, 2 StartUp.cs files etc. They'd need to be kept in sync. Maybe we could address that in a subsequent PR?

src/OmniSharp.Stdio/Logging/StdioLoggerProvider.cs
```diff
@@ -0,0 +1,21 @@
+using System;
+using Microsoft.Framework.Logging;
+
+namespace OmniSharp.Stdio.Logging
+{
+    internal class StdioLoggerProvider : ILoggerProvider
+    {
+
```
davidfowl: Extra space!

src/OmniSharp.Stdio/Program.cs
```diff
@@ -0,0 +1,87 @@
...
+                if (arg == "-s")
+                {
+                    enumerator.MoveNext();
+                    applicationRoot = Path.GetFullPath((string)enumerator.Current);
+                }
+                else if (arg == "-v")
+                {
+                    logLevel = LogLevel.Verbose;
+                }
+                else if (arg == "--hostPID")
```
davidfowl: Useless for stdio

src/OmniSharp.Stdio/Program.cs
```diff
@@ -0,0 +1,87 @@
...
+            });
+
+            Console.WriteLine(new EventPacket()
+            {
+                Event = "started"
+            });
+
+            while (!appShutdownService.ShutdownRequested.IsCancellationRequested)
+            {
+                var line = Console.ReadLine();
```
davidfowl: Do what the other program.main does and wait on ctrl+c

src/OmniSharp.Stdio/Protocol/Packet.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        public int Seq { get; set; }
+
+        public string Type { get { return _type; } }
+
+        public override string ToString() {
+            return JsonConvert.SerializeObject(this, new JsonSerializerSettings() { ContractResolver = LowercaseContractResolver.Instance });
+        }
+
+        private class LowercaseContractResolver : DefaultContractResolver
```
davidfowl: Nooo, use the CamelCase one

src/OmniSharp.Stdio/Protocol/RequestPacket.cs
```diff
@@ -0,0 +1,39 @@
+﻿using System;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace OmniSharp.Stdio.Protocol
+{
+    public class RequestPacket : Packet
+    {
+        private readonly JObject obj;
```
davidfowl: _obj

src/OmniSharp.Stdio/project.json
```diff
@@ -0,0 +1,20 @@
+﻿{
+    "version": "1.0.0-*",
+    "dependencies": {
+        "OmniSharp": "1.0.0-*",
+        "Microsoft.Framework.Runtime": "1.0.0-beta2",
+        "Microsoft.AspNet.Hosting": "1.0.0-beta2",
+        "Newtonsoft.Json": "6.0.8"
+    },
+    "commands": { 
+        "run" : "run"
```
davidfowl: ugh, can we change this to something else. Like:

```JSON
"commands": {
    "OmniSharp.Stdio": "OmniSharp.Stdio"
}
```

jrieken:
Pulling back this PR - I have a solution for this which is sooo much cooler ;-) 

It does pick up @davidfowl crazy idea of asp on stdio. Basically implementing a OWIN server that reads/writes from/to the console. That way we can keep one executable and keep using all fancy ASP and MVC features.

filipw:
sounds awesome!

nosami:
The crazier the better, I reckon!

jrieken:
work is happening here: https://github.com/OmniSharp/omnisharp-roslyn/tree/stdio2. almost done, lacks tests. off for a few days :ski:

nosami:
Heh! I've been liking at it already :) looks like a big improvement to me

jrieken:
new work is in PR #90

PR #87, opened 2015-02-17
added FromBody to /projects action

Yantrio:
How do people want to start e2e testing? Do we ask people to make sure that this actually works via postman/an editor, this should've been caught pre PR-merge :(

### DIFF
--- a/src/OmniSharp/Api/ProjectSystem/OmnisharpController.ProjectSystem.cs
+++ b/src/OmniSharp/Api/ProjectSystem/OmnisharpController.ProjectSystem.cs
@@ -31,7 +31,6 @@ namespace OmniSharp
         }
 
         [HttpPost("/project")]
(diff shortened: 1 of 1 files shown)


cubitouch:
This wasn't working with Postman (before creating test case) so I removed it, sorry about that.
If you have a way to test this using Postman (or anything else) I'm interrested.

nosami:
@cubitouch it would _only_ work in Postman with a POST request when the [FromBody] attribute is present.

cubitouch:
I did, not sure what was not ok... will retry next time :)

nosami:
@Yantrio I don't think we need end to end tests yet. We can just test for the [FromBody] attribute (yuk!) being present on all POST actions?

jchannon:
I'm still having issues with this endpoint even after this fix :(

PR #88, opened 2015-02-17
Automatically use [FromBody] for all complex types

filipw:
This ensures that `[FromBody]` is auto-applied to all complex types. It also respects overriding this, so if you use i.e. `[FromQuery]` on a parameter, that will take precedence.

This will help avoid issues such as #87


nosami:
Nice! Didn't know that you could do that!

nosami:
Thanks :)

Issue #85, opened 2015-02-16
Remove self references from find usages


filipw:
See screenshots

![screenshot 2015-02-15 19 07 49](https://cloud.githubusercontent.com/assets/1710369/6212497/4cb4707a-b5e6-11e4-939f-5e0f33aaca26.png)

![screenshot 2015-02-16 14 15 58](https://cloud.githubusercontent.com/assets/1710369/6212506/5aa20440-b5e6-11e4-8f6f-7242925a4f6b.png)

jrieken:
Let's have a flag ```IncludeDeclaration``` in the request - that makes v1 happy and also full fills this request.

Yantrio:
I also need to de-dupe the results, and/or look into the aspnet vs aspnet core and how that is all returned.

nosami:
IncludeDeclaration is a bad name I think.

I would think of the definition of the symbol, whereas what we really need
is to remove the symbol under the cursor from the usages?

Is there ever a good reason to show the symbol under the cursor when making
a Find Usages request? I can't think of one, but that's what VS does :S

jrieken:
Now that I use findUsages for mark occurrences I actually want the symbol under the cursor.

nosami:
OK. Let's just pick another name for the flag then.
IncludeOriginalLocation??

nosami:
Or just subtract 1 :)

jrieken:
```ExcludeThisLocation``` is my word - it's sort of aligned with https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp/Models/FindUsagesRequest.cs#L9. Tho not saying that was a good choise ;-)

nosami:
Sounds better to me

filipw:
forgot to mention the dupes but I think everyone knows that :) 
this can happen not only between aspnet5/aspnetcore5 but also when the same file is linked under multiple projects

![screenshot 2015-02-16 19 16 35](https://cloud.githubusercontent.com/assets/1710369/6217078/7969ccda-b610-11e4-9d18-134d9091c8a1.png)

Yantrio:
I'll take a look at all that, does anyone mind if (whilst we are on the subject of extending this) , I add a property that returns the Indentifier name also?

jrieken:
[...]

PR #89, opened 2015-02-18
Remove self references with a flag

Yantrio:
Title says it all


nosami:
Needs tests. 

We also need to change TestHelpers.CreateSimpleWorkspace to add each document to 2 different projects to mimic what happens with AspNet5.

Yantrio:
I'll get on that now

Yantrio:
I'll close this PR until I have everything else ready, this is a bigger task than I thought originally. (see #85 for reference)

PR #90, opened 2015-02-23
Support for standard in/out

jrieken:
This PR adds support for standard in/out (in addition to http) to Omnisharp. In contrast to #69 it's implemented as owin server which is driven by standard in. That means you run Omnisharp either with Kestrel (as a web server) or Stdio. The following changes are made with this PR
* add a new flag ```--stdio``` to Omnisharp which means read from/write to the console
* adds Omnisharp.Stdio which implements a owin server that reads single lines from Console.In, interprets them as v8-debugger style JSON message, and generates corresponding messages
* adds a special logger which writes event-packets to Console.Out when logging
* adds Omnisharp.Stdio.Tests which tests the SharedTextWriter and the StdioResponseStream


nosami:
Looks like a huge improvement to me. No idea how to test this thing though :)

jrieken:
@nosami - there are a few tests for the tricky bits (stream, shared writer) but the none for the rest. it's more or less glue code imo.

nosami:
Yup. Happy to merge this once you've fixed @davidfowl nitpicks

davidfowl:
@jrieken needs a rebase

src/OmniSharp.Stdio/StdioResponseStream.cs
```diff
@@ -0,0 +1,187 @@
...
+        {
+            lock (_lock)
+            {
+                CheckDisposed();
+                _disposed = true;
+            }
+
+            try
+            {
+                _writer.Write(",");
```
davidfowl: Should this really be in the Stream implementation?

src/OmniSharp.Stdio/Services/ISharedTextWriter.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+using System.IO;
+using System.Threading.Tasks;
+
+namespace OmniSharp.Stdio.Services
+{
+    public interface ISharedTextWriter
```
davidfowl: This interface feels a bit over engineered to me. Why does it take a callback?

jrieken: That way the control over the lock is encapsulated, esp for error cases.

jrieken:
@davidfowl rebased now

src/OmniSharp.Stdio/StdioResponseStream.cs
jrieken: do you mean the ```,``` or all of it (the rest of the object)? With the comma there is the case that when nothing has been written it's not needed

jrieken:
@davidfowl - as requested I removed the ingenious response stream at the cost of a MemoryStream :disappointed: Also removed the stdio test project which now become pretty much useless.

src/OmniSharp.Stdio/Services/SharedTextWriter.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            lock(_lock)
+            {
+                _writer.WriteLine(value);
+            }
+        }
+
+        public Task WriteLineAsync(object value)
+        {
+            return Task.Factory.StartNew(() => WriteLine(value));
```
davidfowl: What's the point of this?

src/OmniSharp/Program.cs
```diff
@@ -55,14 +57,19 @@ public void Main(string[] args)
...
             }
 
-            Environment = new OmnisharpEnvironment(applicationRoot, serverPort, hostPID, logLevel);
+            Environment = new OmnisharpEnvironment(applicationRoot, serverPort, hostPID, logLevel, transportType);
 
             var config = new Configuration()
              .AddCommandLine(new[] { "--server.urls", "http://localhost:" + serverPort });
 
             var serviceCollection = HostingServices.Create(_serviceProvider, config);
+            serviceCollection.AddSingleton<ISharedTextWriter, SharedConsoleWriter>();
```
davidfowl: This actually doesn't work.

src/OmniSharp.Stdio/Services/SharedTextWriter.cs
jrieken: This is used by the logger so that it doesn't block the rest of the app

src/OmniSharp/Program.cs
jrieken: It does - with the right workaround in ```Startup.cs```

davidfowl:
:shipit:

nosami:
:+1:

PR #96, opened 2015-02-24
limit fuzzy matching 

nosami:
Limit fuzzy matching so that it only matches when the first characters match. This can improve the performance tremendously (especially when you are at the root level of a method and are matching on a single char)

It was returning far too many results previously.


Yantrio:
Looks good to me :+1:

Issue #19, opened 2014-12-14
create endpoint `/reloadsolution`


nosami:
Reloads the current Workspace with the files on disk. Useful after making a big change to the solution such as deleting a folder or switching git branches etc.

cubitouch:
Wanna help on that.
I understand what you need, but is there already an endpoint doing something similar ?

nosami:
There isn't an endpoint for this yet.

It should mimic what the old endpoint
https://github.com/OmniSharp/omnisharp-server/blob/master/OmniSharp/ReloadSolution/ReloadSolutionModule.cs#L10
in omnisharp-server did, i.e. reinitialise the workspace and projects.

It would be useful when switching branch during a git checkout for example
- any time when the solution substantially changes.

cubitouch:
Thanks, I will work on that this week :)

nosami:
:+1:

cubitouch:
Should this return a response ?

nosami:
It should just return true if successful.

TheRealPiotrP:
@DustinCampbell seems this was fixed in #101. Can we close this?

DustinCampbell:
AFAICT, #101 was never merged.

PR #101, opened 2015-02-25
Reloadsolution

cubitouch:
Adding a "/reloadsolution" endpoint to solve this issue : https://github.com/OmniSharp/omnisharp-roslyn/issues/19

### DIFF
--- a/src/OmniSharp/Api/ProjectSystem/OmnisharpController.ProjectSystem.cs
+++ b/src/OmniSharp/Api/ProjectSystem/OmnisharpController.ProjectSystem.cs
@@ -67,7 +67,7 @@ namespace OmniSharp
         {
             foreach (var projectSystem in _projectSystems)
             {
-                projectSystem.Initalize();
+                projectSystem.Reload();
             }
             return true;
         }
--- a/src/OmniSharp/Api/ProjectSystem/OmnisharpController.ProjectSystem.cs
+++ b/src/OmniSharp/Api/ProjectSystem/OmnisharpController.ProjectSystem.cs
@@ -1,5 +1,4 @@
-﻿using System;
-using System.Collections.Generic;
+﻿using System.Collections.Generic;
 using Microsoft.AspNet.Mvc;
 using OmniSharp.AspNet5;
 using OmniSharp.Models;
@@ -62,7 +61,6 @@ namespace OmniSharp
         }
 
         [HttpPost("/reloadsolution")]
-        [HttpGet("/reloadsolution")]
         public bool ReloadSolution()
         {
             foreach (var projectSystem in _projectSystems)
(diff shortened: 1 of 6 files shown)


src/OmniSharp/Api/ProjectSystem/OmnisharpController.ProjectSystem.cs
```diff
@@ -56,5 +60,16 @@ public ProjectInformationResponse CurrentProject(Request request)
                 AspNet5Project = aspNet5ProjectItem
             };
         }
+
+        [HttpPost("/reloadsolution")]
+        [HttpGet("/reloadsolution")]
```
Yantrio: Why do we have post and get?

filipw:
there should probably be a restart or reload method on the project system interface. I think it's not a safe assumption to think that `Initialize` will be OK to call over and over

Yantrio:
I agree with @filipw is this open to memory leaks?

nosami:
Doesn't seem to work. 

I asked @cubitouch to submit it so I could test. It was an experiment :)

cubitouch:
So, I will start over following that list :

- Reload "Workspaces" objects
- Reload "Context" objects
- Reload "Projects" collection
- Reload "ProjectMappings" collection

Please let me know if I'm forgetting something, that I can manage that before you review the PR.

nosami:
@cubitouch sounds good. At the _very_ minimum, your code should compile ;) 

After that, I would expect adding new files, projects, deleting files, should just work after firing the reloadsolution command.

src/OmniSharp/Api/ProjectSystem/OmnisharpController.ProjectSystem.cs
davidfowl: Why do we have a GET

src/OmniSharp/AspNet5/AspNet5ProjectSystem.cs
```diff
@@ -416,7 +416,7 @@ private void TriggerDependeees(string path)
             }
         }
 
-        private void Initialize()
+        private void Startup()
```
davidfowl: What is Startup? That's a bad name.

cubitouch: @davidfowl I don't know more than you... There was a public Initalize() (without the 'i') method on the Interface which I renamed Initialize() (as it should be, IMHO). But there was already a private method with that name so I renamed it too (the Startup one).

What name would you like me to use ? Do you want me to rollback these changes and let the Initalize() method like that ?

nosami:
I've been testing this. It doesn't work reliably yet. Sometimes it works just like you'd expect, but other times it causes intellisense to stop working for the entire solution.

cubitouch:
Is this totally random or some cases just do not work ?
Do you have an idea what can cause the intellisense stop working properly ?

nosami:
I have no idea what is causing the issue because I didn't debug. I spent quite a while testing this though. 

Please test this yourself.

troydai:
@cubitouch this has been stalled for more than a year. Is this PR still valid?

cubitouch:
@troydai no I don't think so, I close it.

PR #105, opened 2015-03-02
updated k runner to beta3 so we are not stuck in the past

Yantrio:
apparently, beta1 is a little outdated


Yantrio:
Seems to hang the CI after tests are run, will try to reproduce locally

Yantrio:
We seem to be blocked by https://github.com/aspnet/Testing/issues/60

Issue #107, opened 2015-03-02
stdio returns invalid json in some cases


jrieken:
When a controller crashes or doesn't write anything to the response stream we will generate JSON like this: ```..."Body":,"Seq":1...```

PR #108, opened 2015-03-02
tackle #107, add (and fix) tests, check for valid request packet

jrieken:
learnt that ```JsonConvert.DeserializeObject``` does not complain about something like this: ```{"KeyWithOutValue":,"KeyWithValue":true}```


nosami:
Looks good. Just realised however that travis isn't running the stdio tests :)

jrieken:
@nosami - pushed another change to run the tests on travis

davidfowl:
LGTM

PR #109, opened 2015-03-04
Stop server support

sriramgd:
From cubitouch's gist. Tested on ubuntu and windows from omnisharp-vim.

### DIFF
--- a/src/OmniSharp/Api/Server/OmnisharpController.cs
+++ b/src/OmniSharp/Api/Server/OmnisharpController.cs
@@ -1,0 +1,20 @@
+﻿using Microsoft.AspNet.Mvc;
+using Microsoft.Framework.Runtime;
+
+namespace OmniSharp.Api.Server
+{
+    public partial class OmnisharpController : Controller
+    {
+        protected readonly IApplicationShutdown _applicationShutdown;
+        public OmnisharpController(IApplicationShutdown applicationShutdown)
+        {
+            _applicationShutdown = applicationShutdown;
+        }
+
+        [HttpPost("/stopserver")]
+        public void StopServer()
+        {
+            _applicationShutdown.RequestShutdown();
+        }
+    }
+}


src/OmniSharp/Api/Server/OmnisharpController.cs
```diff
@@ -0,0 +1,20 @@
...
+        protected readonly IApplicationShutdown _applicationShutdown;
+        public OmnisharpController(IApplicationShutdown applicationShutdown)
+        {
+            _applicationShutdown = applicationShutdown;
+        }
+
+        [HttpPost("/stopserver")]
+        public void StopServer()
+        {
+            _applicationShutdown.RequestShutdown();
```
davidfowl: Do a Task.Run here:

```C#
Task.Run( _applicationShutdown.RequestShutdown);
```

src/OmniSharp/Api/Server/OmnisharpController.cs
```diff
@@ -0,0 +1,20 @@
+﻿using Microsoft.AspNet.Mvc;
+using Microsoft.Framework.Runtime;
+
+namespace OmniSharp.Api.Server
+{
+    public partial class OmnisharpController : Controller
+    {
+        protected readonly IApplicationShutdown _applicationShutdown;
```
nosami: this should be private

src/OmniSharp/Api/Server/OmnisharpController.cs
```diff
@@ -0,0 +1,20 @@
+﻿using Microsoft.AspNet.Mvc;
+using Microsoft.Framework.Runtime;
+
+namespace OmniSharp.Api.Server
+{
+    public partial class OmnisharpController : Controller
```
nosami: This should be a new controller as it has different dependencies to OmniSharpController. Should be
```public class StopServerController```

sriramgd:
Hey guys, Could you decline this request. After some more testing, I notice that kestrel throws a 'EBUSY resource busy or locked' sometimes on ubuntu. The server stops, but the exception is unhandled.

nosami:
Sure. Thanks anyway :)

Issue #97, opened 2015-02-25
Add navigate up/down by method


nosami:
Utility method to make navigating up and down by method faster. The response should be the same format as the go to definition response.

If there is no further method up or down, then the response should contain the location of the first or last method. If there are no methods at all, return the current location to make client side integration as simple as possible.

sriramgd:
If no one else if working on it, can I take this?

nosami:
sure! thanks

sriramgd:
https://github.com/OmniSharp/omnisharp-roslyn/pull/110

nosami:
fixed by #110

PR #110, opened 2015-03-04
Navigate up and down. 

sriramgd:
For https://github.com/OmniSharp/omnisharp-roslyn/issues/97

### DIFF
--- a/src/OmniSharp/Api/Navigation/OmnisharpController.NavigateUpDown.cs
+++ b/src/OmniSharp/Api/Navigation/OmnisharpController.NavigateUpDown.cs
@@ -1,0 +1,74 @@
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Microsoft.AspNet.Mvc;
+using OmniSharp.Models;
+
+namespace OmniSharp
+{
+    public partial class OmnisharpController
+    {
+        [HttpPost("navigateup")]
+        public async Task<NavigateResponse> NavigateUp(Request request)
+        {
+            return await Navigate(request, IsCloserNodeUp);
+        }
+
+        [HttpPost("navigatedown")]
+        public async Task<NavigateResponse> NavigateDown(Request request)
+        {
+            return await Navigate(request, IsCloserNodeDown);
+        }
+
+        private async Task<NavigateResponse> Navigate(Request request, Func<FileMemberElement, FileMemberElement, Request, bool> IsCloserNode)
+        {
+            var stack = new List<FileMemberElement>(await StructureComputer.Compute(_workspace.GetDocuments(request.FileName)));
+            var response = new NavigateResponse();
+            //Retain current line in case we dont need to navigate.
+            response.Line = request.Line;
+            response.Column = request.Column;
+
+            FileMemberElement closestNode = null;
+            FileMemberElement thisNode = null;
+            while (stack.Count > 0)
+            {
+                var node = stack[0];
+                stack.Remove(node);
+                var isCloserNode = IsCloserNode(node, closestNode, request);
+                if (isCloserNode)
+                {
+                    closestNode = node;
+                }
+                if (node.Location.Line == request.Line)
+                {
+                    thisNode = node;
+                }
+                stack.AddRange(node.ChildNodes);
+            }
+
+            //If there is a closest node, use its line and column.
+            //or if we are on the last node, adjust column.
+            //if we are above the first or below the last node, do nothing.
+            if (closestNode != null)
+            {
+                response.Line = closestNode.Location.Line;
+                response.Column = closestNode.Location.Column;
+            }
+            else if (thisNode != null)
+            {
+                response.Column = thisNode.Location.Column;
+            }
+            return response;
+        }
+
+        private bool IsCloserNodeUp(FileMemberElement candidateClosestNode, FileMemberElement closestNode, Request request)
+        {
+            return ((candidateClosestNode.Location.Line < request.Line) && (closestNode == null || candidateClosestNode.Location.Line > closestNode.Location.Line));
+        }
+
+        private bool IsCloserNodeDown(FileMemberElement candidateClosestNode, FileMemberElement closestNode, Request request)
+        {
+            return ((candidateClosestNode.Location.Line > request.Line) && (closestNode == null || candidateClosestNode.Location.Line < closestNode.Location.Line));
+        }
+    }
+}
--- a/src/OmniSharp/Models/NavigateResponse.cs
+++ b/src/OmniSharp/Models/NavigateResponse.cs
@@ -1,0 +1,8 @@
+namespace OmniSharp.Models
+{
+    public class NavigateResponse
+    {
+        public int Line { get; set; }
+        public int Column { get; set; }
+    }
+}
--- a/tests/OmniSharp.Tests/NavigateUpDownFacts.cs
+++ b/tests/OmniSharp.Tests/NavigateUpDownFacts.cs
@@ -1,0 +1,748 @@
+﻿using System.Threading.Tasks;
+using OmniSharp.Models;
+using Xunit;
+
+namespace OmniSharp.Tests
+{
+    enum NavigateDirection
+    {
+        UP,
+        DOWN
+    }
+
+    public class NavigateUpDownFacts
+    {
+        [Fact]
+        public async Task NavigateUp_ReturnsCorrectPositionFromMethodBodyToMethodName()
+        {
+            const string fileContent = @"using System;
+                using System.Collections.Generic;
+                using System.Threading.Tasks;
+                using Microsoft.AspNet.Mvc;
+                using OmniSharp.Models;
+
+                namespace OmniSharp
+                {
+                    public class NavigateController
+                    {
+                        [HttpPost(""navigateup"")]
+                        public async Task<NavigateResponse> %NavigateUp(Request request)
+                        {
+                            return await Navi$gate(request);
+                        }
+
+                        private async Task<NavigateResponse> Navigate(Request request) {
+                            return new NavigateResponse();
+                        }
+                    }";
+
+            await AssertPosition(fileContent, NavigateDirection.UP);
+        }
+
+        [Fact]
+        public async Task NavigateUp_ReturnsCorrectPositionFromMethodNameToPreviousMethodName()
+        {
+            const string fileContent = @"using System;
+                using System.Collections.Generic;
+                using System.Threading.Tasks;
+                using Microsoft.AspNet.Mvc;
+                using OmniSharp.Models;
+
+                namespace OmniSharp
+                {
+                    public class NavigateController
+                    {
+                        [HttpPost(""navigateup"")]
+                        public async Task<NavigateResponse> %NavigateUp(Request request)
+                        {
+                            return await Navigate(request);
+                        }
+
+                        [HttpPost(""navigatedown"")]
+                        public async $Task<NavigateResponse> NavigateDown(Request request)
+                        {
+                            return await Navigate(request);
+                        }
+
+                        private async Task<NavigateResponse> Navigate(Request request) {
+                            return new NavigateResponse();
+                        }
+                    }
+                }";
+
+            await AssertPosition(fileContent, NavigateDirection.UP);
+        }
+
+        [Fact]
+        public async Task NavigateUp_ReturnsCorrectPositionFromMethodNameToPreviousField()
+        {
+            const string fileContent = @"using System;
+                using System.Collections.Generic;
+                using System.Threading.Tasks;
+                using Microsoft.AspNet.Mvc;
+                using OmniSharp.Models;
+
+                namespace OmniSharp
+                {
+                    public class NavigateController
+                    {
+                        private string %text;
+
+                        [HttpPost(""navigateup"")]
+                        public async Task<NavigateResponse> $NavigateUp(Request request)
+                        {
+                            return await Navigate(request);
+                        }
+
+                        [HttpPost(""navigatedown"")]
+                        public async Task<NavigateResponse> NavigateDown(Request request)
+                        {
+                            return await Navigate(request);
+                        }
+
+                        private async Task<NavigateResponse> Navigate(Request request) {
+                            return new NavigateResponse();
+                        }
+                    }
+                }";
+
+            await AssertPosition(fileContent, NavigateDirection.UP);
+        }
+
+        [Fact]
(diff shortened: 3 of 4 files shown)


sriramgd:
This is ready for review.

src/OmniSharp/Api/Navigation/OmnisharpController.NavigateUpDown.cs
```diff
@@ -0,0 +1,74 @@
...
+                {
+                    break;
+                }
+                stack.AddRange(node.ChildNodes);
+            }
+            return new ObjectResult(ret);
+        }
+
+        [HttpPost("navigateup")]
+        public async Task<IActionResult> NavigateUp(Request request)
```
nosami: put the Controller actions above the private methods.... it's a convention we have

src/OmniSharp/Api/Navigation/OmnisharpController.NavigateUpDown.cs
```diff
@@ -0,0 +1,74 @@
...
+                qf.Column = node.Location.Column;
+                retVal = true;
+            }
+            return retVal;
+        };
+
+        private async Task<IActionResult> Navigate(Request request, bool toBreak, Func<Request, FileMemberElement, GotoDefinitionResponse, bool> navigateFunc)
+        {
+            var stack = new List<FileMemberElement>(await StructureComputer.Compute(_workspace.GetDocuments(request.FileName)));
+            var ret = new GotoDefinitionResponse();
```
nosami: Shouldn't really be returning a GotoDefinitionResponse here (although we need an object with the same signature)

nosami: ret is a bad name.

src/OmniSharp/Api/Navigation/OmnisharpController.NavigateUpDown.cs
```diff
@@ -0,0 +1,74 @@
...
+                stack.Remove(node);
+                //Need to call the passed in up or down delegate
+                var found = navigateFunc(request, node, ret);
+                if (toBreak && found)
+                {
+                    break;
+                }
+                stack.AddRange(node.ChildNodes);
+            }
+            return new ObjectResult(ret);
```
nosami: Just return the object instead of the ObjectResult. It will make it simpler to test and we need POCOs for the stdout interface I believe.

src/OmniSharp/Api/Navigation/OmnisharpController.NavigateUpDown.cs
```diff
@@ -0,0 +1,74 @@
...
+using OmniSharp.Models;
+using System;
+
+namespace OmniSharp
+{
+    public partial class OmnisharpController
+    {
+        private Func<Request, FileMemberElement, GotoDefinitionResponse, bool> NavigateUpFunc = (request, node, qf) =>
+        {
+            bool retVal = false;
```
nosami: retVal is a bad name

src/OmniSharp/Api/Navigation/OmnisharpController.NavigateUpDown.cs
```diff
@@ -0,0 +1,74 @@
...
+            if (node.Location.Line < request.Line || (node.Location.Line == request.Line && node.Location.Column < request.Column))
+            {
+                qf.Line = node.Location.Line;
```
nosami: what's qf?

src/OmniSharp/Api/Navigation/OmnisharpController.NavigateUpDown.cs
```diff
@@ -0,0 +1,74 @@
...
+            return await Navigate(request, toBreak, NavigateUpFunc);
+        }
+
+        [HttpPost("navigatedown")]
+        public async Task<IActionResult> NavigateDown(Request request)
+        {
+            bool toBreak = true;
+            return await Navigate(request, toBreak, NavigateDownFunc);
+        }
+
```
nosami: delete this line

src/OmniSharp/Api/Navigation/OmnisharpController.NavigateUpDown.cs
```diff
@@ -0,0 +1,74 @@
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Microsoft.AspNet.Mvc;
+using OmniSharp.Models;
+using System;
```
nosami: `using System` should be first

sriramgd:
Updated based on code review. Refactored not to depend on any implicit ordering of nodes. Now both navigate up and down are symmetric and work the same way.
Tested with properties, methods and classes.

davidfowl:
Where are the unit tests?

src/OmniSharp/Api/Navigation/OmnisharpController.NavigateUpDown.cs
```diff
@@ -0,0 +1,75 @@
...
+                response.Column = closestNode.Location.Column;
+            }
+            else if (thisNode != null)
+            {
+                response.Column = thisNode.Location.Column;
+            }
+            return response;
+        }
+
+        private Func<FileMemberElement, FileMemberElement, Request, bool> IsCloserNodeUpFunc = (candidateClosestNode, closestNode, request) =>
```
davidfowl: Why re these funcs instead of methods?

src/OmniSharp/Models/NavigateResponse.cs
```diff
@@ -0,0 +1,9 @@
+namespace OmniSharp.Models
+{
+    public class NavigateResponse
+    {
+        public string FileName { get; set; }
```
jrieken: will you ever end up in a different file? Or should this be a a _code pointer_ thing (line and position)? Iff so it should have a different name

src/OmniSharp/Api/Navigation/OmnisharpController.NavigateUpDown.cs
```diff
@@ -0,0 +1,75 @@
...
+                if (node.Location.Line == request.Line)
+                {
+                    thisNode = node;
+                }
+                stack.AddRange(node.ChildNodes);
+            }
+
+            //If there is a closest node, use its line and column.
+            //or if we are on the last node, adjust column.
+            //if we are above the first or below the first node, do nothing.
```
jrieken: how about (optionally, defined by request) starting at the other end and you are the top/bottom most node?

Yantrio:
Do we want this for all filemembers, or just methods/properties ?

src/OmniSharp/Api/Navigation/OmnisharpController.NavigateUpDown.cs
sriramgd: Added unit tests - NavigateUpDownFacts.cs.
I could use methods, but I would then need either a delegate definition to pass the methods in as parameters to the Navigate method or an if-else statement in the Navigate method.

src/OmniSharp/Models/NavigateResponse.cs
sriramgd: This will not end up in a different file. Does not try follow a partial, since there is no real order for the partial. Is there any use case you had in mind?

sriramgd:
Updated based on second review. Unit tests added. Ready for another review.

src/OmniSharp/Api/Navigation/OmnisharpController.NavigateUpDown.cs
sriramgd: Since this will be additive and not a breaking update, could we do this as a separate PR?

src/OmniSharp/Api/Navigation/OmnisharpController.NavigateUpDown.cs
davidfowl: You can pass a method group. You don't have to define them as Func<...> to pass them as functions. The compiler does that for you. Turn these into methods.

tests/OmniSharp.Tests/NavigateUpDownFacts.cs
```diff
@@ -0,0 +1,845 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using OmniSharp.Filters;
+using OmniSharp.Models;
+using Xunit;
+
+namespace OmniSharp.Tests
+{
+    enum NavigateDirection {UP, DOWN}
```
davidfowl: This looks weird. Can you expand this so that it looks like:

```
enum NavigateDirection 
{
    UP, 
    DOWN
}
```

src/OmniSharp/Models/NavigateResponse.cs
jrieken: Maybe not have the FileName than?

nosami:
Nitpick - these tests would be clearer if we had another symbol in the source (similar to $) to show where the cursor will finish up.

nosami:
Also, am I correct in thinking that the cursor finishes up on the method name, not the first character of the line? Is that what we want?

nosami:
I added some keybindings for this over at https://github.com/nosami/omnisharp-demo/commit/9f374dbd1658e63d6071fb3d407a927fa18f1a72

Works well!

src/OmniSharp/Api/Navigation/OmnisharpController.NavigateUpDown.cs
sriramgd: Done. Didnt know that before, thanks.

src/OmniSharp/Models/NavigateResponse.cs
sriramgd: Good point, fixed.

tests/OmniSharp.Tests/NavigateUpDownFacts.cs
sriramgd: Done.

sriramgd:
Tests refactored to use a source and target marker. I prefer ending up on the definition of the class/method/property/field. Maybe the client can move to start of line etc. as required? Comments squashed in git (I think).

Ready for third review.

tests/OmniSharp.Tests/NavigateUpDownFacts.cs
```diff
@@ -0,0 +1,751 @@
...
+using System.Threading.Tasks;
+using OmniSharp.Filters;
+using OmniSharp.Models;
+using Xunit;
+
+namespace OmniSharp.Tests
+{
+    enum NavigateDirection 
+    {
+      UP,
```
nosami: Indentation is off.

tests/OmniSharp.Tests/NavigateUpDownFacts.cs
```diff
@@ -0,0 +1,751 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using OmniSharp.Filters;
```
nosami: System.Collections.Generic, System.Linq and OmniSharp.Filters aren't used.

tests/OmniSharp.Tests/NavigateUpDownFacts.cs
```diff
@@ -0,0 +1,751 @@
...
+    enum NavigateDirection 
+    {
+      UP,
+      DOWN
+    }
+
+    public class NavigateUpDownFacts
+    {
+
+        private string foo = "test";
```
nosami: what's this? :) delete the line above and below too

tests/OmniSharp.Tests/NavigateUpDownFacts.cs
```diff
@@ -0,0 +1,751 @@
...
+            {
+                return await controller.NavigateUp(request);
+            }
+            else 
+            {
+                return await controller.NavigateDown(request);
+            }
+        }
+
+        private async Task AssertPosition(string fileContent, NavigateDirection navigateDirection, int finalCursorColumnOffset = 1)
```
nosami: I don't see why this is needed. 

Should always be 0 or some constant. It shouldn't ever change. Looks as if it's masking some bug in the test code, because I can see that this code works.

tests/OmniSharp.Tests/NavigateUpDownFacts.cs
```diff
@@ -0,0 +1,751 @@
...
+      UP,
+      DOWN
+    }
+
+    public class NavigateUpDownFacts
+    {
+
+        private string foo = "test";
+
+        private async Task<NavigateResponse> SendRequest(OmnisharpWorkspace workspace, string fileName, string fileContent, NavigateDirection upOrDown)
```
nosami: not critical, but I'd prefer it if test helper methods came after the tests.

tests/OmniSharp.Tests/NavigateUpDownFacts.cs
```diff
@@ -0,0 +1,751 @@
...
+            using Microsoft.AspNet.Mvc;
+            using OmniSharp.Models;
+
+            namespace OmniSharp
+            {
+                public class NavigateController
+                {
+                    private string text;
+
+                    public string M%oreText
```
nosami: Why are some of the these percent markers one character in?

tests/OmniSharp.Tests/NavigateUpDownFacts.cs
sriramgd: Done.

tests/OmniSharp.Tests/NavigateUpDownFacts.cs
sriramgd: Done, sorry missed that.

tests/OmniSharp.Tests/NavigateUpDownFacts.cs
sriramgd: Was testing property navigation using this file :) Removed.

tests/OmniSharp.Tests/NavigateUpDownFacts.cs
sriramgd: Right. Now I realize I went adjusting for a bug instead of fixing the cause. Fixed.

tests/OmniSharp.Tests/NavigateUpDownFacts.cs
sriramgd: Was a bug in the test code. Fixed.

tests/OmniSharp.Tests/NavigateUpDownFacts.cs
sriramgd: Done.

sriramgd:
Ready for fourth :) review.

nosami:
Looks good to me. Thanks for this :)

PR #111, opened 2015-03-06
Support text changes for rename and code actions

jrieken:
This PR adds support for text changes to the ```rename``` and ```code actions``` route such that they can return a list of text changes instead of the new buffer. The idea is that with text changes editor can update without flickering and without losing scroll positions etc. The changes go like this:

* Add ```WantsTextChanges``` flag to the ```CodeActionRequest``` and ```RenameRequest``` as well as adding a ```Changes``` field to their responses
* Move around and reuse the text change logic we already have to formatting
* Modify the rename and code actions controllers such that they either produce text changes or new buffers
* Update ```RenameFacts``` to also check for valid text changes


nosami:
LGTM

jrieken:
:+1:

jrieken:
thanks

Issue #113, opened 2015-03-06
rename explodes when executed on a none-source symbol


jrieken:
Have code like this:

```C#
               using System;
                public class Program
                {
                    public static void Main()
                    {
                        Console.Write(1);
                    }
                }
```

And rename ```Write``` and :boom: you get

```
  System.ArgumentException : Symbol "Write" is not from source.
      Stack Trace:
            at Microsoft.CodeAnalysis.Rename.ConflictEngine.ConflictResolver.ResolveConflictsAsync(RenameLocationSet renameLocationSet, String originalText, String replacementText, OptionSet optionSet, CancellationToken cancellationToken)
            at Microsoft.CodeAnalysis.Rename.Renamer.<RenameSymbolAsync>d__1.MoveNext()
         --- End of stack trace from previous location where exception was thrown ---
            at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
            at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
            at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
            at OmniSharp.OmnisharpController.<Rename>d__25.MoveNext() in C:\Users\jrieken\Code\omnisharp-roslyn\src\OmniSharp\Api\Refactoring\OmnisharpController.Rename.cs:line 30
         --- End of stack trace from previous location where exception was thrown ---
            at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
            at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
```

PR #114, opened 2015-03-06
fix for #113

jrieken:
Catches ```ArgumentException``` from the renamer and fills a error message. Added a test which tries to rename ```Console.Write```


nosami:
Nice!

PR #118, opened 2015-03-11
Findsymbols

nosami:
Adds /findsymbols endpoint to return all symbols in the workspace.

### DIFF
--- a/src/OmniSharp/Api/Navigation/OmnisharpController.FindSymbols.cs
+++ b/src/OmniSharp/Api/Navigation/OmnisharpController.FindSymbols.cs
@@ -1,8 +1,10 @@
-using System.Collections.Generic;
+using System;
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Mvc;
 using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.FindSymbols;
+using OmniSharp.Extensions;
 using OmniSharp.Models;
 
 namespace OmniSharp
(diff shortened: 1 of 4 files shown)


aeschli:
@nosami - Can we add a search pattern as input, to limit the number of symbols returned? I'd suggest to do it similar to the CompletionText for AutoComplete (prefix match, substring match and camelcase).

nosami:
@jrieken I never saw that before. Can you send me a link? I'll use that if it works.

I just had a look through here https://github.com/dotnet/roslyn/tree/b6484300dfafb43af0c27e542ec457a7583e1aa8/src/Workspaces/Core/Portable/FindSymbols but didn't see anything.

nosami:
@aeschli sure!

nosami:
I just found this internal method https://github.com/dotnet/roslyn/blob/b6484300dfafb43af0c27e542ec457a7583e1aa8/src/Workspaces/Core/Portable/FindSymbols/SymbolTree/SymbolTreeInfo.cs#L160

Is that the one?

jrieken:
@nosami - I was thinking of this one: https://github.com/dotnet/roslyn/blob/b6484300dfafb43af0c27e542ec457a7583e1aa8/src/Workspaces/Core/Portable/FindSymbols/SymbolFinder_Declarations.cs#L260

nosami:
Thanks! Will try it out tonight.

nosami:
The SymbolFinder api was indeed a good choice :) Works well.

I added a new /findsymbolswithfilter endpoint. Not sure how you guys feel about this, but I couldn't find a way to make this backwards compatible (nullable request doesn't help) - maybe I'm just being dense here.

It now also returns types as well as methods etc. I'm thinking of leaving it this way and not bothering with a separate endpoint for types. What do you thing?

nosami:
Oh... and as a side effect of doing this, I think I've also fixed the intellisense bug..... hit it when testing this code.

jrieken:
Kudus for fixing the IntelliSense bug. Regarding ```/findSymbolsWithFilter``` it don't really mind, tho I would prefer a specialized request. Isn't [FromBody] smart enough to create an empty object (all members null) when there is no POST body at all?

nosami:
No. Well, it didn't seem that way anyway. It was returning a 404 from the existing editor code when I had the request parameter in there. 

I tried `string filter` too instead of a request object. Same.

jrieken:
I guess you already tried ```MyRequest? req=null```. Pretty sure there must be a way to achieve this? Right @davidfowl? Otherwise I am OK with another route

nosami:
No... I forgot about that :S Let me try quickly.

nosami:
Ok... That works :D Gimme a minute :)

src/OmniSharp/Api/Navigation/OmnisharpController.FindSymbols.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        private QuickFix ConvertSymbol(ISymbol symbol, Location location)
+        {
+            var lineSpan = location.GetLineSpan();
+            var path = lineSpan.Path;
+            var documents = _workspace.GetDocuments(path);
+
+            return new QuickFix
+            {
+                Text = new SnippetGenerator().GenerateSnippet(symbol),
```
jrieken: does this mean we have the snippet placeholder in the text?

nosami: No. See the test output here https://github.com/OmniSharp/omnisharp-roslyn/blob/findsymbols/tests/OmniSharp.Tests/FindSymbolsFacts.cs#L44

I actually thought about renaming this class when I wrote this - the SnippetGenerator has 2 purposes :)

jrieken:
:+1:

Issue #106, opened 2015-03-02
Should OmniSharp be in charge of file watching?


jrieken:
I wonder if OmniSharp should be watching files that contribute to the state of a project. Such files are  ```*.cs```-files but also ```project.json```, ```*.proj``` etc. Basically, everything that is loaded (or required) to build up the project state during ```IProjectSystem#Initialize```. My reasoning for doing this in OmniSharp is the following:
* file watching is hard and if it's in OmniSharp everyone can benefit
* editors usually don't know what files a project is build up from but make qualified guesses (glob-patterns etc)
* it would make routes like ```/reloadSolution``` not needed anymore

Also, there is a differentiation between files that are _open in the editor_ and those that are not. For instance, editor plugins send buffer updates for ```*.cs``` files but not for ```project.json```. So maybe, once a file has been receiving updates via the ```/updatebuffer``` or ```/changebuffer``` route it should not be watched any more. 

Another question to which I don't know an answer: How good is the mono file watcher?

nosami:
I think in a perfect world, we would handle this on the server. 

However, the mono file watcher is currently terrible in 3.12 (but is currently being worked on). @davidfowl had to remove some file watching code because CPU usage frequently spiked to 100%

Also, I don't see why the editor needs to know what project a file belongs to. None of the editors are project aware. 

Although the editors don't update the server with `project.json` today, there is no reason why that couldn't happen.

It's a little frustrating. I agree with you, but the reason I added `/reloadsolution` in the first place was to deal with deficiencies in FSW. It would be really nice to not have to put all this stuff in the editor.

davidfowl:
I added https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp/Services/IFileSystemWatcher.cs a while back so we could get started on something (the interface might need some clean up). 

So lets do this:

- Implement IFileSystemWatcher and all of the logic within each project system (I think some things are there already) and get that working end to end.
- Have the editor watch files in the short term, and send a /fileschanged event for any file add/remove/change (basically a file system watcher replacement). This is ok because the server will filter events based on what it watched.
- When we have a viable replacement, we swap the IFileSystemWatcher implementation.

jrieken:
Great - I will start by adding a ```filesChanged``` route which drives a special file watcher implementation. Next step is to establish a way to communicate back the effects of such changes, like new compile errors when removing a reference etc.

jrieken:
[...]

PR #119, opened 2015-03-11
Files changed

jrieken:
This PR covers the ground work for #106 and I would like get it out before it gets too big. Until now, it contains these changes

* replace the (flaky) file event watcher with a ```filesChanged``` route which drives the ```IFileWatcher```
* Add an IEventEmitter interface which allows OmniSharp to talk back to an editor (such that it can communicate the effects of changed files)
* Listen for some workspace events (project modifications) and send those events (in the shape of ```ProjectInformationResponse```-objects) to the connected editor
* Make ```ProjectInformationResponse``` also include the source files of a project such that an editor can use this information, for instance requesting diagnostics for more than the open files.

What still needs to be done is refining the file events (adding types ```added```, ```changed```, and ```deleted```) and make the ASP project system smarter such that it picks up new files.

### DIFF
--- a/src/OmniSharp/Roslyn/ProjectEventForwarder.cs
+++ b/src/OmniSharp/Roslyn/ProjectEventForwarder.cs
@@ -1,0 +1,111 @@
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
+using OmniSharp.AspNet5;
+using OmniSharp.Models;
+using OmniSharp.MSBuild;
+using OmniSharp.Services;
+
+namespace OmniSharp.Roslyn
+{
+    public class ProjectEventForwarder
+    {
+        private readonly AspNet5Context _aspnet5Context;
+        private readonly MSBuildContext _msbuildContext;
+        private readonly OmnisharpWorkspace _workspace;
+        private readonly IEventEmitter _emitter;
+        private readonly ISet<SimpleWorkspaceEvent> _queue = new HashSet<SimpleWorkspaceEvent>();
+        private readonly object _lock = new object();
+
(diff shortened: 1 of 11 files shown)


nosami:
I think right now, I'd be really happy if we could just add a new cs file to the workspace :) Could you add that in and we can add the other stuff later?

Some guy on jabbr was just complaining that he has to restart sublime to add a new file :P

src/OmniSharp/Roslyn/ProjectEventForwarder.cs
```diff
@@ -0,0 +1,88 @@
...
+                        Task.Factory.StartNew(async () =>
+                        {
+                            await Task.Delay(500);
+
+                            lock (_lock)
+                            {
+                                _queue.Remove(fileNameAndEventKind);
+
+                                object payload = null;
+                                if (fileNameAndEventKind.Item2 != WorkspaceChangeKind.ProjectRemoved)
```
nosami: Tuples suck in C#. What is Item2? Could we use an anonymous object instead?

davidfowl: Yah or make a type.

src/OmniSharp/Roslyn/ProjectEventForwarder.cs
```diff
@@ -0,0 +1,111 @@
...
+            _aspnet5Context = aspnet5Context;
+            _msbuildContext = msbuildContext;
+            _workspace = workspace;
+            _emitter = emitter;
+            _workspace.WorkspaceChanged += OnWorkspaceChanged;
+        }
+
+        private void OnWorkspaceChanged(object source, WorkspaceChangeEventArgs args)
+        {
+
```
davidfowl: Extra space.

jrieken:
@davidfowl - squashed and without extra space

PR #124, opened 2015-03-13
Fixed issues with build

markrendle:
When I tried to `build.sh` the latest master to use in OmniSharp-Sublime I found:

* the `OmniSharp` build failed because of a missing namespace in `Program.cs`
* when running the server from Sublime it failed because it needed `Microsoft.Composition 1.0.30` and the bundled version was `1.0.27`

This fixes those two problems.

### DIFF
--- a/src/OmniSharp/Program.cs
+++ b/src/OmniSharp/Program.cs
@@ -2,6 +2,7 @@
 using System.Diagnostics;
 using System.IO;
 using System.Threading;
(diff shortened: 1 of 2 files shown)


nosami:
Damn.... i removed a bunch of unused usings the other day.. that must have been needed. I'll test this shortly and merge if good. Thanks!

nosami:
cheers! :+1:

PR #125, opened 2015-03-13
fix windows launch bug

jrieken:
With the dnx adoption we broke launching on windows, because ```File.Exists``` isn't so smart to figure out the ```klr``` is ```klr.exe``` when executed.


davidfowl:
Goof catch!

PR #127, opened 2015-03-16
Create an Appveyor CI build

nosami:
So... this seems like a good start. It's not perfect, but it's on a par with our Travis CI build for linux I think.


nosami:
Having a Windows build shows that the TextChangesAreSortedLastFirst test is failing on Windows.

Issue #131, opened 2015-03-17
Support project wide code check


jrieken:
When changing code I often break other files that depend on the file I am currently editing, like tests. IMO it would be beneficial if OmniSharp gives me a change to check the whole project at once. I am thinking of ```/codecheck``` without a filename.

nosami:
Sounds good to me.

jrieken:
on a second though we can also use the event emitter to send back diagnostics whenever buffers update. that way Omnisharp has a chance to be smart and delay validation when documents constantly change. but maybe that's a different story

nosami:
fixed by #133

PR #133, opened 2015-03-17
#131 /codecheck checks the workspace if filename is missing

jrieken:


tests/OmniSharp.Tests/DiagnosticsFacts.cs
```diff
@@ -0,0 +1,44 @@
...
+        {
+            var workspace = TestHelpers.CreateSimpleWorkspace(new Dictionary<string, string>
+            {
+                { "a.cs", "class C { int n = true; }" }
+            });
+
+            var controller = new OmnisharpController(workspace, null);
+            var quickFixes = await controller.CodeCheck(new Request() { FileName = "a.cs" });
+
+            Assert.Equal(2, quickFixes.QuickFixes.Count());
```
nosami: why is it 2?

jrieken: That's a good question ;-) There are two errors because we have two projects, but then we have logic in place to de-dup error messages. I think it's a regression which we never saw due to be lack of a test (https://github.com/OmniSharp/omnisharp-roslyn/blame/master/src/OmniSharp/Models/QuickFix.cs#L36) 

I wonder if @Yantrio  knows if equality of a quick fix really requires the list of projects or we should use a special comparator to de-dup quick fixes.

nosami: I was gonna take a look at this the other night, but never got around to it. 

I think that all the other code that relies on Equals and GetHashCode is covered by tests now so we should just be able to remove it and see.

jrieken: Seems to be fine

nosami:
LGTM... Cheers

jrieken:
thanks

PR #147, opened 2015-03-30
trying to fix #128

### LINKED ISSUE #128: Failing test on Windows
We've got a test that fails on Windows but not nix

`FormattingFacts.TextChangesAreSortedLastFirst`

For some reason its returning 8 spaces as the text change in windows and only a single space on nix.

Had a quick peek around but not familiar with the code so may be one for @jrieken :)

jrieken:


jrieken:
needs a different angle

PR #149, opened 2015-04-02
Signature Help

jrieken:
This PR add a new route ```/signatureHelp``` which can be used to get information about signatures at a given position. It will return a list of signatures if invoked inside a constructor call or method call. Each signature represents an overload with its specific comment and parameters. The active signature and active parameter are denoted by the result object.

### DIFF
--- a/src/OmniSharp/Api/Signatures/OmnisharpController.SignatureHelp.cs
+++ b/src/OmniSharp/Api/Signatures/OmnisharpController.SignatureHelp.cs
@@ -59,7 +59,13 @@ namespace OmniSharp
             }
             else if (!symbolInfo.CandidateSymbols.IsEmpty)
             {
+
+                // collect types of invocation to select overload
+                var typeInfos = invocation.ArgumentList.Arguments
+                    .Select(argument => semanticModel.GetTypeInfo(argument.Expression));
+
                 // method has overloads
+                var bestScore = int.MinValue;
                 var signatures = new List<SignatureHelpItem>();
                 foreach (var symbol in symbolInfo.CandidateSymbols)
                 {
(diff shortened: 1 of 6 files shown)


src/OmniSharp/Models/SignatureHelp.cs
```diff
@@ -0,0 +1,13 @@
+using System.Collections.Generic;
+
+namespace OmniSharp.Models
+{
+    public class SignatureHelp
+    {
+        public IEnumerable<SignatureHelpItem> Signatures { get; set; }
+
+        public int ActiveSignature { get; set; }
```
nosami: Maybe this might be a little easier to work with if we had `bool Active` on SignatureHelpItem?

src/OmniSharp/Models/SignatureHelp.cs
```diff
@@ -0,0 +1,13 @@
...
+        
+        public int ActiveParameter { get; set; }
```
nosami: Same.

nosami:
Minor niggle - not too important. Happy to merge this if you want.

src/OmniSharp/Api/Signatures/OmnisharpController.SignatureHelp.cs
```diff
@@ -0,0 +1,182 @@
...
+{
+    public partial class OmnisharpController
+    {
+        [HttpPost("signatureHelp")]
+        public async Task<SignatureHelp> GetSignatureHelp(Request request)
+        {
+            foreach (var document in _workspace.GetDocuments(request.FileName))
+            {
+                var response = await GetSignatureHelp(document, request);
+                if (response != null)
```
davidfowl: What happens to overloads that may exist in one project but not the other?

jrieken:
@nosami I made it use two indices because it makes it easy to consume. Instead of iterating you can simply do something like ```response[activeSignature][activeParameter]```. That will give you the most interesting information with ease.

jrieken:
working on the thing with different projects now

jrieken:
@davidfowl  overloads from different projects are now handled

nosami:
Looks good

PR #150, opened 2015-04-03
Adding fixusings support

sriramgd:
1. Add missing usings
2. Send a dictionary of key-list for ambiguous usings
3. Remove redundant usings

TODO: 
1. Remove unused using statements
2. Sort using statements
3. If possible, replace RemoveRedundantUsings by something better from the framework

### DIFF
--- a/src/OmniSharp/Api/Refactoring/OmnisharpController.FixUsings.cs
+++ b/src/OmniSharp/Api/Refactoring/OmnisharpController.FixUsings.cs
@@ -1,0 +1,27 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.AspNet.Mvc;
+using OmniSharp.Models;
+
+namespace OmniSharp
+{
+    public partial class OmnisharpController
+    {
+        [HttpPost("fixusings")]
+        public async Task<FixUsingsResponse> FixUsings(Request request)
+        {
+            var response = new FixUsingsResponse(string.Empty, new List<QuickFix>());
+            var document = _workspace.GetDocument(request.FileName);
+
+            if (document != null)
+            {
+                var semanticModel = await document.GetSemanticModelAsync();
+                response = await FixUsingsWorker.AddMissingUsings(request.FileName, document, semanticModel);
+            }
+
+            return response;
+        }
+    }
+}
--- a/src/OmniSharp/Models/FixUsingsResponse.cs
+++ b/src/OmniSharp/Models/FixUsingsResponse.cs
@@ -1,0 +1,16 @@
+using System.Collections.Generic;
+
+namespace OmniSharp.Models
+{
+    public class FixUsingsResponse
+    {
+        public FixUsingsResponse(string buffer, IEnumerable<QuickFix> ambiguous)
+        {
+            Buffer = buffer;
+            AmbiguousResults = ambiguous;
+        }
+
+        public string Buffer { get; private set; }
+        public IEnumerable<QuickFix> AmbiguousResults { get; private set; }
+    }
+}
--- a/src/OmniSharp/Workers/Refactoring/FixUsings.cs
+++ b/src/OmniSharp/Workers/Refactoring/FixUsings.cs
@@ -1,0 +1,179 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.FindSymbols;
+using OmniSharp.Models;
+
+namespace OmniSharp
+{
+        //TODO:
+        //1. Remove unused using statements
+        //2. Sort using statements
+        //3. Remove redundant using statements
+        //4. Replace code which checks accessibility by Roslyn method if it is made available
+    public class FixUsingsWorker
+    {
+        public static async Task<FixUsingsResponse> AddMissingUsings(string fileName, Document document, SemanticModel semanticModel)
+        {
+            var compilationUnitSyntax = await AddUnambiguousUsings(fileName, document, semanticModel);
+            document = document.WithSyntaxRoot(compilationUnitSyntax);
+            semanticModel = await document.GetSemanticModelAsync();
+            return await GetAmbiguousUsings(fileName, document, semanticModel, compilationUnitSyntax);
+        }
+
+        private static async Task<CompilationUnitSyntax> AddUnambiguousUsings(string fileName, Document document, SemanticModel semanticModel)
+        {
+            var root = (await document.GetSyntaxTreeAsync()).GetRoot();
+            var unresolvedTypes = root.DescendantNodes()
+                .OfType<SimpleNameSyntax>()
+                .Where(x => semanticModel.GetSymbolInfo(x).Symbol == null &&
+                        semanticModel.GetSymbolInfo(x).CandidateReason != CandidateReason.OverloadResolutionFailure);
+            var compilationUnitSyntax = (CompilationUnitSyntax)root;
+            var usingsToAdd = new HashSet<string>();
+            var unresolvedSet = new HashSet<string>();
+            var candidateUsings = new List<ISymbol>();
+
+            if (compilationUnitSyntax != null)
+            {
+                usingsToAdd.UnionWith(GetUsings(root));
+
+                foreach (var unresolvedType in unresolvedTypes)
+                {
+                    candidateUsings = GetCandidateUsings(document, unresolvedType);
+                    //If there is only one candidate - add it to the usings list
+                    if (candidateUsings.Count() == 1)
+                    {
+                        var candidateNameSpace = candidateUsings[0].ContainingNamespace.ToString();
+                        var candidateName = SyntaxFactory.ParseName(candidateNameSpace);
+                        var usingToAdd = SyntaxFactory.UsingDirective(candidateName).NormalizeWhitespace()
+                            .WithTrailingTrivia(SyntaxFactory.Whitespace(Environment.NewLine));
+                        var usingToAddText = usingToAdd.GetText().ToString();
+
+                        if (!usingsToAdd.Contains(usingToAddText.Trim()))
+                        {
+                            compilationUnitSyntax = compilationUnitSyntax.AddUsings(usingToAdd);
+                            usingsToAdd.Add(usingToAddText.Trim());
+                        }
+                    }
+                }
+            }
+
+            return compilationUnitSyntax;
+        }
+
+
+        private static async Task<FixUsingsResponse> GetAmbiguousUsings(string fileName, Document document, SemanticModel semanticModel, CompilationUnitSyntax compilationUnitSyntax)
+        {
+            var root = (await document.GetSyntaxTreeAsync()).GetRoot();
+            var unresolvedTypes = root.DescendantNodes()
+                .OfType<SimpleNameSyntax>()
+                .Where(x => semanticModel.GetSymbolInfo(x).Symbol == null &&
+                        semanticModel.GetSymbolInfo(x).CandidateReason != CandidateReason.OverloadResolutionFailure);
+            var usingsToAdd = new HashSet<string>();
+            var unresolvedSet = new HashSet<string>();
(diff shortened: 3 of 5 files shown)


Yantrio:
can we switch the bracing from k+r? apart from that, it looks good in theory to me.

mat-mcloughlin:
Is it possible to make the response to match the old server so that we don't break backwards compatibility? 

https://github.com/OmniSharp/omnisharp-server/blob/master/OmniSharp/CodeActions/FixUsingsResponse.cs

src/OmniSharp/Api/Refactoring/OmnisharpController.FixUsings.cs
```diff
@@ -0,0 +1,127 @@
...
+            var compilationUnitSyntax = (CompilationUnitSyntax)(root);
+            var usingsToAdd = new HashSet<string>();
+            response.UnresolvedDict = new Dictionary<string, List<string>>();
+
+            if (compilationUnitSyntax != null) {
+                foreach (var unresolvedType in unresolvedTypes) {
+                    var candidateUsings = SymbolFinder.FindDeclarationsAsync(document.Project, 
+                            unresolvedType.Identifier.ValueText, false, SymbolFilter.All, CancellationToken.None).Result;
+
+                    var isAmbiguousSymbol = (candidateUsings.Count() > 1);
```
Yantrio: this isn't needed, can be done inline

src/OmniSharp/Api/Refactoring/OmnisharpController.FixUsings.cs
```diff
@@ -0,0 +1,127 @@
...
+                    foreach (var candidateUsing in candidateUsings) {
+                        var candidateNameSpace = GetFullNameSpace(candidateUsing, ignoreName: true);
+                        var candidateName = SyntaxFactory.ParseName(candidateNameSpace);
+                        var usingToAdd = SyntaxFactory.UsingDirective(candidateName).NormalizeWhitespace()
+                            .WithTrailingTrivia(SyntaxFactory.Whitespace("\n"));
+                        var usingToAddText = usingToAdd.GetText().ToString();
+                        if (!isAmbiguousSymbol) {
```
Yantrio: candidateUsings.Any() could do better here

src/OmniSharp/Api/Refactoring/OmnisharpController.FixUsings.cs
```diff
@@ -0,0 +1,127 @@
...
+using Microsoft.CodeAnalysis.Simplification;
+using OmniSharp.Models;
+
+namespace OmniSharp
+{
+    public partial class OmnisharpController
+    {
+        //TODO: 
+        //1. Remove unused using statements
+        //2. Sort using statements
```
Yantrio: does this todo list need updating/removing?

src/OmniSharp/Api/Refactoring/OmnisharpController.FixUsings.cs
nosami: Don't think so... needs to check for > 1, not > 0

src/OmniSharp/Api/Refactoring/OmnisharpController.FixUsings.cs
```diff
@@ -0,0 +1,127 @@
...
+                            }
+                        }
+                    }
+                }
+            }
+
+            return compilationUnitSyntax;
+        }
+
+        private async Task<CompilationUnitSyntax> RemoveRedundantUsings(CompilationUnitSyntax compilationUnitSyntax,
```
nosami: This is just removing dupes?

tests/OmniSharp.Tests/FixUsingsFacts.cs
```diff
@@ -0,0 +1,188 @@
...
+
+        private async Task AssertUnresolvedReferences(string fileContents, Dictionary<string, List<string>> expectedUnresolvedDict) 
+        {
+            var response = await RunFixUsings(fileContents);
+            foreach (var k in response.UnresolvedDict.Keys) {
+                var count = 0;
+                foreach (var x in response.UnresolvedDict[k]) {
+                    Assert.Contains(k, expectedUnresolvedDict.Keys);
+                    Assert.Equal(x, (expectedUnresolvedDict[k])[count]);
+                    //System.Console.WriteLine("Checked " + k + "->" + x);
```
nosami: delete

nosami:
Very happy to see this PR. 

It absolutely must use the [same response](https://github.com/OmniSharp/omnisharp-server/blob/master/OmniSharp/CodeActions/FixUsingsResponse.cs) as omnisharp v1.

Also... not sure there's much benefit from having remove usings if it's only removing dupes. We should probably delete that code and address that in a separate PR.

src/OmniSharp/Api/Refactoring/OmnisharpController.FixUsings.cs
```diff
@@ -0,0 +1,127 @@
...
+            {
+                var semanticModel = await document.GetSemanticModelAsync();
+                var compilationUnitSyntax = await AddMissingUsings(document, semanticModel, response);
+                await RemoveRedundantUsings(compilationUnitSyntax, response);
+            }
+
+            return response;
+        }
+
+        private async Task<CompilationUnitSyntax> AddMissingUsings(Document document,
```
davidfowl: Logic like this is usually moved into the https://github.com/OmniSharp/omnisharp-roslyn/tree/master/src/OmniSharp/Workers folder. The controllers are a thin layer that expose an end point calling into those workers.

src/OmniSharp/Api/Refactoring/OmnisharpController.FixUsings.cs
```diff
@@ -0,0 +1,127 @@
...
+                await RemoveRedundantUsings(compilationUnitSyntax, response);
+            }
+
+            return response;
+        }
+
+        private async Task<CompilationUnitSyntax> AddMissingUsings(Document document, 
+                SemanticModel semanticModel, FixUsingsResponse response) 
+        {
+            var root = (await document.GetSyntaxTreeAsync()).GetRoot();
```
davidfowl: This code needs some comments

src/OmniSharp/Models/FixUsingsResponse.cs
```diff
@@ -0,0 +1,11 @@
+using System.Collections.Generic;
+
+namespace OmniSharp.Models
+{
+    public class FixUsingsResponse
+    {
+        public string FileContents { get; set; }
+        public Dictionary<string, List<string>> UnresolvedDict {get; set;}
```
davidfowl: Is that really the property name? Dict?

davidfowl:
This code needs to be formatted using the format function in omnisharp :smile:

nosami:
@sriramgd Are you planning on working on this some more? If not, I can pick it up where you left off.

sriramgd:
@nosami I am. Will get in the updated PR by tomorrow evening. (Sorry about the break, was on vacation).

src/OmniSharp/Api/Refactoring/OmnisharpController.FixUsings.cs
sriramgd: Done

src/OmniSharp/Api/Refactoring/OmnisharpController.FixUsings.cs
sriramgd: Yes, removed.

tests/OmniSharp.Tests/FixUsingsFacts.cs
sriramgd: Deleted.

src/OmniSharp/Api/Refactoring/OmnisharpController.FixUsings.cs
sriramgd: Moved.

src/OmniSharp/Models/FixUsingsResponse.cs
sriramgd: Removed and replaced with response matching v1 server.

sriramgd:
Updated, ready for second review. Seems the travis build fails due to a previous merge.

davidfowl:
Gonna fix the build in a sec

src/OmniSharp/Api/Refactoring/OmnisharpController.FixUsings.cs
```diff
@@ -0,0 +1,32 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.AspNet.Mvc;
+using OmniSharp.Models;
+
+namespace OmniSharp
+{
+    public partial class OmnisharpController
+    {
+        //TODO:
```
davidfowl: Move these todos, to the worker.

src/OmniSharp/Workers/Refactoring/FixUsings.cs
```diff
@@ -0,0 +1,113 @@
+/* using Microsoft.CodeAnalysis.CSharp;
```
davidfowl: Dead code?

src/OmniSharp/Workers/Refactoring/FixUsings.cs
```diff
@@ -0,0 +1,113 @@
...
+{
+    public class FixUsingsWorker
+    {
+        public static async Task<FixUsingsResponse> AddMissingUsings(string fileName, Document document, SemanticModel semanticModel)
+        {
+            var root = (await document.GetSyntaxTreeAsync()).GetRoot();
+            var unresolvedTypes = root.DescendantNodes()
+                .OfType<SimpleNameSyntax>()
+                .Where(x => semanticModel.GetSymbolInfo(x).Symbol == null);
+            var compilationUnitSyntax = (CompilationUnitSyntax)(root);
```
davidfowl: `(CompilationUnitSyntax)root;` Get rid of the extra parenthesis

nosami:
I just tried this out. It seems to almost be working, but there are quite a few issues with it I think.

Removing `using OmniSharp.Models;` from a controller and then running fix usings results in 2 usings being added back

```
using OmniSharp.Models;
using OmniSharp;
```
There are also lots of ambiguous messages when there should be none (see screenshot)
screen shot 2015-04-17 at 13 54 20

Lots of namespaces in the OmniSharp project don't get imported, e.g. Xunit. I tried this on lots of different classes. I've also seen it import namespaces that are already included.

sriramgd:
Yes, this implementation is broken. Too many false positives for the symbols. I had only tested against the unit tests which were quite trivial. Will make better tests and look at fixing this.

sriramgd:
I am blocked on this for now. The solution I have requires being able to update the Document with a CompilationUnitSyntax before doing the ambiguity check. However it seems like document.UpdateSyntaxRoot is not available in Roslyn now (looks like it used to be a while back). I have asked a question about this on SO. In the meantime, anyone else who knows better or has other ideas could take this forward.

sriramgd:
Actually I did get a response on SO. Was missing something obvious. Will update PR in sometime.

src/OmniSharp/Workers/Refactoring/FixUsings.cs
sriramgd: Done

src/OmniSharp/Workers/Refactoring/FixUsings.cs
sriramgd: Removed

src/OmniSharp/Api/Refactoring/OmnisharpController.FixUsings.cs
sriramgd: Moved

sriramgd:
I have updated the implementation. This should do a better job at
1. Not including existing usings
2. Not flagging non-ambiguous symbols as ambiguous.

If this is accepted, we should replace any code which can be replaced by relevant APIs which Roslyn exposes in the future.

nosami:
Just had a chance to try this again. It's definitely better, but there are still some bugs.

```cs
public class MyClass
{
    public MyClass()
    {
        var l = new List<string>();
    }
}
```

This code errored when trying to import usings.


Also..  does this code work with linq syntax? i.e. `from s in list where.....`. I couldn't test because of the previous error.

sriramgd:
Sorry, had been busy for a while.
Seems like my implementation does not work well. Please close my PR in case anyone else wants to take this on.

nosami:
Closing in favour of #291

PR #154, opened 2015-04-13
updated xunit commands for new runner

nosami:
This switches xunit arguments to be compatible with the new aspnet.xunit.runner.

`--test` becomes `-method`

I've also added a fixture `-class` command that we had in v1.

If you attempt to run a single test when the cursor is at the class level, then the fixture is ran instead.

### DIFF
--- a/src/OmniSharp/AspNet5/AspNet5TestCommandProvider.cs
+++ b/src/OmniSharp/AspNet5/AspNet5TestCommandProvider.cs
@@ -18,7 +18,7 @@ namespace OmniSharp.AspNet5
         {
             _context = context;
             var aspNet5Paths = new AspNet5Paths(env, options, loggerFactory);
-            _dnx = aspNet5Paths.Dnx ?? aspNet5Paths.K;
+            _dnx = aspNet5Paths.Dnx + " ." ?? aspNet5Paths.K;
         }
 
         public string GetTestCommand(TestContext testContext)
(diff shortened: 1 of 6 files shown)


src/OmniSharp/AspNet5/AspNet5TestCommandProvider.cs
```diff
@@ -28,35 +28,41 @@ public string GetTestCommand(TestContext testContext)
...
+                    }
+                    else if (symbol is INamedTypeSymbol)
+                    {
+                        arguments = " -class " + containingNamespace +
+                            symbol.Name;
+                    }
                     break;
             }
-            return testCommand;
+            return "k test" + arguments;
```
davidfowl: k isn't right anymore, use the same logic restore has

src/OmniSharp/AspNet5/AspNet5TestCommandProvider.cs
```diff
@@ -28,35 +38,41 @@ public string GetTestCommand(TestContext testContext)
...
+                    }
+                    else if (symbol is INamedTypeSymbol)
+                    {
+                        arguments = " -class " + containingNamespace +
+                            symbol.Name;
+                    }
                     break;
             }
-            return testCommand;
+            return _dnx + " test" + arguments;
```
davidfowl: Did you test this with dnx?

nosami: No. Will do tomorrow. Looks like `k test` needs to be `dnx . test` now?

nosami:
This seems to be working for me now with dnx projects. I tested it with the KRuntime solution and was able to successfully run single tests, fixtures or assemblies.

The command it generates looks like this

`/Users/jason/.dnx/runtimes/dnx-mono.1.0.0-beta5-11427/bin/dnx . test -method Microsoft.Framework.Runtime.Tests.ServiceProviderFacts.NonRegisteredServicesCanBeIEnumerableResolved`

nosami:
OK.... I broke a bunch of tests. I need to come back to this when I have some more time.

PR #155, opened 2015-04-14
Unity

nosami:
Adds support for loading Unity solution files.

### DIFF
--- a/src/OmniSharp/MSBuild/MSBuildProjectSystem.cs
+++ b/src/OmniSharp/MSBuild/MSBuildProjectSystem.cs
@@ -130,7 +130,7 @@ namespace OmniSharp.MSBuild
 
         }
 
-        private Guid UnityTypeGuid(string projectName)
+        public static Guid UnityTypeGuid(string projectName)
         {
             var md5 = MD5.Create();
             var bytes = Encoding.UTF8.GetBytes(projectName);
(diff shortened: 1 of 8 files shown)


src/OmniSharp/MSBuild/MSBuildProjectSystem.cs
```diff
@@ -124,6 +130,22 @@ public void Initalize()
 
         }
 
+        public static Guid UnityTypeGuid(string projectName)
+        {
+            var md5 = MD5.Create();
```
jrieken: add ```using```-block?

src/OmniSharp/MSBuild/SolutionFile/SolutionFile.cs
```diff
@@ -17,7 +17,7 @@ internal sealed partial class SolutionFile
         private readonly IEnumerable<ProjectBlock> projectBlocks;
         private readonly IEnumerable<SectionBlock> globalSectionBlocks;
 
-        public SolutionFile(
+        internal SolutionFile(
```
davidfowl: ?

src/OmniSharp/MSBuild/SolutionPicker.cs
```diff
@@ -0,0 +1,40 @@
...
+                        return new SolutionPickerResult(unitySolution);
+                    }
+                    return new SolutionPickerResult(null, "Could not determine solution file");
+                default:
+                    return new SolutionPickerResult(null, "Could not determine solution file");
+            }
+        }
+    }
+
+    public class SolutionPickerResult
```
davidfowl: Move this to another file

src/OmniSharp/MSBuild/SolutionFile/SolutionFile.cs
nosami: I had a compilation error without that. Inconsisistent accessiblility because of SectionBlock and ProjectBlock.

PR #157, opened 2015-04-14
use the full XML markup instead simple value

### LINKED ISSUE #152: XML comments often off
Our XML documentation provider misses some XML child elements. See the screen shot below

image
 
and how the value misses the actual name of the class

jrieken:
fixes #152 by sending the whole xml doc comment to client and not only their text node descentents

### DIFF
--- a/src/OmniSharp/Roslyn/XmlDocumentationProvider.cs
+++ b/src/OmniSharp/Roslyn/XmlDocumentationProvider.cs
@@ -26,7 +26,7 @@ namespace OmniSharp.Roslyn
                     {
                         if (e.Attribute("name") != null)
                         {
-                            comments[e.Attribute("name").Value] = e.Value;
+                            comments[e.Attribute("name").Value] = e.ToString();
                         }
                     }
                 }


nosami:
We shouldn't be returning raw xml.

davidfowl:
Yes, I'm not sure what the client is supposed to do with this response.

jrieken:
I am open for suggestions. Today's problem is the comments are totally broken. For instance if the mark up reads like this:

```xml
        <member name="M:TestNamespace.TestClass.GetZero">
            <summary>
            The GetZero method.
            </summary>
            <example> 
            This sample shows how to call the <see cref="M:TestNamespace.TestClass.GetZero"/> method.
            <code>
            class TestClass 
            {
                static int Main() 
                {
                    return GetZero();
                }
            }
            </code>
            </example>
        </member>
```
We produce this string: ```The GetZero method.This sample shows how to call the method.class TestClass { static int Main(){return GetZero();}}```. It gets worst when you have a method with parameters ```<param name="foo">Int32</param>``` or exceptions. For samples are here: https://msdn.microsoft.com/en-us/library/cc837134.aspx. 

So unless we come up with a way to generate meaning full plain text from those XML comments I opt for sending XML to the client and let them handle it.

nosami:
https://github.com/OmniSharp/omnisharp-server/blob/d40bbb7a60a1e3274a124e991bcb6586371ee8c8/OmniSharp/Documentation/DocumentationConverter.cs ?

nosami:
The other editors currently expect plain text, so this would be a breaking change.

If you need xml for formatting purposes, then we need to parameterise it.

jrieken:
closing because @nosami has something in the works

PR #159, opened 2015-04-15
forward ProjectInformation events

jrieken:
When project state in Roslyn changes we already emit an event. This PR makes sure we also emit an event when project information outside roslyn (like commands in ASP projects) changes.


nosami:
LGTM

jrieken:
thanks

src/OmniSharp/AspNet5/AspNet5ProjectSystem.cs
```diff
@@ -105,6 +109,11 @@ public void Initalize()
                         project.Commands = val.Commands;
                         project.ProjectSearchPaths = val.ProjectSearchPaths;
 
+                        this._emitter.Emit(EventTypes.ProjectChanged, new ProjectInformationResponse()
```
davidfowl: this.? nooooo remove the this

jrieken: ops... not easy switching between TypeScript and C# all day long ;-)

PR #162, opened 2015-04-15
Plain text docs

nosami:
OK. This might work. Haven't tested it for reals yet ;)

### DIFF
--- a/src/OmniSharp/Services/DocumentationConverter.cs
+++ b/src/OmniSharp/Services/DocumentationConverter.cs
@@ -47,7 +47,7 @@ namespace OmniSharp.Documentation
                                     break;
                                 case "exception":
                                     ret.Append(lineEnding);
-                                    ret.Append(GetCref(xml["cref"]));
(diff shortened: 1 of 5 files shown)


src/OmniSharp/Services/DocumentationConverter.cs
```diff
@@ -0,0 +1,129 @@
...
+                                    ret.Append(lineEnding);
+                                    break;
+                                case "example":
+                                    ret.Append(lineEnding);
+                                    ret.Append("Example:");
+                                    ret.Append(lineEnding);
+                                    break;
+                                case "exception":
+                                    ret.Append(lineEnding);
+                                    ret.Append(GetCref(xml["cref"]));
```
jrieken: @nosami - we should trim the result from ```GetCref``` because otherwise it reads ```Exception : some text, some text``` (whitesapce before ```:```)

jrieken: apart from that it looks good. tested with library code and my own code :+1:

jrieken:
:+1:

PR #171, opened 2015-04-20
add logging for MSBuild project system

jrieken:
adds log forwards of warnings and errors for the MSBuild project system and emit project diagnostics messages to editors


nosami:
Looks good to me.

PR #172, opened 2015-04-20
send message when no or not the correct runtime is around

jrieken:


src/OmniSharp/AspNet5/AspNet5Paths.cs
```diff
@@ -39,9 +45,19 @@ public class AspNet5Paths
...
+        public string GlobalJson
+        {
+            get
+            {
+                var globalJson = Path.Combine(RootDirectory, "global.json");
+                return File.Exists(globalJson) ? globalJson : null;
+            }
+        }
+
         private string GetRuntimePath()
```
davidfowl: I would make this return a result and emit from the outside.

nosami:
LGTM

PR #173, opened 2015-04-21
send PackageRestore[Started|Finished] messages

jrieken:


davidfowl:
LGTM waiting until CI passes

PR #177, opened 2015-04-21
add a watchdog that kills stuck dnu|kpm restore processes

jrieken:
On mono you often and on windows you sometimes see dnu restore (or kpm restore) hanging forever. This PR adds a watchdog that kills the restore command if no progress was reported for >20 seconds. Then it will try one more time.

### DIFF
--- a/src/OmniSharp/AspNet5/PackagesRestoreTool.cs
+++ b/src/OmniSharp/AspNet5/PackagesRestoreTool.cs
@@ -48,56 +48,91 @@ namespace OmniSharp.AspNet5
 
                 lock (projectLock)
                 {
+                    var exitCode = -1;
                     _emitter.Emit(EventTypes.PackageRestoreStarted, new PackageRestoreMessage() { FileName = project.Path });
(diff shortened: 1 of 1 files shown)


src/OmniSharp/AspNet5/PackagesRestoreTool.cs
```diff
@@ -48,56 +48,91 @@ public void Run(Project project)
...
+            };
+            restoreProcess.ErrorDataReceived += (sender, e) =>
+            {
+                _logger.WriteError(e.Data);
+                lastSignal = DateTime.UtcNow;
+            };
+
+            restoreProcess.BeginOutputReadLine();
+            restoreProcess.BeginErrorReadLine();
+            restoreProcess.WaitForExit();
```
davidfowl: Why not just WaitForExit with a timeout?

jrieken: And how long should I wait? that depends on the packages I would say.

Issue #174, opened 2015-04-21
PathTooLongException


egamma:
I'm getting a path too long exception on an ASP.vnext sample project. What puzzles me is that I could work with this project before.

```
[INFORMATION:OmniSharp.AspNet5.AspNet5Paths] Looking for sdk version in 'c:\Users\egamma\Projects\tries\PartsUnlimited\global.json'.
[INFORMATION:OmniSharp.AspNet5.AspNet5Paths] Using runtime 'C:\Users\egamma\.dnx\runtimes\dnx-clr-win-x86.1.0.0-beta4-11566'.
[INFORMATION:OmniSharp.AspNet5.AspNet5ProjectSystem] Scanning 'c:\Users\egamma\Projects\tries\PartsUnlimited' for ASP.NET 5 projects
System.IO.PathTooLongException: The specified path, file name, or both are too long. The fully qualified file name must be less than 260 characters, and the directory name must be less than 248 characters.
   at System.IO.Path.NormalizePath(String path, Boolean fullCheck, Int32 maxPathLength, Boolean expandShortPaths)
   at System.IO.Path.GetDirectoryName(String path)
   at System.IO.FileSystemEnumerableIterator`1..ctor(String path, String originalUserPath, String searchPattern, SearchOption searchOption, SearchResultHandler`1 resultHandler, Boolean checkHost)
   at System.IO.DirectoryInfo.EnumerateFileSystemInfos(String searchPattern, SearchOption searchOption)
   at Microsoft.Framework.FileSystemGlobbing.Abstractions.DirectoryInfoWrapper.<EnumerateFileSystemInfos>d__3.MoveNext()
   at System.Collections.Generic.List`1.InsertRange(Int32 index, IEnumerable`1 collection)
   at Microsoft.Framework.FileSystemGlobbing.Internal.MatcherContext.Match(DirectoryInfoBase directory, String parentRelativePath)
   at Microsoft.Framework.FileSystemGlobbing.Internal.MatcherContext.Match(DirectoryInfoBase directory, String parentRelativePath)
   at Microsoft.Framework.FileSystemGlobbing.Internal.MatcherContext.Match(DirectoryInfoBase directory, String parentRelativePath)
   at Microsoft.Framework.FileSystemGlobbing.Internal.MatcherContext.Match(DirectoryInfoBase directory, String parentRelativePath)
[...]

PR #181, opened 2015-04-22
don't use matcher when project pattern is **/project.json

jrieken:
This PR is a pragmatic fix for #174 as it doesn't use the Matcher when the project pattern is ```**/project.json``` which is the default. If you author that flag you are supposed to know what you are doing but the default behaviour should not crash OmniSharp


davidfowl:
This is a good stop gap fix until we figure out a better solution

PR #182, opened 2015-04-23
add flag to enable package restore, add timeout for package restore

jrieken:


src/OmniSharp/config.json
```diff
@@ -1,7 +1,9 @@
 ﻿{
     "aspnet5": {
         "alias": "default",
-        "projects": "**/project.json"
+        "projects": "**/project.json",
+        "enablePackageRestore": false,
```
davidfowl: True by default and let the command line options override it

PR #183, opened 2015-04-23
don't emit error when there no project.json files, hint to installation ...

jrieken:
...path if someone really want to DNX

### DIFF
--- a/src/OmniSharp/AspNet5/AspNet5Paths.cs
+++ b/src/OmniSharp/AspNet5/AspNet5Paths.cs
@@ -17,8 +17,6 @@ namespace OmniSharp.AspNet5
         private readonly IOmnisharpEnvironment _env;
         private readonly OmniSharpOptions _options;
         private readonly ILogger _logger;
-        private readonly IEventEmitter _emitter;
-        public string RuntimePath { get; private set; }
         public string Dnx { get; private set; }
         public string Dnu { get; private set; }
         public string Klr { get; private set; }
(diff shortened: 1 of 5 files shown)


src/OmniSharp/AspNet5/AspNet5Paths.cs
```diff
@@ -27,23 +26,21 @@ public class AspNet5Paths
...
             _logger = loggerFactory.Create<AspNet5Paths>();
-            _emitter = emitter;
 
             RuntimePath = GetRuntimePath();
-            Dnx = FirstPath(RuntimePath, "dnx", "dnx.exe");
-            Dnu = FirstPath(RuntimePath, "dnu", "dnu.cmd");
-            Klr = FirstPath(RuntimePath, "klr", "klr.exe");
-            Kpm = FirstPath(RuntimePath, "kpm", "kpm.cmd");
-            K   = FirstPath(RuntimePath, "k", "k.cmd");
+            Dnx = FirstPath(RuntimePath.Value, "dnx", "dnx.exe");
```
davidfowl: Can't this be null?

jrieken: RuntimePath cannot, RuntimePath.Value can

davidfowl: Ah we check for null

MrTomWhite:
Do we need to leave this branch open? It wouldn't hurt to delete this and a few other old feature branches.

Issue #29, opened 2014-12-18
Endpoint list


nosami:
## v1 endpoints

- [ ] /addreference
- [ ] /addtoproject
- [x] /autocomplete
- [ ] /buildcommand
- [ ] /buildtarget
- [x] /checkalivestatus
- [x] /checkreadystatus
- [x] /codecheck
- [x] /codeformat
- [x] /currentfilemembersasflat
- [x] /currentfilemembersastree
- [ ] /currentfiletopleveltypes
- [x] /findimplementations
- [x] /findsymbols

~~- [ ] /findtypes~~
- [x] /findusages
- [ ] /fixcodeissue
- [ ] /fixusings
- [ ] /getcodeactions

~~- [ ] /getcodeissues~~
- [ ] /getoverridetargets
- [x] /gotodefinition
- [x] /gotofile
- [x] /gotoregion
- [ ] /lookupalltypes
- [ ] /reloadsolution
- [ ] /removefromproject
- [x] /rename
- [x] /runcodeaction
- [ ] /runoverridetarget

~~- [ ] /semanticerrors~~
- [x] /stopserver

~~- [ ] /syntaxerrors~~
- [x] /typelookup
- [x] /updatebuffer

## new endpoints

- [x] /changebuffer
- [x] /formatRange
- [x] /formatAfterKeystroke
- [x] /signatureHelp

CoreyKaylor:
Noticing when trying to use omnisharp-emacs with omnisharp-roslyn that there are errors for for /semanticerrors and saw it crossed out here. Is there a plan or workaround for this beyond not using flycheck w/roslyn?

nosami:
`/semanticerrors', '/syntaxerrors' and '/codeissues' have been superseded by `/codecheck` which performs all 3 actions.

For now, you should probably manually patch it... see here
https://github.com/nosami/omnisharp-demo/blob/master/elpa/omnisharp-20141219.950/omnisharp.el#L980

and https://github.com/nosami/omnisharp-demo/blob/master/elpa/omnisharp-20141219.950/omnisharp.el#L128-L133

CoreyKaylor:
Much better, thanks for the pointer.

Issue #16, opened 2014-12-14
/stopserver


nosami:
stops the server

cubitouch:
I can take that, Application.Exit() is sufficient or you need something more elaborate ?

nosami:
Think you need to do this
https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp/Program.cs#L118

cubitouch:
Thanks, I'm on it :)

nosami:
👍

filipw:
i bet kestrel will throw exceptions here, mark my words :smile:

cubitouch:
Should it return a response ?

davidfowl:
Don't call application, exit. Use the IApplicationShutdown interface.

cubitouch:
System.NullReferenceException:
![image](https://cloud.githubusercontent.com/assets/7486151/6218985/71264080-b624-11e4-953e-a252e1f9b4ce.png)

@davidfowl do you have an idea ? Throwing this when I run a "_applicationShutdown.RequestShutdown();"

cubitouch:
I will move on something else in the next few days, if someone need to get what I did the other day, in case the kestrel exception is fixed -> https://gist.github.com/cubitouch/99e5d61573c74789b498

sriramgd:
cubitouch's code stops the server okay when I tested by sending the /stopserver command from omnisharp-vim via python on ubuntu and windows. The only change I had to make was to use HttpPost since we use that now in omnisharp-vim.
Am I missing something?
cubitouch, would it be possible to submit the PR? If not I can submit it.

cubitouch:
@sriramgd Please create the PR, glad my gist could help ;)

nosami:
Oh! That's good news. The bug in Kestrel must be fixed now. 

@sriramgd All requests should be POSTs

david-driscoll:
Stop server landed with #302

PR #190, opened 2015-04-27
Added /stopserver. Kestrel throws exception

sgwill:
Probably WIP because of Kestrel bug.

Add gist @cubitouch, updated to POST instead of GET from #16.

This does shut down the DTH, but Kestrel throws an exception:

```
System.NullReferenceException was unhandled
Message: An unhandled exception of type 'System.NullReferenceException' occurred in Microsoft.AspNet.Server.Kestrel.dll
Additional information: Object reference not set to an instance of an object.
```

Update #29 when merged...

### DIFF
--- a/src/OmniSharp/Api/Server/OmnisharpController.cs
+++ b/src/OmniSharp/Api/Server/OmnisharpController.cs
@@ -3,18 +3,18 @@ using Microsoft.Framework.Runtime;
 
 namespace OmniSharp.Api.Server
 {
-	public partial class OmnisharpController : Controller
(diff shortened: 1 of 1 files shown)


nosami:
You are correct, we were waiting for an upstream fix in Kestrel.

However, this shouldn't affect us at all when using the stdio interface.

nosami:
Can anyone tell me if this works with stdio? If so, I say we get it merged in.

willl:
As #302 was merged recently, we should probably close this PR.

PR #191, opened 2015-04-28
Added support for watching project.lock.json files

### LINKED ISSUE #189: Send RefreshDependencies message when project.lock.json changes
This allows restore be be run outside of the editor

davidfowl:
- When project.lock.json files change, refresh dependencies
in the DTH. This will allow a workflow where package restore
can happen outside of the editor but the editor can still reflect
those changes that were made.
- Made some fixes where files weren't being watched for project references that were outside of the initially scanned projects.
Fixes #189


jrieken:
:+1:

PR #194, opened 2015-04-29
Feature highlighting

Alxandr:
Adds a highlighting endpoint.

@Yantrio was bugging me about PRs and all that -.-


src/OmniSharp/Api/Highlighting/OmnisharpController.Highlighting.cs
```diff
@@ -0,0 +1,68 @@
...
+            }
+
+            var walker = new HighlightSyntaxWalker(model);
+            foreach (var node in nodes)
+            {
+                walker.Visit(node);
+            }
+
+            var regions = walker.Regions;
+            if (request.Lines != null && request.Lines.Length != 0)
```
davidfowl: Do we need to check this twice?

Alxandr: If you mean the `request.Lines` check, it could be stored in a `bool` I just don't think that'd buy you much.

davidfowl: DRY?

src/OmniSharp/Models/HighlightResponse.cs
```diff
@@ -0,0 +1,10 @@
+namespace OmniSharp.Models
+{
+    public class HighlightResponse
+    {
+        public int Line { get; set; }
+        public int Start { get; set; }
+        public int End { get; set; }
+        public string Kind { get; set; }
```
davidfowl: Should this just be an enum?

src/OmniSharp/Workers/Highlighting/HighlightSyntaxWalker.cs
```diff
@@ -0,0 +1,193 @@
...
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Recommendations;
+using Microsoft.CodeAnalysis.Text;
+using OmniSharp.Models;
+
+namespace OmniSharp
+{
+    public class HighlightSyntaxWalker : CSharpSyntaxWalker
+    {
+        readonly SemanticModel _model;
```
davidfowl: private

src/OmniSharp/Models/HighlightResponse.cs
Alxandr: It definitely *can* be an enum. Though, the way it works today is more like CSS classes, it space-separates multiple "kinds" into a string. The original implementation actually just used css classes. Not to mention that I just `ToString()` the kind I get from Roslyn, so I don't know all of the kinds one might get.

davidfowl:
We should use this

http://source.roslyn.io/#Microsoft.CodeAnalysis.Workspaces/Classification/Classifier.cs

Alxandr:
I like this :-) 
Will investigate.

Alxandr:
Wow, @davidfowl, that API was really awesome. I now have highlighting inside interpolated strings!

tests/OmniSharp.Tests/HighlightFacts.cs
```diff
@@ -0,0 +1,130 @@
...
+            var workspace = TestHelpers.CreateSimpleWorkspace(new Dictionary<string, string>
+            {
+                { "a.cs", "namespace N1\n{\nclass C1 { int n = true; }\n}" }
+            });
+
+            var controller = new OmnisharpController(workspace, null);
+            var regions = await controller.Highlight(new HighlightRequest() { FileName = "a.cs", Lines = new[] { 2 } });
+            
+            ValidateRegions(regions,
+                new Region("keyword", 2, 0, 5),
```
nosami: these tests are incomprehensible

nosami: i am not a computer.

davidfowl: hhahahahahahahahah

src/OmniSharp/Models/HighlightRequest.cs
```diff
@@ -0,0 +1,15 @@
+namespace OmniSharp.Models
+{
+    public class HighlightRequest : Request
+    {
+        /// <summary>
+        ///   Spesifies which lines to highlight.
```
davidfowl: Specifies

src/OmniSharp/Api/Highlighting/OmnisharpController.Highlighting.cs
```diff
@@ -0,0 +1,36 @@
...
+using OmniSharp.Models;
+
+namespace OmniSharp
+{
+    public partial class OmnisharpController
+    {
+        [HttpPost("highlight")]
+        public async Task<IEnumerable<HighlightResponse>> Highlight(HighlightRequest request)
+        {
+            var document = _workspace.GetDocument(request.FileName);
```
davidfowl: Doesn't look this like handles documents in multiple projects.

tests/OmniSharp.Tests/HighlightFacts.cs
Alxandr: I thought I cleaned them up fairly nicely.

How would you rather have them?

src/OmniSharp/Api/Highlighting/OmnisharpController.Highlighting.cs
Alxandr: Apparently my Roslyn/OmniSharp knowledge is missing here. What do you mean handle documents in multiple projects? Is it like if you have one file that's included in multiple projects?

nosami: When a project contains multiple frameworks, then we have one document per framework. `_workspace.GetDocuments` will return them for you.

tests/OmniSharp.Tests/HighlightFacts.cs
nosami: Something like `AssertSyntax("namespace", "keyword")` ?

Yantrio: what if he named the variables so you can see what is going on?
```
ValidateRegions(regions,
               new Region("keyword", line: 2, startCh: 0,  endCh: 5)...
```

Alxandr: If you add something simple like that it's impossible to test contextual keywords, string interpolation (which by the way I already test), difference between interface names and class names etc.

src/OmniSharp/Api/Highlighting/OmnisharpController.Highlighting.cs
jrieken: @Alxandr With DNX you can define different projects (check project.json), basically for core_clr and desktop_clr. Internally, Omnisharp maps that to different Roslyn projects. Therefore you can have multiple documents for the same filepath. So depending on the containing project of a document a line of code might be actually doing something or not. VS15 has a nice representation for this:

image

nosami:
Looks like this might actually work :)

Yantrio:
Is that atom @nosami?

nosami:
@Yantrio yeah!

AndersNS:
Very nice. :o

jrieken:
It still uses ```getDocument```. Can we make it handle multiple documents as discussed earlier and then merge?

Alxandr:
@jrieken we discuessed handling multiple documents, but came to no conclusion as to how it should work.

davidfowl:
Start with the dropdown list?

nosami:
Using a dropdown list is the simplest solution. We just need to specify the target framework(project) to the request and choose the matching document.

Alxandr:
We would also need an api (if one doesn't exist already) to get the list of the frameworks to select from then.

src/OmniSharp/Models/HighlightRequest.cs
```diff
@@ -0,0 +1,15 @@
...
+        ///   Specifies which lines to highlight.
+        ///   If none are given, highlight the entire
+        ///   file.
+        /// </summary>
+        /// <remarks>
+        ///   This is 0 indexed.
+        /// </remarks>
+        public int[] Lines { get; set; }
+    }
+}
```
jrieken: Should not be zero indexed since everything else in OmniSharp is 1-offset based

jrieken:
wrt to multiple document I see two solutions: 

* return a response object collection per framework
* include the framework in the ```HighlightResponse``` object

my personal preference would be the 2nd since that doesn't make editors merge the result by hand it is also similar to other cases where we add the framework to data, see: https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp/Models/FileMemberElement.cs#L14

jrieken:
@Alxandr - the list of frameworks is part of the data ```/projects``` returns.

nosami:
3rd option, include a filter on the request - if the filter isn't specified return highlighting per framework.

I expect these responses would get pretty big. We need symbols for the whole document on initialisation.

jrieken:
Let combine option 2 and 3: The request allows to filter by project/framework and the highlight objects contain a list of containing projects/frameworks

Alxandr:
I agree. This seems like a good strategy. I have examns this friday, so I'll get it done during the weekend probably :).

jrieken:
great

Alxandr:
Sorry it was a bit later than expected, but apparently we had a trip to Sweeden planned O_o.

Anyways, did a rebase of master too.

.gitignore
```diff
@@ -26,4 +26,5 @@ nuget.exe
 *.ipch
 *.sln.ide
 debugSettings.json
-buildlog
\ No newline at end of file
+buildlog
+project.lock.json
```
jrieken: unsure about that... @davidfowl

src/OmniSharp/Api/Highlighting/OmnisharpController.Highlighting.cs
```diff
@@ -0,0 +1,47 @@
...
+using Microsoft.AspNet.Mvc;
+using Microsoft.CodeAnalysis.Classification;
+using Microsoft.CodeAnalysis.Text;
+using OmniSharp.Models;
+
+namespace OmniSharp
+{
+    public partial class OmnisharpController
+    {
+        [HttpPost("highlight")]
```
jrieken: should it be called ```highlights```?

src/OmniSharp/Models/HighlightResponse.cs
```diff
@@ -0,0 +1,24 @@
+using Microsoft.CodeAnalysis.Classification;
+using Microsoft.CodeAnalysis.Text;
+
+namespace OmniSharp.Models
+{
+    public class HighlightResponse
+    {
+        public LinePosition Start { get; set; }
+        public LinePosition End { get; set; }
```
jrieken: I like using positions (Line,Column) but for the sake of consistency we should stick with ```StartLine```, ```StartColumn```, ```EndLine```, and ```EndColumn``` - it's used like that in other places.

nosami:
Closing as this was superceded by #251

PR #198, opened 2015-04-30
Act like VS

davidfowl:
Allow msbuild based projects to set the tools version (msbuild tools verson) and the visual studio version.


nosami:
LGTM

PR #199, opened 2015-04-30
Add support for language services for scriptcs

filipw:
This introduces a 3rd project system - `ScriptCsProjectSystem`. It's ifdef-ed since it will not work on Core.

It finds CSX files in the directory and uses scriptcs assemblies to retrieve all the necessary information about the runtime - what usings are in the context, what DLLs are references and all other scriptcs-specific things. 

It then creates scripting projects in the workspace. After that, all Omnisharp endpoints work normally against the incoming requests as if you were working with any other type of project.

Example:

screenshot 2015-04-30 12 21 08

Note: I currently suppress `CS1024` error for scriptcs projects - this is to prevent treating `#load` directive as error (which is valid in scriptcs). Roslyn will introduce it later too, so this is temporary.

### DIFF
--- a/src/OmniSharp/project.json
+++ b/src/OmniSharp/project.json
@@ -25,8 +25,8 @@
                "ScriptCs.Hosting.KpmTemp": "0.14.1"
             },
             "frameworkAssemblies": {
-				"System.Globalization": "",
-				"System.Reflection":  "",
+                "System.Globalization": "",
+                "System.Reflection":  "",
                 "Microsoft.Build": "",
                 "Microsoft.Build.Engine": "",
                 "Microsoft.Build.Framework": ""
--- a/src/OmniSharp/project.json
+++ b/src/OmniSharp/project.json
@@ -22,7 +22,7 @@
         "aspnet50": {
             "dependencies": {
                "ICSharpCode.NRefactory": "6.0.0-alpha3",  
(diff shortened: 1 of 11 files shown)


filipw:
after this comes in, the editors will need to catch up to treat CSX files as csharp. This has already been done for brackets https://github.com/OmniSharp/omnisharp-brackets/blob/73a2ac98d6793e43442737988498e29a4ab67c50/main.js#L65

adamralph:
woot!

Yantrio:
LGTM :+1: if @davidfowl / @nosami want to sign off on the formatting im happy for it to be merged.

src/OmniSharp/project.json
```diff
@@ -21,10 +21,12 @@
     "frameworks": {
         "aspnet50": {
             "dependencies": {
-               "ICSharpCode.NRefactory": "6.0.0-alpha3"  
+               "ICSharpCode.NRefactory": "6.0.0-alpha3",  
+               "ScriptCs.Hosting.KpmTemp": "0.14.1"
             },
             "frameworkAssemblies": {
-                "System.Globalization": "",
+				"System.Globalization": "",
```
nosami: indent

nosami:
Cheers @filipw :dancers:

PR #201, opened 2015-04-30
Support LangVersion property on csproj

monoman:
Builds OK, but new functionality is untested as installing Visual Studio 5 RC messed things a little bit...

### DIFF
--- a/src/OmniSharp/MSBuild/ProjectFile/ProjectFileInfo.cs
+++ b/src/OmniSharp/MSBuild/ProjectFile/ProjectFileInfo.cs
@@ -4,6 +4,7 @@ using System.IO;
 using System.Linq;
 using System.Runtime.Versioning;
 using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.Framework.Logging;
 
 #if ASPNET50
@@ -27,6 +28,8 @@ namespace OmniSharp.MSBuild.ProjectFile
         public string ProjectFilePath { get; private set; }
 
         public FrameworkName TargetFramework { get; private set; }
+
+        public LanguageVersion? SpecifiedLanguageVersion { get; private set; }
 
         public string ProjectDirectory
         {
@@ -52,7 +55,7 @@ namespace OmniSharp.MSBuild.ProjectFile
         {
             var projectFileInfo = new ProjectFileInfo();
             projectFileInfo.ProjectFilePath = projectFilePath;
-            
+
 #if ASPNET50
             if (!PlatformHelper.IsMono)
             {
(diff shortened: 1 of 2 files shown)


monoman:
travis-ci is failing because of a bug in Mono that timeouts sometimes in dependencies downloading...

src/OmniSharp/MSBuild/ProjectFile/ProjectFileInfo.cs
```diff
@@ -175,5 +180,24 @@ public static ProjectFileInfo Create(MSBuildOptions options, ILogger logger, str
...
+                {
+                    case "ISO-1": return LanguageVersion.CSharp1;
+                    case "ISO-2": return LanguageVersion.CSharp2;
+                    case "3": return LanguageVersion.CSharp3;
+                    case "4": return LanguageVersion.CSharp4;
+                    case "5": return LanguageVersion.CSharp5;
+                    case "6": return LanguageVersion.CSharp6;
+                }
+            }
+            return LanguageVersion.CSharp5;
```
davidfowl: Why aren't we defaulting to C# 6? Is this the regular msbuild behavior? Have you looked at the MSbuildWorkspace in the roslyn code base?

monoman: Just guessed that for most existing projects C# 5 is the default. Only
VS2015 may default to C# 6. The right thing probably would be to see which
Framework is being targeted but I didn't want to make it more complex.

Another alternative would be to have a Nullable<LanguageVersion> property
and let Roslyn choose the default if not specified in the csproj, by not
setting parsing options as was the code before.

Rafael Teixeira
O..:.)oooo

nosami: That sounds like the right thing to do to me. Just don't specify LanguageVersion at all if it's not specified by the csproj file.

davidfowl: Yep Make it return `LanguageVersion?`

monoman: Updated PR. It builds and so pass tests, but I'll try to test on VSCode
proper, too.

Rafael Teixeira
O..:.)oooo

monoman:
Wrong rebase

monoman:
Redone as single commit

monoman:
Finally was able to test my changes, manually doing what #PR203 does to be able to compile with Beta4 and debug on VS2015RC, it works fine. 

On Beta3 things still bork, probably due to version mismatches on dependencies, but that is the state of master itself, irrespective of this PR.

So I hope this can get merged. @davidfowl @nosami

monoman:
Just to be clear my branch named 'debug' has this same patch but targets beta4, it compiles flawlessly with beta3 here and beta4 (and roslyn RC2) there, but here I can't make it to run properly, there it runs but something about beta4 breaks communication over stdio so vscode starts omnisharp ok (which parses my sln/csproj based project with the LangVersion property nicely) but then says it could not communicate with omnisharp, seems like beta4 changed some things that affect how stdio would redirect requests/responses. Also there some dependency for the test projects is inappropriately versioned so the build script fails when trying to run the tests, but jumping over it to the dnu publish step the generated build starts fine and again, only fails to communicate with vscode, but works over http which I test on the browser with http://localhost:2000/projects.

ianbattersby:
Just to say I stole @monoman's branch for my own ends and made the following changes for *nix build: https://github.com/ianbattersby/omnisharp-roslyn/compare/8cdd89afa3c94ba31628e0b44616090b793efd3a...a1f507ebf690a4ad605898a77dd3bf60d1a70e88

nosami:
Looks good to me. Thanks @monoman

PR #204, opened 2015-05-03
Fixed typelookup to return the response object

david-driscoll:
This will fix the reflection tool used to gen typescript information.


src/OmniSharp.Stdio/project.lock.json
```diff
@@ -0,0 +1,2282 @@
+{
```
davidfowl: Lets leave this out of source control for now.

davidfowl:
I think this needs a do over

david-driscoll:
All I gotta say is don't `drink and git` :grin:

PR #203, opened 2015-05-01
Let's upgrade to dnx!

Yantrio:
DNX

PR #205, opened 2015-05-03
Set warningsAsErrors to true and fixed warnings

davidfowl:


tests/OmniSharp.Tests/BufferManagerFacts.cs
```diff
@@ -10,7 +9,7 @@ namespace OmniSharp.Tests
     public class BufferManagerFacts
     {
         [Fact]
-        public async Task UpdateBufferIgnoresVoidRequests()
+        public void UpdateBufferIgnoresVoidRequests()
```
nosami: Returning `void` here instead of `async Task` was causing XUnit to hang indefinitely when ran on mono. Did that ever get fixed?

nosami: (even with `-parallel none`

davidfowl: It's fixed in the latest. I guess we can wait until we get beta4 done.

Yantrio: it works for me in beta4 using the latest mono CI build, just gotta wait for a nice new shiny mono release.

monoman:
Well on the 'debug' branch of my fork I have this and #201 and something like #203, and with the new xunit.runner.dnx, so build.cmd now works including tests, all running on beta4 and being debuggable inside VS2015RC. Of course in Windows, in Mono I'm not sure how it goes.

Hope you enjoy

PR #206, opened 2015-05-03
added comment about CS1024

filipw:
Added comment from PR #199 to the source code.

The `#load` directive, valid for script files (mentioned here https://github.com/dotnet/roslyn/issues/122) is currently not recognized by Roslyn so [CS1024](https://msdn.microsoft.com/en-us/library/ww43bwhw%28v=vs.90%29.aspx) need to be suppressed


nosami:
cheers @filipw!

PR #211, opened 2015-05-07
Added gotofile endpoint

filipw:
Returns a list of all files in the current workspace - same as the functionality in omnisharp-server.


nosami:
Nice!

Issue #175, opened 2015-04-21
rename creates duplicate changes


jrieken:
when rename is performed on a public symbol which exists in two frameworks/projects you get changes for each document which logically is the same file.

mat-mcloughlin:
Is this just as simple as running a distinct on the response or is it more complicated?

PR #216, opened 2015-05-12
deduplicating renames

jrieken:
fix for #175

### DIFF
--- a/src/OmniSharp/Models/LinePositionSpanTextChange.cs
+++ b/src/OmniSharp/Models/LinePositionSpanTextChange.cs
@@ -76,10 +76,14 @@ namespace OmniSharp.Models
                 && EndLine == other.EndLine
                 && EndColumn == other.EndColumn;
         }
-        
+
         public override int GetHashCode()
         {
-            return NewText.GetHashCode() + 23 * StartLine + 23 * StartColumn + 23 * EndLine + 23 * EndColumn;
+            return NewText.GetHashCode()
+                * (23 + StartLine)
+                * (29 + StartColumn)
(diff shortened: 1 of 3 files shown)


src/OmniSharp/Models/LinePositionSpanTextChange.cs
```diff
@@ -61,5 +61,26 @@ public class LinePositionSpanTextChange
...
+            return NewText == other.NewText
+                && StartLine == other.StartLine
+                && StartColumn == other.StartColumn
+                && EndLine == other.EndLine
+                && EndColumn == other.EndColumn;
+        }
+        
+        public override int GetHashCode()
+        {
+            return NewText.GetHashCode() + 23 * StartLine + 23 * StartColumn + 23 * EndLine + 23 * EndColumn;
```
nosami: This isn't a good hashcode function. Easy to create dupes.

nosami: Line 1, Column 10 or Line 10, Column 1 ?

jrieken:
@nosami - pushed a better hash function. btw - is there a way to tell travis not to use mono 4.0.1

nosami:
Looks much better. I'll test it out later.

You can do pretty much anything you want on Travis. See here https://github.com/OmniSharp/omnisharp-server/blob/master/.travis.yml#L1-L7. It's even possible to run build against a matrix of mono versions. 
The only restriction that I'm aware of is that you can't use Windows ;)

nosami:
Looks good!

PR #218, opened 2015-05-13
Tell travis to use mono 3.12.0

jrieken:


nosami:
Build has failed 3 times but will merge anyway. It has to be better than 4.01 ;)

Issue #214, opened 2015-05-12
CSX crashes OmniSharp when dependent file not found


jrieken:
Have code like below but don't have the referenced file. OmniSharp crashed hard.

```
#load "model.csx"
Console.WriteLine("Hello World");
var i = 99;
Console.WriteLine(i * 8);
```

error:

```
System.IO.FileNotFoundException: Could not find file 'c:\Users\jrieken\AppData\Local\Temp\_monacodata_9888\workspaces\ConsoleApp%20-%20Copy\model.csx'.
File name: 'c:\Users\jrieken\AppData\Local\Temp\_monacodata_9888\workspaces\ConsoleApp%20-%20Copy\model.csx'
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.FileStream.Init(String path, FileMode mode, FileAccess access, Int32 rights, Boolean useRights, FileShare share, Int32 bufferSize, FileOptions options, SECURITY_ATTRIBUTES secAttrs, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
   at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
   at System.IO.StreamReader..ctor(String path, Encoding encoding, Boolean detectEncodingFromByteOrderMarks, Int32 bufferSize, Boolean checkHost)
   at System.IO.StreamReader..ctor(String path, Encoding encoding)
   at System.IO.File.InternalReadAllLines(String path, Encoding encoding)
   at System.IO.File.ReadAllLines(String path)
   at ScriptCs.FileSystem.ReadFileLines(String path)
   at ScriptCs.FilePreProcessor.ParseFile(String path, FileParserContext context)
   at ScriptCs.LoadLineProcessor.ProcessLine(IFileParser parser, FileParserContext context, String line)
   at ScriptCs.Contracts.DirectiveLineProcessor.ProcessLine(IFileParser parser, FileParserContext context, String line, Boolean isBeforeCode)
   at ScriptCs.FilePreProcessor.<>c__DisplayClassd.<ParseScript>b__a(ILineProcessor x)
   at System.Linq.Enumerable.Any[TSource](IEnumerable`1 source, Func`2 predicate)
[...]

PR #219, opened 2015-05-14
Added exception handling to project initialization

filipw:
- wrapped the call to `Initialize()` on project system collection in try/catch since any unhandled exception that spills from project systems (file IO but also other) will crash the server. now at least other project systems have a chance to take over
 
 - before, the only CSX exception handled was `InvalidDirectiveUseException`, but really any exception should be caught, and the processing should move to the next CSX file


davidfowl:
LGTM

filipw:
Thanks, this fixes #214

PR #223, opened 2015-05-16
Do not report namespace for non-regular C# syntax

### LINKED ISSUE #222: Type lookup should not report namespace for CSX files
The scripting semantics define that CSX files don't have namespaces - namespace is always null for all the types defined in Interactive/Script code.

Omnisharp does not recognize this and report `.TypeName` instead of just `TypeName`

csx-typelookup

filipw:
For `SourceCodeKind.Interactive` and `SourceCodeKind.Script`, do not report a namespace as part of type lookup.

Fixes #222.

Also added a helper for creating a CSX workspace which we can use for more tests in the future.


nosami:
Looks good. Cheers!

PR #224, opened 2015-05-18
Trick to make Roslyn keep the indentation for format on enter

jrieken:
when having this
```
1:........
2:........// this is a comment
3:........<cursor>
```

we don't want the last line to be removed. OmniSharp will tell Roslyn to format only line 2 but since all is trivia lines 1 to 3 will be formatted - that includes removing the indentation for line 3. The ideal would be to ask Roslyn after the formatting operation what the proper indentation for line #3 is but that API is internal. So, this PR does a simple trick by adding ```;``` at the position in which format on enter was triggered. The outcome will be:

```
1:
2:........// this is a comment
3:........<cursor>
```

### DIFF
--- a/src/OmniSharp/Api/Formatting/OmnisharpController.Format.cs
+++ b/src/OmniSharp/Api/Formatting/OmnisharpController.Format.cs
@@ -40,7 +40,7 @@ namespace OmniSharp
             var lines = (await document.GetSyntaxTreeAsync()).GetText().Lines;
             var start = lines.GetPosition(new LinePosition(request.Line - 1, request.Column - 1));
             var end = lines.GetPosition(new LinePosition(request.EndLine - 1, request.EndColumn - 1));
-            var changes = await Formatting.GetFormattingChangesForRange(_workspace, Options, document, start, end);
+            var changes = await Formatting.GetFormattingChangesForRange(Options, document, start, end);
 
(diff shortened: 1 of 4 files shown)


jrieken:
@nosami - I will add tests and then it read for review

nosami:
Does this trick work for expanding {<enter>} ?

jrieken:
no, it will only format the line you have just completed. that means it put the ```{``` on the next line but doesn't do anything for ```}```. Your scenario sound interesting. Let me repeat:

```
void M(){}
```

press enter so that it becomes

```
void M(){
<cursor>}
```

and format on enter should make it into

```
void M()
{
....<cursor>
}

Sound reasonable. I could check if the current node is a block and format that - instead of the line. Let me think about it...

jrieken:
This is hard without knowing the indent at position. Ideally we would have this flow: (1) format on enter -> make format changes, (2) for the new line ask for the proper indentation - make indent change. Until then formal on newline character will be flaky... We can either wait for https://github.com/dotnet/roslyn/issues/3079 and go with a our known -most likely incorrect- implementation of format on enter (like ```number_of_containing_blocks * indentSize```)

nosami:
@jrieken What's the status of this?

jrieken:
it has issues.... I will close and tackle again when I got smarter :smile:

PR #235, opened 2015-06-07
DNX runtime algorithm fixes:

wasker:
1. Both %HOME% and %USERPROFILE% are searched.
2. Environment variables are expanded to accomodate default value of
DNX_HOME.

### DIFF
--- a/src/OmniSharp/AspNet5/AspNet5Paths.cs
+++ b/src/OmniSharp/AspNet5/AspNet5Paths.cs
@@ -50,7 +50,8 @@ namespace OmniSharp.AspNet5
 
             foreach (var location in GetRuntimeLocations())
             {
-                var paths = GetRuntimePathsFromVersionOrAlias(versionOrAlias, location);
+                //  Need to expand variables, because DNX_HOME variable might include %USERPROFILE%.
+                var paths = GetRuntimePathsFromVersionOrAlias(versionOrAlias, Environment.ExpandEnvironmentVariables(location));
 
                 foreach (var path in paths)
                 {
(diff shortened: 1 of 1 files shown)


nosami:
```cs
var home = Environment.GetEnvironmentVariable("HOME") ?? Environment.GetEnvironmentVariable("USERPROFILE");
```

Could this be fixed by switching the order of these two variables? It seems very unlikely that a linux user would have $USERPROFILE set to anything.

wasker:
In my case both are set, on Windows system and both point to different places, so the algorithm completely fails to find DNX. The root cause though is that DNX_HOME uses environment variable in the value and the algorithm doesn't expand the variables.

wasker:
Seems that travis-ci has issues. Is this known?

nosami:
Yeah.. I just restarted it.

wasker:
I wonder on how to redo the checks on this PR again? All should be green now and ready to merge.

nosami:
Looks good to me. I'll kick Travis again. 

Want to see if @davidfowl has anything to say about it before merging. Thanks!

nosami:
@wasker do you know that if you are using vscode or atom, that you can make the editor point to your working version by pointing the OMNISHARP env var to one of [these](https://github.com/OmniSharp/omnisharp-roslyn/tree/master/scripts) files depending on platform?

wasker:
There're other tricks that are available to make Code/Atom work, but I believe it's better to make Omnisharp a bit more flexible in reading user's environment - like this patch does - rather than rely on workarounds.

nosami:
I wasn't suggesting that ;) I was merely suggesting a way for you to be able to use your code while you wait :)

wasker:
Oh, thanks, I'm good. :)

nosami:
I just took another look at this. Looks pretty safe. Thanks!

nosami:
Should be incorporated into Atom in the next hour or so.

nosami:
in omnisharp-atom v0.12.5

wasker:
Great, thanks a lot!

PR #238, opened 2015-06-15
Magic

nosami:
This PR is to add VS refactorings to omnisharp.

Please note! This isn't ready to be merged yet.


nosami:
I spoke to @DustinCampbell just now about us redistributing the Roslyn Features dll. He was "deeply unconcerned" about this as it is on a completely separate feed :)

DustinCampbell:
LOL. It should be a long term plan, but you shouldn't be blocked by this either. Here's the way I see this happening:

1. We put up the Roslyn nightly Features NuGets on MyGet. This should actually happen very soon. The change has gone through a couple of iterations in code review to get the shape of the packages right and is almost there.
2. We ship official Features NuGets at VS 2015 RTM. Those should match the versions that DNX snaps to at the same time.
3. DNX needs to be updated to not pin a specific version of Roslyn. That will make it easier for us to continue to evolve feature-level APIs and for OmniSharp to pick up nightlies.

nosami:
Sounds like a plan :+1:

shanselman:
LOL :shipit:

On Tue, Jun 16, 2015 at 3:37 PM, Jason Imison <notifications@github.com>
wrote:

> Sounds like a plan [image: :+1:]
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/OmniSharp/omnisharp-roslyn/pull/238#issuecomment-112594165>
> .
>



-- 
Scott Hanselman
Donate to fight diabetes: http://hnsl.mn/fightdiabetes

src/OmniSharp/Models/RunCodeActionResponse.cs
```diff
@@ -5,6 +5,6 @@ namespace OmniSharp.Models
     public class RunCodeActionResponse
     {
         public string Text { get; set; }
-        public IEnumerable<LinePositionSpanTextChange> Changes { get; set; }
+        public IEnumerable<ModifiedFileResponse> Changes { get; set; }
```
nosami: @jrieken I changed the public signature here to allow for Roslyn fixes that generate new files. Is this ok with you?

jrieken: opting for a /v2/runcodeaction route where we make this change and also use string identfiers instead of index-based identifiers

PR #246, opened 2015-06-21
Handle unauthorized access exceptions when searching for projects

### LINKED ISSUE #192: Omnisharp tries (and crashes) to find project.json in folders it has no permission to access
When Omnisharp is trying to find the project.json it looks at various folders. On OSX, it will crash with the following exception:
```
System.UnauthorizedAccessException: Access to the path "/.DocumentRevisions-V100" is denied.
at System.IO.Directory+c__Iterator0.MoveNext () [0x00000] in :0 
at System.IO.Directory+c__Iterator0.MoveNext () [0x00000] in :0 
at OmniSharp.AspNet5.AspNet5ProjectSystem.ScanForProjects () [0x00000] in :0 
at OmniSharp.AspNet5.AspNet5ProjectSystem.Initalize () [0x00000] in :0 
at OmniSharp.Startup.Configure (IApplicationBuilder app, ILoggerFactory loggerFactory, IOmnisharpEnvironment env, ISharedTextWriter writer) [0x00000] in :0 
at (wrapper managed-to-native) System.Reflection.MonoMethod:InternalInvoke (System.Reflection.MonoMethod,object,object[],System.Exception&)
at System.Reflection.MonoMethod.Invoke (System.Object obj, BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x00000] in :0
```

Bug was found in atom, only happens if there is no project.json at the root level and Omnisharp instead goes searching for one.

Should just skip the folders it doesn't have permission to access.

Andorbal:
This is an attempt to fix #192 by ignoring (and logging) any paths that are unauthorized while searching for project.json files.  I added a check for paths that are too long as well.  I've tested this on Windows and Linux with unauthorized paths and without.

### DIFF
--- a/src/OmniSharp/Utilities/DirectoryEnumerator.cs
+++ b/src/OmniSharp/Utilities/DirectoryEnumerator.cs
@@ -25,11 +25,11 @@ namespace OmniSharp.Utilities
             while (directoryStack.Any())
             {
                 var current = directoryStack.Pop();
-                
+
                 try
                 {
                     allFiles = allFiles.Concat(GetFiles(current, pattern));
-                    
+
                     foreach (var subdirectory in GetSubdirectories(current))
                     {
                         directoryStack.Push(subdirectory);
@@ -48,7 +48,7 @@ namespace OmniSharp.Utilities
         {
             try
             {
-                return Directory.EnumerateFiles(path, pattern, SearchOption.TopDirectoryOnly);
+                return Directory.GetFiles(path, pattern, SearchOption.TopDirectoryOnly);
             }
             catch (PathTooLongException)
             {
--- a/src/OmniSharp/Utilities/DirectoryEnumerator.cs
+++ b/src/OmniSharp/Utilities/DirectoryEnumerator.cs
@@ -15,41 +15,58 @@ namespace OmniSharp.Utilities
             _logger = loggerFactory.CreateLogger<DirectoryEnumerator>();
         }
 
-        public IEnumerable<string> SafeEnumerateFiles(string path, string searchPattern)
+        public IEnumerable<string> SafeEnumerateFiles(string target, string pattern = "*.*")
         {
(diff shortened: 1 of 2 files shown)


nosami:
I'd worry about getting a StackOverflowException if the directory being searched was sufficiently large and deeply nested.

Something like this https://gist.github.com/jaywick/a7a730009d57ca67b506 looks safer to me.

src/OmniSharp/Utilities/DirectoryEnumerator.cs
```diff
@@ -0,0 +1,52 @@
...
+            while (directoryStack.Any())
+            {
+                var current = directoryStack.Pop();
+
+                try
+                {
+                    var files = Directory.GetFiles(current, pattern);
+                    allFiles = allFiles.Concat(files);
+
+                    foreach (var subdirectory in Directory.EnumerateDirectories(current))
```
davidfowl: This will throw with path to long exceptions if current.Length > MAX_PATH.

```
Directory.EnumerateFiles(_env.Path, "project.json", SearchOption.AllDirectories)
```

That actually works better in those scenarios.

nosami: `Directory.EnumerateFiles(_env.Path, "project.json", SearchOption.AllDirectories)` is the code we are trying to replace because there's no way to catch UnauthorizedAccessException without discarding all results

davidfowl: I get it, I'm just calling out things that will fail when you make this code change

nosami: This is handled by `catch (PathTooLongException)` though?

Andorbal:
I verified this works:

image

nosami:
nice! thanks :)

PR #247, opened 2015-06-23
created a v2 version of code actions

nosami:
As requested by @jrieken, I created a v2 endpoint for refactorings.


nosami:
This supercedes #238

nosami:
OK. I think this is good to merge now. 

I intend to work some more on this, but this is already much better than what we have currently.

In a future PR, I will work on integrating with MEF better (so less reflection) and having a plugin folder for things like code-cracker and http://vsrefactoringessentials.com/

PR #251, opened 2015-06-29
Feature highlighting

david-driscoll:
This is based off of #194 as I was getting things working in omnisharp-atom.


src/OmniSharp/Models/v1/HighlightRequest.cs
```diff
@@ -0,0 +1,89 @@
...
+        //    If none are given, highlight for all the projects.
+        /// </summary>
+        public string[] ProjectNames { get; set; }
+        /// <suimmary>
+        ///   Request comment classifications
+        /// </summary>
+        /// <remarks>
+        /// Defaults to true
+        /// </remarks>
+        public bool WantComments { get; set; } = true;
```
nosami: An array of token kinds would be better for this.

src/OmniSharp/Models/v1/HighlightResponse.cs
```diff
@@ -0,0 +1,95 @@
...
+                EndLine = linePos.End.Line + 1,
+                // Caution, Character is 1-based.
+                StartColumn = linePos.Start.Character,
+                EndColumn = linePos.End.Character,
+                Kind = span.ClassificationType,
+                Projects = projects
+            };
+        }
+
+        public int CompareTo(HighlightResponse other)
```
nosami: return other.StartLine.CompareTo(StartLine) * other.StartColumn.CompareTo(StartColumn) ? not 100% but seem like it would do the same thing as this method

src/OmniSharp/Api/v1/Highlighting/OmnisharpController.Highlighting.cs
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        class ClassifiedResult
+        {
+            public ClassifiedSpan Span { get; set; }
+            public TextLineCollection Lines { get; set; }
+            public string Project { get; set; }
+        }
+
+        private IEnumerable<ClassifiedSpan> FilterSpans(HighlightRequest request, IEnumerable<ClassifiedSpan> spans)
```
nosami: This could could pretty much disappear if we put the classification types in the request.

src/OmniSharp/Models/v1/HighlightResponse.cs
```diff
@@ -0,0 +1,95 @@
...
+namespace OmniSharp.Models
+{
+    public class HighlightResponse : IComparable<HighlightResponse>
+    {
+        public int StartLine { get; set; }
+        public int StartColumn { get; set; }
+        public int EndLine { get; set; }
+        public int EndColumn { get; set; }
+        public string Kind { get; set; }
+        public IEnumerable<string> Projects { get; set; }
```
nosami: Projects? You're returning highlighting for an entire project(s)?

src/OmniSharp/Models/v1/HighlightRequest.cs
```diff
@@ -0,0 +1,89 @@
...
+    {
+        /// <summary>
+        ///   Specifies which lines to highlight.
+        ///   If none are given, highlight the entire
+        ///   file.
+        /// </summary>
+        public int[] Lines { get; set; }
+        /// <summary>
+        ///   Specifies which projects to highlight for.
+        //    If none are given, highlight for all the projects.
```
nosami: Is this not very slow?

src/OmniSharp/Models/v1/HighlightResponse.cs
nosami: How long does that take on the omnisharp-roslyn solution?

src/OmniSharp/Models/v1/HighlightRequest.cs
```diff
@@ -0,0 +1,89 @@
...
+        /// </summary>
+        public string[] ProjectNames { get; set; }
```
nosami: It seems wrong to not be able to do this per-document

nosami: Ignore this^^ I see you can format by project or document.

src/OmniSharp/Api/v1/Highlighting/OmnisharpController.Highlighting.cs
```diff
@@ -0,0 +1,97 @@
...
+using Microsoft.CodeAnalysis.Classification;
+using Microsoft.CodeAnalysis.Text;
+using OmniSharp.Models;
+
+namespace OmniSharp
+{
+    public partial class OmnisharpController
+    {
+        [HttpPost("highlight")]
+        public async Task<IEnumerable<HighlightResponse>> Highlight(HighlightRequest request)
```
nosami: I know we return IEnumerable from other controllers, but it's not good as it makes it impossible to extend the response in the future.

A response with a property of IEnumerable is better.

src/OmniSharp/Api/v1/Highlighting/OmnisharpController.Highlighting.cs
```diff
@@ -0,0 +1,97 @@
...
+                    {
+                        Span = span,
+                        Lines = text.Lines,
+                        Project = project
+                    }));
+            }
+
+            return results
+                .GroupBy(z => z.Span.TextSpan.ToString())
+                .Select(a => HighlightResponse.FromClassifiedSpan(a.First().Span, a.First().Lines, a.Select(z => z.Project)));
```
nosami: rename z and a

src/OmniSharp/Models/v1/HighlightResponse.cs
```diff
@@ -0,0 +1,95 @@
+using System;
+using System.Collections.Generic;
+using Microsoft.CodeAnalysis.Classification;
+using Microsoft.CodeAnalysis.Text;
+
+namespace OmniSharp.Models
+{
+    public class HighlightResponse : IComparable<HighlightResponse>
```
nosami: Is the IComparable interface needed? I can't see it being used.

src/OmniSharp/Models/v1/HighlightResponse.cs
```diff
@@ -0,0 +1,95 @@
...
+
+        internal static HighlightResponse FromClassifiedSpan(ClassifiedSpan span, TextLineCollection lines, IEnumerable<string> projects)
+        {
+            var linePos = lines.GetLinePositionSpan(span.TextSpan);
+
+            return new HighlightResponse
+            {
+                StartLine = linePos.Start.Line + 1,
+                EndLine = linePos.End.Line + 1,
+                // Caution, Character is 1-based.
```
nosami: Looks like it's zero based - https://github.com/dotnet/roslyn/blob/f306fa9c328e476738eaab67db560b135f22a416/src/Compilers/Core/Portable/Text/LinePosition.cs#L38

src/OmniSharp/Models/v1/HighlightRequest.cs
david-driscoll: What would that array be a type of? just string?  I thought about an array at first, but we only get strings back from classifier, hence I went with this request approach.

nosami: Yeah, I think so. Strong types aren't helping us at all here.

nosami: Can always use an enum to depict the string if it makes you more comfortable :)

src/OmniSharp/Models/v1/HighlightResponse.cs
david-driscoll: I've flipped around the code based on that.

src/OmniSharp/Api/v1/Highlighting/OmnisharpController.Highlighting.cs
david-driscoll: done

src/OmniSharp/Api/v1/Highlighting/OmnisharpController.Highlighting.cs
david-driscoll: done

nosami:
You didn't gain much by using an enum here. I forgot that C# enums can't have string values.

Maybe something like this would be better? http://stackoverflow.com/a/1343517/316874

david-driscoll:
@nosami what are you thinking, create a class, to handle the delegates?

nosami:
No.... I was thinking that this code

```cs
else if (classification == HighlightClassification.PreprocessorKeyword)
    spans = spans.Where(x => x.ClassificationType != "preprocessor keyword");
```

was a little annoying and could be fixed if the string representation for PreProcessorKeyword was the string "preprocessor keyword".

It's not a big deal though, so merging anyway. I'll publish it as 1.0.6

Issue #226, opened 2015-05-20
Support <DefineConstants> from *.csproj -file


jrieken:

jrieken:
related #202

nosami:
Looks like this can be closed as #252 should fix

PR #252, opened 2015-06-30
Read DefineConstants from MsBuild file

filipw:
While Omnisharp doesn't support msbuild configuration switches at the moment - and always defaults to `Debug` configuration, this PR at least introduces the capability of reading its `<DefineConstants>` into the `CSharpCompilationOptions` which would allow the language services to properly respect them.

Based on #226


nosami:
Cheers Filip! I can knock a couple of issues off now :)

PR #256, opened 2015-07-01
Added new parameter to allow excluding project files from response.

jtbm37:
Hi

This PR relates to https://github.com/OmniSharp/omnisharp-server/pull/202
We need to exclude the project files from the response as they are not used by the clients.

### DIFF
--- a/src/OmniSharp/Api/v1/ProjectSystem/OmnisharpController.ProjectSystem.cs
+++ b/src/OmniSharp/Api/v1/ProjectSystem/OmnisharpController.ProjectSystem.cs
@@ -1,6 +1,7 @@
 ﻿using Microsoft.AspNet.Mvc;
 using OmniSharp.Dnx;
 using OmniSharp.Models;
+using OmniSharp.Models.v1;
 using OmniSharp.MSBuild;
 using OmniSharp.ScriptCs;
 
@@ -24,11 +25,15 @@ namespace OmniSharp
 
         [HttpPost("/projects")]
         [HttpGet("/projects")]
-        public WorkspaceInformationResponse ProjectInformation()
+        public WorkspaceInformationResponse ProjectInformation(ProjectInformationRequest request)
         {
+            var includeSourceFiles = true;
+            if (request != null)
+                includeSourceFiles = request.IncludeSourceFiles;
+
             return new WorkspaceInformationResponse
             {
-                MSBuild = new MsBuildWorkspaceInformation(_msbuildContext),
+                MSBuild = new MsBuildWorkspaceInformation(_msbuildContext, includeSourceFiles),
                 Dnx = new DnxWorkspaceInformation(_dnxContext),
                 ScriptCs = _scriptCsContext
             };
--- a/src/OmniSharp/Api/v1/ProjectSystem/OmnisharpController.ProjectSystem.cs
+++ b/src/OmniSharp/Api/v1/ProjectSystem/OmnisharpController.ProjectSystem.cs
@@ -27,13 +27,9 @@ namespace OmniSharp
         [HttpGet("/projects")]
         public WorkspaceInformationResponse ProjectInformation(ProjectInformationRequest request)
         {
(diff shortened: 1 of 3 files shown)


src/OmniSharp/Api/v1/ProjectSystem/OmnisharpController.ProjectSystem.cs
```diff
@@ -24,11 +25,15 @@ public class ProjectSystemController
 
         [HttpPost("/projects")]
         [HttpGet("/projects")]
-        public WorkspaceInformationResponse ProjectInformation()
+        public WorkspaceInformationResponse ProjectInformation(ProjectInformationRequest request)
         {
+            var includeSourceFiles = true;
```
nosami: It would be better if this default value was set on the request

src/OmniSharp/Api/v1/ProjectSystem/OmnisharpController.ProjectSystem.cs
```diff
@@ -24,11 +25,15 @@ public class ProjectSystemController
...
+            if (request != null)
```
nosami: Don't need to test for the request being null

src/OmniSharp/Models/v1/MsBuildWorkspaceInformation.cs
```diff
@@ -6,11 +6,20 @@ namespace OmniSharp.Models
...
         {
             SolutionPath = msbuildContext.SolutionPath;
 
-            Projects = msbuildContext.Projects.Values.Select(p => new MSBuildProject(p));
+            Projects = msbuildContext
+                .Projects
+                .Values
+                .OrderBy(x => x.AssemblyName)
+                .Select(p => {
+                    var prj = new MSBuildProject(p);
```
nosami: rename prj to project

src/OmniSharp/Api/v1/ProjectSystem/OmnisharpController.ProjectSystem.cs
jtbm37: If clients are already making use of this action method and they do not send any parameters it will throw an NRE (NullReferenceException). Are you ok with that ?

src/OmniSharp/Api/v1/ProjectSystem/OmnisharpController.ProjectSystem.cs
jtbm37: I will rename it to `excludeSourceFiles` as well. I wanted to preserve the default behaviour which was to include the source files.

src/OmniSharp/Api/v1/ProjectSystem/OmnisharpController.ProjectSystem.cs
nosami: Yeah. We don't check anywhere else.

nosami:
Ah, sorry... I just realised that ProjectInformationRequest is a new parameter, so the request will need a null check. I'll add it back in.

src/OmniSharp/Api/v1/ProjectSystem/OmnisharpController.ProjectSystem.cs
```diff
@@ -24,11 +25,11 @@ public class ProjectSystemController
...
         [HttpPost("/projects")]
         [HttpGet("/projects")]
-        public WorkspaceInformationResponse ProjectInformation()
+        public WorkspaceInformationResponse ProjectInformation(ProjectInformationRequest request)
         {
             return new WorkspaceInformationResponse
             {
-                MSBuild = new MsBuildWorkspaceInformation(_msbuildContext),
+                MSBuild = new MsBuildWorkspaceInformation(_msbuildContext, request.ExcludeSourceFiles),
                 Dnx = new DnxWorkspaceInformation(_dnxContext),
```
jrieken: Why only for MSBuild projects?

PR #259, opened 2015-07-07
Add Dnx frameworks

david-driscoll:
This gives us a list of the frameworks the selected project is loaded for, as well as the `FriendlyName`  (for display) and short name for querying about specific projects (`project.Name + ShortName`).


src/OmniSharp/Models/v1/DnxProject.cs
```diff
@@ -12,6 +13,7 @@ public class DnxProject
         public IList<string> Configurations { get; set; }
         public IList<string> ProjectSearchPaths { get; set; }
         public IList<string> Frameworks { get; set; }
+        public IList<DnxFramework> DnxFrameworks { get; set; }
```
nosami: Not so sure about having both Frameworks and DnxFrameworks here. I think I'd be happier with just

```cs
public IList<DnxFramework> Frameworks { get; set; }
```

and ditching the old IList<string> - but that would mean breaking backwards compatibility /cc @jrieken

david-driscoll:
I did not change the existing Frameworks list, because it could be used elsewhere.

src/OmniSharp/Models/v1/DnxProject.cs
david-driscoll: Agreed, but I didn't want to break anyone intentionally.  I'm not sure if it's a highly used property yet for editors though.

nosami: It's used by @vscode which is why I cc'd @jrieken to see if he'd be ok with changing this.

src/OmniSharp/Api/v1/Highlighting/OmnisharpController.Highlighting.cs
```diff
@@ -48,7 +48,7 @@ public partial class OmnisharpController
                 }
                 else
                 {
-                    foreach (var line in request.Lines)
+                    foreach (var line in request.Lines.Where(z => z <= text.Lines.Count))
```
nosami: This predicate shouldn't be required.

david-driscoll: It's just a guard if the editor sends too many lines.  This was was something I saw in atom when copying and pasting an entire files contents.

nosami: The document should be updated with the content from the request before it gets to the controller, so I'm not sure how it could ever be different.

nosami: maybe ignore request.Lines completely and loop over the document lines instead.

david-driscoll: What about just stitching them together with `.Join`?

```
    var linesToClassify = request.Lines.Join(
        text.Lines,
        line => line,
        line => line.LineNumber,
        (requestLine, line) => line.Span);
    foreach (var lineSpan in linesToClassify)
    {
        foreach (var span in await Classifier.GetClassifiedSpansAsync(document, lineSpan))
        {
            spans.Add(span);
        }
    }
```

nosami:
I'm going to merge this as we need it. We can discuss removing the duplicate Frameworks property with @jrieken later.

jrieken:
+1

nosami:
Hey @jrieken, you're alive! :)

OK to do this then?

```diff
-         public IList<string> Frameworks { get; set; }
-        public IList<DnxFramework> DnxFrameworks { get; set; }
+        public IList<DnxFramework> Frameworks { get; set; }
```

jrieken:
Yeah, back from my ES6 excursion. I'll update OmniSharp in VSCode this week - before I leave for vacation ;-)

jrieken:
Change looks good to me!

nosami:
Cool.. I'll do that now then

PR #260, opened 2015-07-07
Beta5

nosami:
So... this seems to work - but - I had to to disable refactorings because I don't have a signed version of the Roslyn features assemblies for -rc3


tests/OmniSharp.Tests/TestHelpers.cs
```diff
@@ -169,9 +170,10 @@ public static OmnisharpWorkspace AddProjectToWorkspace(OmnisharpWorkspace worksp
             return workspace;
         }
 
-        private static Assembly AssemblyFromType(Type type)
+        private static MetadataReference MetadataReferenceFromType(Type type)
         {
-            return type.GetTypeInfo().Assembly;
+            var assembly = type.GetTypeInfo().Assembly;
+            return MetadataReference.CreateFromFile(assembly.Location);
```
nosami: assembly.Location only works on desktop CLR.

src/OmniSharp/Startup.cs
```diff
@@ -109,8 +108,8 @@ public static OmnisharpWorkspace CreateWorkspace()
         {
             var assemblies = MefHostServices.DefaultAssemblies;
 #if DNX451
-            assemblies = assemblies.AddRange(RoslynCodeActionProvider.MefAssemblies);
-            assemblies = assemblies.AddRange(NRefactoryCodeActionProvider.MefAssemblies);
+//             assemblies = assemblies.AddRange(RoslynCodeActionProvider.MefAssemblies);
+//             assemblies = assemblies.AddRange(NRefactoryCodeActionProvider.MefAssemblies);
```
nosami: Had to comment the refactoring code out. Need to wait until we can get a signed compatible version of the Roslyn features dlls for -rc3

src/OmniSharp/Services/MetadataFileReferenceCache.cs
```diff
@@ -26,18 +26,20 @@ public MetadataReference GetMetadataReference(string path)
...
 
-            var metadata = _cache.GetOrSet(cacheKey, ctx =>
-            {
+            var metadata = _cache.Get<AssemblyMetadata>(cacheKey);
+
+            if (metadata == null) {
                 _logger.LogVerbose(string.Format("Cache miss {0}", path));
 
-                ctx.AddExpirationTrigger(new FileWriteTimeTrigger(path));
+                // HOWTODO? ctx.AddExpirationTrigger(new FileWriteTimeTrigger(path));
```
nosami: Not sure how to handle cache expiration trigger.

src/OmniSharp/Settings/FromBodyApplicationModelConvention.cs
```diff
@@ -15,8 +15,8 @@ public void Apply(ApplicationModel application)
                     foreach (var parameter in action.Parameters)
                     {
                         if (parameter.BindingInfo?.BindingSource != null ||
-                            parameter.Attributes.OfType<IBindingSourceMetadata>().Any() ||
-                            ValueProviderResult.CanConvertFromString(parameter.ParameterInfo.ParameterType))
+                            parameter.Attributes.OfType<IBindingSourceMetadata>().Any())
+                            //|| ValueProviderResult.CanConvertFromString(parameter.ParameterInfo.ParameterType))
```
nosami: Need to figure out where this method went. Do we even need it?

src/OmniSharp/Services/MetadataFileReferenceCache.cs
```diff
@@ -26,18 +26,21 @@ public MetadataReference GetMetadataReference(string path)
...
 
-            var metadata = _cache.GetOrSet(cacheKey, ctx =>
+            var metadata = _cache.Get<AssemblyMetadata>(cacheKey);
+
+            if (metadata == null)
             {
                 _logger.LogVerbose(string.Format("Cache miss {0}", path));
 
-                ctx.AddExpirationTrigger(new FileWriteTimeTrigger(path));
+                // HOWTODO? ctx.AddExpirationTrigger(new FileWriteTimeTrigger(path));
```
davidfowl: cache.Set takes options.

src/OmniSharp/Settings/FromBodyApplicationModelConvention.cs
davidfowl: /cc @rynowak

src/OmniSharp/Utilities/AnalyzerAssemblyLoader.cs
```diff
@@ -0,0 +1,189 @@
...
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Reflection.Metadata;
+using System.Reflection.PortableExecutable;
+using Roslyn.Utilities;
+
+namespace Microsoft.CodeAnalysis
+{
+    public class AnalyzerAssemblyLoader : IAnalyzerAssemblyLoader
```
davidfowl: HMMMM

davidfowl: There must be a default one.

src/OmniSharp/Settings/FromBodyApplicationModelConvention.cs
filipw: it is now this private, you can just copy it straight up https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.WebApiCompatShim/Conventions/WebApiParameterConventionsApplicationModelConvention.cs#L32

filipw: the method is needed to not apply `[FromBody]` to string, int or any other string-convertible parameters - only complex types should come from body

davidfowl: There might be a better way to do this using application model instead of a filter.

nosami: Yeah... I meant specifically this part `ValueProviderResult.CanConvertFromString` - everything is a POST, so do we even need to check if it's a simple type?

src/OmniSharp/Utilities/AnalyzerAssemblyLoader.cs
nosami: Nope. I saw a super simple one in the Roslyn tests, but it didn't work for analyzers that are spread across multiple assemblies.

src/OmniSharp/Settings/FromBodyApplicationModelConvention.cs
filipw: yeah technically it's not needed because omnisharp doesn't use anything bound from route or querystring atm. 
But in the future if anyone would introduce something like this for example (for whatever reason), it would be failing:

```csharp
       [HttpPost("v{version:int}/findusages")]
       public async Task<QuickFixResponse> FindUsages(int version, FindUsagesRequest request)
```
But of course there's YOLO approach and you can just remove it for now and it will be brought back only if it's ever needed :)

rynowak: Looks like this is here to mimic WebAPI2's behavior? IMO it's best for APIs to be explicit. You could also go the route of `[FromBody]` by default and have anything else be explicit.

nosami:
[Roslyn switched to .net 4.6](https://github.com/dotnet/roslyn/commit/5d7308d9f515bcaafc69964bd6a56e356b9f84bb) some time back for the pieces deemed to be used only by VS. This means that we can't use the Features dll until mono supports .net 4.6 :(

nosami:
@DustinCampbell has created this ticket https://github.com/dotnet/roslyn/issues/3998 to make the Roslyn Features layer portable which should also unblock this.

borgdylan:
Mono will support .NET 4.6 once MSFT releases teh changes in reference source...something they are taking too long to do (just releasing code they already have). Having omnisharp on Beta 5 would be better as with beta 4 restore seems to hang for me which means I can only consume omnisharp in a "built by other people" form.

nosami:
@borgdylan try this `export MONO_THREADS_PER_CPU=2000` or use the build.sh script to do the restore. The build script temporarily switches to the latest dnu to restore.

borgdylan:
I managed to install the v 1.1 tarball into vscide. It's working OK.

nosami:
Seems like beta7 or later should be used when this branch is resumed. Would be nice to get RefactoringEssentials into the mix and drop NR6-alpha3

Just took a look at the DNX runtime to see which version of Roslyn it's using. Beta-7 is using a [specific version](https://github.com/aspnet/dnx/blob/1.0.0-beta7/src/Microsoft.Dnx.Runtime/project.json#L33) which will likely hinder adoption of Roslyn from the nightly feed.

nosami:
I'm unlikely to work on this in the foreseeable future - I don't write C# as part of day job any more, so would be awesome if someone could pick this up at some point.

david-driscoll:
I'm looking at picking it up after after plugins is finished, the nightly's
should now have the portable assemblies. I'm going to look at how we can
unpin roslyn, perhaps by generating our own dnx versions.

nosami:
@david-driscoll That's awesome news! Thinking about this a little more, it's probably possible to use the Features assembly from the nightlies with the 1.1-pinned version that DNX uses of CodeAnalysis

david-driscoll:
That's my hope, that it'll just work side by side, but if it won't then
I'll investigate how to setup travis to CI build us our own custom dnx,
that can.  Seeing as omnisharp hosts in own runtime, this shouldn't be a
big concern.

nosami:
There is a newer version of Refactoring Essentials available here https://github.com/icsharpcode/RefactoringEssentials/tree/latestroslyn that is built against the Roslyn nightlies.

troydai:
@nosami we're moving on to `dotnet` now. Can we close this PR?

troydai:
@nosami , I close this PR since it has been 6 month since the last comments. We already moved on to the `dotnet-cli` and the `master` branch is already on beta8.

PR #262, opened 2015-07-08
Replace IList<string> Frameworks with IList<DnxFramework> Frameworks,…

david-driscoll:
… remove DnxFrameworks

See some comments on #259 for more backstory.

cc @nosami @jrieken


nosami:
@david-driscoll I totally forgot that you did this :) I made the same change in #269

PR #267, opened 2015-07-17
Add Nuget package search and version end points

david-driscoll:
cc @nosami @davidfowl @jrieken


nosami:
Merged. If anyone has any issues with this code, it can be fixed later.

PR #271, opened 2015-07-21
Added package sources endpoint

david-driscoll:


src/OmniSharp/Api/v1/NuGet/OmnisharpController.Package.cs
```diff
@@ -51,6 +77,10 @@ public partial class OmnisharpController
                 var tasks = new List<Task<IEnumerable<SimpleSearchMetadata>>>();
                 var repositoryProvider = new OmniSharpSourceRepositoryProvider(projectPath);
                 var repos = repositoryProvider.GetRepositories().ToArray();
+                if (request.Sources.Any()) {
```
nosami: js brace ;)

Issue #272, opened 2015-07-22
Design Time Host on Windows not Starting


brad-jones:
Continuing on from the issue I posted here: https://github.com/OmniSharp/omnisharp-atom/issues/450

brad-jones:
Just cloned the master branch and built it.
When I run _omnisharp.cmd --stdio_, I get:

```json
{"Event":"log","Body":{"LogLevel":"INFORMATION","Name":"OmniSharp.Startup","Message":"Omnisharp server running using stdio at location 'C:\\Users\\Brad\\Documents\\Projects\\dotneteloquent' on host -1."},"Seq":1,"Type":"event"}
{"Event":"log","Body":{"LogLevel":"INFORMATION","Name":"OmniSharp.Dnx.DnxPaths","Message":"Looking for sdk version in 'C:\\Users\\Brad\\Documents\\Projects\\dotneteloquent\\global.json'."},"Seq":2,"Type":"event"}
{"Event":"log","Body":{"LogLevel":"INFORMATION","Name":"OmniSharp.Dnx.DnxPaths","Message":"Using runtime 'C:\\Users\\Brad\\.dnx\\runtimes\\dnx-clr-win-x64.1.0.0-beta5'."},"Seq":3,"Type":"event"}
{"Event":"log","Body":{"LogLevel":"INFORMATION","Name":"OmniSharp.Dnx.DnxProjectSystem","Message":"Scanning 'C:\\Users\\Brad\\Documents\\Projects\\dotneteloquent' for DNX projects"},"Seq":4,"Type":"event"}
{"Event":"log","Body":{"LogLevel":"INFORMATION","Name":"OmniSharp.Dnx.DnxProjectSystem","Message":"Found project 'C:\\Users\\Brad\\Documents\\Projects\\dotneteloquent\\tests\\DotNetEloquent.Tests\\project.json'."},"Seq":5,"Type":"event"}
{"Event":"log","Body":{"LogLevel":"INFORMATION","Name":"OmniSharp.Dnx.DnxProjectSystem","Message":"Found project 'C:\\Users\\Brad\\Documents\\Projects\\dotneteloquent\\src\\DotNetEloquent\\project.json'."},"Seq":6,"Type":"event"}
[...]

PR #274, opened 2015-07-22
Fixes OmniSharp/omnisharp-roslyn#272

brad-jones:
On my system I found that 4 seconds was not long enough for DHT to start.
Time for a new computer perhaps...

Anyway I have bumped the timeout to 10 seconds (FYI: my system on average took
6 seconds to start DHT).

I have also added in a check to see if the timeout has indeed been reached.


nosami:
Looks good to me. Thanks!

PR #280, opened 2015-07-31
Fix usings

sriramgd:
Second try on https://github.com/OmniSharp/omnisharp-roslyn/pull/150

The code can be optimized for speed (performance), for now to test may need to increase timeout from client editor.

1. Add missing usings
2. Send a dictionary of key-list for ambiguous usings
3. Remove redundant usings
TODO: 
1. Remove unused using statements
2. Sort using statements
3. If possible, replace RemoveRedundantUsings by something better from the framework

### DIFF
--- a/src/OmniSharp/Workers/Refactoring/FixUsings.cs
+++ b/src/OmniSharp/Workers/Refactoring/FixUsings.cs
@@ -1,0 +1,276 @@
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CodeActions;
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.CodeRefactorings;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using OmniSharp.Models;
+using OmniSharp.Services;
+
+namespace OmniSharp
+{
+    public class FixUsingsWorker
+    {
+        public static async Task<FixUsingsResponse> FixUsings(OmnisharpWorkspace workspace, string fileName, Document document, SemanticModel semanticModel, bool wantsTextChanges)
+        {
+            var codeActionProvider = new RoslynCodeActionProvider();
+            var ambiguous = await AddMissingUsings(fileName, workspace, document, semanticModel, codeActionProvider);
+            await OrganizeUsings(fileName, workspace, document, semanticModel, codeActionProvider);
+            document = workspace.GetDocument(fileName);
+            var compilationUnitSyntax = (await AddLinqQuerySyntax(fileName, workspace, document, semanticModel));
+            document = document.WithSyntaxRoot(compilationUnitSyntax);
+
+            var response = new FixUsingsResponse();
+            response.AmbiguousResults = ambiguous;
+            if (!wantsTextChanges)
+            {
+                // return the new document
+                var docText = await document.GetTextAsync();
+                response.Buffer = docText.ToString();
+            }
+            else
+            {
+                // return the text changes
+                var changes = await workspace.CurrentSolution.GetDocument(document.Id).GetTextChangesAsync(document);
+                response.Changes = await LinePositionSpanTextChange.Convert(document, changes);
+            }
+
+            return response;
+        }
+
+        private static async Task<List<QuickFix>> AddMissingUsings(string fileName, OmnisharpWorkspace workspace, Document document, SemanticModel semanticModel, ICodeActionProvider codeActionProvider)
+        {
+            bool processMore = true;
+            var ambiguousNodes = new List<SimpleNameSyntax>();
+            var ambiguous = new List<QuickFix>();
+
+            while (processMore)
+            {
+                bool updated = false;
+
+                var syntaxNode = (await document.GetSyntaxTreeAsync()).GetRoot();
+                var nodes = syntaxNode.DescendantNodes()
+                    .OfType<SimpleNameSyntax>()
+                    .Where(x => semanticModel.GetSymbolInfo(x).Symbol == null && !ambiguousNodes.Contains(x)).ToList();
+
+                foreach (var node in nodes)
+                {
+                    document = workspace.GetDocument(fileName);
+                    semanticModel = await document.GetSemanticModelAsync();
+                    var sourceText = (await document.GetTextAsync());
+                    var diagnostics = semanticModel.GetDiagnostics();
+                    var location = node.Identifier.Span;
+                    var usingOperations = new Dictionary<string, CodeActionOperation>();
+                    //Restrict diagnostics only to missing usings
+                    var pointDiagnostics = diagnostics.Where(d => d.Location.SourceSpan.Contains(location) &&
+                            (d.Id == "CS0246" || d.Id == "CS1061" || d.Id == "CS0103")).ToImmutableArray();
+
+                    if (pointDiagnostics.Any())
+                    {
+                        var pointdiagfirst = pointDiagnostics.First().Location.SourceSpan;
+                        if (pointDiagnostics.Any(d => d.Location.SourceSpan != pointdiagfirst))
+                        {
+                            continue;
+                        }
+                        var usingActions = GetUsingActions(document, codeActionProvider, pointDiagnostics, "using");
+                        foreach (var action in usingActions)
+                        {
+                            var operations = await action.GetOperationsAsync(CancellationToken.None).ConfigureAwait(false);
+                            if (operations != null)
+                            {
+                                foreach (var codeOperation in operations)
+                                {
+                                    usingOperations.Add(action.Title, codeOperation);
+                                }
+                            }
+                        }
+
+                        if (usingOperations.Count() == 1)
+                        {
+                            //Only one operation - apply it
+                            usingOperations.Single().Value.Apply(workspace, CancellationToken.None);
+                            updated = true;
+                        }
+                        else if (usingOperations.Count() > 1)
+                        {
+                            //More than one operation - ambiguous
+                            ambiguousNodes.Add(node);
+                            var unresolvedText = node.Identifier.ValueText;
+                            var unresolvedLocation = node.GetLocation().GetLineSpan().StartLinePosition;
+                            ambiguous.Add(new QuickFix
+                            {
+                                Line = unresolvedLocation.Line + 1,
+                                Column = unresolvedLocation.Character + 1,
+                                FileName = fileName,
+                                Text = "`" + unresolvedText + "`" + " is ambiguous"
+                            });
+                        }
+                    }
+                }
+
+                processMore = updated;
+            }
+
+            return ambiguous;
+        }
+
+        private static async Task OrganizeUsings(string fileName, OmnisharpWorkspace workspace, Document document, SemanticModel semanticModel, ICodeActionProvider codeActionProvider)
+        {
+            //Remove unneccessary usings
+            var syntaxNode = (await document.GetSyntaxTreeAsync()).GetRoot();
+            var nodes = syntaxNode.DescendantNodes().Where(x => x is UsingDirectiveSyntax);
+
+            foreach (var node in nodes)
+            {
+                document = workspace.GetDocument(fileName);
+                semanticModel = await document.GetSemanticModelAsync();
+                var sourceText = (await document.GetTextAsync());
+                var diagnostics = semanticModel.GetDiagnostics();
+                var location = node.Span;
+                var actions = new List<CodeAction>();
+                //Restrict diagnostics only to unneccessary and duplicate usings
+                var pointDiagnostics = diagnostics.Where(d => d.Location.SourceSpan.Contains(location) &&
+                        (d.Id == "CS0105" || d.Id == "CS8019")).ToImmutableArray();
+
+                if (pointDiagnostics.Any())
+                {
+                    var pointdiagfirst = pointDiagnostics.First().Location.SourceSpan;
+                    if (pointDiagnostics.Any(d => d.Location.SourceSpan != pointdiagfirst))
+                    {
+                        continue;
+                    }
+                    var usingActions = GetUsingActions(document, codeActionProvider, pointDiagnostics, "Remove Unnecessary Usings");
+                    foreach (var action in usingActions)
+                    {
+                        var operations = await action.GetOperationsAsync(CancellationToken.None).ConfigureAwait(false);
+                        if (operations != null)
+                        {
+                            foreach (var codeOperation in operations)
+                            {
+                                if (codeOperation != null)
+                                {
+                                    codeOperation.Apply(workspace, CancellationToken.None);
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+
+            //Sort usings
+            //TODO: Sorting usings is not available as a public service in Roslyn yet.
+            //When it is, replace the NRefactory based provider with the Roslyn one.
+            var nRefactoryProvider = new NRefactoryCodeActionProvider();
+            var sortActions = new List<CodeAction>();
+            var refactoringContext = await GetRefactoringContext(document, sortActions);
+            if (refactoringContext != null)
+            {
+                foreach (var refactoring in nRefactoryProvider.Refactorings)
+                {
+                    if (refactoring.ToString() != "ICSharpCode.NRefactory6.CSharp.Refactoring.SortUsingsAction")
+                    {
+                        continue;
+                    }
+                    await refactoring.ComputeRefactoringsAsync(refactoringContext.Value);
+                }
+
+                foreach (var action in sortActions)
+                {
+                    var operations = await action.GetOperationsAsync(CancellationToken.None).ConfigureAwait(false);
+                    if (operations != null)
+                    {
+                        foreach (var codeOperation in operations)
+                        {
+                            if (codeOperation != null)
+                            {
+                                codeOperation.Apply(workspace, CancellationToken.None);
+                            }
+                        }
+                    }
+                }
+            }
+
+            return;
+        }
+
+        private static async Task<CodeRefactoringContext?> GetRefactoringContext(Document document, List<CodeAction> actionsDestination)
+        {
+            var firstUsing = (await document.GetSyntaxTreeAsync()).GetRoot().DescendantNodes().Where(n => n is UsingDirectiveSyntax);
+            if (firstUsing.Count() == 0)
+            {
+                return null;
+            }
+            var location = firstUsing.First().GetLocation().SourceSpan;
+
+            return new CodeRefactoringContext(document, location, (a) => actionsDestination.Add(a), CancellationToken.None);
+        }
+
+        private static List<CodeAction> GetUsingActions(Document document, ICodeActionProvider codeActionProvider,
+                ImmutableArray<Diagnostic> pointDiagnostics, string actionPrefix)
+        {
+
+            var actions = new List<CodeAction>();
+            var context = new CodeFixContext(document, pointDiagnostics.First().Location.SourceSpan, pointDiagnostics, (a, d) => actions.Add(a), CancellationToken.None);
+            var providers = codeActionProvider.CodeFixes;
+
+            //Disable await warning since we dont need the result of the call. Else we need to use a throwaway variable.
+#pragma warning disable 4014
+            foreach (var provider in providers)
+            {
+                provider.RegisterCodeFixesAsync(context);
+            }
+#pragma warning restore 4014
+
+            return actions.Where(a => a.Title.StartsWith(actionPrefix)).ToList();
+        }
+
+        private static async Task<CompilationUnitSyntax> AddLinqQuerySyntax(string fileName, OmnisharpWorkspace workspace, Document document, SemanticModel semanticModel)
+        {
+            var syntaxNode = (await document.GetSyntaxTreeAsync()).GetRoot();
+            var compilationUnitSyntax = (CompilationUnitSyntax)syntaxNode;
+            var usings = GetUsings(syntaxNode);
+            if (HasLinqQuerySyntax(semanticModel, syntaxNode) && !usings.Contains("using System.Linq;"))
+            {
+                var linqName = SyntaxFactory.QualifiedName(SyntaxFactory.IdentifierName("System"),
+                        SyntaxFactory.IdentifierName("Linq"));
+                var linq = SyntaxFactory.UsingDirective(linqName).NormalizeWhitespace()
+                            .WithTrailingTrivia(SyntaxFactory.Whitespace(Environment.NewLine));
+                compilationUnitSyntax = compilationUnitSyntax.AddUsings(linq);
+            }
+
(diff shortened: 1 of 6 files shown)


sriramgd:
The tests fail with nullreference exception, though the tests I added for this PR passes. The failure seems random - different tests fail each time I do a whitespace commit.

tests/OmniSharp.Tests/FixUsingsFacts.cs
```diff
@@ -0,0 +1,348 @@
...
+    public class class1 
+    {
+        public void method1()()
+        {
+            var s = new StringBuilder();
+        }
+    }
+}";
+            string expectedFileContents = @"using System.Text;"  + 
+            Environment.NewLine +
```
nosami: Why are you adding Enviroment.NewLine here? Better as 1 string rather than concatenating?

sriramgd: I am following the way I am adding missing usings. The code which adds the using is:
    var usingToAdd = SyntaxFactory.UsingDirective(candidateName).NormalizeWhitespace()
            .WithTrailingTrivia(SyntaxFactory.Whitespace(Environment.NewLine));

The tests which are failing are not the ones I added, but different/random ones. The tests all pass locally.

nosami: Yeah. Don't worry about those. I know how to fix.

nosami: You shouldn't be using Environment.NewLine in either place. It should get the newline character from the configuration.

src/OmniSharp/Workers/Refactoring/FixUsings.cs
```diff
@@ -0,0 +1,201 @@
...
+                .Where(x => x.Kind() == SyntaxKind.QueryExpression || x.Kind() == SyntaxKind.QueryBody || x.Kind() == SyntaxKind.FromClause || x.Kind() == SyntaxKind.LetClause || x.Kind() == SyntaxKind.JoinClause || x.Kind() == SyntaxKind.JoinClause || x.Kind() == SyntaxKind.JoinIntoClause || x.Kind() == SyntaxKind.WhereClause || x.Kind() == SyntaxKind.OrderByClause || x.Kind() == SyntaxKind.AscendingOrdering || x.Kind() == SyntaxKind.DescendingOrdering || x.Kind() == SyntaxKind.SelectClause || x.Kind() == SyntaxKind.GroupClause || x.Kind() == SyntaxKind.QueryContinuation).Any();
+        }
+
+        private static HashSet<string> GetUsings(SyntaxNode root)
+        {
+            var usings = new HashSet<string>();
+            root.DescendantNodes().OfType<UsingDirectiveSyntax>()
+                .ToList().ForEach(x =>
+                        {
+                            if (!usings.Contains(x.ToString().Trim()))
```
nosami: HashSet should automatically remove dupes. There is no need for this.

Something like this would be better I think.

```cs
var usings = root.DescendantNodes().OfType<UsingDirectiveSyntax>().Select(using => using.ToString().Trim());
return new HashSet<string>(usings);
```

src/OmniSharp/Workers/Refactoring/FixUsings.cs
```diff
@@ -0,0 +1,201 @@
...
+                        {
+                            compilationUnitSyntax = compilationUnitSyntax.AddUsings(usingToAdd);
+                            usingsToAdd.Add(usingToAddText.Trim());
+                        }
+                    }
+                }
+
+                //Handle linq with query syntax
+                var linqName = SyntaxFactory.QualifiedName(SyntaxFactory.IdentifierName("System"), SyntaxFactory.IdentifierName("Linq"));
+                var linqUsingText = "using " + linqName.ToString() + ";";
```
nosami: Is this better than just `using System.Linq;` ? I'm not so sure.

src/OmniSharp/Workers/Refactoring/FixUsings.cs
```diff
@@ -0,0 +1,201 @@
...
+                .OfType<SimpleNameSyntax>()
+                .Where(x => semanticModel.GetSymbolInfo(x).Symbol == null &&
+                        semanticModel.GetSymbolInfo(x).CandidateReason != CandidateReason.OverloadResolutionFailure);
+
+        }
+
+        private static bool HasLinqQuerySyntax(SemanticModel semanticModel, SyntaxNode syntaxNode)
+        {
+            return syntaxNode.DescendantNodes()
+                .Where(x => x.Kind() == SyntaxKind.QueryExpression || x.Kind() == SyntaxKind.QueryBody || x.Kind() == SyntaxKind.FromClause || x.Kind() == SyntaxKind.LetClause || x.Kind() == SyntaxKind.JoinClause || x.Kind() == SyntaxKind.JoinClause || x.Kind() == SyntaxKind.JoinIntoClause || x.Kind() == SyntaxKind.WhereClause || x.Kind() == SyntaxKind.OrderByClause || x.Kind() == SyntaxKind.AscendingOrdering || x.Kind() == SyntaxKind.DescendingOrdering || x.Kind() == SyntaxKind.SelectClause || x.Kind() == SyntaxKind.GroupClause || x.Kind() == SyntaxKind.QueryContinuation).Any();
```
nosami: this line is _way_ too long :)

nosami:
There are Roslyn codefixes for adding / sorting and removing using statements. It should be possible to re-use the same code I think.

nosami:
I just realised that you probably haven't seen the `v2/getcodeactions` endpoint in action yet as it's not been implemented in vim. 

This allows you to add a using for the current symbol under the cursor. It also sorts and removes unused usings (as separate actions - the cursor has to be on a using statement for this to work). 

I would imagine that you could loop over the unresolved symbols and run the CodeAction to add a using statement for each one - it might reduce the amount of code that you have to write.

sriramgd:
Took a look at getcodeactions. Agree that it should be possible to run codeaction over each unresolved symbol. Will rewrite the code using this and resubmit.

sriramgd:
@nosami, I am not sure why, but when I apply the codeactions in a loop - only the latest one seems to apply for the workspace. I will look at this some more, was wondering if you had come across something like this. If this does not work, I will still use code actions but apply the changes in a more manual way.

nosami:
Every change to a solution object results in a new solution object being created as everything is immutable. 

I think there is a call Workspace.TryApplyChanges or something like that?

sriramgd:
I tried Workspace.TryApplyChanges too, run into the same issue. Only the last iteration is applied. Actually on each iteration the change is applied to the original workspace, so the last one sticks. I took a look at the Roslyn source, and couldnt really tell why this happens.
I have asked about this on so and msdn, hoping someone will help.
http://stackoverflow.com/questions/31832756/how-to-apply-a-list-of-applychangesoperation-on-a-roslyn-workspace
https://social.msdn.microsoft.com/Forums/vstudio/en-US/0d139c6d-33d7-4911-9984-d0295d02bc98/how-to-apply-a-list-of-applychangesoperation-on-a-roslyn-workspace?forum=roslyn

tests/OmniSharp.Tests/FixUsingsFacts.cs
```diff
@@ -0,0 +1,342 @@
...
+                Assert.Equal(qfList[i].FileName, expectedQuickFix.FileName);
+                Assert.Equal(qfList[i].Text, expectedQuickFix.Text);
+                i++;
+            }
+        }
+
+        private async Task<FixUsingsResponse> RunFixUsings(string fileContents)
+        {
+            var workspace = TestHelpers.CreateSimpleWorkspace(fileContents, fileName);
+            var controller = new OmnisharpController(workspace, null);
```
nosami: You should pass in options here. Instead of null, use `new FakeOmniSharpOptions()`. You can modify the FakeOmniSharpOptions class to allow you to specify the desired newline character.

src/OmniSharp/Workers/Refactoring/FixUsings.cs
```diff
@@ -0,0 +1,238 @@
...
+
+            foreach (var node in nodes)
+            {
+                document = workspace.GetDocument(fileName);
+                semanticModel = await document.GetSemanticModelAsync();
+                var sourceText = (await document.GetTextAsync());
+                var diagnostics = semanticModel.GetDiagnostics();
+                var location = node.Span;
+                var actions = new List<CodeAction>();
+                var pointDiagnostics = diagnostics.Where(d => d.Location.SourceSpan.Contains(location)).ToImmutableArray();
```
nosami: This should be filtered further to only include the relevant diagnostics for this location.

src/OmniSharp/Workers/Refactoring/FixUsings.cs
```diff
@@ -0,0 +1,238 @@
...
+                foreach (var node in nodes)
+                {
+                    document = workspace.GetDocument(fileName);
+                    semanticModel = await document.GetSemanticModelAsync();
+                    var sourceText = (await document.GetTextAsync());
+                    var diagnostics = semanticModel.GetDiagnostics();
+                    var location = node.Identifier.Span;
+                    var actions = new List<CodeAction>();
+                    var usingOperations = new Dictionary<string, CodeActionOperation>();
+                    var pointDiagnostics = diagnostics.Where(d => d.Location.SourceSpan.Contains(location)).ToImmutableArray();
```
nosami: This should be filtered further to only include the relevant diagnostics for this location.

src/OmniSharp/Workers/Refactoring/FixUsings.cs
```diff
@@ -0,0 +1,238 @@
...
+                                if (action.Title.StartsWith("using"))
+                                {
+                                    var operations = await action.GetOperationsAsync(CancellationToken.None).ConfigureAwait(false);
+                                    if (operations != null)
+                                    {
+                                        foreach (var codeOperation in operations)
+                                        {
+                                            if (codeOperation != null)
+                                            {
+                                                usingOperations.Add(action.Title, codeOperation);
```
nosami: This code is 10 levels deep. I think you should consider refactoring this.

src/OmniSharp/Workers/Refactoring/FixUsings.cs
```diff
@@ -0,0 +1,238 @@
...
+
+                        if (usingOperations.Count() == 1)
+                        {
+                            //Only one operation - apply it
+                            usingOperations.Single().Value.Apply(workspace, CancellationToken.None);
+                            updated = true;
+                        }
+                        else if (usingOperations.Count() > 1)
+                        {
+                            if (usingOperations.Keys.Any(k => k.StartsWith("using Microsoft.CodeAnalysis.CSharp")))
```
nosami: I don't understand this line. Why is it here?

src/OmniSharp/Workers/Refactoring/FixUsings.cs
```diff
@@ -0,0 +1,238 @@
...
+
+                processMore = updated;
+            }
+
+            return ambiguous;
+        }
+
+        private static async Task RemoveDuplicateUsings(string fileName, OmnisharpWorkspace workspace, Document document, SemanticModel semanticModel, ICodeActionProvider codeActionProvider)
+        {
+
```
nosami: superfluous blank line.

src/OmniSharp/Workers/Refactoring/FixUsings.cs
```diff
@@ -0,0 +1,238 @@
...
+                                usings.Add(x.ToString().Trim());
+                            }
+                        });
+            return usings;
+        }
+
+        private static bool HasLinqQuerySyntax(SemanticModel semanticModel, SyntaxNode syntaxNode)
+        {
+            return syntaxNode.DescendantNodes()
+                .Where(x => x.Kind() == SyntaxKind.QueryExpression
```
nosami: Instead of `Where().Any()` you could just use `Any(x => ....)`

src/OmniSharp/Workers/Refactoring/FixUsings.cs
nosami: i.e. only look for `Unnecessary using directive.`. I think diagnostics have an ID that you could use.

src/OmniSharp/Workers/Refactoring/FixUsings.cs
```diff
@@ -0,0 +1,238 @@
...
+                        var context = new CodeFixContext(document, pointDiagnostics.First().Location.SourceSpan, pointDiagnostics, (a, d) => actions.Add(a), CancellationToken.None);
+                        var providers = codeActionProvider.CodeFixes;
+
+                        foreach (var provider in providers)
+                        {
+                            var register = provider.RegisterCodeFixesAsync(context);
+
+                            foreach (var action in actions)
+                            {
+                                if (action.Title.StartsWith("using"))
```
nosami: I think that instead of looping over the providers and then actions, you should create an instance of the CodeAction that you want to use (will probably need reflection) and keep it around between requests.

src/OmniSharp/Workers/Refactoring/FixUsings.cs
```diff
@@ -0,0 +1,238 @@
...
+                                {
+                                    var operations = await action.GetOperationsAsync(CancellationToken.None).ConfigureAwait(false);
+                                    if (operations != null)
+                                    {
+                                        foreach (var codeOperation in operations)
+                                        {
+                                            if (codeOperation != null)
```
nosami: Can this be null?

tests/OmniSharp.Tests/FixUsingsFacts.cs
```diff
@@ -0,0 +1,342 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+using OmniSharp.Models;
+using Xunit;
+ 
+namespace OmniSharp.Tests
+{ 
+    public class FixUsingsFacts
+    {
+
```
nosami: extra blank line

nosami:
A few things still to change, but this is looking much better now. Thanks!

Also, we need to be able to return the changes that were made to the document instead of the full text of the modified buffer. See [here](https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp/Api/v1/Refactoring/OmnisharpController.CodeActions.cs#L57-L68) for an example. Vim and Sublime are the only editors left that use the full text.

src/OmniSharp/Workers/Refactoring/FixUsings.cs
sriramgd: Done

src/OmniSharp/Workers/Refactoring/FixUsings.cs
sriramgd: Done

src/OmniSharp/Workers/Refactoring/FixUsings.cs
sriramgd: Split

tests/OmniSharp.Tests/FixUsingsFacts.cs
sriramgd: Using fake options, unit tests work locally but fail on Appveyor not sure how to fix it. The error seems to start at:   OmniSharp.Tests.FormattingFacts.TextChangesAreSortedLastFirst_MultipleLines [FAIL]
      System.NullReferenceException : Object reference not set to an instance of an object.
      Stack Trace:
            at Microsoft.CodeAnalysis.Workspace.get_Options()
            at Microsoft.CodeAnalysis.Host.AbstractSyntaxTreeFactoryService..ctor(HostLanguageServices languageServices)

src/OmniSharp/Workers/Refactoring/FixUsings.cs
sriramgd: Filtered using IDs

src/OmniSharp/Workers/Refactoring/FixUsings.cs
sriramgd: Filtered.

src/OmniSharp/Workers/Refactoring/FixUsings.cs
sriramgd: Have reduced the nesting. There was no need to loop over each action/operation for each provider. Looping over the providers separately.

src/OmniSharp/Workers/Refactoring/FixUsings.cs
sriramgd: Usings which start with using Microsoft.CodeAnalysis.CSharp and Microsoft.CodeAnalytics.VisualBasic (for using Microsoft.CodeAnalysis.<language>.Syntax), cause ambiguity; so resolved them all to CSharp.

src/OmniSharp/Workers/Refactoring/FixUsings.cs
sriramgd: Done

src/OmniSharp/Workers/Refactoring/FixUsings.cs
sriramgd: No, removed.

sriramgd:
@nosami, I have incorporated most of the review comments (I think). The builds still fail - appveyor on a GetOptions call and Travis due to non-existent project.lock.json. The tests work locally, and I am using the feature locally for editing and it works ok.

src/OmniSharp/Workers/Refactoring/FixUsings.cs
nosami: If it's ambiguous, it's ambiguous isn't it? I don't think we should decide which using statement it should be here.

src/OmniSharp/Models/v2/FixUsingsResponse.cs
```diff
@@ -0,0 +1,14 @@
+using System.Collections.Generic;
+
+namespace OmniSharp.Models.V2
```
nosami: This file should be removed.

src/OmniSharp/Workers/Refactoring/FixUsings.cs
```diff
@@ -0,0 +1,240 @@
...
+            return;
+        }
+
+        private static List<CodeAction> GetUsingActions(Document document, ICodeActionProvider codeActionProvider,
+                ImmutableArray<Diagnostic> pointDiagnostics, string actionPrefix)
+        {
+
+            var actions = new List<CodeAction>();
+            var context = new CodeFixContext(document, pointDiagnostics.First().Location.SourceSpan, pointDiagnostics, (a, d) => actions.Add(a), CancellationToken.None);
+            var providers = (IEnumerable<CodeFixProvider>)codeActionProvider.CodeFixes;
```
nosami: I don't think this cast is needed.

src/OmniSharp/Workers/Refactoring/FixUsings.cs
```diff
@@ -0,0 +1,240 @@
...
+                    }
+                }
+
+                processMore = updated;
+            }
+
+            return ambiguous;
+        }
+
+        private static async Task RemoveDuplicateUsings(string fileName, OmnisharpWorkspace workspace, Document document, SemanticModel semanticModel, ICodeActionProvider codeActionProvider)
```
nosami: This method should be renamed to RemoveUsings or something similar. It isn't just removing duplicates any more.

src/OmniSharp/Workers/Refactoring/FixUsings.cs
```diff
@@ -0,0 +1,240 @@
...
+        }
+
+        private static async Task<CompilationUnitSyntax> AddLinqQuerySyntax(string fileName, OmnisharpWorkspace workspace, Document document, SemanticModel semanticModel)
+        {
+            var syntaxNode = (await document.GetSyntaxTreeAsync()).GetRoot();
+            var compilationUnitSyntax = (CompilationUnitSyntax)syntaxNode;
+            var usings = GetUsings(syntaxNode);
+            if (HasLinqQuerySyntax(semanticModel, syntaxNode) && !usings.Contains("using System.Linq;"))
+            {
+                System.Console.WriteLine("Adding linq");
```
nosami: Delete this line :)

tests/OmniSharp.Tests/FixUsingsFacts.cs
```diff
@@ -0,0 +1,431 @@
...
+            }
+        }
+
+        private async Task<FixUsingsResponse> RunFixUsings(string fileContents)
+        {
+            var workspace = TestHelpers.CreateSimpleWorkspace(fileContents, fileName);
+
+            var fakeOptions = new FakeOmniSharpOptions();
+            fakeOptions.Options = new OmniSharpOptions();
+            fakeOptions.Options.FormattingOptions = new FormattingOptions() {NewLine = "\r\n"};
```
nosami: Pretty sure that the line ending should be `\n` here. Git converts the files to \n on commit. I think this will make the the tests pass in appveyor too.

src/OmniSharp/Models/v2/FixUsingsResponse.cs
sriramgd: Removed.

src/OmniSharp/Workers/Refactoring/FixUsings.cs
sriramgd: Removed

src/OmniSharp/Workers/Refactoring/FixUsings.cs
sriramgd: Renamed as RemoveUsings

src/OmniSharp/Workers/Refactoring/FixUsings.cs
sriramgd: Deleted :)

tests/OmniSharp.Tests/FixUsingsFacts.cs
sriramgd: Updated to \n. Build still fails, not sure what I am missing.

sriramgd:
@nosami, 
1. Sort usings is not as simple as I thought. The correct code is in roslyn but is marked internal, and it would be a lot of code to bring down since it includes dependencies which are internal too: https://github.com/dotnet/roslyn/blob/0382e3e3fc543fc483090bff3ab1eaae39dfb4d9/src/Features/CSharp/Organizing/CSharpOrganizingService.Rewriter.cs
2. There seems to be something wrong with the way the newlines are generated from the codeactions in appveyor builds.
 - I ensured my file ends with \n, and the fakeoption is set to \n too - this passes on my box and travis but fails on appveyor since it expects \r\n as line ending.
- If I make the file windows formatted and set fakeoption to be \r\n - the tests pass on appveyor but fail on travis and my box.
Not sure what to do - the latest checkin has \n as line ending and fake option.
3. The tests pass on travis, but fail on a project.lock.json file.

nosami:
There is a `sort usings` code action in NRefactory. You should be able to invoke it with the cursor on one of the using statements.

If you can get that working, I can take a look at the tests if you'd like.

nosami:
For the appveyor line endings, take a look [here](https://github.com/ember-cli/ember-cli/blob/master/appveyor.yml#L5).

Seems like appveyor isn't checking the files out with autocrlf true

sriramgd:
@sami, I have added sort usings. Let me know if I should strip out \r from the fix using unit tests to get past the appveyor error.

nosami:
Closing in favour of #291 . Thanks for persevering with this :+1: This is my favourite feature :)

Issue #275, opened 2015-07-23
dnx, dnu and DTH will be renamed


victorhurdugaci:
We are in the process of making a mega rename in DNX. The issue detailing the renames is here: https://github.com/aspnet/dnx/issues/1855

I wanted to start a discussion here because the names will impact OmniSharp and VSCode.

Ideally, the changes to OmniSharp should be backward compatible (support both the old and the new names) and the code supporting both names will go out before the renames. That way, no user will be impacted. 

Thoughts? Concerns? Suggestions?

victorhurdugaci:
cc @sayedihashimi

victorhurdugaci:
Btw, I think I'll have a dnx build with the new names by the end of week

sayedihashimi:
Adding @chrisdias as well

nosami:
Hard to say for certain, but I think it will be a low impact for OmniSharp. It looks like we might just need to change the way that we [launch DTH](https://github.com/OmniSharp/omnisharp-roslyn/blob/6f4a6175132d2b6a370583658742fff280c7ce5f/src/OmniSharp/Dnx/DesignTimeHostManager.cs#L44) if we detect that the user's application is running on beta6. 

We might have some small changes to make when OmniSharp itself moves to beta6 but that's not important.

We use some of the DTH namespaces for IncomingMessages and OutgoingMessages but they're just DTOs and aren't relevant at runtime.

nosami:
There will be a small change needed in VSCode needed to update to the latest omnisharp-roslyn. @jrieken knows about it.

davidfowl:
This will be a tiny change on the omnisharp side as @nosami already mentioned

victorhurdugaci:
The DNX renames are done: https://github.com/aspnet/announcements/issues/51

nosami:
Fixed in v1.1.1 of omnisharp-roslyn... and released in v0.23.1 of omnisharp-atom

PR #281, opened 2015-07-31
React to DesignTimeHost rename

Alxandr:
Allows omnisharp to work with both older and newer versions of DNX


nosami:
fixes https://github.com/OmniSharp/omnisharp-roslyn/issues/275

PR #284, opened 2015-08-09
Add flag to allow update buffer to read from disk

nosami:
I need this in emacs when a buffer is reverted from disk. I don't get any file changed events and just need the server to have the same version of the file.


david-driscoll:
Maybe add a test to `UpdateBufferFilterFacts`?  It's pretty minor I know... :+1:

nosami:
Thought about it ... but that would require having files on disk to test which is a bit of a pain and not worth it IMO. It's going to be tested from emacs though.

p.s. this code doesn't work :)

nosami:
On second thoughts, I should totally write a test ;)

david-driscoll:
Is this in a place where it could be tested with atom?

nosami:
Yeah.... it's working for me now in emacs. No idea what's up with the tests though. Not related to my changes but I'm investigating.

PR #289, opened 2015-08-20
Conditionally apply changes for rename server side.

nosami:
Added a new flag to the request. Ideally, we would have used
WantsTextChanges to both return text changes and also to not apply
them at immediately at the server.

The assumption here is that new text changes will be returned to the
server as they are applied to the editor.

### DIFF
--- a/src/OmniSharp/Api/v1/Refactoring/OmnisharpController.Rename.cs
+++ b/src/OmniSharp/Api/v1/Refactoring/OmnisharpController.Rename.cs
@@ -73,8 +73,15 @@ namespace OmniSharp
                     }
                 }
 
-                // Attempt to update the workspace
-                if (_workspace.TryApplyChanges(solution))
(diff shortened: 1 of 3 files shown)


jrieken:
@nosami - Should we do the same for code actions? https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp/Api/v2/Refactoring/OmnisharpController.CodeActions.cs#L67

jrieken:
maybe we can do a flag which makes [this](https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp/Roslyn/OmniSharpWorkspace.cs#L103) return ```false```. I think that will do the job. I'll try...

nosami:
@jrieken This was actually weird. I thought that code actions would need it too, but seems like it doesn't.

jrieken:
funky

PR #290, opened 2015-08-24
Allow to send file specific edits with a request

jrieken:
Simliar to requests containing a new [buffer](https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp/Models/v1/Request.cs#L11) this PR adds support for attaching incremental buffer updates to a request. This allows you get around an explicit call to ```/changebuffer``` and also has the neat side effect that clients don't have to process requests one after the other but send them in parallel - given they know what they do ;-)


nosami:
Looks good to me. Think I'll try and adopt this code myself. Thanks!

PR #291, opened 2015-08-26
Sriramgd fixusing

nosami:
WIP. Continuation of #280

### DIFF
--- a/src/OmniSharp/Workers/Refactoring/FixUsings.cs
+++ b/src/OmniSharp/Workers/Refactoring/FixUsings.cs
@@ -1,0 +1,276 @@
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CodeActions;
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.CodeRefactorings;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using OmniSharp.Models;
+using OmniSharp.Services;
+
+namespace OmniSharp
+{
+    public class FixUsingsWorker
+    {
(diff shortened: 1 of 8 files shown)


src/OmniSharp/Workers/Refactoring/FixUsings.cs
```diff
@@ -0,0 +1,212 @@
...
+        {
+            var actions = new List<CodeAction>();
+            var context = new CodeFixContext(_document, pointDiagnostics.First().Location.SourceSpan, pointDiagnostics, (a, d) => actions.Add(a), CancellationToken.None);
+            var providers = codeActionProvider.CodeFixes;
+
+            //Disable await warning since we dont need the result of the call. Else we need to use a throwaway variable.
+#pragma warning disable 4014
+            foreach (var provider in providers)
+            {
+                provider.RegisterCodeFixesAsync(context);
```
nosami: @sriramgd If this call is awaited, the code breaks. Did you see this? Any idea why?

sriramgd: @nosami, That is because the async method returns a Task which we need to ignore. But ignoring the returned object causes the compiler to throw a warning. So another option is to do an await and read the return value into a throwaway variable. I had that in place before, but used the pragma to be explicit. http://stackoverflow.com/a/12548725

nosami: Hmm... but it's an async method. It _should_ be awaited.

sriramgd: Perhaps something like: https://github.com/dotnet/roslyn/blob/03e3cfbfaa95a8d697beb4fbb7dd8fbc8f221386/src/EditorFeatures/Test/RenameTracking/RenameTrackingTestState.cs#L178

nosami:
Could someone please finish this (maybe @sriramgd)? I'm not going to have time to work on it any time soon.

There is currently a bug with it. Ambiguous using locations are the locations from `before` new usings are added or removed instead of `after`

sriramgd:
@nosami, I will take this.

sriramgd:
@nosami , I have updated the code with ambigous using and linq query syntax on https://github.com/sriramgd/omnisharp-roslyn/tree/sriramgd-fixusing
I dont have access to the master so cant push into the feature branch for this PR. Do you need me to open a new PR, or can the merge be done from my branch?

nosami:
closing in favour of #304 , #305

PR #293, opened 2015-08-29
Plugin system with MEF

david-driscoll:


src/OmniSharp.Abstractions/Roslyn/OmniSharpWorkspace.cs
```diff
@@ -13,6 +17,10 @@ public class OmnisharpWorkspace : Workspace
 
         public BufferManager BufferManager { get; private set; }
 
+        public CompositionHost PluginHost { get; private set; }
+
+        public static OmnisharpWorkspace Instance{ get; private set; }
```
david-driscoll: I don't like having to have a static property.  Does anyone have any idea how to insert an export into an MEF CompositionHost?

I want to force the value in from the startup class if possible.

src/OmniSharp/project.json
```diff
@@ -4,7 +4,9 @@
         "warningsAsErrors": true
     },
     "dependencies": {
+        "OmniSharp.Abstractions": "1.0.0-*",
         "OmniSharp.Stdio": "1.0.0-*",
+        "OmniSharp.Roslyn.CSharp": "1.0.0-*",
```
david-driscoll: This is just temporary, to force a reference to the assembly, for true plugins we'd have a set of folders to scan for assemblies.

src/OmniSharp/Api/v1/Navigation/OmnisharpController.GotoDefinition.cs
```diff
@@ -16,59 +18,16 @@ public partial class OmnisharpController
         [HttpPost("gotodefinition")]
         public async Task<GotoDefinitionResponse> GotoDefinition(GotoDefinitionRequest request)
         {
-            var quickFixes = new List<QuickFix>();
-
-            var document = _workspace.GetDocument(request.FileName);
-            var response = new GotoDefinitionResponse();
-            if (document != null)
+            var definitions = _workspace.PluginHost.GetExports<IGotoDefintion>();
+            foreach (var instance in definitions)
```
nosami: This pattern is going to be the same for all request handlers. Would be nice to abstract it away somehow.

nosami: I feel like this whole class could disappear and replaced with RegisterHandler<GotoDefinitionRequest, GotoDefinitionResponse>("gotodefinition") somewhere.

src/OmniSharp.Abstractions/Endpoints/IGotoDefinition.cs
```diff
@@ -0,0 +1,11 @@
+using System.Threading.Tasks;
+using OmniSharp.Models;
+
+namespace OmniSharp.Endpoints
+{
+    public interface IGotoDefintion
+    {
+        Task<bool> IsApplicableTo(GotoDefinitionRequest request);
```
nosami: This interface could be generic I think.

Maybe

```
public interface IRequestHandler<TRequest, TResponse>
{
    bool IsApplicableTo(Language language);
   Task<TResponse> Handle(TRequest request);
}
```

src/OmniSharp.Roslyn.CSharp/Services/RoslynGotoDefinition.cs
```diff
@@ -0,0 +1,79 @@
...
+using OmniSharp.Models;
+
+namespace OmniSharp.Roslyn.CSharp.Services
+{
+    [Export(typeof(IGotoDefintion))]
+    class RoslynGotoDefinition : IGotoDefintion
+    {
+        public Task<bool> IsApplicableTo(GotoDefinitionRequest request)
+        {
+            return Task.FromResult(OmnisharpWorkspace.Instance.GetDocument(request.FileName) != null);
```
nosami: how about ```return extension.Equals("cs")```

nosami: The request object seems too heavyweight for this method. Should just be ```Language``` or ```Extension``` for file extension I think. I can't think of a situation where we need more than that.

src/OmniSharp.Abstractions/Endpoints/IGotoDefinition.cs
nosami: With the right [attributes](https://msdn.microsoft.com/en-us/library/ee155691(v=vs.110).aspx), it might be possible to avoid this method call altogether. Just fetch the gotodefinition handler than can handle C#

src/OmniSharp/Api/v1/Navigation/OmnisharpController.GotoDefinition.cs
david-driscoll: We could probably just insert everything as custom middleware, I'll look at that, would let us remove most of our endpoints (perhaps not all) from the mvc controller.

src/OmniSharp.Roslyn.CSharp/Services/RoslynGotoDefinition.cs
david-driscoll: I thought about `return extension.Equals("cs")` but then `ScriptCs` and `Cake` crossed my mind, which would also be using the Roslyn Services.

nosami: `return ["cs", "csx", "cake"].Contains(extension)` :)

src/OmniSharp.Abstractions/Endpoints/IGotoDefinition.cs
david-driscoll: So we could, but should we?  I suppose it's more `Functional`, lol.

We could do something like...

    public static class OmniSharpEndpoints
    {
        public static Type GotoDefinition {get;} = typeof(Func<GotoDefinitionRequest, Task<GotoDefintionResponse>>)
    }

    class RoslynServices
    {
        [OmniSharpEndpoint(OmniSharpEndpoints.GotoDefinition, Languages.CSharp)]
        public async Task<GotoDefinitionResponse> GotoDefinition(GotoDefinitionRequest request) { ... }
    }

That just feels verbose on the C# side, but may be easier on the F# side of things?  Could we support both?  And additionally have....

    [OmniSharpEndpoint(typeof(IGotoDefinition), Languages.CSharp")]
    class RoslynServices : IGotoDefinition { ... }

Then there is just an extra check to find the correct service, for the given language (which could be thrown behind a cache).

I'm not against the former, it would enforce some extra rules on our part to make sure we
a) never change the delegates
b) make sure we document any changes to the delegates up front.

Note: `OmniSharpEndpointAttribute` would derive from `ExportAttribute` and implement metadata attribute.  On the import side, we would import that metadata with the language.

david-driscoll: The one issue I see with the delegate route is that we can't initialize any properties from the MEF host.  So if we need any external services, they will all have to be done on demand in each delegate.

Also the workspace will have to stay static, or become part of the delegate itself, so we can hand it.

nosami: Erm, where did the talk of delegates come from? :)

nosami: Thinking about this more, the IsApplicableTo logic shouldn't be here at all. It should be at the top level (what types of files does this plugin handle?)

src/OmniSharp.Roslyn.CSharp/CSharpLanguage.cs
```diff
@@ -0,0 +1,15 @@
...
+using Microsoft.CodeAnalysis;
+using OmniSharp.Mef;
+
+namespace OmniSharp.Roslyn.CSharp.Services
+{
+    public class LanguagePredicate
+    {
+        private static readonly string[] ValidCSharpExtensions = { "cs", "csx", "cake" };
+        [OmniSharpLanguage(LanguageNames.CSharp)]
+        public Func<string, Task<bool>> IsApplicableTo { get; } = filePath => Task.FromResult(ValidCSharpExtensions.Any(extension => filePath.EndsWith(extension)));
```
david-driscoll: This was impossible to do as a method, and it's specifically something that isn't supported by the PCL version of MEF.  Delegate Properties on the other hand work just fine.

src/OmniSharp/Middleware/Endpoint/EndpointHandler.cs
```diff
@@ -0,0 +1,201 @@
...
+        private readonly Type _responseType;
+        private readonly OmnisharpWorkspace _workspace;
+
+        public EndpointHandler(OmnisharpWorkspace workspace, LanguagePredicateHandler languagePredicateHandler, CompositionHost host, EndpointMapItem item)
+        {
+            EndpointName = item.EndpointName;
+            _host = host;
+            _languagePredicateHandler = languagePredicateHandler;
+            _interfaceType = item.InterfaceType;
+            _workspaceDelegateType = WorkspaceFuncType.MakeGenericType(typeof(OmnisharpWorkspace), item.RequestType, TaskType.MakeGenericType(item.ResponseType));
```
david-driscoll: Support for `(OmnisharpWorkspace workspace, TRequest) => TResponse`

src/OmniSharp/Middleware/Endpoint/EndpointHandler.cs
```diff
@@ -0,0 +1,201 @@
...
+            _delegateType = FuncType.MakeGenericType(item.RequestType, TaskType.MakeGenericType(item.ResponseType));
```
david-driscoll: Support for `(TRequest) => TResponse`

src/OmniSharp/Middleware/Endpoint/EndpointHandler.cs
```diff
@@ -0,0 +1,201 @@
...
+        private readonly Type _requestType;
+        private readonly Type _responseType;
+        private readonly OmnisharpWorkspace _workspace;
+
+        public EndpointHandler(OmnisharpWorkspace workspace, LanguagePredicateHandler languagePredicateHandler, CompositionHost host, EndpointMapItem item)
+        {
+            EndpointName = item.EndpointName;
+            _host = host;
+            _languagePredicateHandler = languagePredicateHandler;
+            _interfaceType = item.InterfaceType;
```
david-driscoll: Support for implementation by interface.

src/OmniSharp.Roslyn.CSharp/CSharpLanguage.cs
nosami: Why the Task?

david-driscoll: If the plugin is out of process, then we might be waiting for stdio to give us a result, figured this would make that logic easier.  Though thinking about it, maybe it won't be needed.

src/OmniSharp.Abstractions/Endpoints.cs
```diff
@@ -0,0 +1,49 @@
...
+    {
+        public static EndpointMapItem[] AvailableEndpoints = {
+            EndpointMapItem.Create<GotoDefinition, GotoDefinitionRequest, GotoDefinitionResponse>("/gotodefinition"),
+        };
+
+        public interface GotoDefinition
+        {
+            Task<GotoDefinitionResponse> GotoDefinition(GotoDefinitionRequest request);
+        }
+
```
nosami: Could this be

```cs
        public static EndpointMapItem[] AvailableEndpoints = {
            EndpointMapItem.Create<GotoDefinition>("/gotodefinition"),
        };

        public interface Handler<TRequest, TResponse>
        {
            Task<TResponse> Handle(TRequest request);
        }
```

david-driscoll:
keeps getting better!

* Moved the following to their own projects
  * Dnx
  * MsBuild (currently only supporting dnx451)
  * ScriptCs (currently only supporting dnx451)
* Shuffled unit tests with the moves
* Skeleton of how Oop Plugins might work
* Oop Plugins are aligned with the OmniSharp stdio interface, so in theory, OmniSharp [dnxcore] could start OmniSharp [dnx451] to add MSBuild / ScriptCs support.

davidfowl:
Uh interesting 😀

david-driscoll:
Feedback is welcome. :surfer:

DustinCampbell:
This looks really promising. I'm trying to sort out how best to factor our OmniSharp in VS Code to avoid bringing along the "kitchen sink". Having a supported plug-in model would make that easier.

What is your thinking around deployment? Would each OmniSharp plug-in be a separate NuGet package?

The code changes thus far look reasonable. Where could I help out?

david-driscoll:
So I hadn't yet got to deployment, but in reality it would be nice if we could ship them as NuGet Packages.

What I have so far in the PR is:

* `OmniSharp.MsBuild`
  * Depends on
    * `OmniSharp.Abstractions`
    * `OmniSharp.Roslyn.CSharp`
* `OmniSharp.Dnx`
  * Depends on
    * `OmniSharp.Abstractions`
    * `OmniSharp.Roslyn.CSharp`
* `OmniSharp.ScriptCs`
  * Depends on
    * `OmniSharp.Abstractions`
    * `OmniSharp.Roslyn.CSharp`

Right now `OmniSharp` then depends on all of those, so they get loaded up by Dnx.  Today we can pick and choose which ones to load into MEF, but I would prefer something where we could dynamically load the assemblies, based on what we send to the server.  The editor, might know that it's dealing with a dnx or msbuild solution, so we only really need to load in the specific solution.  There could also be mixed solutions, with C#, ScriptCs and F# at the same time (theoretically) so we'd need to be able to load more than one plugin system at a time as well.

What might be interesting is if we could somehow load in the NuGet package inside of `Program` before we bootstrap `Startup`.

This is just my thoughts on what we could possibly do.  Basically add in command line options for where we can get plugins from.  They could be a directory off disc or it could be a NuGet feed or we could default to nuget.org.  If using nuget, restore those packages into the OmniSharp install location itself.  If it's a directory on disc, hopefully we could just leave them in place.  Now the problem is once we restore them, I have _no_ idea how to get (or if it's possible) Dnx to side load those additional nuget packages, since they won't be part of the original project[.lock].json.  Or is there something we can add to the `HostingContext` that would let us add additional assemblies?

We _could_ have an OmniSharp bootstrapper, that does the nuget magic and then writes a project.json out.  That still seems like a lot of work, and would add an extra process to proxy through.

david-driscoll:
Also thoughts, on breaking out the plugin distribution / loading into another issue?

That way we get what we have today, working well, get that merged in.  Then deal with how do we best load in additional user plugins.   Ideally with this merged in, existing editors would not have to make any changes to consume the refactored server.

nosami:
Yeah... plugin distribution should be a separate issue.

src/OmniSharp.Abstractions/project.json
```diff
@@ -0,0 +1,54 @@
...
+    "Microsoft.Framework.Runtime.Interfaces": "1.0.0-beta4",
+    "Microsoft.Framework.Logging.Console": "1.0.0-beta4",
+    "Microsoft.Framework.Caching.Memory": "1.0.0-beta4",
+    "Microsoft.Framework.FileSystemGlobbing": "1.0.0-beta4",
+    "Microsoft.Composition": "1.0.30"
+  },
+  "frameworks": {
+    "dnx451": {
+      "frameworkAssemblies": {
+        "System.Runtime": "4.0.0.0",
```
davidfowl: Get rid of these versions

src/OmniSharp.Abstractions/project.json
```diff
@@ -0,0 +1,54 @@
...
+        "System.Linq": "4.0.0.0",
+        "System.Text.Encoding":"4.0.0.0",
+        "System.IO":"4.0.0.0",
+        "System.Reflection":"4.0.0.0",
+        "System.Security":"4.0.0.0"
+      }
+    },
+    "dnxcore50": {
+      "dependencies": {
+        "System.Runtime": "4.0.20-beta-22816",
```
davidfowl: We should make these versions consistent

src/OmniSharp.Plugins/OopResponse.cs
```diff
@@ -0,0 +1,47 @@
+using System;
+using Newtonsoft.Json.Linq;
+
+namespace OmniSharp.Plugins
+{
+    class OopResponse
```
davidfowl: Change the name of this to plugin response

src/OmniSharp.Plugins/OutOfProcessPlugin.cs
```diff
@@ -0,0 +1,131 @@
...
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using OmniSharp.Models.v1;
+using OmniSharp.Services;
+using OmniSharp.Stdio.Protocol;
+using OmniSharp.Stdio.Services;
+
+namespace OmniSharp.Plugins
+{
+    public class OutOfProcessPlugin : IProjectSystem, IDisposable
```
davidfowl: All plugins are out of process right? Call it a plugin?

src/OmniSharp.Plugins/OutOfProcessPlugin.cs
```diff
@@ -0,0 +1,131 @@
...
+        {
+            while (!_cancellation.IsCancellationRequested)
+            {
+                var line = await _process.StandardOutput.ReadLineAsync();
+                if (line == null)
+                {
+                    break;
+                }
+
+                var ignored = Task.Factory.StartNew((Action)(() =>
```
davidfowl: Why are you casting to an action.

src/OmniSharp.Plugins/OutOfProcessPlugin.cs
```diff
@@ -0,0 +1,131 @@
...
+                            _writer.WriteLine(new EventPacket()
+                            {
+                                Event = "error",
+                                Body = response.Message,
+                            });
+                            return;
+                        }
+
+                        Action<string> requestHandler = null;
+                        if (!_requests.TryGetValue((int)response.Request_seq, out requestHandler))
```
davidfowl: Why are you casting an int to and int?

src/OmniSharp.Plugins/OutOfProcessPlugin.cs
```diff
@@ -0,0 +1,131 @@
...
+                        {
+                            throw new ArgumentException("invalid seq-value");
+                        }
+
+                        requestHandler((string)response.BodyJson);
```
davidfowl: Same question here. Isn't BodyJson already a string?

src/OmniSharp.Plugins/OutOfProcessPlugin.cs
```diff
@@ -0,0 +1,131 @@
...
+        }
+
+        public void Dispose()
+        {
+            _cancellation.Cancel();
+        }
+
+        public void Initalize(IConfiguration configuration)
+        {
+            Task.Run(() => Run());
```
davidfowl: Fire an Forget for a specific reason?

src/OmniSharp.Roslyn.CSharp/Services/FindSymbolsService.cs
```diff
@@ -1,18 +1,27 @@
...
+    [OmniSharpEndpoint(typeof(RequestHandler<FindSymbolsRequest, QuickFixResponse>), LanguageNames.CSharp)]
+    public class FindSymbolsService : RequestHandler<FindSymbolsRequest, QuickFixResponse>
     {
-        [HttpPost("findsymbols")]
-        public async Task<QuickFixResponse> FindSymbols(FindSymbolsRequest request = null)
+        private OmnisharpWorkspace _workspace;
+
+        [ImportingConstructor]
+        public FindSymbolsService(OmnisharpWorkspace workspace) {
+_workspace = workspace;
```
davidfowl: Formatting?

src/OmniSharp.Roslyn.CSharp/Services/GotoDefinitionService.cs
```diff
@@ -1,20 +1,30 @@
...
-        public async Task<GotoDefinitionResponse> GotoDefinition(GotoDefinitionRequest request)
+        private OmnisharpWorkspace _workspace;
+
+        [ImportingConstructor]
+        public GotoDefinitionService(OmnisharpWorkspace workspace)
+        {
+            _workspace = workspace;
+        }
+
+        async Task<GotoDefinitionResponse> RequestHandler<GotoDefinitionRequest, GotoDefinitionResponse>.Handle(GotoDefinitionRequest request)
```
davidfowl: private? Public? internal?

src/OmniSharp/Api/v1/Navigation/OmnisharpController.FindUsages.cs
```diff
@@ -15,41 +15,48 @@ public partial class OmnisharpController
         [HttpPost("findusages")]
         public async Task<QuickFixResponse> FindUsages(FindUsagesRequest request)
         {
-            var document = _workspace.GetDocument(request.FileName);
```
davidfowl: Is this an unrelated change to the plugin model changes?

src/OmniSharp/Api/v1/ProjectSystem/OmnisharpController.ProjectSystem.cs
```diff
@@ -1,66 +1,83 @@
...
-        public WorkspaceInformationResponse ProjectInformation(ProjectInformationRequest request)
+        public async Task<WorkspaceInformationResponse> ProjectInformation(ProjectInformationRequest request)
         {
-            return new WorkspaceInformationResponse
+            var tasks = new List<Task<Action>>();
+            var response = new WorkspaceInformationResponse();
+
+            foreach (var projectSystem in _projectSystems)
+            {
+                tasks.Add(projectSystem.GetInformationModel(request)
```
davidfowl: Why are you using continue with?

src/OmniSharp/Api/v1/ProjectSystem/OmnisharpController.ProjectSystem.cs
```diff
@@ -1,66 +1,83 @@
...
-            DnxProject dnxProjectItem = null;
+            var tasks = new List<Task<Action>>();
+            var response = new ProjectInformationResponse();
 
-            if (msBuildContextProject != null)
+            foreach (var projectSystem in _projectSystems)
             {
-                msBuildProjectItem = new MSBuildProject(msBuildContextProject);
+                tasks.Add(projectSystem.GetProjectModel(request.FileName)
+                    .ContinueWith((project) =>
```
davidfowl: Same question here.

src/OmniSharp/Middleware/Endpoint/EndpointHandler.cs
```diff
@@ -0,0 +1,281 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNet.Http;
+using Microsoft.Framework.Logging;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using OmniSharp.Mef;
+using OmniSharp.Models;
+using OmniSharp.Plugins;
+
+namespace OmniSharp.Middleware.Endpoint
```
davidfowl: Way too many classes in this file. Break it up.

src/OmniSharp/Middleware/EndpointMiddleware.cs
```diff
@@ -0,0 +1,78 @@
...
+using Newtonsoft.Json;
+using OmniSharp.Mef;
+using OmniSharp.Middleware.Endpoint;
+using OmniSharp.Models;
+using OmniSharp.Plugins;
+using OmniSharp.Services;
+
+namespace OmniSharp.Middleware
+{
+    // You may need to install the Microsoft.AspNet.Http.Abstractions package into your project
```
davidfowl: Yes this is true :smile:

src/OmniSharp/Startup.cs
```diff
@@ -119,11 +116,70 @@ public static OmnisharpWorkspace CreateWorkspace()
             return new OmnisharpWorkspace(MefHostServices.Create(assemblies));
         }
 
+        public static CompositionHost ConfigurePluginHost(IServiceProvider serviceProvider,
+                              OmnisharpWorkspace workspace,
```
davidfowl: Fix formatting

src/OmniSharp.Abstractions/project.json
david-driscoll: done, pinned `22816` for now.

src/OmniSharp.Plugins/OutOfProcessPlugin.cs
david-driscoll: Done on all counts.

src/OmniSharp.Roslyn.CSharp/Services/GotoDefinitionService.cs
david-driscoll: Was implemented explicitly, moved to public.

src/OmniSharp/Api/v1/Navigation/OmnisharpController.FindUsages.cs
david-driscoll: This was an odd duck... I was getting a new exception, it appears to be related to something in Roslyn itself.  It fails on both the current version of the server and the PR, I'll roll this back.

src/OmniSharp/Api/v1/ProjectSystem/OmnisharpController.ProjectSystem.cs
david-driscoll: I was concerned that there could be a threading issue, adding the items into the dictionary.

src/OmniSharp/Middleware/EndpointMiddleware.cs
david-driscoll: copy and paste fail :sheep:

david-driscoll:
holycrap! All tests passed on Travis finally! (the tests are flaky on windows, and I have no idea why, but they are currently flaky on appveyor as well)

cc @davidfowl @DustinCampbell @nosami

david-driscoll:
Merged with master to bring things up to date, also fixed an intellisense bug that has been bothering me.

src/OmniSharp.Abstractions/Models/v1/FilesChangedRequest.cs
```diff
@@ -0,0 +1,8 @@
+using System.Collections.Generic;
+using OmniSharp.Mef;
+
+namespace OmniSharp.Models
+{
+    [OmniSharpEndpoint("/filesChanged", typeof(IEnumerable<Request>), typeof(object), TakeOne = true)]
```
nosami: Feels wrong to have this metadata here (and the other requests)

src/OmniSharp.Roslyn.CSharp/Services/Buffer/ChangeBufferService.cs
```diff
@@ -0,0 +1,26 @@
+using System.Composition;
+﻿using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
+using OmniSharp.Mef;
+using OmniSharp.Models;
+
+namespace OmniSharp.Roslyn.CSharp.Services.Buffer
+{
+    [OmniSharpHandler(typeof(RequestHandler<ChangeBufferRequest, object>), LanguageNames.CSharp)]
```
nosami: Seems wrong to have this twice `typeof(RequestHandler<ChangeBufferRequest, object>)` Can we use the route name here instead? `/changebuffer` etc.

david-driscoll: We could, but then we'd lose some of the runtime analysis that Mef can give us, since it validates that the ContractType (the handler) matches the type that the attribute is attached to.

src/OmniSharp.Abstractions/Models/v1/FilesChangedRequest.cs
david-driscoll: So this is worth some discussion.

I did have a static collection before and it worked fairly well, but then I got thinking about how we could have editors extend OmniSharp with extra features that don't necessarily fit as part of OmniSharp proper.

The big example I have in my head, that I'd like to take a stab at in the future, is Continuous Testing. Using the attribute based approach here lets us add arbitrary endpoints for the editor to call through a custom plugin that editor supports or ships with.

nosami:
What does isMergeable mean? https://github.com/OmniSharp/omnisharp-roslyn/blob/plugins/mef/src/OmniSharp/Middleware/Endpoint/EndpointHandler.cs#L163

Is that used for batching requests?

nosami:
EndpointHandler seems too complex to me although I could be overlooking something.

nosami:
What's isSpreadable for? https://github.com/OmniSharp/omnisharp-roslyn/blob/plugins/mef/src/OmniSharp/Middleware/Endpoint/EndpointHandler.cs#L154

src/OmniSharp.Roslyn.CSharp/Services/Buffer/ChangeBufferService.cs
nosami: There is no way to do this other than by adding redundant information?

david-driscoll: It _may_ be possible, but it would outside how Mef seems to do things typically.  In theory we could take the string, check it against the list of end points, and then reflect over the type to ensure that it implements the correct interface.
It _may_ be possible, but it would outside how Mef seems to do things typically.  In theory we could take the string, check it against the list of end points, and then reflect over the type to ensure that it implements the correct interface.

nosami:
Everything looks good to me except for EndpointHandler, which looks more complicated than it needs to be. Not sure why need to use reflection at all for this - MEF should handle that

Could you do something like this (pseudo code-ish)? I've mentioned this before but it got lost in the trail of dust :)
```cs

/// Handles every route / request
public class OmniSharpController<TRequest, TResponse>
{
     public OmniSharpController(Dictionary<routeinfo, IHandler> handlers)
    {
    }

    public TResult Handle(whatever routeInfo, TRequest request)
    {
        return _handlers[routeInfo].Handle(request);
    } 
}
```

nosami:
Having said that, if nobody else comments here you should just merge it anyway (assuming it works!)

david-driscoll:
I could break the endpoint handler into a set of static helper classes for some of the things.

* `_isSpreadable` probably could use a better name.  Basically there is a subset of requests that are special.
  * First there are mergable requests (below)
  * The other are requests that aren't specific to a language, `/project`, `/projects`, `/packagesearch` are examples of requests that don't really belong to a specific language.  So these get picked up and which even language returns a result first, is the response that is sent back.
`_isMergable` are for requests that can be stitched together, for things like getting all diagnostics, or finding symbols.  These will delegate to all plugins and then `IMergableResponse` allows everything to be merged for sending the final response to the client.

src/OmniSharp.Roslyn.CSharp/Services/Buffer/ChangeBufferService.cs
nosami: This would look worse in F# where you don't expect to see so much type cluttering.

nosami:
> The other are requests that aren't specific to a language, /project, /projects, /packagesearch are examples of requests that don't really belong to a specific language. So these get picked up and which even language returns a result first, is the response that is sent back.

This doesn't sound right to me. I think you need to know at request time enough information to know which handler will run.

src/OmniSharp.Roslyn.CSharp/Services/Buffer/ChangeBufferService.cs
david-driscoll: F# can also do this form, not sure if it's any easier to do in F# or not.

````
        [OmniSharpHandler(typeof(Func<UpdateBufferRequest, Task<object>>), LanguageNames.CSharp)]
        public Func<UpdateBufferRequest, Task<object>> UpdateBuffer { get; } = (request) => { return Task.FromResult<object>(true); };
```

david-driscoll:
I can see if I can factor them out, there have been some changes since it went in.

src/OmniSharp.Roslyn.CSharp/Services/Buffer/ChangeBufferService.cs
nosami: I'm not saying it can't be done, just that it would look weird.

In my head, every handler should implement IHandler and have a Handle method.

david-driscoll:
made a small change, and then took some time this weekend to get linux working in my VM, and using this branch everything seems to be working.

image

nosami:
Confirmed working with vim over HTTP

nosami:
Epic! :fireworks: :dancer: :dancers: :dash:

PR #295, opened 2015-09-05
Feature/add to project

mat-mcloughlin:
Just setting up this PR for adding in the `addtoproject` and `removefromproject` endpoints. 

Hoping for some initial feedback but don't pull it in yet


jchannon:
Where we at with this shizzle?

troydai:
@mat-mcloughlin is this change still needed?

willl:
@troydai I think the code is a bit outdated (and is possibly incomplete), but the change was to add/remove files to/from an MS Build based project. I think this is something users are still using MS Build projects are likely to still want.

filipw:
@mat-mcloughlin hey Matt now that there is a `dev` branch (see [here](https://github.com/OmniSharp/omnisharp-roslyn/blob/dev/CONTRIBUTING.md#workflow)) this PR will need to be re-opened against `dev` branch instead of `master`

so closing this one and once you rebase please re-open against `dev`, thanks

PR #302, opened 2015-09-18
Added /stopserver

mispencer:
Works in stdio, unlike https://github.com/OmniSharp/omnisharp-roslyn/pull/190.

Run needs to returns for RequestShutdown to shutdown the server in stdio mode.

RequestShutdown is requested in the background to avoid an error which sometime occurs if it is ran in an action.

### DIFF
--- a/src/OmniSharp.Stdio/StdioServer.cs
+++ b/src/OmniSharp.Stdio/StdioServer.cs
@@ -26,40 +26,43 @@ namespace OmniSharp.Stdio
             _next = next;
             _cancellation = new CancellationTokenSource();
 
-            var ignored = Run();
+            Run();
         }
 
-        private async Task Run()
+        private void Run()
         {
-            _writer.WriteLine(new EventPacket()
+            Task.Factory.StartNew(async () =>
             {
-                Event = "started"
-            });
-
-            while (!_cancellation.IsCancellationRequested)
-            {
-                var line = await _input.ReadLineAsync();
-                if (line == null)
+                _writer.WriteLine(new EventPacket()
                 {
-                    break;
-                }
-
-                var ignored = Task.Factory.StartNew(async () =>
-                {
-                    try
-                    {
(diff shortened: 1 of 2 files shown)


src/OmniSharp/Api/OmnisharpController.cs
```diff
@@ -9,11 +12,23 @@ public partial class OmnisharpController
     {
         private readonly OmnisharpWorkspace _workspace;
         private readonly OmniSharpOptions _options;
+        private readonly IApplicationShutdown _applicationShutdown;
 
-        public OmnisharpController(OmnisharpWorkspace workspace, IOptions<OmniSharpOptions> optionsAccessor)
+        public OmnisharpController(OmnisharpWorkspace workspace, IOptions<OmniSharpOptions> optionsAccessor, IApplicationShutdown applicationShutdown)
```
david-driscoll: It would be awesome if you could move this into it's own controller/class (and own file) similar to [ProjectSystem](https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp/Api/v1/ProjectSystem/OmnisharpController.ProjectSystem.cs).

1) you then you wouldn't have to change all the tests by inserting the FakeApplicationShutdown).
2) It would then be _much_ easier to merge into the [Plugins branch](https://github.com/OmniSharp/omnisharp-roslyn/tree/plugins/mef) which is now one class per one endpoint.

david-driscoll:
Do we need any unit tests against it?  (I'm not sure how easy/hard/useful that would be though)

mispencer:
Okay, it's been made its own controller.

I could write a test if desired, but I don't know how useful it would be.

nosami:
No need for a test here.

david-driscoll:
rock on, works for me, I'll add it to the client soon :pizza: 

image

thinca:
:+1:

PR #304, opened 2015-09-29
Sriramgd fixusing

sriramgd:
Fixed ambiguous using order, added linq query syntax support.


nosami:
rejected in favour of #305

PR #305, opened 2015-09-29
Fixusings

nosami:


nosami:
continuation from #304

nosami:
@sriramgd I don't have time to test this... but if you're sure that the ambiguous locations are now correct, I'll merge

nosami:
not tested, but looking at this code, it doesn't appear to be correct

```cs
var ambiguous = await GetAmbiguousUsings();
await AddMissingUsings();
await RemoveUsings();
await SortUsings();
await TryAddLinqQuerySyntax();
```

the ambiguous locations are going to be incorrect after adding and removing using statements?

nosami:
:cocktail: :dancer: Yay! Thanks @sriramgd :)

PR #308, opened 2015-10-03
Always launch DTH using mono (on non windows)

davidfowl:
- Today we choose whatever default the user has as the thing to launch for *everything*. This is normally ok but DTH doesn't work on coreclr cross platform as yet (because of missing socket support). This change fixes up the runtime to be mono before launching the design time host. This change can be removed once things work with core clr again.

### DIFF
--- a/src/OmniSharp/Dnx/DnxSdk.cs
+++ b/src/OmniSharp/Dnx/DnxSdk.cs
@@ -1,0 +1,184 @@
+
+using System;
+using System.Text;
+
+namespace OmniSharp.Dnx
+{
+    public class DnxSdk
+    {
+        public static string GetRuntimeNameFromFullPath(string runtimePath)
+        {
+            // Treat the folder name as the runtime name
+            // ~/.dnx/runtimes/{runtimename}/
+
+            var slash = runtimePath.TrimEnd('/').LastIndexOf('/');
+
+            return runtimePath.Substring(slash + 1);
+        }
+
+        public static string GetFullName(string version, string flavor, string os, string arch)
+        {
+            return GetRuntimeName(flavor, os, arch) + $".{version}";
+        }
+
+        public static bool TryParseFullName(string fullName, out string flavor, out string os, out string arch, out string version)
+        {
+            flavor = null;
+            os = null;
+            arch = null;
+            version = null;
+
+            var tokenBuilder = new StringBuilder();
+
+            // e.g. dnx-clr-win-x64.1.0.0-dev
+            // e.g dnx-mono.1.0.0-dev
+            var state = DnxNameState.Dnx;
+            foreach (var ch in fullName)
+            {
+                if (ch == '-')
+                {
+                    if (state == DnxNameState.Dnx)
+                    {
+                        var runtime = tokenBuilder.ToString();
+
+                        if (!string.Equals(runtime, "dnx", StringComparison.OrdinalIgnoreCase))
+                        {
+                            return false;
+                        }
+
+                        state = DnxNameState.Flavor;
+                        tokenBuilder.Clear();
+                    }
+                    else if (state == DnxNameState.Flavor)
+                    {
+                        flavor = tokenBuilder.ToString();
+
+                        if (!IsValidFlavor(flavor))
+                        {
+                            return false;
+                        }
+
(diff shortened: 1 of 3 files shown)


src/OmniSharp/Dnx/DnxSdk.cs
```diff
@@ -0,0 +1,184 @@
...
+
+            return runtimePath.Substring(slash + 1);
+        }
+
+        public static string GetFullName(string version, string flavor, string os, string arch)
+        {
+            return GetRuntimeName(flavor, os, arch) + $".{version}";
+        }
+
+        public static bool TryParseFullName(string fullName, out string flavor, out string os, out string arch, out string version)
```
davidfowl: I hate regex that much

src/OmniSharp/Dnx/DnxSdk.cs
```diff
@@ -0,0 +1,184 @@
...
+using System.Text;
+
+namespace OmniSharp.Dnx
+{
+    public class DnxSdk
+    {
+        public static string GetRuntimeNameFromFullPath(string runtimePath)
+        {
+            // Treat the folder name as the runtime name
+            // ~/.dnx/runtimes/{runtimename}/
```
davidfowl: May or may not have a trailing slash

src/OmniSharp/Dnx/DnxSdk.cs
tugberkugurlu: and we love you for that reason :heart:

nosami: This is actually nuts. Regex is the best tool for the job here. Would reduce to one line and be easier to read too.

davidfowl: Nah too many special cases (actually just one, mono). If you send me the regex I'll use it

david-driscoll: ScriptCs code ahead

```csharp
using System;
using System.Text.RegularExpressions;
using System.Linq;

var monoRegex = new Regex("dnx-(?:(?<flavor>mono)|(?<flavor>(?:clr)|(?:coreclr))-(?<os>.*?)-(?<arch>.*?))\\.(?<version>.*)");

var tests = new[] {
    "dnx-clr-win-x86.1.0.0-dev",
    "dnx-clr-win-x64.1.0.0-dev",
    "dnx-coreclr-win-arm.1.0.0-dev",
    "dnx-coreclr-win-x86.1.0.0-dev",
    "dnx-coreclr-win-x64.1.0.0-dev",
    "dnx-coreclr-linux-x64.1.0.0-dev",
    "dnx-coreclr-darwin-x64.1.0.0-dev",
    "dnx-coreclr-darwin-x64.1.0.0-dev",
    "dnx-mono.1.0.0-dev"
};

foreach (var test in tests)
{
    var result = monoRegex.Match(test);

    if (result.Success)
    {
        Console.WriteLine(test);
        Console.Write(" Flavor: " + result.Groups["flavor"].Value);
        Console.Write(" OS: " + result.Groups["os"].Value);
        Console.Write(" Arch: " + result.Groups["arch"].Value);
        Console.Write(" Version: " + result.Groups["version"].Value);
        Console.WriteLine(string.Empty);
    }
}

```

david-driscoll: Which results in

image

davidfowl: holy regex batman.

nosami: :)

davidfowl: Can we change this after the merge? That regex hurts my brain even tho I can see it works :smile:

nosami: Sure. I think the regex can be simplified though

frapontillo:
Awesome, can't wait for the next beta!

nosami:
@frapontillo You should be able to get it working now in theory. Clone omnisharp-roslyn somewhere, build with build.sh or build.cmd..... set OMNISHARP environment variable to /path/to/cloned/folder, launch vscode or $EDITOR

Issue #287, opened 2015-08-12
Don't show <global namespace> in type lookup


david-driscoll:
see https://github.com/OmniSharp/omnisharp-atom/issues/275

PR #310, opened 2015-10-05
Do not show <global namespace> in type lookup

filipw:
Fixes https://github.com/OmniSharp/omnisharp-roslyn/issues/287
Fixes https://github.com/OmniSharp/omnisharp-atom/issues/275


nosami:
On a roll :)

Issue #266, opened 2015-07-14
formattingOptions.tabSize does not work


yilei:
I'm using OmniSharp-Atom, I set Atom's indent size to two spaces, and changed the formattingOptions.tabSize to two in ~/.atom/packages/omnisharp-atom/node_modules/omnisharp-client/roslyn/packages/OmniSharp/1.0.0/root/config.json

But when I type ';' at the end of a line, it auto-indents my code using 4 spaces.

Filing this as suggested by https://twitter.com/jasonimison/status/620833046992916480

solarsailer:
Same here (using Atom).

I was using Omnisharp-Sublime before, and one way to change the configuration was to edit a config.json file ([see here](https://github.com/OmniSharp/omnisharp-sublime#omnisharpserver-settings)).

Is there a way to change the config like this with Roslyn? A dotfile or a setting in Atom to target a custom configuration? It's not just for the indent, though. I want to select my warnings, my code format, etc. :)

Figbash:
Pretty sure this is the same issue with hard tabs.  If you turn off soft tabs in atom prefs and work on a file with hard tabs, omnisharp will replace the tabs with spaces whenever you add the ; at the end of a line.  There is an option in config.json called "useTabs" but changing it does nothing.

shadowmint:
Hm... you've closed this issue, but how do you actually *set* these settings? They're not exposed in the omnisharp settings.

Do you literally just edit the config.json deeeeeeeeeep within the OmniSharp package and redo it everytime you install an update?

david-driscoll:
I'll see if I can work on the ui fairly soon. I think that there is a
config file you can have in your solution directory that will have the same
behavior as config.json
On Wed, Oct 14, 2015 at 5:49 AM Doug <notifications@github.com> wrote:

> Hm... you've closed this issue, but how do you actually *set* these
> settings? They're not exposed in the omnisharp settings.
>
> Do you literally just edit the config.json deeeeeeeeeep within the
> OmniSharp package and redo it everytime you install an update?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/OmniSharp/omnisharp-roslyn/issues/266#issuecomment-147993739>
> .
>
-- 
Thanks,
David

PR #311, opened 2015-10-05
Added IndentationSize to formatting options

filipw:
Fixes https://github.com/OmniSharp/omnisharp-roslyn/issues/266

Currently, omnisharp options expose `tabSize` which some of the users have been trying to use to control indentation (see https://github.com/OmniSharp/omnisharp-roslyn/issues/266).

However, it's only applicable to code formatting with Roslyn (and indeed it works with omnisharp now) when tabs are enabled - to determine the size of the tab. 

For regular indentation, Roslyn exposes another setting called `Microsoft.CodeAnalysis.Formatting.FormattingOptions.IndentationSize` which is the principal way of controlling how large the indentation should be.


nosami:
Very cool :)

PR #312, opened 2015-10-06
correctly flow formatting options

filipw:
I noticed another bug in formatting code - the formatting options are not flown correctly into `FormatAsync` when using `GetFormattingChangesForRange`.

On the other hand, when using `GetFormattedDocument`, the options are applied ([see here](https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp/Workers/Format/Formatting.cs#L80)).

This looks like a clear bug, because when formatting **the entire document**, the options (i.e. the new indentation size I added yesterday or use of tabs) are applied as expected, but on stroke or range formatting they are not falling back to the defaults, resulting in two different types of behavior in the document :)

This PR fixes it and correctly applies formatting also on stroke/range:
format

### DIFF
--- a/src/OmniSharp/Options/FormattingOptions.cs
+++ b/src/OmniSharp/Options/FormattingOptions.cs
@@ -2,6 +2,15 @@ namespace OmniSharp.Options
 {
     public class FormattingOptions 
     {
+        public FormattingOptions()
+        {
+            //just defaults
+            NewLine = "\n";
+            UseTabs = false;
+            TabSize = 4;
+            IndentationSize = 4;
+        }
+
         public string NewLine { get; set; }
         
         public bool UseTabs { get; set; }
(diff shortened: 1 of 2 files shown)


src/OmniSharp/Options/FormattingOptions.cs
```diff
@@ -2,6 +2,15 @@ namespace OmniSharp.Options
 {
     public class FormattingOptions 
     {
+        public FormattingOptions()
+        {
+            //just defaults
```
filipw: the defaults I here because when we now flow the options into formatter, it was using the default state of this object - so indentation was 0 which was causing some tests to fail

nosami:
Looks good to me. Thanks!

filipw:
i think that's 4 PRs in 12h :)

nosami:
Want to make it 5 and fix format on Enter key? :)

filipw:
what's the problem with formatting on enter?

nosami:
doesn't work properly at all, so it isn't even enabled.  Needs to handle cases like these (| is the cursor) `{|}`,  `{ someText|}`, `(x => |)` etc

nosami:
@jrieken worked on this originally, but it was never completed. Maybe he can provide a better answer. 
I think the tricky part was finding the correct indentation for a given block of code. 

There is some code to do this in Roslyn's features assembly, but internal.

jrieken:
Yes, the problem is that format on enter will remove indent, event for the line you have just created by hitting enter. So the trick is to track the cursor somehow and to figure out the right indent...

PR #318, opened 2015-10-23
Fixed up coding guidelines after the plugins merge, this lets the tes…

david-driscoll:
…t show all files that are failing before asserting so that they can all be fixed before the next run


nosami:
thanks!

PR #320, opened 2015-10-27
Add Visual Studio projects where they're missing

SamB:
This happened as a result of opening and building the solution.


nosami:
Personally, I'd prefer it the xproj files were added to .gitignore. They are only needed by VS and are generated on demand like you've shown :)

SamB:
@nosami: VS does not like it when I delete all of the .xproj files that the solution tells it should be there, and does not appear to be able to recreate them.  (Build does nothing; it does not even add "Build" to the "Show output from:" dropdown in the "Output" pane.)

If I bring back just `OmniSharp.xproj`, Build then works, but does not recreate the rest of the .xproj files.

So, I'm afraid your idea will not work very well.  I did notice some EOL issues in the originally submitted commit, though, so I'm revising it to fix that.

nosami:
Oh. OK. I'm mistaken then. I don't use VS. I thought I heard they were
generated.

nosami:
Thanks!

SamB:
@nosami: And yeah, I figured you didn't use VS. Given the purpose of OmniSharp, that isn't particularly surprising ;-).

davidfowl:
xproj magic

PR #321, opened 2015-10-30
Plugin bootstrap

david-driscoll:
This is a little utility that is inserted as part of startup, that does the following:

* Captures any defined plugins
  * `--plugins` adds a folder of plugins
  * `--plugin-name` loads a plugin via nuget
* Creates a temporary `project.json` file, with all the plugins as dependencies.
* Runs `dnu restore` on the temporary `project.json` file.
  * This lets us load plugins by name, or from disk as dlls/source
* Returns the path to the temporary `project.json` so that it can be started.

Still need to test on my Linux VM.

cc @davidfowl @dustincampbell is there a better way to go about this, perhaps getting dnx to load assemblies.  I know we had talked about this before and there were many different issues floated with dependencies getting loaded, etc.   This seems to be the best of both worlds at the moment, we can easily load packages with dnu restore, for example, once we publish them to nuget.


nosami:
Like we discussed, you don't need `dnu` or a bootstrap script to load plugins.

david-driscoll:
My only concern is, with this solution we get a ton of free wins.

1. Plugin distribution via nuget, we can just treat them as class libraries.
2. User defined plugins, editors can let users define plugin drop folders. We can offer an easy plugin drop folder (`.omnisharp` like vscode), or even a plugins array property on `omnisharp.json`, and then they just load via nuget.

nosami:
Can an end user install a plugin by dropping a dll(s) / something else into a ./plugins folder? Without source available?

nosami:
My concern is that none of this is necessary, and running `dnu restore` at startup sounds slooow. It already takes an age to start up.

david-driscoll:
Those are fair comments, and I've worked on ways of improving the repeat startup time (only restore once) which is helping.  I envisioned this in about 30 minutes and got a proof of concept working in a few hours, I didn't spend days toiling away at the problem. :zap: :bulb:  

I'm not against a different solution, it just seems like then we're forwarding the pain onto anyone that wants to make a plugin, they have to package it in a special way (perhaps differently than how `dnu publish` would behave).  Then we have to deal with how to load them (download, folder scan, unzip, whatever), and add in all that work (not against it, just this POC started working right away).

I'd still love to have a conversation with with @davidfowl and @dustincampbell to see what alternatives we have to loading dlls with dnx.  Me and @davidfowl are apparently playing tag at the moment, lol.  


Perhaps we need another Microsoft meeting?, it's been a few months since the last time.
cc @sayedihashimi @shanselman

PR #322, opened 2015-11-01
Ignore omnisharp.tar.gz

thinca:


Yantrio:
is it maybe better to ignore the entire artifacts directory?

nosami:
@Yantrio the artefacts folder is already ignored. This file ends up in the root folder.

@thinca Thanks!

davidfowl:
It should probably end up in artifacts

nosami:
Good point

PR #329, opened 2015-11-06
[Don't merge yet]: Fix strong naming support in dnx projects

davidfowl:
- Flow compilation options to the workspace

There's a missing feature with respect to strong naming support for dnx projects. VsCode is still based off 1.5.5 so I made a branch to enable fixing this specific thing. VSCode can't take 1.6.0 because it still has a bunch of bugs that need to be worked out (I've filed some myself). 

This is mostly a for code review. I'll send out another one based on 1.6.0 soon. The port should be trivial.

/cc @jrieken


natemcmaster:
:+1:

PR #335, opened 2015-11-10
Fix string escaping on exception messages for Stdio

ianbattersby:
Stdio passes exception messages through as an `error` event-type but fails to escape the string, which causes an error on the client-side when parsing the JSON.


david-driscoll:
Awesome!

Issue #341, opened 2015-11-21
Support Protobuild modules


hach-que:
I'm planning on submitting a PR for this.

[Protobuild](http://protobuild.org/) is a cross-platform project generator for .NET.  It generates the appropriate C# project and solution files for any platform, from any platform.  It is used by projects like [MonoGame](http://monogame.net/), where it's not practical to perform platform checks at runtime for either performance or complexity reasons.

I want to merge in support for Protobuild modules into OmniSharp so that they work in Visual Studio Code without users having to run the Protobuild executable.

PR #342, opened 2015-11-21
Initial support for Protobuild modules

hach-que:
Addresses #341.  If you open a folder with Protobuild.exe inside it, it now automatically runs Protobuild with the host platform chosen.  When users edit definition files, Protobuild is automatically re-run.

One thing I'm not sure how to address yet is prompting the user what platform they want to generate for.  I don't know if there's an API to interact with the editor in such a way to prompt the user for an option.


david-driscoll:
This will probably be something that's more appropriate as a plugin, which is something we're building today.

hach-que:
There are some things in here which require modification to the MSBuild project system (so that it doesn't run twice, because Protobuild puts C# projects and solutions on disk).  So I guess that's my only concern is that the plugin will need those modifications to be present in upstream in order to work properly.

filipw:
This should be closed for now as all new PR should go against `dev` branch (https://github.com/OmniSharp/omnisharp-roslyn/blob/dev/CONTRIBUTING.md#workflow) - plus there is a question whether this belongs in the core (probably doesn't and should become a plugin).

Feel free to open a discussion about Protobuild in a separate issue or on Slack/Jabbr. Thanks

Issue #356, opened 2015-12-01
NotSupportedException on Emacs: [Microsoft.AspNet.Diagnostics.ErrorHandlerMiddleware] An unhandled exception has occurred: Must be able aggregate the response to spread them out across all plugins for /updatebuffer


dgalizzi:
First: It's working on omnisharp-server but not with omnisharp-roslyn. I'm manually running the server. The function for checkalivestatus works but others don't (I tried autocomplete and /typelookup).

I tried debugging it, a few things I found:
* The exception thrown is at middleware.endpoint.endpointhandler.cs:170
* Same file on line 129 is returning null (model.FileName is fine):

> var language = _languagePredicateHandler.GetLanguageForFilePath(model.FileName ?? string.Empty); 

* Tried forcing language to LanguageNames.CSharp but then I got another exception: error : 

>[Microsoft.AspNet.Diagnostics.ErrorHandlerMiddleware] An unhandled exception has occurred: C# does not support /updatebuffer System.NotSupportedException: C# does not support /updatebuffer  at OmniSharp.Middleware.Endpoint.EndpointHandler`2.<HandleSingleRequest>d__18.MoveNext() in [...]\omnisharp-roslyn\src\OmniSharp\Middleware\Endpoint\EndpointHandler.cs:line 163

I tried on both windows and linux, got the same. I'm using the latest code on master branch. I'm using emacs with the spacemacs configuration and the csharp layer which is using omnisharp 20150709.1045.

Thanks!

david-driscoll:
cc @ianbattersby Have any idea about this issue?  I know you've been working on emacs a little bit lately.

ianbattersby:
I suspect this might relate to the issue see before with the newer Roslyn branch that when emacs doesn't send the file name the server logic differs because it can't establish the language from the file extension.

I'll try and dig that out for reference. I'm likely not seeing this myself because my in-development version sends the file name with every request.

I'll try and get some time tomorrow to play with the current master branch and see if I can replicate however, rather than hypothising :)

ianbattersby:
https://github.com/OmniSharp/omnisharp-roslyn/issues/324#issuecomment-153787076

david-driscoll:
What request is trying to make where it's not setting a filename?

To resolve this one, I'm not against defaulting the language to C#, instead of throwing an exception, I just want to make sure it'll work.

dgalizzi:
[...]

PR #358, opened 2015-12-01
Modified OmniSharp to work as it used to to avoid breaking emacs and vim

david-driscoll:


dgalizzi:
Tried this by running the Omnisharp.cmd on windows and it solves #356 . But tried the same (Omnisharp.sh) on linux and didn't (doesn't throw the exception but I'm not getting any response).

dgalizzi:
Got it to work on linux! I added -s [path to folder] and worked. Still I didn't have to do that in Windows or with omnisharp-server.

david-driscoll:
Good to hear... I'll see if I can make some changes so that it won't break vim or emacs

david-driscoll:
rather doesn't break them, without any modifications to those plugins

david-driscoll:
@dgalizzi If you get a chance to test out this branch again, instead this time, hopefully you shouldn't have to make any changes to the emacs config to make it work.

davidfowl:
?

PR #362, opened 2015-12-11
Design Time Host changes to improve socket connection and monitoring

wolfspider:
Original socket connection was exceeding it's timeout checking the connection before the socket was bound. Socket status is checked and logged asynchronously.

### DIFF
--- a/src/OmniSharp.Dnx/Utilities/SocketConnection.cs
+++ b/src/OmniSharp.Dnx/Utilities/SocketConnection.cs
@@ -8,71 +8,71 @@ namespace OmniSharp
 {
     public class SocketConnection
     {
-        public ILogger Logger { get; private set; }
-        public string ContextFlag { get; private set; }
-        
+        public ILogger Logger { get; }
+        private string ContextFlag { get; }
+
         public bool dthConnectSuccess = false;
-        
-        public SocketConnection( Socket socket, int port, string contextFlag, ILogger logger )
+
+        public SocketConnection(Socket socket, int port, string contextFlag, ILogger logger)
         {
             Logger = logger;
             ContextFlag = contextFlag;
-            Initialize(socket, port);        
+            Initialize(socket, port);
         }
-        
-        private void Initialize( Socket socket, int port)
+
+        private void Initialize(Socket socket, int port)
         {
-                    
+
             socket.NoDelay = true;
             socket.SetSocketOption(SocketOptionLevel.Socket, SocketOptionName.KeepAlive, true);
-                    
+
             SocketAsyncEventArgs connect = new SocketAsyncEventArgs();
-                    
+
             connect.Completed += new EventHandler<SocketAsyncEventArgs>(SocketEventArg_Completed);
             connect.RemoteEndPoint = new IPEndPoint(IPAddress.Loopback, port);
             connect.UserToken = socket;
-                    
+
             Thread.Sleep(500);
-                try
-                {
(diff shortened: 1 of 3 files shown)


src/OmniSharp.Dnx/Utilities/SocketConnection.cs
```diff
@@ -0,0 +1,78 @@
...
+                   {
+                       Logger.LogError("Connection Timed Out");
+                       throw new SocketException((int)e.SocketError);
+                   }
+                  break;
+            }
+        }
+        
+}
+
```
troydai: Please format the code here.

src/OmniSharp.Dnx/Utilities/SocketConnection.cs
```diff
@@ -0,0 +1,78 @@
...
+using System.Net.Sockets;
+using System.Threading;
+using Microsoft.Framework.Logging;
+
+namespace OmniSharp
+{
+    public class SocketConnection
+    {
+        public ILogger Logger { get; private set; }
+        public string ContextFlag { get; private set; }
```
troydai: Both properties are initialized in constructor only. You can remove the `private set;` here.

troydai: Can `ContextFlag` be a private field only?

src/OmniSharp.Dnx/Utilities/SocketConnection.cs
```diff
@@ -0,0 +1,78 @@
...
+                    
+            Thread.Sleep(500);
+                try
+                {
+                socket.ConnectAsync(connect);                
+                }
+                catch (SocketException)
+                {
+                // this happens when the DTH isn't listening yet
+                Logger.LogVerbose("[SocketConnection]: Socket Connection failed");
```
troydai: What about LogError?

david-driscoll:
Travis is mean.

david-driscoll:
Travis is being mean, only because the travis.yml is wrong...  I'll fix that tonight, otherwise this looks okay after the feedback.

david-driscoll:
@wolfspider if you could pull in the travis change (it's now in master), it should then pass the build.

wolfspider:
@david-driscoll Thanks, I'll give it a try shortly

troydai:
@wolfspider @david-driscoll since we're moving towards `dotnet-cli` I'm wondering if we still want to take this change.

wolfspider:
@troydai from my perspective I was not able to use VSCode at all without many retries unless I patched this code in. I originally started looking into this because I found one other person who experienced the same problem while looking for answers. The connection time also speeds up when launching VSCode using this code. If the DTH is being replaced altogether in an upcoming release soon then I don't see a need for it however. If its not a common issue for other people then we probably don't need it as I was even expecting more occurrences of the issue to roll in. If there is going to be some time before dotnet-cli lands in VSCode which is the main entry point for the Mac users I think it would be good to have it included.

troydai:
@wolfspider thanks for the clarification. I'm fine with merge this. But DTH is going to be replaced altogether. However I am not able to comment on schedule. You can try out the `dev` branch and you will see it is fairly close.

@david-driscoll how do you like this PR?

filipw:
I think this PR is now obsolete following yesterday's merge of dotnet-cli

troydai:
@wolfspider I'll close this PR since we're moving on to the `dotnet` now. Thanks.

PR #370, opened 2015-12-18
Update the dependencies on ASP.NET and CoreCLR packages to rc2-*

troydai:
* A few tests are skipped and need some investigation
* Need some help with the `build.sh`.


david-driscoll:
@troydai found an issue getting omnisharp to start with the scripts in `/scripts` folder.  I've got some changes that should fix everything, and I'll be pushing those later today.

david-driscoll:
cc @troydai 

Any idea how to handle this error?

`failed to locate libcoreclr with error libunwind-x86_64.so.8: cannot open shared object file: No such file or directory`?

troydai:
This: https://github.com/aspnet/dnx/blob/dev/.travis.yml#L18
Install the `libunwind` package for Travis.

david-driscoll:
It's there, and installed, hence my confusion.  I double checked the build log and it installed successfully.

troydai:
Interesting, let me check the log.

troydai:
This is weird, cause missing `libunwin8` is the only known reason for this error. https://github.com/dotnet/coreclr/issues/984

troydai:
I suspect it could be a dnvm issue. Here's my thought. The build didn't fail on the first installation of coreclr DNX: https://github.com/OmniSharp/omnisharp-roslyn/blob/update-rc2/build.sh#L18. Instead if failed at the `dnvm use -r coreclr` right after restore.

troydai:
How about try change the `dnvm use 1.0.0-rc2-16337 -runtime coreclr` to `dnvm install 1.0.0-rc2-16337 -r coreclr`

troydai:
Let's see if it works. Otherwise, I'll try ping a few guys on the team in charge of *nix and dnvm.

CoreyKaylor:
I couldn't get my build scripts to work for coreclr + linux without matching the other settings to match the dnx projects. sudo: required, mono: 4.0.5, etc.

CoreyKaylor:
https://github.com/aspnet/dnx/blob/dev/.travis.yml#L2

CoreyKaylor:
It didn't make a lot of sense to me, but it did work after I changed everything else to match.

troydai:
@BrennanConroy do you know why it fails to install dnx coreclr on linux/mac in this case for Travis?

BrennanConroy:
Yes, you are trying to install an rc2 version from nuget https://github.com/OmniSharp/omnisharp-roslyn/pull/370/files#diff-0b83f9dedf40d7356e5ca147a077acb4R21

troydai:
@BrennanConroy in previous case, we tried `dnvm use 1.0.0-rc2-16337 -r coreclr` it throws the same exception.

src/OmniSharp.MSBuild/Analyzers/SimpleAnalyzerAssemblyLoader.cs
```diff
@@ -0,0 +1,21 @@
+﻿#if DNX451
+using System;
+using System.Reflection;
+using Microsoft.CodeAnalysis;
+
+namespace OmniSharp.MSBuild.Analyzers
+{
+    public class SimpleAnalyzerAssemblyLoader : IAnalyzerAssemblyLoader
```
davidfowl: ?

troydai: I didn't find a public implementation of it. `AnalyzerFileReference` now requires an `IAnalyzerAssemblyLoader`.

troydai:
There are 14 tests failed on coreclr-linux only (they pass on Windows coreclr instead). I investigate one of them and it seems to be Roslyn related: https://github.com/OmniSharp/omnisharp-roslyn/tree/update-rc2-nixfailure.

I'm not going to dig deeper for now, since I'd like to focus on `OmniSharp.DotNet` migration. Perhaps we can skip these 14 tests for now and try to make a smaller repo for `Roslyn` team to fix.

troydai:
@david-driscoll my latest commits fix the tests. Now test cases all pass on both platforms. However the following scripts of building packages failed. Could you please take a look at the remaining part of the `build.sh`.

Note: doesn't use coreclr DNX to build CLR target framework on *nix.

troydai:
@david-driscoll I'm actually working on fixing the `build.sh` now...

troydai:
All green! for the first time!

troydai:
Ping @davidfowl @david-driscoll

david-driscoll:
Awesome, I'll try to look at this tonight.  I'm now mostly back from vacation, lol.

troydai:
@david-driscoll Welcome, back :)

david-driscoll:
Couple tweaks in the scripts, I'm working on them right now.

I need a `sh` command to get the latest runtime directory.  I have the windows `cmd` version, but I need a similar one for `sh`. :smile: 

    for /F "delims=" %%I in ('dir %USERPROFILE%\.dnx\runtimes\dnx-clr-win-x86.*-rc2-* /b /ad /on') do set RUNTIME=%%I

david-driscoll:
So I've run into something that I feel like is a blocker, there is a work around but it might kill performance when running large diagnostics.

Basically this code from [`StdioServer.cs`](https://github.com/OmniSharp/omnisharp-roslyn/blob/update-rc2/src/OmniSharp.Stdio/StdioServer.cs#L65-L79) is falling over.

```csharp
    var ignored = Task.Factory.StartNew(async () =>
    {
        try
        {
            await HandleRequest(line, application);
        }
        catch (Exception e)
        {
            _writer.WriteLine(new EventPacket()
            {
                Event = "error",
                Body = JsonConvert.ToString(e.ToString(), '"', StringEscapeHandling.Default)
            });
        }
    });
```

My feeling is that [Kestrel](https://github.com/aspnet/KestrelHttpServer) is reusing `Response`s between requests, and indeed from time to time you see Responses trying to overwrite each other.  What happens is either the request writes two json values to Body, which produces output that looks like `{ ... Body: { ... my response ... }{ ... my other resposne ...} ... }`.  This is almost certianly related to the performance work that is being done, but it's causing us issues here.

There is a workaround, simply eliminate the task, but it basically means we can only handle one request at time.  If I implement the workaround things start working, but there is a noticeable delay when running full diagnostics and editing code.

```csharp
    try
    {
        await HandleRequest(line, application);
    }
    catch (Exception e)
    {
        _writer.WriteLine(new EventPacket()
        {
            Event = "error",
            Body = JsonConvert.ToString(e.ToString(), '"', StringEscapeHandling.Default)
        });
    }
```

cc: @troydai @davidfowl @benaadams @halter73 @lodejard @DustinCampbell 

I feel this is a blocker for moving forward with rc2, while the workaround works the performance feels the hit.  It may be possible to rewrite the `StdioServer` to work around a problem (`HttpClient` maybe?).  That said it has been working great for us for a long time now.

halter73:
@david-driscoll  This is something @lodejard and I were talking about the other day.

We faced the exact same problem when building SignalR to run on top of IIS using OWIN/katana . Our ServerSentEvents transport would responses like `{ ... data: { ... message... }{ ... other message ...} ... }` when customers tried to invoke client methods from multiple non-synchronized threads. We were able to work around this issue without limiting writes to a single thread or locking by using a [TaskQueue](https://github.com/SignalR/SignalR/blob/dev/src/Microsoft.AspNet.SignalR.Core/Infrastructure/TaskQueue.cs).

We are considering several options. One expanding a lock where we copy response data into our internal data structures inside of `WriteAsync`. This is the simplest solution, but I was seeing a ~3% performance degradation in our plaintext benchmark which does no concurrent writes. We could also use a `TaskQueue` in Kestrel itself, I haven't tested the perf impact of this yes, but I'm afraid it would be even worse. Obviously neither would guarantee any sort of ordering, but it would make writes atomic.

The last option is to simply attempt to recognize when concurrent writes are occurring and throw and exception if that is the case. This is the solution I was leaning toward. My feeling is the vast majority of web applications will not attempt to do concurrent writes, since it's impossible to know the order of the data in the response if you do so. Things like SignalR will have to find a way to sequence their writes, just like they used to on IIS.

david-driscoll:
Thanks @halter73 for the quick reply.  I'm not so concerned with ordering, I know for example getting all diagnostics for a solution will take longer than getting the type info for a symbol.

The problem is that `HandleRequest` calls `CreateContext`, which appears to be returning an `HttpContext` that is still in flight (and therefore the `HttpResponse` and `HttpRequest` objects, I assume anyway).  The `Body` of the response is then swapped, and written twice, or in the case of the previous request, never written to at all.

I'm not seeing much facility for us to know when, where or if that swap will happen.  Granted I haven't looked all that hard yet.

david-driscoll:
Also one odd thing... which is more coreclr related... if I try to run on the CoreCLR, with my work computer, the anti-virus removes the exe from memory, which causes the app to crash.  That was confusing.

benaadams:
@david-driscoll Request and Response are set after the Request headers have been parsed https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNet.Server.Kestrel/Http/FrameOfT.cs#L76-L77

benaadams:
For Response you can replace it as part of OnStarting; but to replace both currently you may have to an UpgradeFeature as per websockets https://github.com/aspnet/WebSockets/blob/dev/src/Microsoft.AspNet.WebSockets.Server/WebSocketMiddleware.cs#L112

halter73:
@benaadams I don't think the issue here are writes not writing to the correct response body, I'm pretty sure the only problem are that writes aren't atomic. I just submitted a PR here: https://github.com/aspnet/KestrelHttpServer/pull/562 that addresses the issue.

benaadams:
@halter73 isn't that a bit weird? If you write to any stream in `System.IO` concurrently, you will mashup your data and there is no synchronization applied (e.g. MemoryStream https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/IO/MemoryStream.cs#L540-L613); its up the caller to synchronize their writes - which is normally automatic by waiting for one to complete before doing another.

@david-driscoll another approach would be to use a filter?

halter73:
Note: the linked PR might save you if you are using something like JsonSerializer to write your messages, since that will make multiple calls to `Stream.Write`.

benaadams:
@david-driscoll testing the theory, if you changed the code to this does it work?
```csharp
var writeLock = new object();

while (!_cancellation.IsCancellationRequested)
{
    var line = await _input.ReadLineAsync();
    if (line == null)
    {
        break;
    }

    var ignored = Task.Factory.StartNew(async () =>
    {
        try
        {
            lock (writeLock)
            {
                await HandleRequest(line, application);
            }
        }
        catch (Exception e)
        {
            _writer.WriteLine(new EventPacket()
            {
                Event = "error",
                Body = JsonConvert.ToString(e.ToString(), '"', StringEscapeHandling.Default)
            });
        }
    });
}
```

troydai:
@david-driscoll about the script to get the latest runtime directory. I'm not sure how important it is for now, since the toolings are being transformed to `dotnet cli`. In time the only DNX sdk you want to keep the is the `1.0.0-rc1-update1` (or may be `update2` if there is going to be one).

david-driscoll:
If rc2 bits can run on `1.0.0-rc1-update1` then we can pin on that, I just wasn't fully sure.  If that's the case then things can be shifted around fairly easily.  We still have the Kestrel issue to deal with @benaadams I'll try out that code tomorrow and see what behavior I get.

david-driscoll:
@troydai attempting to run on rc1-update1 returns...

```
fail: OmniSharp.Dnx.DnxProjectSystem[0]
      Entry point was not found.
fail: OmniSharp.Dnx.DnxProjectSystem[0]
      Entry point was not found.
fail: OmniSharp.Dnx.DnxProjectSystem[0]
      Entry point was not found.
info: OmniSharp.ScriptCs.ScriptCsProjectSystem[0]
      Detecting CSX files in 'D:\Development\Omnisharp\omnisharp-roslyn'.
info: OmniSharp.ScriptCs.ScriptCsProjectSystem[0]
      Could not find any CSX files
fail: OmniSharp.Dnx.DnxProjectSystem[0]
      Entry point was not found.
info: OmniSharp.Startup[0]
      Solution has finished loading
fail: OmniSharp.Dnx.DnxProjectSystem[0]
      Entry point was not found.
```

david-driscoll:
I think I figured out the cause of the problem, which might actually be an issue on the OmniSharp side instead of on the Kestrel side of things.  I'll do some more testing and push the changes later.

david-driscoll:
So my work around is to pass in a unique features collection for each request.  I'm worried it's not quiet the "correct" way to make things work, but it does get past this issue we're having at the moment.

david-driscoll:
I hope this should be pretty stable.

@ianbattersby would you be able to test and see if this works with emacs?  Or if we've broken anything in this branch before we merge it friday/saturday?

DustinCampbell:
I'm trying to get this working with VS Code. I've got the CoreCLR OmniSharp loading properly. However, the `DnxProjectSystem` never gets included in the MEF composition. So, specifying a path to a code base on the command-line to `OmniSharp.Host` doesn't result in any projects or files getting added to the `OmniSharpWorkspace`, so nothing really works. Is this known? How is DnxProjectSystem supposed to get into the MEF composition?

david-driscoll:
I remember Troy said that the Assembly had to be referenced manually
because there is no way to load assemblies like that in CoreCLR.

DustinCampbell:
> I remember Troy said that the Assembly had to be referenced manually
because there is no way to load assemblies like that in CoreCLR.

Where does it need to be referenced manually? I'm not clear on how this is supposed to work now. Does it actually work in Atom for you?

troydai:
It's not about CoreCLR. It's the dotnet. Today, for the project build with dotnet cli the traditional `LibraryManager` is not there for searching for all referenced assembly. 

We do just have replacement mechanism now. I'll adopt it in my dotnet branch.

david-driscoll:
@DustinCampbell @troydai @davidfowl @OmniSharp/all 

Anyone had a chance to do some more testing here, thing we're good to go?  We might lose a few of the refactorings for the interim, but we stand to gain those back later.

I've got [deployment](https://github.com/OmniSharp/omnisharp-roslyn/releases/tag/v1.7.0-beta21) working, and I've got a version sitting in the master of https://github.com/OmniSharp/omnisharp-node-client that now downloads the OmniSharp runtime on demand as needed.

DustinCampbell:
@troydai -- got it -- thanks!

Mpdreamz:
Did a quick pass over in Atom and things look good :+1: Also tested out omnisharp-node-client's download of the runtime which works a treat.

david-driscoll:
Alright... moment of truth... lets merge this!

PR #374, opened 2016-01-05
Update dotnet

troydai:


build-cli.py
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env python
```
Mpdreamz: Nothing against python per se but this introduces a python dep for something that could be replaced in the build.cmd/sh at first glance.

troydai: Thanks for point out. I’ll eventually turn this script in to ps1 and sh.

However, I find it’s nice to have one script across the platform. If we can agree on something that will be great.

src/OmniSharp.CliHost/Program.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+using OmniSharp.Host;
+
+namespace OmniSharp.CliHost
+{
+    public class Program
+    {
+        public static void Main(string[] args)
+        {
+            Console.WriteLine($"OmniSharp-CLI: {string.Join(" ", args)}");
```
davidfowl: Remove

troydai:
I'll begin working on this branch tomorrow.

troydai:
Move to https://github.com/OmniSharp/omnisharp-roslyn/pull/398

PR #384, opened 2016-01-22
Fixed bootstrap so that it only returns a folder, if a bootstrap inst…

david-driscoll:
…ance needs to be created


filipw:
@david-driscoll is this still needed?

david-driscoll:
nah, we can dump it for now.

PR #386, opened 2016-01-23
Upgrade DNX from 16388 to 16401

troydai:
Upgrade the DNX only.


troydai:
Close. Doesn't help with the System.Runtime.Extension 4.1.0.0 missing issue. The issue is introduced by a inconsistent dependency and a bug in the dnx. I'll try to address it in my other PR.

PR #389, opened 2016-01-25
Update SharedTextWriter

troydai:
Using a BlockedCollection instead of spin off a thread for each write action.


david-driscoll:
@troydai this ready for some extra testing and merging?

troydai:
Passed Stdio tests, but not sure how sufficient the tests are. Made change because I notice 2 dozens of threads queued up for writing. Used the change for a while in VS Code. Feels good to me.

PR #397, opened 2016-02-05
Attempting to fix unused architecture and runtime options in global.json

hal-ler:
Thank you guys for the great work, just wanted to chime in with a small patch attempting to fix: https://github.com/OmniSharp/omnisharp-vscode/issues/12.

Breaking changes: "alias" now has its own field in "global.json/sdk" and "version" takes precedence over "alias". I chose this design as "alias" can always default to "default". "version" on the other hand shows that the user knows what they are doing.

"architecture" and "runtime" default to the existing values ("x86" and "clr"/"mono"). I even added the ability to set Linux as the OS (eg. coreclr-linux-x64), but could not figure out Darwin using a simple mechanism.

Istvan Haller


hal-ler:
Closed since the DNX tools are deprecated now.

PR #406, opened 2016-02-24
Pin the dependencies and use single feed

troydai:
In addition, remove empty test projects to reduce restore cost.


troydai:
@david-driscoll all passed. I'm going to merge now.

PR #411, opened 2016-03-03
Update README.md

troydai:


README.md
```diff
@@ -41,6 +41,22 @@ _Update path to fit your environment_
 }
 ```
 
+### For Linux / Unix
+
+If you're running OmniSharp on Linux or Unit, you need to set `DOTNET_REFERENCE_ASSEMBLIES_PATH` for projects target desktop CLR. You can use following bash script to set `DOTNET_REFERENCE_ASSEMBLIES_PATH` to mono's reference assemblies folder
```
willl: I think you mean Unix rather than "Unit" :)

Also, you should use "for projects targetting the desktop CLR" rather than "for projects target desktop CLR".

troydai: :) Thank you for point that out. I’ll update. This is the reason why my English teachers never liked me.

Issue #396, opened 2016-02-04
Suggestion: Cross-platform build script


troydai:
I’m updating the build tool chain from DNX to dotnet-cli now. To enable the build on different platforms, I have to write both bash script and ps1/batch script for cross-platform build. This is a long-term maintenance cost. I’m wondering if we can exploring the possibility of using a unified cross-plaform script language or tool to drive the build.

david-driscoll:
One of my friends was looking to get involved and wanted to build a Cake script, then we would just bootstrap the cake build for windows and osx.

I know he has made some progress, but there might be some issues with the exports in `build.sh` today.

Personally I like the idea of a Cake script, as long as it can fill our needs.  It does offer some more abstractions over what say ScriptCS has today, specifically with building and running tasks, etc.  It also keeps us in the ".NET" world so to speak.  The only gotcha I see is that in order for the build to complete the travis environment will still need to have mono installed.

cc @RLittlesII

troydai:
I'm interested in the progress of the project. Do we already have a running prototype/version or it is still in development? The major concerns about using a newly developed tool is it will be lack of tooling support and takes time to mature.

RLittlesII:
I have started development of the cake script.  I have been active in the cake gitter channel around some of the environment variable concerns.  I haven't pushed any changes to my fork yet.  Sounds like the need is sooner than later.  I know there is DNU support in Cake, there is currently a PR in for dotnetcli.  I would be glad to talk through the requirements of the new script.

devlead:
:+1: @RLittlesII just ping me @patriksvensson or @gep13 or raise an issue if you encounter any obstacles/concerns with @cake-build

hal-ler:
I have made a pull request to make it possible to set the install path for dotnetcli as a command line argument instead of being forced to use the environment. This functionality already existed in the linux script, but not the powershell one. Using this change I was able to quickly create a Cake script which builds the dnxcore50 variant on Windows (and should work for Linux as-well, but still need to test).

The only other environment variables which I can see right now, are the ones related to the reference assemblies, which can also be set in the bootstrapping linux shell script.

https://github.com/dotnet/cli/pull/1701

troydai:
@hal-ler I just pinged a few guys on the cli team to review and sign off the change.

RLittlesII:
[...]

PR #416, opened 2016-03-05
Creating a portable Cake build script.

hal-ler:
This is my initial suggestion of a Cake-based build script  #396.

Updated to use build.json from current build system.

Usage:
build.(ps1|sh) are used to retrieve the Cake bootstrapping script and to launch it. build.cake includes the portable build script itself.

Cake info:
Cake currently has no direct support for .NET Core CLI (https://github.com/cake-build/cake/pull/674). For now I call the tools as external processes, which works well, especially since I can generate output logs to make it functionally equivalent to the current build script.
While Cake is cross-platform, it is not built against dnxcore50, so it requires mono to work on Unix :(.

Definitions:
-Local build: build only against local runtime (or closest match)
  1) If local default runtime is in build.json, then only build against it
  2) Find first match using version-less OS name between default runtime and list in build.json (win10->win7, osx.10.10->osx.10.11, etc.)
  3) Report error if none found, cannot build locally

Targets:
-target=Default : Same as Local target
-target=Local : Full local build with restore and test
-target=All : Full build for all runtimes listed in build.json
-target=Quick : Equivalent to "-quick" build (only performs local build, no restore, no test)
-target=Install : Quick + install artifacts into folder configured in build.cake (similar to "-install")
Options:
-configuration={Release|Debug} : Configuration to use for build and publish
-test-configuration={Release|Debug} : Configuration to use for test
-install-path={Path} : Destination folder for install target

- [x] Check and fix on Unix.
- [x] Make it the default on Windows (for Appveyor).
- [x] Make it the default on Unix (for Travis).
- [x] Implement dnx451 build path.
- [x] Integrate with CI.
- [x] Document code.


hal-ler:
Any ideas about getting dnx451 to work without installing VS (https://github.com/dotnet/cli/issues/1545)?

hal-ler:
Redirected build.sh to the new script to test it on Travis.

hal-ler:
Uncovered https://github.com/dotnet/cli/issues/1725. Solved it by making version "latest" explicit.

hal-ler:
Sorry about the CI spam guys, but I did not have access to a machine with working dnx451 over the week-end and assumed that solving XUnit is simpler. Now it works for both build paths on both Windows (tested locally) and Unix (tested on Travis).

troydai:
Hi @hal-ler should be rebase this on the latest dev?

hal-ler:
@troydai Yes, just trying out one more commit locally.

hal-ler:
@troydai Rebased on pushed updates to bring it in line with current build system and the build.json file. Interesting features include local build, which selects the best candidate runtime from build.json to build it for local machine only. It also has an additional test to run OmniSharp (with local runtime) after publishing to see if the executable truly works.

.gitignore
```diff
@@ -6,6 +6,7 @@ TestResults/
 _ReSharper.*/
 packages/
 artifacts/
+fakeinstall/
```
troydai: How about we move this to the `artifacts` folder so that all the output are centralized in one place?

build-cake.sh
```diff
@@ -0,0 +1,22 @@
...
+# https://github.com/dotnet/cli/issues/531
+if [ -z "$DOTNET_REFERENCE_ASSEMBLIES_PATH" ]; then
+    if [ $(uname) == Darwin ] && [ -d "/Library/Frameworks/Mono.framework/Versions/Current/lib/mono/xbuild-frameworks" ]; then
+        export DOTNET_REFERENCE_ASSEMBLIES_PATH="/Library/Frameworks/Mono.framework/Versions/Current/lib/mono/xbuild-frameworks"
+    elif [ -d "/usr/local/lib/mono/xbuild-frameworks" ]; then
+        export DOTNET_REFERENCE_ASSEMBLIES_PATH="/usr/local/lib/mono/xbuild-frameworks"
+    elif [ -d "/usr/lib/mono/xbuild-frameworks" ]; then
+        export DOTNET_REFERENCE_ASSEMBLIES_PATH="/usr/lib/mono/xbuild-frameworks"
+    fi
+fi
```
troydai: This is no longer needed. I fixed https://github.com/dotnet/cli/issues/531. :beers:

build-cake.sh
```diff
@@ -0,0 +1,22 @@
...
+
+if [ $(uname) == Darwin ]; then
+    export OSSTRING=osx
+elif [ $(uname) == Linux ]; then
+    export OSSTRING=linux
+fi
+#curl -Lsfo cake-bootstrap.sh http://cakebuild.net/bootstrapper/$OSSTRING
+#chmod +x cake-bootstrap.sh
+./cake-bootstrap.sh "$@"
```
troydai: Why check in this file instead of download dynamically? Is it customized here?

build.cake
```diff
@@ -0,0 +1,408 @@
...
+                    return builder;
+                }
+            };
+            xunitSettings.ExcludeTrait("category", new[] { "failing" });
+            XUnit2($"{instanceFolder}/{project}.dll", xunitSettings);
+        }
+    }
+});
+
+Task("OnlyPublish")
```
troydai: I think we should have two kinds of `Publish` target here:
1. `LocalPublish` which only publish the application for current platform. This is for local build and testing.
2. `FullPublish` which publishes for all platforms. This is for CI.

troydai:
Added a few comments. `PublishProject` can be deleted, too.

build.cake
hal-ler: The system does it implicitly by running the RestrictToLocalRuntime task at the start of the plan, but it makes sense to make it specifically for Publish. It is only Publish that is affected by this anyway.

build-cake.sh
hal-ler: I was blocked on https://github.com/cake-build/cake/issues/737 and https://github.com/cake-build/cake/pull/734. The latter was merged, but not yet released (at least not on GitHub). I created a PR (https://github.com/cake-build/cake/pull/756) to fix the former. Another option is to hard-code the NUGET_EXE environment variable in this script.

Long term PR https://github.com/cake-build/cake/pull/605 would also be beneficial, as custom arguments are not allowed otherwise (just learned this today).

.gitignore
hal-ler: It is just a temporary place-holder for a proper install path. What path would you prefer on Windows (similar to ~/.omnisharp/local)?

build-cake.sh
hal-ler: Great!

hal-ler:
Thanks for the feed-back!

.gitignore
troydai: We can designate the `%USERPROFILE%\.omnisharp\local` / `$HOME/.omnisharp/local` as the output by default. And we can also allow user to override the path in `build.json`. Try not to have default value hardcoded.

build-cake.sh
troydai: Got it. Thanks.

troydai:
Looking good. @david-driscoll Do you want to take a look as well?

build.cake
```diff
@@ -0,0 +1,408 @@
...
+    }
+    // On all platforms use TAR.GZ for Unix runtimes
+    else
+    {
+        var tarFile = outputFile.AppendExtension("tar.gz");
+        // Use 7z to create TAR.GZ on Windows
+        if (IsRunningOnWindows())
+        {
+            var tempFile = outputFile.AppendExtension("tar");
+            var exitCode = StartProcess("7z", new ProcessSettings
```
troydai: Ensure the `7z` exists in this environment.
https://github.com/OmniSharp/omnisharp-roslyn/issues/431

build.cake
```diff
@@ -0,0 +1,408 @@
...
+    .IsDependentOn("Publish")
+    .IsDependentOn("RestrictToLocalRuntime")
+    .Does(() =>
+{
+    var project = buildPlan.MainProject;
+    foreach (var framework in buildPlan.Frameworks)
+    {
+        // Skip testing mono executables
+        if (!IsRunningOnWindows() && !framework.Equals("dnxcore50"))
+            continue;
```
troydai: I notice that this is your coding style. However generally we add curly braces in all conditions. This is the convention here in `OmniSharp`.

build-cake.ps1
```diff
@@ -0,0 +1,5 @@
+# Hack until dotnet/cli##1701 is accepted
```
david-driscoll: Could we go back to `build.ps1` and `build.sh`, I know underlying they will bootstrap cake for me, but from the outside I don't really care that cake is the build system, I just want to `build`.  Feels like an extra 5 characters for nothing (I know I know tab completion etc :dragon_face: )

david-driscoll: Oh I see you preserved the old build files, I'm fine with getting rid of them to replace with Cake assuming all the features are replaced, we should have just one set of build files.

hal-ler: Sure, the different naming was mostly to make rebase-ing sane.

david-driscoll:
It builds, I like it.  Couple minor tweaks then we can probably :shipit: Next step is to look at merging something to master fairly soon, but I just haven't had time to sit down and test with a real project all that much.

build.cake
hal-ler: Sure!

build.cake
hal-ler: Best way I found to do it is a try/catch block and a warning message if it did not find it. Default build is also Local so people should not trigger it, unless they explicitly want it.

build-cake.sh
hal-ler: Found that setting NUGET_EXE solves it for now. Also got some responses on the Cake repository, so I will keep an eye out and simplify it when updates are pushed.

.gitignore
hal-ler: Okay, I added a command line option for the install-path.

hal-ler:
Cake bootstrapper temporarily broken on Windows. Fix should appear soon.

hal-ler:
Cake bootstrapper looks fixed, but can only test it in the evening.

david-driscoll:
Looks good, going to try a merge locally to see if there are any problems since github is saying it's out of date but it isn't saying there are conflicts.

hal-ler:
I'm still working out the Cake bootstrapping bug with the guys over in their repository. You can test it on non-Windows platforms though.

david-driscoll:
interestingly `Invoke-WebRequest https://raw.githubusercontent.com/cake-build/bootstrapper/develop/build.ps1 -OutFile cake-bootstrap.ps1` seems to work just fine.

hal-ler:
Yes, they use slightly modified scripts to build their own stuff, so they did not see the breakage on CI when they updated the official script at `/develop/res/scripts/build.ps1`

hal-ler:
After waiting several days to get the official bootstrapping bugs fixed, I decided to get around it, but without having to maintain our own set of scripts.
For bootstrapping I use the scripts from the main Cake repo. These are tested and guaranteed to work, but require an existing package.config file to select the appropriate Cake version.
I replicated the mechanism to retrieve package.config from their official bootstrapping scripts in build.{ps1|sh}.
For now I download the package.config file from their "website" repo. Normally it is generated on the fly, using a link on their website, but they are facing some bugs with that functionality right now. Looking back at update history suggests that they keep it up to date.

hal-ler:
I also added an additional DotNetVersion option to build.json, which allows the user to select any given CLI version.

hal-ler:
Final solution to bootstrapping based on recommendation from Cake devs. The bootstrapping scripts are now part of the repository. As new custom behavior, I removed the default target and configuration options described in these scripts. They are set in build.cake and  we don't want duplicate defaults. The defaults can also migrate to build.json at some point with this setup.

Ubuntu breaks until .NET CLI is updated, but the other systems pass. @troydai @david-driscoll I would say that this is a good first candidate from my part, if you guys are also ok with it. I can add features in the future as the need pops up.

troydai:
For the Ubuntu build, please wait for next Tuesday. Let me try to get CLI fix the build.

.gitignore
```diff
@@ -35,5 +35,7 @@ buildlog
 *.nuget.targets
 
 # Build folder
-.build/
-.dotnet/
+.dotnet
+
+# Cake build
+tools/
```
troydai: make it `.tools/`

build.cake
```diff
@@ -0,0 +1,567 @@
...
+    public string ArtifactsFolder { get; set; }
+    public string DotNetFolder { get; set; }
+    public string DotNetInstallScriptURL { get; set; }
+    public string DotNetChannel { get; set; }
+    public string DotNetVersion { get; set; }
+    public string[] Frameworks { get; set; }
+    public string[] Rids { get; set; }
+    public string MainProject { get; set; }
+}
+var buildPlan = DeserializeJsonFromFile<BuildPlan>($"{workingDirectory}/build.json");
```
troydai: What's the effect of deserialize failure? Is the build terminated here?

build.cake
```diff
@@ -0,0 +1,567 @@
...
+var sourceFolder = $"{workingDirectory}/src";
+var testFolder = $"{workingDirectory}/tests";
+
+var artifactFolder = $"{workingDirectory}/{buildPlan.ArtifactsFolder}";
+var publishFolder = $"{artifactFolder}/publish";
+var logFolder = $"{artifactFolder}/logs";
+var packageFolder = $"{artifactFolder}/package";
+
+/// <summary>
+///  Retrieve the default local RID from .NET CLI.
```
troydai: Remove the extra space.

```
/// Retrieve ...
```

build.cake
```diff
@@ -0,0 +1,567 @@
...
+string MatchLocalRuntimeID(string dotnetcli, BuildPlan buildPlan)
+{
+    var localRuntime = GetLocalRuntimeID(dotnetcli);
+    if (buildPlan.Rids.Contains(localRuntime))
+    {
+        return localRuntime;
+    }
+    else
+    {
+        foreach (var runtime in buildPlan.Rids)
```
troydai: Can use a simple Linq to achieve this:

```
return buildPlan.Rids.First(rid => {
    // conditions
});
```

build.cake
```diff
@@ -0,0 +1,567 @@
...
+    var process = StartAndReturnProcess(dotnetcli, 
+        new ProcessSettings
+        { 
+            Arguments = "--version",
+            RedirectStandardOutput = true
+        });
+    process.WaitForExit();
+    if (process.GetExitCode() != 0)
+    {
+        throw new Exception("Failed to get run dotnet --version");
```
troydai: I think there should be a __prerequisit__ target to detect if `dotnet` is on path. The check will be run once and then we can safely execute `dotnet` commands afterwards.

build.cake
```diff
@@ -0,0 +1,567 @@
...
+var publishFolder = $"{artifactFolder}/publish";
+var logFolder = $"{artifactFolder}/logs";
+var packageFolder = $"{artifactFolder}/package";
+
+/// <summary>
+///  Retrieve the default local RID from .NET CLI.
+/// </summary>
+/// <param name="dotnetcli">Full path to .NET CLI binary</param>
+/// <returns>Default RID</returns>
+string GetLocalRuntimeID(string dotnetcli)
```
troydai: The implementation of this is not stable. We don't guarantee `dotnet --version` output's format. Use `PlatformServices` instead.

build.cake
```diff
@@ -0,0 +1,567 @@
...
+/// <summary>
+///  Match the local RID with the ones specified in build.json.
+///  Return exact match if found.
+///  Return first OS match (without version number) otherwise.
+///  Report error if no OS match found.
+/// </summary>
+/// <param name="dotnetcli">Full path to .NET CLI binary</param>
+/// <param name="buildPlan">BuildPlan from build.json</param>
+/// <returns>Matched RID</returns>
+string MatchLocalRuntimeID(string dotnetcli, BuildPlan buildPlan)
```
troydai: I don't think we need this. The list of `RID` in the build plan is just the officially supported platforms. If a user decide to build `OmniSharp` on a platform other than those listed, we should allow them.
/cc @david-driscoll

build.cake
```diff
@@ -0,0 +1,567 @@
...
+});
+
+/// <summary>
+///  Restore required NuGet packages.
+/// </summary>
+Task("Restore")
+    .IsDependentOn("BuildEnvironment")
+    .Does(() =>
+{
+    var exitCode = StartProcess(dotnetcli,
```
troydai: Shall we consider timeout?

build.cake
```diff
@@ -0,0 +1,567 @@
...
+    if (exitCode != 0)
+    {
+        throw new Exception("Failed to restore.");
+    }
+});
+
+/// <summary>
+///  Build Test projects.
+/// </summary>
+Task("TestBuild")
```
troydai: I prefer the name `BuildTest`

troydai:
@hal-ler it will be more convenient for us to collaborate if you squash your changes and move them onto a branch in this repo. I'd love to help make a few changes to the cake script but I can't do so if the change it in your fork.

build.cake
hal-ler: Yes, exception.

build.cake
hal-ler: But `dotnet restore` uses the RIDs specified in `Omnisharp/project.json`. The build fails if it targets other platforms explicitely (for example even `win10`). Also this code is meant to allow for local-only builds, which is nice to support without the user having to manually edit the `build.json` file.

hal-ler:
@troydai Sure, but I don't think I have the rights to create a branch in the repo.

build.cake
troydai: If user do a local build, which means he only build for current platform. `dotnet restore` doesn't need a `--runtime` parameter. We can simply omit it. In this situation, the build process shouldn't build for any other platform, because it is user's choice to build for local platform.

Imagine it works like this:

```
$ .\build.sh   # user choose to build locally, then only one platform is build, test and published.
$ .\build.sh --all # user choose to build for all platforms in build.json RIDs list. This is more like a CI scenario
```

david-driscoll:
I've created a branch `cake-build` from `dev` if you point you pull request there (you might have to make a new one) we can merge it in.

build.cake
hal-ler: Unfortunately it is not available as Cake is not a .NET Core project yet :(

build.cake
hal-ler: This is just a helper function so I cannot have prerequisites on it, but I added the check in BuildEnvironment and made sure that everything using the command depends on it. Eventually an extra option can be added into build.json to supress the InstallDir option and just use the default folder (when the appropriate install script version is used).

build.cake
hal-ler: Done.

build.cake
hal-ler: But doesn't `dotnet restore` take the rids defined in `project.json`? Which is defined for OmniSharp and cannot be overridden on the command line? I know from experience that `dotnet publish --runtime win10-x64` fails.

build.cake
hal-ler: Done.

.gitignore
hal-ler: Done.

hal-ler:
I closed this pull request to create  the other one on the cake-build branch.

build.cake
troydai: You have a point. It feels like something should be improved on `dotnet restore`.

PR #417, opened 2016-03-05
Build publish task in a .NET program and update build

troydai:


troydai:
@david-driscoll @DustinCampbell I move majority of the building logic into the `PublishProject` tool project. Therefore this is no need to replicate the building steps between `bash` and `batch`. 

@hal-ler is working on `Cake` build script. And I believe that should be the eventual form our build scripts. Much of the logic in this PR can be morphed into `Cake` targets.

hal-ler:
@troydai Looks great! The Cake script is also effectively ready, just waiting on PR dotnet/cli#1701 to default the windows path and check on CI.

troydai:
@hal-ler I plan to merge this later today once @david-driscoll green light it. And we can rebase your Cake scripts.

tools/PublishProject/Program.cs
```diff
@@ -0,0 +1,102 @@
...
+            return 0;
+        }
+
+        private static void Package(string publishOutput,
+                                    string packageOutput,
+                                    string projectName,
+                                    string rid,
+                                    string framework)
+        {
+            var zipFilePath = Path.Combine(packageOutput, $"{projectName}-{rid}-{framework}.zip");
```
david-driscoll: Are we wanting zips for all platforms now?  I know traditionally for *nix/OSX we've gone with tarballs.

I don't have a preference, not sure if anyone else cares either.

troydai: @DustinCampbell will this impact how OmniSharp is shipped with VS Code?
@david-driscoll Not sure if `System.IO.Compression` support compress into tar. But not relying on external tools for packaging does simplify the build process.

build.sh
```diff
@@ -2,67 +2,62 @@
...
-    if [ -d ~/.omnisharp/local ]; then
-        echo "Removing local omnisharp ..."
-        rm -rf ~/.omnisharp/local
-    fi
+if [ "$TRAVIS" == true ]; then
+    echo "Installing dotnet from beta channel for CI environment ..."
 
-    mkdir -p ~/.omnisharp/local
+    $_dotnet_install_script -c beta -d "$work_dir/.dotnet"
+    dotnet="$work_dir/.dotnet/bin/dotnet"
```
troydai: Question for @david-driscoll: it is really necessary to download `dotnet` to the `.dotnet` folder? The install script sets up the symbol link correctly so that assuming `dotnet` is on the path is recommended.

tools/PublishProject/Program.cs
DustinCampbell: No problem for me. I can react to it in VS Code.

david-driscoll: I'd like to see it a tarball if possible, mainly because it's generally what you see for those environments, I know they can unzip too but... lets be good citizens of the community. :smile: 

Actually doing some research I'd be concerned if using System.IO.Compression was going to keep around attributes (I'm not sure it will!).  It is something we definitely need to test, because I don't want the first instruction to be `unzip` and then `chmod`.
Not sure how relevante but see: http://lists.ximian.com/pipermail/mono-list/2009-February/041428.html

build.sh
david-driscoll: I was trying to follow along with the proposed guidelines I saw on https://github.com/dotnet/cli/issues/1022, about using a specific version of `dotnet` to build an application.  Today it isn't a concern as we will always want the latest but it could be a problem somepoint in the future.

I won't object to removing it for now though.

troydai: I see. 

I'll remove the `.dotnet` specific thing. As soon as `dotnet-cli` comes up with a proper version management we can begin adopt that solution. Currently, since we stay on the current edge build, there isn't a reason to trade off the simplicity.

tools/PublishProject/Program.cs
troydai: ok. I'll open up a new issue and let's upgrade the zip part after this PR goes in,

build.sh
hal-ler: The install script needs root access to install it in the PATH. It also cannot modify the PATH on Linux because the environment is destroyed on exit.

david-driscoll:
As it is it builds, I say :shipit:.  I think the zip file issue will be a problem going forward to do a release, but that can be a different PR.

tools/PublishProject/Program.cs
troydai: https://github.com/OmniSharp/omnisharp-roslyn/issues/421

@david-driscoll will add more details to it. I'll keep the this `zipping` logic intact for now.

troydai:
Can't :shipit: for now. The `--runtime` option is removed from `dotnet restore` command. I'll make an update tonight to address the change.

build.sh
troydai: On Linux the install script doesn't add it to PATH. It create symbol link in `/usr/local/bin`

troydai: Ugr I just realize even this needs root.

troydai:
@david-driscoll all passed now. I'll merge to `dev` soon

PR #433, opened 2016-03-16
Add support for DotNetProjectSystem.GetInformationModel(...)

DustinCampbell:
This is necessary for the DotNetProjectSystem to handle the /projects end point.

cc @troydai, @david-driscoll

### DIFF
--- a/src/OmniSharp.DotNet/Models/DotNetWorkspaceInformation.cs
+++ b/src/OmniSharp.DotNet/Models/DotNetWorkspaceInformation.cs
@@ -1,9 +1,22 @@
-﻿using System.Collections.Generic;
+﻿using Microsoft.DotNet.ProjectModel;
+using System.Collections.Generic;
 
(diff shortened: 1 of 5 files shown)


src/OmniSharp.DotNet/Models/DotNetWorkspaceInformation.cs
```diff
@@ -1,9 +1,22 @@
-﻿using System.Collections.Generic;
+﻿using Microsoft.DotNet.ProjectModel;
+using System.Collections.Generic;
```
troydai: Put `System.*` at the beginning.

troydai:
:shipit:

src/OmniSharp.DotNet/Models/DotNetWorkspaceInformation.cs
DustinCampbell: Oops! Thanks!

PR #437, opened 2016-03-18
Update README.md

troydai:


david-driscoll:
this should merge just fine, not worrying about the build breaks.

PR #460, opened 2016-03-23
Add EntryPoint property to DotNetProjectInformation

DustinCampbell:


DustinCampbell:
Actually, this doesn't work like I'd hoped. I'll submit a different change.

PR #462, opened 2016-03-23
Add more properties to DotNetProjectInformation

DustinCampbell:
Two new properties:

1. TargetFramework
2. EmitEntryPoint

cc @troydai

### DIFF
--- a/src/OmniSharp.DotNet/Models/DotNetProjectInformation.cs
+++ b/src/OmniSharp.DotNet/Models/DotNetProjectInformation.cs
@@ -9,11 +9,15 @@ namespace OmniSharp.DotNet.Models
         {
             this.Path = projectContext.RootProject.Path;
             this.Name = projectContext.ProjectFile.Name;
+            this.TargetFramework = new DotNetFramework(projectContext.TargetFramework);
 
             var outputPaths = projectContext.GetOutputPaths(configuration);
             this.CompilationOutputPath = outputPaths.CompilationOutputPath;
             this.CompilationOutputAssemblyFile = outputPaths.CompilationFiles.Assembly;
             this.CompilationOutputPdbFile = outputPaths.CompilationFiles.PdbPath;
+
+            var compilationOptions = projectContext.ProjectFile.GetCompilerOptions(targetFramework: projectContext.TargetFramework, configurationName: configuration);
+            this.EmitEntryPoint = compilationOptions.EmitEntryPoint;
 
             var sourceFiles = new List<string>();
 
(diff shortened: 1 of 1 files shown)


src/OmniSharp.DotNet/Models/DotNetProjectInformation.cs
```diff
@@ -9,12 +9,16 @@ public DotNetProjectInformation(ProjectContext projectContext, string configurat
...
             this.Path = projectContext.RootProject.Path;
             this.Name = projectContext.ProjectFile.Name;
+            this.TargetFramework = new DotNetFramework(projectContext.TargetFramework);
 
             var outputPaths = projectContext.GetOutputPaths(configuration);
             this.CompilationOutputPath = outputPaths.CompilationOutputPath;
             this.CompilationOutputAssemblyFile = outputPaths.CompilationFiles.Assembly;
             this.CompilationOutputPdbFile = outputPaths.CompilationFiles.PdbPath;
 
+            var compilationOptions = projectContext.ProjectFile.GetCompilerOptions(targetFramework: projectContext.TargetFramework, configurationName: configuration);
```
filipw: maybe this extension method would be better https://github.com/dotnet/cli/blob/7c6a2ad8b359e94ea35cf1fbe69b2d24da43bd34/src/Microsoft.DotNet.Compiler.Common/ProjectContextExtensions.cs#L20?

DustinCampbell: Why?

filipw: looks like it's gonna include the warning suppressions CLI is trying to suppress by default, so you wouldn't want them showing up in language services I guess

DustinCampbell: Sure, but we're only using it to get at a single property that this doesn't affect. Wouldn't it be very broken if I have to call an extension method to get the "right" values?

DustinCampbell: Note that the DotNetProjectSystem already suppress the very same warnings as https://github.com/dotnet/cli/blob/7c6a2ad8b359e94ea35cf1fbe69b2d24da43bd34/src/Microsoft.DotNet.Compiler.Common/DefaultCompilerWarningSuppresses.cs.

Look here: https://github.com/OmniSharp/omnisharp-roslyn/blob/dev/src/OmniSharp.DotNet/DotNetProjectSystem.cs#L322-L324

filipw: ah

troydai:
:shipit:

DustinCampbell:
@david-driscoll -- are you OK if I go ahead and merge this?

david-driscoll:
yeah, should be fine.

Issue #431, opened 2016-03-14
[dev] Build fails on Windows if 7z isn't on path


panopticoncentral:
PublishProject also crashes, so it should probably handle the situation gracefully as well.

DustinCampbell:
This was fixed long ago.

PR #464, opened 2016-03-23
Update build scripts

troydai:
1. Enable stop packing through modify `build.json`
2. Skip install `dotnet` on Windows. Use the local one.

### DIFF
--- a/tools/PublishProject/Program.cs
+++ b/tools/PublishProject/Program.cs
@@ -2,6 +2,7 @@
 using System.Diagnostics;
 using System.IO;
 using System.IO.Compression;
+using System.Linq;
 using System.Text.RegularExpressions;
 using Microsoft.Extensions.PlatformAbstractions;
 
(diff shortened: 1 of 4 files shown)


tools/PublishProject/Program.cs
```diff
@@ -80,69 +83,80 @@ public static int Main(string[] args)
...
-            if (runtimeString.Contains("osx-")) {
-                return;
+            // Rename framework name to mono for non-windows platforms
+            if (runtimeString.Contains("ubuntu-") || runtimeString.Contains("osx-"))
+            {
+                buildIdentifier.Replace("dnx451", "mono");
             }
             
-            // Simplify Ubuntu to Linux
-            runtimeString = runtimeString.Replace("ubuntu", "linux");
```
troydai: @hal-ler @david-driscoll what's the original intention here? Ubuntu != Linux since the implementation packages are different among distros.

tools/PublishProject/Program.cs
```diff
@@ -80,69 +83,80 @@ public static int Main(string[] args)
...
-            var buildIdentifier = $"{runtimeString}-{framework}";
-            // Linux + dnx451 is renamed to Mono
-            if (runtimeString.Contains("linux-") && framework.Equals("dnx451"))
-                buildIdentifier ="linux-mono";
-            // No need to package OSX + dnx451
-            else if (runtimeString.Contains("osx-") && framework.Equals("dnx451"))
```
troydai: @hal-ler I add it back. DNX451 is needed on OSX for MSBuild scenarios.

tools/PublishProject/Program.cs
```diff
@@ -80,69 +83,80 @@ public static int Main(string[] args)
...
+
                 if (PlatformServices.Default.Runtime.OperatingSystemPlatform == Platform.Windows)
                 {
+                    // if the build is running on Windows, use 7z
+                    var probeStartInfo = new ProcessStartInfo("where", "7z") { UseShellExecute = false };
+                    var probe = Process.Start(probeStartInfo);
+                    if (!probe.WaitForExit(1000) || probe.ExitCode != 0)
+                    {
+                        throw new InvalidOperationException($"7z is not installed on this machine. It is required for packaging.");
+                    }
```
troydai: https://github.com/OmniSharp/omnisharp-roslyn/issues/431

tools/PublishProject/Program.cs
hal-ler: Just to keep with the proposed naming convention.

tools/PublishProject/Program.cs
hal-ler: But is it needed to have both platforms, since it builds against mono? The previous packaging just had omnisharp-mono for both platforms (without OS designation).

tools/PublishProject/Program.cs
troydai: @david-driscoll thoughts?

tools/PublishProject/Program.cs
troydai: I get your point. But I'm not sure if `Ubuntu-Mono` and `OSX-Mono` is exchangeable. Did we test this before?

david-driscoll: In the past the mono build has been cross-platform, that is anything that can run mono can run the mono `exe`

david-driscoll: I do run some tests that run for both linux and mac with `omnisharp-node-client` which defaults to mono today, and they haven't been failing, so it should be fine.

troydai: Ok. I'll tune this logic.

troydai:
Unnecessary.

PR #467, opened 2016-03-24
[cake-build] Initial version of portable Cake build script.

hal-ler:
This is my initial suggestion of a Cake-based build script #396.

Updated to use build.json from current build system.

Usage:
build.(ps1|sh) are used to retrieve the Cake bootstrapping script and to launch it. build.cake includes the portable build script itself.

Cake info:
Cake currently has no direct support for .NET Core CLI (cake-build/cake#674). For now I call the tools as external processes, which works well, especially since I can generate output logs to make it functionally equivalent to the current build script.
While Cake is cross-platform, it is not built against dnxcore50, so it requires mono to work on Unix :(.

Definitions:
-Local build: build only against local runtime (or closest match)
1) If local default runtime is in build.json, then only build against it
2) Find first match using version-less OS name between default runtime and list in build.json (win10->win7, osx.10.10->osx.10.11, etc.)
3) Report error if none found, cannot build locally

Targets:
-target=Default : Same as Local target
-target=Local : Full local build with restore and test
-target=All : Full build for all runtimes listed in build.json
-target=Quick : Equivalent to "-quick" build (only performs local build, no restore, no test)
-target=Install : Quick + install artifacts into folder configured in build.cake (similar to "-install")
Options:
-configuration={Release|Debug} : Configuration to use for build and publish
-test-configuration={Release|Debug} : Configuration to use for test
-install-path={Path} : Destination folder for install target


david-driscoll:
@hal-ler sorry, I had to flip around the builds because osx isn't behaving, this should be building after the other builds finish.

hal-ler:
@troydai 
To continue the discussion on `dotnet restore`.

From what I understand, the requirements are:
-Cross-compile against a set of pre-defined runtimes described in `build.json` (custom build configuration file) and/or `OmniSharp\project.json`.
-Compile against the runtime of the local machine, which might not be in `OmniSharp\project.json` (such as win10-x64).
-Portable library is a nice option, but self-contained packages are still required.

We are looking for the official long-term solution for this problem.

My current solution looks for the closest match between the local runtime and the list described in `build.json` (which is identical to `OmniSharp\project.json`). I use this runtime for the local build, which works on Windows and OSX for now, but I would prefer a less hackish solution.

Pinging @anurse as he seems knowledgeable about this behavior in `NuGet` / `dotnet restore`.

hal-ler:
Browsing the NuGet repository suggests to me that `RestoreArgs.Runtimes` is still used as an extension over the runtimes described in `project.json`. Unfortunately that part of RestoreArgs was disabled as a command-line option. Maybe it could be re-introduced with updated documentation?

In NuGet.Core / NuGet.Commands / RestoreCommand / RequestFactory / RestoreArgs.cs 
`request.RequestedRuntimes.UnionWith(Runtimes);`

In NuGet.Core / NuGet.Commands / Utility / RequestRuntimeUtility.cs:
`runtimes.UnionWith(request.Project.RuntimeGraph.Runtimes.Keys);`
`runtimes.UnionWith(request.RequestedRuntimes);`

david-driscoll:
Getting this in. and we'll get it all fixed up to move to dev shortly

PR #476, opened 2016-03-24
Add Cake build script

david-driscoll:


david-driscoll:
@hal-ler merged with `dev`, CI will probably take a little bit.

hal-ler:
Me and my optimism... We might need to add back the following to build.sh.

`# Handle to many files on osx `
`if [ "$TRAVIS_OS_NAME" == "osx" ] || [ `uname` == "Darwin" ]; then`
`    ulimit -n 4096`
`fi`

dmccaffery:
Should have used condo :-P, although it still has a dependency on mono because sake has a dependency on mono -- i plan to break that dependency to only rely on coreclr soon :)

david-driscoll:
So far looking good, I'll pull it down later and mess with it, otherwise it looks pretty good.

david-driscoll:
I made a couple of tweaks, so I'm going to let that run and check in tonight / tomorrow.

@hal-ler thanks a bunch for sticking with us through this crazy transition to `dotnet cli`, the effort you've brought to this is greatly appreciated!

build.cake
```diff
@@ -0,0 +1,581 @@
...
+    if (buildPlan.Rids.Contains(localRuntime))
+    {
+        return localRuntime;
+    }
+    else
+    {
+        return buildPlan.Rids.First(runtime => {
+            var localRuntimeWithoutVersion = Regex.Replace(localRuntime, "(\\d|\\.)*-", "-");
+            var runtimeWithoutVersion = Regex.Replace(runtime, "(\\d|\\.)*-", "-");
+            return localRuntimeWithoutVersion.Equals(runtimeWithoutVersion);
```
hal-ler: You can use this to restrict to local RIDs by using Find instead of First.

david-driscoll: Then it won't build both x86 and x64 versions on windows.  I guess I could use that to further limit it for linux and mac, at least until (if?) x86 versions become a thing.

hal-ler: Sorry, I meant Where not Find (still getting used to Linq). It would keep all suffixed variants including -x64/-86 or OS version number.

hal-ler: Sorry, forgot about how I designed the regex.

david-driscoll: No worries, you raised a good point! :+1:

PR #479, opened 2016-03-25
make DotnetVersion optional, unpin dotnetversion.  Added setup task t…

david-driscoll:
…o better restrict types.  Reacted to --version -> --info from dotnet

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -12,7 +12,6 @@ var testConfiguration = Argument("test-configuration", "Debug");
 var installFolder = Argument("install-path", IsRunningOnWindows() ?
                         $"{EnvironmentVariable("USERPROFILE")}/.omnisharp/local" : "~/.omnisharp/local");
 
-
 // Working directory
 var workingDirectory = (new CakeEnvironment()).WorkingDirectory;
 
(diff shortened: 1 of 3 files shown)


build.cake
```diff
@@ -249,21 +269,31 @@ Task("BuildEnvironment")
...
     DownloadFile($"{buildPlan.DotNetInstallScriptURL}/{installScript}", scriptPath);
-    if (!IsRunningOnWindows())
+    string installArgs = "";
+
+    if (IsRunningOnWindows())
+    {
+        installArgs = $"{buildPlan.DotNetChannel}";
+        if (!String.IsNullOrEmpty(buildPlan.DotNetVersion))
+            installArgs = $"{installArgs} -version {buildPlan.DotNetVersion}";
+        if (!buildPlan.UseSystemDotNetPath)
```
david-driscoll: I wish these were consistent between platforms... *cough* *cough*

david-driscoll:
cc @troydai @DustinCampbell @hal-ler

hal-ler:
You can also use "Latest" as the version string to unpin, without having to add an additional check in build.cake..

build.cake
hal-ler: Oh I just assumed that the guys followed standard *NIX argument naming convention where full argument names use "--". But it looks like they also recently added the Windows convention with "-version" and such.

troydai:
:shipit:

PR #485, opened 2016-04-01
Updated to new frameworks and CLI + extra sugaring in build script

hal-ler:
Updated to the latest CLI and new framework names/packages.

Updating to the most recent CLI requires updating the version of `"Microsoft.Extensions.DependencyModel"`, which brings with itself dependencies on a newer NETStandard library. Since changes were required to the `project.json`, I decided to also update the framework monikers.

Changes:

- dnx451->net451

- dnxcore50->netcoreapp1.0 for OmniSharp and tests and appropriate netstandard for the others

- netstandard projects depend on `NETStandard.Library` (and other libraries if needed)

- netcoreapp1.0 projects depend on `NETStandard.Library` and `Microsoft.DotNet.ProjectModel` (and other libraries if needed)
    - NETCore.App brings extra unneeded dependencies including conflicting CodeAnalysis

- Changed `NETStandard.Library`, `Microsoft.DotNet.*`, `Microsoft.Extensions.DependencyModel` , `System.*` dependency versions to use wild-cards for now (see below)

- Fixed issues rising from new libraries

- Tried to eliminate some of the unnecessary dependencies in the `project.json` files

- Clean-up of `build.cake`
    - I also removed Cake addins as they are third-party components targeting .NET 4.5, thus might not be available on a .NET Core version of Cake

- Updated `.tools` path to match the recent change in moving `cake-bootstrap.*` into `scripts`

- New utility target in `build.cake` named `SetPackageVersions`

Things to fix long-term:
- Eliminate imports for netstandard, especially dnxcore50

     - Newtonsoft.Json only targets dnxcore50, this should change as soon as rc2 is released

     - `Microsoft.Extensions.DependencyModel` pulls in `NuGet.Packaging`, which pulls in `System.Security.Cryptography.Hashing.Algorithms`, which only targets `dnxcore50` @troydai do you know any plans about this one?

**SetPackageVersions**

It uses a new file called `depversion.json` as a central dictionary to replace all the dependency versions from `project.json` files across the different folders with the one specified within the file. This can be used to quickly migrate to new package versions. Dependencies without entries in `depversion.json` are left alone.

This target is not executed as part of the build, but can be called separately, by using `-target=SetPackageVersions` with `build.{ps1|sh}`.

For now `depversion.json` contains the packages that I specified with wild-card in this PR. This allows the wildcards to be quickly replaced by any version you guys prefer. Long-term the other dependencies can also be added here.

### DIFF
--- a/build.json
+++ b/build.json
@@ -1,30 +1,30 @@
 {
     "UseSystemDotNetPath" : "false",
     "DotNetFolder": ".dotnet",
-    "DotNetInstallScriptURL": "https://raw.githubusercontent.com/dotnet/cli/78e60bcb8b06c7787a9eb0356cf9ab3f1e76b86e/scripts/obtain",
+    "DotNetInstallScriptURL": "https://raw.githubusercontent.com/dotnet/cli/rel/1.0.0/scripts/obtain",
     "DotNetChannel": "beta",
-    "DotNetVersion": "1.0.0.001897",
+    "DotNetVersion": "Latest",
     "BuildToolsFolder": ".tools",
     "ArtifactsFolder": "artifacts",
     "TestProjects": {
         "OmniSharp.Bootstrap.Tests": [
-            "dnxcore50",
-            "dnx451"
+            "netcoreapp1.0",
+            "net451"
         ],
         "OmniSharp.MSBuild.Tests": [
-            "dnx451"
+            "net451"
         ],
         "OmniSharp.Roslyn.CSharp.Tests": [
-            "dnxcore50"
+            "netcoreapp1.0"
         ],
         "OmniSharp.Stdio.Tests": [
-            "dnxcore50",
-            "dnx451"
+            "netcoreapp1.0",
+            "net451"
         ]
     },
     "Frameworks": [
-        "dnx451",
-        "dnxcore50"
+        "net451",
+        "netcoreapp1.0"
     ],
     "Rids": [
         "osx.10.11-x64",
--- a/src/OmniSharp.Abstractions/project.json
+++ b/src/OmniSharp.Abstractions/project.json
@@ -13,7 +13,8 @@
     "Microsoft.Extensions.Configuration.Binder": "1.0.0-rc2-20143",
     "Microsoft.Extensions.FileSystemGlobbing": "1.0.0-rc2-20143",
     "Microsoft.Extensions.Logging.Console": "1.0.0-rc2-20143",
-    "Microsoft.Extensions.PlatformAbstractions": "1.0.0-rc2-20143"
+    "Microsoft.Extensions.PlatformAbstractions": "1.0.0-rc2-20143",
+    "System.Reflection.Metadata": "1.2.0"
   },
   "frameworks": {
     "net451": {
--- a/src/OmniSharp.Abstractions/project.json
+++ b/src/OmniSharp.Abstractions/project.json
@@ -11,14 +11,12 @@
     "Microsoft.Extensions.Caching.Memory": "1.0.0-rc2-20143",
     "Microsoft.Extensions.Configuration": "1.0.0-rc2-20143",
     "Microsoft.Extensions.Configuration.Binder": "1.0.0-rc2-20143",
-    "Microsoft.Extensions.DependencyModel": "1.0.0-beta-001718",
     "Microsoft.Extensions.FileSystemGlobbing": "1.0.0-rc2-20143",
     "Microsoft.Extensions.Logging.Console": "1.0.0-rc2-20143",
-    "Microsoft.Extensions.PlatformAbstractions": "1.0.0-rc2-20143",
-    "System.Reflection.Metadata": "1.2.0"
+    "Microsoft.Extensions.PlatformAbstractions": "1.0.0-rc2-20143"
   },
   "frameworks": {
-    "dnx451": {
+    "net451": {
       "frameworkAssemblies": {
         "System.Runtime": "",
         "System.Threading": "",
@@ -33,14 +31,17 @@
         "System.Xml.Linq": ""
       }
     },
-    "dnxcore50": {
-      "imports": "portable-net45+win8",
+    "netstandard1.3": {
+      "imports": [
+        "dotnet5.4",
+        "dnxcore50",
+        "portable-net45+win8"
+      ],
       "dependencies": {
-        "NETStandard.Library": "1.0.0-rc2-23910",
-        "System.Dynamic.Runtime": "4.0.11-rc2-23911",
-        "System.Diagnostics.Process": "4.1.0-rc2-23911",
-        "System.IO.FileSystem.Watcher": "4.0.0-rc2-23911",
-        "System.Threading.Thread": "4.0.0-rc2-23911"
+        "NETStandard.Library": "1.5.0-*",
+        "System.Diagnostics.Process": "4.1.0-*",
+        "System.IO.FileSystem.Watcher": "4.0.0-*",
+        "System.Threading.Thread": "4.0.0-*"
       }
     }
   }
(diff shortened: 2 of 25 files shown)


hal-ler:
@troydai @david-driscoll This is my own personal hack to try and get it to work. My understanding of the `project.json` system is still limited so I might have made some major mistakes, but I think there are some useful bits here.

build.json
```diff
@@ -1,30 +1,30 @@
...
-            "dnx451"
+            "netcoreapp1.0",
+            "net451"
         ]
     },
     "Frameworks": [
-        "dnx451",
-        "dnxcore50"
+        "net451",
+        "netcoreapp1.0"
```
troydai: I thought we decide to keep our application as standalone, in that case we shouldn't use this tfm.

src/OmniSharp.Abstractions/project.json
```diff
@@ -11,14 +11,12 @@
     "Microsoft.Extensions.Caching.Memory": "1.0.0-rc2-20143",
     "Microsoft.Extensions.Configuration": "1.0.0-rc2-20143",
     "Microsoft.Extensions.Configuration.Binder": "1.0.0-rc2-20143",
-    "Microsoft.Extensions.DependencyModel": "1.0.0-beta-001718",
     "Microsoft.Extensions.FileSystemGlobbing": "1.0.0-rc2-20143",
     "Microsoft.Extensions.Logging.Console": "1.0.0-rc2-20143",
-    "Microsoft.Extensions.PlatformAbstractions": "1.0.0-rc2-20143",
-    "System.Reflection.Metadata": "1.2.0"
```
troydai: This is added intentionally to ensure Roslyn pick up the right version.

src/OmniSharp.Abstractions/project.json
```diff
@@ -33,14 +31,17 @@
...
+        "dnxcore50",
+        "portable-net45+win8"
+      ],
       "dependencies": {
-        "NETStandard.Library": "1.0.0-rc2-23910",
-        "System.Dynamic.Runtime": "4.0.11-rc2-23911",
-        "System.Diagnostics.Process": "4.1.0-rc2-23911",
-        "System.IO.FileSystem.Watcher": "4.0.0-rc2-23911",
-        "System.Threading.Thread": "4.0.0-rc2-23911"
+        "NETStandard.Library": "1.5.0-*",
```
troydai: Either we use wildcard everywhere or we use fix version every where..

src/OmniSharp.Bootstrap/project.json
```diff
@@ -13,19 +13,21 @@
...
+    "net451": {
       "frameworkAssemblies": {
         "System.Runtime": ""
       }
     },
-    "dnxcore50": {
-      "imports": "portable-net45+win8",
+    "netstandard1.5": {
+      "imports": [
+        "dotnet5.6",
```
troydai: Needed?

src/OmniSharp.Bootstrap/project.json
```diff
@@ -13,19 +13,21 @@
...
+      ],
       "dependencies": {
-        "NETStandard.Library": "1.0.0-rc2-23910",
-        "System.Dynamic.Runtime": "4.0.11-rc2-23911",
-        "System.Linq.Expressions": "4.0.11-rc2-23911",
-        "System.Security.Cryptography.Algorithms": "4.0.0-rc2-23907",
-        "System.Xml.XDocument": "4.0.11-rc2-23911"
+        "NETStandard.Library": "1.5.0-*",
+        "System.Dynamic.Runtime": "4.0.11-*",
+        "System.Security.Cryptography.Algorithms": "4.1.0-*"
```
troydai: This version need to be pinned down. There is flawed dependency introduced by other sources.

src/OmniSharp.Host/Startup.cs
```diff
@@ -130,12 +130,12 @@ public void ConfigureServices(IServiceCollection services)
...
             Func<RuntimeLibrary, bool> shouldLoad = lib => lib.Dependencies.Any(dep => dep.Name == "OmniSharp.Abstractions" ||
                                                                                        dep.Name == "OmniSharp.Roslyn");
-                       
+
             var assemblies = DependencyContext.Default
                                               .RuntimeLibraries
                                               .Where(shouldLoad)
-                                              .SelectMany(lib => lib.Assemblies)
-                                              .Select(each => loader.Load(each.Name))
+                                              .SelectMany(lib => lib.RuntimeAssemblyGroups.GetDefaultGroup().AssetPaths)
```
troydai: This is not right. I'll send out a fix for this.

troydai:
I'm working on `troy/update` branch. I'll pin the version once I got the test work. I began working on it last night and didn't notice you were working on it. https://github.com/OmniSharp/omnisharp-roslyn/tree/troy/update

troydai:
Most of the "redundant" dependencies look useless but they have their purpose. They were added at some point of the development to mitigate wrongfully introduced dependencies by other packages. They may not be necessarily now, but you have to verify them one by one. Remove one and restore to check the warning. There should be no unexpected package upgrade or downgrade.

build.json
hal-ler: Which is the standalone tfm? I assumed `netcoreapp1.0` is the new `netstandardapp`. The corefx repo can be a bit confusing on this topic.

src/OmniSharp.Abstractions/project.json
hal-ler: I was planning on using the new `SetPackageVersions` command to pin it to a certain version defined as good.

src/OmniSharp.Bootstrap/project.json
hal-ler: The dnxcore50 import might hide the need for it so I don't know, but I would love to get rid of the dnxcore50 import first.

src/OmniSharp.Host/Startup.cs
hal-ler: It seemed to me that they tried deprecating the Assemblies property, but it would be great if you could look into it.

src/OmniSharp.Abstractions/project.json
hal-ler: I removed it due to a package downgrade warning during restore.

`Detected package downgrade: System.Reflection.Metadata from 1.3.0-rc2-23931 to 1.2.0`

hal-ler:
Please ignore the portable name in the branch, I removed those parts from this PR and just wanted to focus on the dependency and CLI update.

troydai:
The test won't work no matter what due to an internal inconsistency between ASP.NET and dotnet CLI. The dotnet packages used by two sides are not aligned cause issues in dependency resolve. OmniSharp relies on both side's packages the problem is unsolvable. 

At this point we're heading to rc2 finish line, all the packages will be aligned eventually. I suggest we disable the tests temporary and leave E2E tests only.

src/OmniSharp.Bootstrap/project.json
hal-ler: Also the new one?

`Detected package downgrade: System.Security.Cryptography.Algorithms from 4.1.0-rc3-24001-00 to 4.0.0-rc2-23907`

troydai: There shouldn't be rc3 packages.

troydai:
I'll suggest not to combine multiple changes in one PR. We either upgrade all the versions and open another PR for __ SetPackageVersions__ work. Combine them together lead to confusions.

src/OmniSharp.Bootstrap/project.json
troydai: good.

troydai:
:shipit: I'll update the `DependencyModel` piece afterwards.

hal-ler:
@troydai Could you please create a local branch that I can redo the PR against? I can do fixed versions there (based on what the wild-cards gave me) and skip the SetPackageVersions bits. That would allow you to fine-tune it before it goes to dev.

troydai:
https://github.com/OmniSharp/omnisharp-roslyn/pull/486

@hal-ler are you able to make changes to `update` branch?

PR #488, opened 2016-04-02
Updated RID logic to remove need for dotnet-info

hal-ler:
New virtual RID within the build script called `default`. When targeting `default`, `dotnet publish` is called with no explicit runtime, thus allowing it to select the best match based on the list specified in `project.json`. When targeting `default` and trying to package, the environment variable `OMNISHARP_PACKAGE_OSNAME` is used instead of the short runtime identifier. `default` should always be in the list of Rids! The targets `Install` and `TestPublished` use `default` as it is the only one guaranteed to work locally.

RestrictToLocal sets the runtime list to `default`.

RestrictToEnvironment was changed to PopulateRuntimes and should populate the list of Rids, based on the OS. For now it sets the runtime list to `default` + `win7-x86` on Windows. Better logic could be used in the future, but this works for now without `dotnet --info`.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -59,58 +59,6 @@ var logFolder = $"{artifactFolder}/logs";
 var packageFolder = $"{artifactFolder}/package";
 
 /// <summary>
-///  Retrieve the default local RID from .NET CLI.
-/// </summary>
-/// <param name="dotnetcli">Full path to .NET CLI binary</param>
-/// <returns>Default RID</returns>
-string GetLocalRuntimeID(string dotnetcli)
-{
-    var process = StartAndReturnProcess(dotnetcli,
-        new ProcessSettings
-        { 
-            Arguments = "--info",
-            RedirectStandardOutput = true
-        });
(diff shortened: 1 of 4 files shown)


hal-ler:
@david-driscoll Win-x86 .NET Core build fails on Win-x64 with `Failed to bind to coreclr`. I could repro it with the last release as-well. Is this expected behavior?

hal-ler:
I based this design on the need to perform builds against the local RID first and foremost (faster CI, less space used on CI machine, easier to build and install locally). Cross-compilation can be added by extending `PopulateRuntimes`, potentially with information from `build.json` (Such as a WinCrossCompileRids array).

david-driscoll:
Looks good!

Needs to be updated (I merged the other changes), if you're around in the next few hours and can update it, awesome! Otherwise I have a rebased branch at https://github.com/OmniSharp/omnisharp-roslyn/tree/portable that I'll merge to dev later.

hal-ler:
Rebased.

PR #490, opened 2016-04-02
Update to remove exception for codeaction at end of line.

### LINKED ISSUE #296: Throws an exception when requesting code actions when cursor is at the end of the line
This might be an atom problem sending through the wrong cursor position but I don't think so.

When you request code actions for a given token but the cursor is at the end of the line it throws the following exception

```
An unhandled exception has occurred: Diagnostic must have span '[1269..1269)'
Parameter name: diagnostics
System.ArgumentException: Diagnostic must have span '[1269..1269)'
Parameter name: diagnostics
  at Microsoft.CodeAnalysis.CodeFixes.CodeFixContext.VerifyDiagnosticsArgument (ImmutableArray`1 diagnostics, TextSpan span) [0x00000] in :0 
  at Microsoft.CodeAnalysis.CodeFixes.CodeFixContext..ctor (Microsoft.CodeAnalysis.Document document, TextSpan span, ImmutableArray`1 diagnostics, System.Action`2 registerCodeFix, Boolean verifyArguments, CancellationToken cancellationToken) [0x00000] in :0 
  at Microsoft.CodeAnalysis.CodeFixes.CodeFixContext..ctor (Microsoft.CodeAnalysis.Document document, TextSpan span, ImmutableArray`1 diagnostics, System.Action`2 registerCodeFix, CancellationToken cancellationToken) [0x00000] in :0 
  at OmniSharp.Api.V2.CodeActionController+d__9.MoveNext () [0x00000] in :0 
```

hal-ler:
Fixes #296 

Based on the following snippet in Roslyn (`CodeFixContext.cs`).

`/// All the diagnostics must have the same <paramref name="span"/>.`


hal-ler:
Best of both worlds now. If you have a precise span, it will only match exactly. If you only have a point-wise location within a span, then it searches for the first match and returns all other diagnostics matching said span.

hal-ler:
Checked in VS Code and the exception is now gone.

david-driscoll:
@troydai  @DustinCampbell any input, otherwise looks good to me

src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/CodeActionHelper.cs
```diff
@@ -47,7 +47,18 @@ public static class CodeActionHelper
             var diagnostics = semanticModel.GetDiagnostics();
             var location = GetTextSpan(request, sourceText);
```
DustinCampbell: Probably better to call this "span" since there's a Roslyn type called "Location" which is being accessed below.

src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/CodeActionHelper.cs
```diff
@@ -47,7 +47,18 @@ public static class CodeActionHelper
...
+            if (!pointDiagnostics.Any())
+            {
+                var firstMatchingDiagnostic = diagnostics.FirstOrDefault(d => d.Location.SourceSpan.Contains(location));
+                // Try to find other matches with the same location as the first approximate match
+                if (firstMatchingDiagnostic != null)
+                {
+                    pointDiagnostics = diagnostics.Where(d => d.Location.SourceSpan.Equals(firstMatchingDiagnostic.Location.SourceSpan)).ToImmutableArray();
+                }
+            }
             if (pointDiagnostics.Any())
```
DustinCampbell: Note that this is an `ImmutableArray<Diagnostic>` and not an `IEnumerable<Diagnostics>`. Since it's a value type, you could test for `pointDiagnostics.Length > 0` and `pointDiagnostics[0].Location.SourceSpan` below if you don't want to box the struct.

src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/CodeActionHelper.cs
```diff
@@ -47,7 +47,18 @@ public static class CodeActionHelper
...
             {
                 return new CodeFixContext(originalDocument, pointDiagnostics.First().Location.SourceSpan, pointDiagnostics, (a, d) => actionsDestination.Add(a), CancellationToken.None);
```
DustinCampbell: Should we pass the span we're testing for rather than the first span of the first diagnostic here? (I'm actually not sure).

hal-ler: All spans within pointDiagnostics should be equal at this point, but we don't know if it was the original span or the firstMatchingDiagnosts one.

DustinCampbell: I see

src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/CodeActionHelper.cs
hal-ler: Didn't know that, just used the old code for reference. I will update to your suggestion then.

src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/CodeActionHelper.cs
hal-ler: Good idea.

hal-ler:
Applied feed-back.

david-driscoll:
I think this is good to go, I haven't tested it in an editor yet, anyone else have any opinions?  @DustinCampbell @troydai

troydai:
looks good to me, too.

david-driscoll:
Thanks again @hal-ler !

PR #493, opened 2016-04-03
Zero based index support

david-driscoll:
This is something I've wanted to do for a while, because it's a pain to be translating to and from zero-based indexes both in the server and in the client.

For historical reasons, the default will remain `1-based` so we don't break anyone.  This PR adds a new flag `--zero-based-indices` (and I've added docs :scream:).

The server internals are all now zero-based, with all existing unit tests passing as expected.  I've added a `JsonConverter` that depending on the global `Configuration` will serialize out as one based the columns, lines, endlines, etc.  The converter also the does same for values coming into the server.

This is a non-breaking change from the outside, but editors that are zero-based by default (atom, vscode, sublime maybe?) can transition to using zero-based indexes if they want to.


src/OmniSharp.Abstractions/Models/v1/ChangeBufferRequest.cs
```diff
@@ -1,4 +1,6 @@
 ﻿
```
troydai: Remove the empty line.

src/OmniSharp.Abstractions/Models/v1/ChangeBufferRequest.cs
```diff
@@ -7,9 +9,13 @@ namespace OmniSharp.Models
     public class ChangeBufferRequest : IRequest
     {
         public string FileName { get; set; }
+        [JsonConverter(typeof(ZeroBasedIndexConverter))]
```
troydai: This line introduce two namespaces into the source code. Can we have a attribute like this to avoid it?

```
namespace Omnisharp.Json
{
    public class ZeroBasedIndexConverterAttribute: JsonConverterAttribute
    {
        public ZeroBasedIndexConverterAttribute: base(typeof(ZeroBasedIndexConverter))
        {
        }
    }    
}
```

troydai:
My only concern is the about the performance. Not sure the perf implication of implementing this as a Newtonsoft.Json converter. Will it greatly increase the allocation?

src/OmniSharp.Abstractions/Models/v1/ChangeBufferRequest.cs
david-driscoll: JsonConverter is sealed, and there is no way, that I'm aware of to extend it so Newtonsoft will pick it up.  That was my first though too.

david-driscoll:
I haven't noticed any big perf problems but frankly the test size is too tiny to tell for sure.

There may be some problems, but I don't know how much more than we already have.

Right now the flow is something like:

`Request -> (convert to zero based) -> Roslyn -> (convert to one based) -> Response`

with this change... with `--zero-based-indices`

` Request -> Roslyn -> Response`

without `--zero-based-indices`

`Request (convert to zero-based) -> Roslyn -> Response (convert to zero based)`

So it's basically the same flow, we're just doing the work in a slightly difference place.  ie Serialization vs in the handlers.

david-driscoll:
anyone else have any comments, otherwise I'll rebase and get this merged in the next day or two.

@DustinCampbell ?

Issue #481, opened 2016-03-28
omnisharp-roslyn server fails to start properly


martin-sweeney:
When using roslyn and vim together in omnisharp the following message is shown whenever the server is started. I am using the automatic starting of the server as decribed in the installation notes on github.

/var/folders/39/w91g26n56ld31s3spxy1cwg00000gn/T/vHyf7fm/5:48: no such file or directory: /Users/jtbmanagementserviceslts/.vim/bundle/omnisharp-vim/omnisharp-roslyn/scripts/Omnisharp
--- Press ENTER to continue ---

[Steps]
- Download from github (using pathogen method)
- install vim-dispatch
- install syntastic
- install youcompleteme
- install ctrl.p
- cd ~/.vim/bundle/omnisharp-vim/omnisharp-roslyn
- run  ./build.sh
- download yeoman / yeoman aspnet generator
- run yo aspnet 
- choose 'web application'
- run  dnu restore
- run  dnu build (2242 errors here - if you use coreclr / mono compiles)
- run dnx web
- vim startup.cs - to load a c# file and vim roslyn fails in background
# Additional information:

 ./install.sh fails when you run from scripts directory in omnisharp-roslyn
# 
# python

$python -h
usage: /usr/local/Cellar/python/2.7.11/Frameworks/Python.framework/Versions/2.7/Resources/Python.app/Contents/MacOS/Python [option]
# 
# My .vimrc

"let g:OmniSharp_server_type='v1'
let g:OmniSharp_server_type = 'roslyn'
" If you are using the omnisharp-roslyn backend, use the following
let g:syntastic_cs_checkers = ['code_checker']
# 
# dnvm settings 

$dnvm list
Active              Version Runtime Architecture OperatingSystem Alias

---

```
            1.0.0-beta4    mono                    linux/osx
      1.0.0-rc1-update1 coreclr          x64          darwin default
      1.0.0-rc1-update1    mono                    linux/osx
      1.0.0-rc1-update1 coreclr          x64          darwin default
      1.0.0-rc1-update1 coreclr          x64           linux
      1.0.0-rc1-update1    mono                    linux/osx
        1.0.0-rc2-16595 coreclr          x64          darwin
        1.0.0-rc2-16595    mono                    linux/osx
 *      1.0.0-rc2-20221 coreclr          x64          darwin
        1.0.0-rc2-20221    mono                    linux/osx
```

david-driscoll:
[...]

PR #494, opened 2016-04-03
Added auto-generated run scripts.

hal-ler:
Brings back old functionality to allow users to easily use the result of a build without having to know the folder layout / framework monikers. (#481)

scripts/OmniSharp{.cmd} calls into .NET 4.5.1 OmniSharp
scripts/OmniSharp.Core{.cmd} calls into .NET Core OmniSharp

The scripts are generated/updated with every build or install. After install they point to the installed path.

TestPublished uses these scripts to test the result of the build.


hal-ler:
Merged in updates from yesterday.

david-driscoll:
I like this, I need more time to review it, but I like it!

build.cake
```diff
@@ -57,6 +58,7 @@ var artifactFolder = $"{workingDirectory}/{buildPlan.ArtifactsFolder}";
 var publishFolder = $"{artifactFolder}/publish";
 var logFolder = $"{artifactFolder}/logs";
 var packageFolder = $"{artifactFolder}/package";
+var scriptFolder = $"{workingDirectory}/scripts";
```
troydai: how about move this into artifacts folder. this keeps the convention that all content under artifacts are disposable.

build.cake
```diff
@@ -130,6 +132,68 @@ void DoArchive(string runtime, DirectoryPath inputFolder, FilePath outputFile)
 }
 
 /// <summary>
+///  Generate the scripts which target the OmniSharp binaries.
+/// </summary>
+/// <param name="outputRoot">The root folder where the publised (or installed) binaries are located</param>
+void CreateRunScript(string outputRoot, string scriptFolder)
```
troydai: i think we put too many logic in the cake file, which make it gradually become unmanageable. we have limited intellisense for cake script. can we build this logic in a separate app or script or task?

build.cake
```diff
@@ -130,6 +132,68 @@ void DoArchive(string runtime, DirectoryPath inputFolder, FilePath outputFile)
...
+{
+    if (IsRunningOnWindows())
+    {
+        var desktopScript = $"{scriptFolder}/OmniSharp.cmd";
+        var coreScript = $"{scriptFolder}/OmniSharp.Core.cmd";
```
troydai: Use `System.IO.Path.Combine` instead of string interpolation for path.

build.cake
```diff
@@ -470,24 +535,34 @@ Task("TestPublished")
...
         // Wait 10 seconds to see if project terminates early with error
         bool exitsWithError = process.WaitForExit(10000);
         if (exitsWithError)
         {
-            throw new Exception($"Failed to run {project} / default / {framework}");
+            throw new Exception($"Failed to run {script}");
+        }
+        // Need to kill script recursively on Windows
+        if (IsRunningOnWindows())
+        {
```
troydai: doesn't `process.Kill()` do the trick?

troydai:
⌚

build.cake
david-driscoll: all the more reason to make an omnisharp plugin that supports cake. :smile:

david-driscoll:
Looks good to me, couple pieces from troy, but otherwise :+1:

build.cake
hal-ler: Makes sense as long as people are made aware of the change.

build.cake
hal-ler: I could not get it to work on Windows, because killing the shell process does not kill the child process.

build.cake
hal-ler: Tasks cannot take arguments, so I might have to play around with explicit state variables to get it working with a separate task, but that can be confusing (difficult to infer the current state within any Task). 

In Cake you can have separate script files, but the way they are loaded is similar to the C include directive. This means that Tasks in separate files use the global state in the main file without having explicit references to it. I would prefer to keep Tasks in the main file, but we can move all helper functions into their own files (they only interact with the build via their arguments due to Mono restrictions anyway).

I could even port the helpers to ScriptCs, but that introduces an additional tooling to the system.

@david-driscoll I could also look into creating one based on the ScriptCs plugin. Do you think that is feasible for someone who is still not versed in the internals of OmniSharp?

david-driscoll: `#load` and such is supported by cake, I'm fine with trying to split things up, as long as it makes sense to.

Making `build.cake` less daunting is not a bad thing.

build.cake
troydai: How does editors terminate OmniSharp process on Windows then?

hal-ler: I see that omnisharp-vscode does it similarly:
https://github.com/OmniSharp/omnisharp-vscode/blob/bc40a7f4bb3b4f1cd9016c8148e0ab71acdce28a/src/omnisharpServer.ts

I would love it if I could learn an alternative though.

troydai: In `dotnet-projectmodel-server` we have a mechanism to let the server terminate itself. Here's how it works: the parent host's pid is sent to the server when it is started, in our case is the editor's pid. The server monitor the editor process and terminate when editor exits. Too big a change here though.

@david-driscoll how does atom deal with this?

hal-ler: I will migrate this ugly code as an extension method together with some wrappers like your Run methods to simplify the core script.

build.cake
hal-ler: I just played around with System.IO.Path.Combine and unfortunately Cake uses `/` on Windows.
This results in weird file paths when combining the two, like `C:/Work/coreclr/omnisharp-roslyn\artifacts\publish\OmniSharp\default\`. I will try to get rid of all the Cake paths used by the script, but this should be kept in mind for future work as-well.

troydai: ok, let's keep consistent then.

hal-ler:
Updated based on feed-back and discussion.

Functional change:

Scripts moved to `artifacts\scripts`

Refactoring:

Due to the lack of IntelliSense support it is not easy for people to use the Cake aliases and their behavior is many times contrary to that of C# core libraries (for example the discrepancies in Path format and Process creation). I feel that it is best to use "pure" C# instead, because the aliases do not save that much code anyway. This also allows future ports to ScriptCs for example. This update attempts to eliminate as many of the Cake aliases as possible.

I replaces all file-system accesses to use ``System.IO` and `System.IO.Path.Combine`. This works well, without much additional code, but unfortunately Cake redefines the classes Path, File and DIrectory, with no way to not import those (or at least I could not find any yet). As a result all operations are prefixed with `System.IO` right now. `using System.IO.Path` and `var a = Combine(b, c);` would work, but I feel that it can be a little confusing.

I used the Run wrappers written lately as a basis for a more generic family of wrappers, which replaces all existing process creation. The code is put into an external helper script now to keep the main script clean. Any feed-back is welcome here as I just want to have a uniform and clean way for us to create processes.

The remaining Cake aliases are:
- Argument
- HasArgument
- IsRunningOnWindows
- Information

hal-ler:
Had to add back the `Zip` alias. Cake currently does not allow referencing system assemblies from the script (in this case `System.IO.Compression.FileSystem`).

hal-ler:
@david-driscoll @troydai Any ideas/opinions on the update?

david-driscoll:
The only thing I can think of, is would it move the new cake files under `scripts`?  That way we can remove a little bit of the clutter from the root directory.

hal-ler:
Sure.

PR #496, opened 2016-04-04
[wip] initial run at cli based plugins

david-driscoll:
Currently this fails to run properly on CoreClr... I can't determine why, other than it the AssemblyLoadContext cannot resolve the AssemblyName for some reason.

The overall goal, once this is done...

* Packages will all have the same name, and the same default plugins
* A new package `omnisharp-bare` will be introduced, that doesn't have any plugins installed by default.
* A new set of archives (zip, nupkg, tar.gz?? I don't really know) will be added to releases as well.
* New command-line argument `--plugin` that allows you to declare plugin load directories.
  ie. `--plugin /path/to/MsBuild, `--plugin /path/to/CakePlugin`, etc.

After this is done and in, the next step is to look at potential out of process plugins, I thought I might attempt it at some point using TypeScript as a target, not because there aren't lots of TypeScript editors out there (there are!) but because I understand the language services a little better, so it'll be easier to build out a solid example of what a "OOP" plugin would look like.


hal-ler:
Subscribing to this as I also plan to use AssemblyLoadContext for a .NET Core version of Cake.

hal-ler:
Played around with it today and the issue is that Trusted Platform Assemblies cannot be loaded with AssemblyLoadContext. They are managed by Assembly.Load instead. However I'm still uncertain about how to list the Trusted Platform Assemblies for the current executable. Maybe DependencyContext.Default.GetDefaultAssemblyNames() could do the trick.

troydai:
@david-driscoll about out-of-process plugins, i tried on my `troy/sap` branch. it is a lot of work. the hards parts are:
1. Flow information into roslyn workspace, especially for compilation options;
2. Flow in and out these event emitted for restore and project operations.
It's doable, but really need to make sure it's stability.

david-driscoll:
Have a new PR #640

PR #502, opened 2016-04-05
Update NuGet feeds

troydai:


hal-ler:
Just a heads up, the build is starting to pull in rc3 parts from AspNetCore.

troydai:
Ha? that is bad. Any idea now what package pulls in `rc3` packages?

hal-ler:
I think AspNetCore just auto-updates to 1.0.0-rc3-20553.

hal-ler:
Is there something like rc2-final that we could pin and be sure that it exists on the aspnet feed?

troydai:
@hal-ler I can't answer that for obviously reason :). We're getting there.

.gitignore
```diff
@@ -35,6 +35,7 @@ scripts/Omnisharp*
 
 # NuGet
 *.nuget.targets
+.packages/
```
troydai: this is not needed. I'll remove it.

PR #503, opened 2016-04-05
Fix omnisharp.json configuration loading

troydai:
Currently, when `omnisharp.json` exists in the project folder, Omnisharp crashes and exits. This is because the base path is set to the `ConfigurationBuilder`, which points to the `ApplicaitonBase` instead of the root folder of the project .


filipw:
yeah noticed the crash too!

troydai:
thanks, man.

PR #511, opened 2016-04-07
Adding .vscode/settings.json into .gitignore

dopare:
Just added `.vscode` folder into ignore list.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,7 @@ nuget.exe
 debugSettings.json
 buildlog
 /.vs
+/.vscode/settings.json
 *.lock.json
 /omnisharp*.tar.gz
 


david-driscoll:
I'm fine with leaving the folder in, I'm sure eventually we'll have some perfectly valid vscode tasks when developing with omnisharp. It's fine to keep it that way locally if you want, but I'm not certain we will always need to ignore it.  cc @DustinCampbell @troydai

DustinCampbell:
If we have specific needs for OmniSharp in the launch.json, I think we'd want to check it in. If we don't, it shouldn't be a big deal to ignore these.

dopare:
Sorry, I haven't thought about shared use of `launch.json` .

I am using latest `omnisharp-roslyn` build in VS code for `omnisharp-rolsyn` development via `settings.json`.
What do you think about ignoring just the `settings.json` ?

troydai:
I think ignore `settings.json` makes more senses. Personalized setting can be saved in user setting instead of workspace setting.

dopare:
Made changes to ignore only `settings.json`.

DustinCampbell:
That sounds like the right fix to me. :+1:

troydai:
:shipit:

troydai:
The shadow copy issue happens more frequently now.

troydai:
Thanks for the contribution, @dopare.

dopare:
@troydai Thank you so much, it's my first step in open source.

troydai:
Cheers!

Issue #469, opened 2016-03-24
OmniSharp should search for a project if neither project.json or global.json exists in the given path


troydai:
The `DotNetProjectSystem` throws when non project is found. https://github.com/OmniSharp/omnisharp-roslyn/blob/dev/src/OmniSharp.DotNet/DotNetProjectSystem.cs#L109

1. The exception type is wrong;
2. Missing projects should not cause error.

DustinCampbell:
Maybe. I'm kind of OK with there needing to be a global.json or project.json at the the given path. However, I'm *really* disliking how the whole project.json selection works in VS Code. I want to gut that completely. :smile:

troydai:
Thanks @willl :)

halter73:
Can we follow the restore logic for finding project.jsons recursively from the root/solution directory.

ASP.NET projects don't have samples in tests in global.json for example.

troydai:
Fixed in https://github.com/OmniSharp/omnisharp-roslyn/commit/c36e8c4e55014bebc5295d70831371262b1ccb46

PR #518, opened 2016-04-13
Use ProjectSearcher to locate initial batch of projects.

troydai:
- If there is a project.json at current location, loads it and its dependencies.
- If there is a global.json at current localtion, use search paths in it to local all project.json files. And add their dependnecies.
- If there is neither of them, recursively looking for all project.json. The recurisive searching will stop at any level when a project.json is located.

Issue #469


src/OmniSharp.DotNet/Projects/ProjectSearcher.cs
```diff
@@ -0,0 +1,82 @@
...
+        
+        private static IEnumerable<string> FindProjectsThroughGlobalJson(string root)
+        {
+            GlobalSettings globalSettings;
+            if (GlobalSettings.TryGetGlobalSettings(root, out globalSettings))
+            {
+                return globalSettings.ProjectSearchPaths
+                                     .Select(searchPath => Path.Combine(globalSettings.DirectoryPath, searchPath))
+                                     .Where(actualPath => Directory.Exists(actualPath))
+                                     .SelectMany(actualPath => Directory.GetDirectories(actualPath))
```
dopare: Is there any chance that project.json may be located not in the first level subdirectory ?

troydai: This is the same logic we use since `DNX`: if global.json presents, we only look the first level.

david-driscoll: In this case I don't think so.  The search paths is fairly specific, you define a set a directories, to look for project directories in.  As far as I know that's fairly specific to how cli manages projects.

dopare: Thanks, its clear for me now.
Another question about using full directory in global.json for linking external projects e.g. MVC project, etc. Should this logic consider this case?

troydai: You don't use full path in global.json. In the case you want include MVC source code, you modify your global.json to be:
``` JSON
{
  "projects": [ "src", "../MVC/src" ]
}
```
_Assuming MVC repo is cloned at the sample level of your web app_.

troydai: In above sample, the relative path will be joined with global.json's directory and correctly locate the MVC samples. This is the same logic we use in DNX and dotnet for expanding global.json search paths.

dopare: As far as I remember `global.json` allowed full directories. Maybe I have missed something.
Thank you for update.

david-driscoll:
So my only concern is deep nesting.

If for example someone has no `global.json`, and in the solution directory, there is no `project.json`.

What happens if they also happen to have a `node_modules` directory, lets make it worst case scenario, their `npm --version < 3`.  Now we're desperately looking for `project.json` in anything under the directory they have us.

Say for example they had some projects under `/src` and `/test`, once we get into `FindProjects` we're now commited to search until we can't search any more.  I see no place that this bails out until the stack is empty.  If no `project..json` in the folder, it'll just push on more any folders in that one.

If we changed `else` to `else if (!result.Any())` then we have a decent bail out scenario, but then we can't handle cases like:

* `src/project/project.json`
* `test/subdira/project/project.json`

Which means we still need to do some deep searching... then things get really confusing if they have additional `project.json` projects deep down that tree (node_modules, junctions, or otherwise).

Also is there a chance we would then hit the wonderful `MAX_PATH` limit on windows, in the C# side of the system?

src/OmniSharp.DotNet/Projects/ProjectSearcher.cs
david-driscoll: you are correct, it may not always be a lookup path, it could very well be an absolute path.

troydai:
We will need to inventing something additionally. One approach could be ignore the well-known folder like `node_modules`.

It is an issue we didn't solve in `dotnet` either. Consider this situation happens to `dotnet restore`.

troydai:
Another approach is to define a max depth switch. Like 5 levels of folder. User can override is in the omnisharp.json.

david-driscoll:
Max depth would be fine, just so we avoid digging too deeply.

tests/OmniSharp.DotNet.Tests/project.json
```diff
@@ -0,0 +1,24 @@
...
+    "compilationOptions": {
+        "warningsAsErrors": true
+    },
+    "dependencies": {
+        "OmniSharp.DotNet": "1.0.0-*",
+        "TestCommon": "1.0.0-*",
+        "xunit": "2.1.0"
+    },
+    "frameworks": {
+        "netstandardapp1.5": {
```
hal-ler: Is it best to use netstandardapp or netcoreapp here? I think it would be best to use the same everywhere to avoid confusing people.

tests/OmniSharp.DotNet.Tests/project.json
```diff
@@ -0,0 +1,24 @@
...
+            "dependencies": {
+                "dotnet-test-xunit": "1.0.0-dev-*",
+                "NETStandard.Library": "1.5.0-rc2-*"
+            },
+            "imports": [
+                "dnxcore50",
+                "portable-net45+win8"
+            ]
+        }
+    },
```
hal-ler: Are runtimes not needed for the tests anymore?

tests/OmniSharp.DotNet.Tests/project.json
```diff
@@ -0,0 +1,24 @@
...
+    "dependencies": {
+        "OmniSharp.DotNet": "1.0.0-*",
+        "TestCommon": "1.0.0-*",
+        "xunit": "2.1.0"
+    },
+    "frameworks": {
+        "netstandardapp1.5": {
+            "dependencies": {
+                "dotnet-test-xunit": "1.0.0-dev-*",
+                "NETStandard.Library": "1.5.0-rc2-*"
```
hal-ler: I would pin the versions as that was the policy for the other projects as-well.

tests/OmniSharp.DotNet.Tests/project.json
troydai: I plan to keep `DotNetSystem` tests portable only for now.

tests/OmniSharp.DotNet.Tests/project.json
troydai: I'll do that.

troydai:
@hal-ler note the change in test related targets and how test projects are compiled

troydai:
Ping @david-driscoll @hal-ler

src/OmniSharp.DotNet/Projects/ProjectSearcher.cs
troydai: Note in this logic absolute path is accept, too. `Path.Combine` will return the absolute path if the second parameter is an absolute path (https://msdn.microsoft.com/en-us/library/fyy7a5kt(v=vs.110).aspx)

david-driscoll:
LGTM @DustinCampbell ?

tests/TestCommon/project.json
```diff
@@ -0,0 +1,9 @@
+{
+    "version": "1.0.0-*",
+    "dependencies": {
+        "NETStandard.Library": "1.5.0-rc2-*"
```
hal-ler: TestCommon is a dependency of DotNet.Tests so I assume the same versions should be pinned.

DustinCampbell:
LGTM

hal-ler:
Happy to see that portable tests are working now. One note from the logs: TestCore seems to also build .NET 4.5.1 binaries for the test folders. It makes no practical difference right now and the code is simpler this way so it is fine as long as people are aware of it.

tests/TestCommon/project.json
troydai: Will upgrade.

troydai:
@david-driscoll something wrong with one of the CI machine. Can I merge this?

PR #519, opened 2016-04-13
added builds for additional runtimes, linux is now an alias for ubuntu

david-driscoll:


david-driscoll:
cc @DustinCampbell this may require a change to how vscode looks up what OS it's running on, you'll have to infer the distro based on similar logic that `cli` / `coreclr` use.

hal-ler:
What will be the logic for populating the the runtime list in the script with the other linux distros? Everything done on Travis/Ubuntu and otherwise stick to default runtime?

DustinCampbell:
Thanks @david-driscoll -- it will change VS Code slightly, but it's a small change and makes things better. At the moment, VS Code will grab the "linux" distro, which fails to launch on RHEL/CentOS.

david-driscoll:
@hal-ler my hope (that needs to be tested) is that even though it was built on `trusty` (travis) that they will run correctly on `RHEL` / `CentOS` because of the different runtime provided.  I'm assuming the binary format of `OmniSharp` doesn't change.

If that isn't the case... then I guess I have to setup some sort of Azure VMs to shuffle the build bits too.

hal-ler:
I think that it should work. Just make sure to also populate the list with the new runtimes (PopulateRuntimes in build.cake).

PR #524, opened 2016-04-14
Run restore target during install, to automatically retry

david-driscoll:
This is an attempt to reduce the common errors caused when building... things that are fixed by just running the command again.

As per https://blog.travis-ci.com/2013-05-20-network-timeouts-build-retries/
Travis will retry any command in the `install` step.  So in this scenario we will also restore.


build.cake
```diff
@@ -252,13 +252,23 @@ Task("Restore")
...
+    p.WaitForExit();
+    var exitCode = p.GetExitCode();
+
+    if (exitCode == 0)
+    {
+        Information("Package restore successful!");
+    }
+    else
     {
+        Error(string.Join("\n", p.GetStandardOutput()));
```
david-driscoll: should we parse GetStandardOutput for any `warn:` items and emit those as information, when it passes.

hal-ler: I was thinking about just using `dotnet restore -v Warning`.

hal-ler:
The Linux build fails because Cleanup creates the `artifacts/logs` folder for the tests and TestAll does not depend on it. There could be a separate Cleanup for tests, but maybe it is not worth the clutter. For OSX I have no idea, it seems that even NuGet is getting hung.

david-driscoll:
Yeah, I had a friend try to build for me on his mac, and it hung too.  It seems like it's a nuget issue of some sort.

david-driscoll:
or nuget and mono.

hal-ler:
Is there a way to get a failure in the retry stage to kill the second part of the build? It is weird to see that a build fails when the last call to `build.sh` finishes successfully. You have to start looking all over the log to find that the tests failed.

hal-ler:
The release of NuGet 3.4.3 kind of coincides with the last succesfull build on OSX (8 days ago). We could try replacing `latest` in the url from `cake-bootstrap` with `v3.3.0` (the previous version on the site).

david-driscoll:
@hal-ler nice catch... looks like things are passing now... I'll have to log a bug with nuget!

scripts/runhelpers.cake
```diff
@@ -151,6 +151,37 @@ ExitStatus Run(string exec, string args, RunOptions runOptions)
...
 
 /// <summary>
+///  Run restore with the given arguments
+/// </summary>
+/// <param name="exec">Executable to run</param>
+/// <param name="args">Arguments</param>
+/// <param name="runOptions">Optional settings</param>
+/// <returns>The exit status for further queries</returns>
+ExitStatus RunRestore(string exec, string args, string workingDirectory)
+{
```
hal-ler: I would refactor this to use Run internally to keep with the uniformity.

`var restoreLog = new List<string>();`
`var exitCode = Run(exec, args, new RunOptions`
 `                   {`
`                        WorkingDirectory = workingDirectory,`
`                        StandardOutputListing = restoreLog`
`                    });`
`if (exitCode != 0)`
`{`
`    Error(string.Join("\n", restoreLog));`
`}`
`return exitCode;`

hal-ler:
I would refactor RunRestore to use Run internally (just to keep with uniformity), but I can do that in a separate PR.

`var restoreLog = new List<string>();`
`var exitCode = Run(exec, args, new RunOptions`
`{`
`WorkingDirectory = workingDirectory,`
`StandardOutputListing = restoreLog`
`});`
`if (exitCode != 0)`
`{`
 `Error(string.Join("\n", restoreLog));`
`}`
`return exitCode;`

PR #525, opened 2016-04-14
Refinement of archive generation

hal-ler:
Proposal for simpler archiving logic. I'm also including existing functionality to make the behavior clear and easy to discuss.

Windows: nothing changes, archives generated with the `win-x64-{framework}` and `win-x86-{framework}` (for non-Local builds) identifiers.

Linux/Users: nothing changes, archives generated with the `linux-64-{framework}` identifiers for the local runtime.

Linux/Travis:
  - archive generated with the `mono` identifier for the local runtime using .NET 4.5.1 as framework.
  - archives generated with the `{osshortname}-x64` identifier for all the runtimes explicitly mentioned in PopulateRuntimes (none for the local one, we expect it to be in the list) using .NET Core as framework.

OSX/User: nothing changes, archive generated with the `osx-x64-{framework}` identifier for the local runtime.

OSX/Travis: nothing changes, archive generated with the `osx-x64` identifier for the local runtime using .NET Core as framework.

The differences are:
  - `linux-mono` -> `mono` to clarify that it works on OSX.
  - No more double archiving for local runtime and explicit Ubuntu. Safeguard to ignore OS variant on Travis/Linux. Users still get archive if they desire, using generic `linux-x64-{framework}` name.
  - No more `linux` alias for .NET Core.

Suggestions? Opinions?

@DustinCampbell @david-driscoll @troydai

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -423,25 +423,24 @@ Task("OnlyPublish")
             }
 
             var buildIdentifier = $"{runtimeShort}-{framework}";
+            // Rename/restrict some archive names on CI
             // Travis/Linux + default + net451 is renamed to Mono
             if (string.Equals(EnvironmentVariable("TRAVIS_OS_NAME"), "linux") && runtime.Equals("default") && framework.Equals("net451"))
             {
-                buildIdentifier ="linux-mono";
+                buildIdentifier ="mono";
             }
-            // No need to archive for other Travis + net451 combinations
+            // No need to archive other Travis + net451 combinations
             else if (EnvironmentVariable("TRAVIS_OS_NAME") != null && framework.Equals("net451"))
+            {
+                continue;
+            }
+            // No need to archive Travis/Linux + default + not(net451) (expect all runtimes to be explicitely named)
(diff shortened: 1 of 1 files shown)


hal-ler:
Used wrong branch sorry.

hal-ler:
Used wrong branch initially, sorry.

david-driscoll:
`linux-mono` -> `mono` is breaking, but that's minor, and it's something I can get behind.

Everything else makes sense to me.

hal-ler:
One question is if the OS versions should be kept for the Linux builds. The Ubuntu 16.04 binary might be incompatible with Ubuntu 14.04. It could lead to another set of breaking changes in the future.

troydai:
os version should be in the builds' names. it is likely the implementation for os platforms will be various.

david-driscoll:
I think we're fine without the os version for the moment, mainly because I don't see any specific ubuntu packages for cli either.

When they react with multiple versions per OS, I think that's the time we should as well.  Either way we can do that as part of another PR.

PR #529, opened 2016-04-15
Handle strong-name key files from msbuild projects

jbevain:
This change adds support for the SignAssembly and AssemblyOriginatorKeyFile properties in MSBuild project.
It allows projects with InternalsVisibleTo attributes for strong named assemblies to have proper error reporting and intellisense.

### DIFF
--- a/src/OmniSharp.MSBuild/ProjectFile/ProjectFileInfo.cs
+++ b/src/OmniSharp.MSBuild/ProjectFile/ProjectFileInfo.cs
@@ -47,6 +47,10 @@ namespace OmniSharp.MSBuild.ProjectFile
 
         public OutputKind OutputKind { get; private set; }
 
+        public bool SignAssembly { get; private set; }
+
+        public string AssemblyOriginatorKeyFile { get; private set; }
+
         public static ProjectFileInfo Create(MSBuildOptions options, ILogger logger, string solutionDirectory, string projectFilePath, ICollection<MSBuildDiagnosticsMessage> diagnostics)
         {
             var projectFileInfo = new ProjectFileInfo();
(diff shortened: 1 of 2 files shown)


src/OmniSharp.MSBuild/ProjectFile/ProjectFileInfo.cs
```diff
@@ -133,6 +137,14 @@ public static ProjectFileInfo Create(MSBuildOptions options, ILogger logger, str
                 {
                     projectFileInfo.DefineConstants = defineConstants.Split(new[] { ';' }, StringSplitOptions.RemoveEmptyEntries).Distinct().ToList();
                 }
+
+                var signAssembly = projectInstance.GetPropertyValue("SignAssembly");
+                if (!string.IsNullOrWhiteSpace(signAssembly))
+                {
+                    projectFileInfo.SignAssembly = Convert.ToBoolean(signAssembly);
```
troydai: I'm a bit concern about the exception here (https://msdn.microsoft.com/en-us/library/86hw82a3(v=vs.110).aspx). Do we want to fail everything in the case of an incorrect string representation or can we just let it slip and send a warning.

I'm fine with both, just make sure we're all aware of the possibility.

troydai:
in addition to my only comments i hope we can have automation to cover this rare scenario. it is not tested frequently so automation coverage is important to stop regression.

expect that, all look good to me. can you rebase the change on the latest dev, it can't be merged because it is out of date now.

src/OmniSharp.MSBuild/ProjectFile/ProjectFileInfo.cs
jbevain: That's definitely a valid concern. In this case, I've reused the pattern that was already there for `AllowUnsafe`.

troydai: i see. well, then we can address similar issue together in another PR. i'm fine with this change now.

jbevain:
@troydai I haven't seen any automation/integration tests for MSBuild projects in the repo. Is that planned?

I've rebased the PR on top of origin/dev just after I pushed the PR.

troydai:
@jbevain it is not planned. we just didn't add tests before.

troydai:
@david-driscoll want to take another look?

david-driscoll:
I'm fine with the change, looks mergable, so lets go for it.

PR #538, opened 2016-04-26
PowerShell doesn't support double slash parameters, it would have bee…

wangzq:
…n failed when you have [CmdletBinding()] in cake-bootstrap.ps1, but since you have "ValueFromRemainingArguments" set to true for the "$ScriptArgs", the "--Experimental" will be captured by it instead.


wangzq:
oops, the comment has a typo: I meant "double dash parameters" instead of "double slash parameters"

hal-ler:
LGTM

PR #540, opened 2016-04-27
Update for API change in ASP.NET

hal-ler:


david-driscoll:
Going to roll this in, Cake / OSX is giving some trouble... trying to resolve that separately.

PR #560, opened 2016-05-23
Set event's delegate type as return type

dopare:
In [PR](https://github.com/OmniSharp/omnisharp-vscode/pull/164) I have added return type information into `completionItemProvider`. But the `ReturnType` of an `event` was returned `null` from omnisharp server.

I have changed to return `event`'s `delegate` type for `ReturnType` like it is implemented in Visual Studio's intellisense.

image

P.S. Waiting for [new icon](https://github.com/Microsoft/vscode/issues/5407) in VS Code to change event's icon.


david-driscoll:
LGTM

dopare:
rebased dev changes

PR #562, opened 2016-05-27
Add capability to discover and run xunit test

troydai:
### Demo

_it's a gif, may take time to load_
mn0xdmo - imgur

### Changes
Add endpoint for test discovery and test execution
Corresponding VSCode change to adapter the capability: https://github.com/OmniSharp/omnisharp-vscode/pull/382


src/OmniSharp.Abstractions/Models/v1/FileMemberElement.cs
```diff
@@ -10,6 +10,8 @@ public class FileMemberElement : IComparable<FileMemberElement>
         public QuickFix Location { get; set; }
 
         public string Kind { get; set; }
+        
+        public ICollection<string> Features { get; } = new List<string>();
```
david-driscoll: So this collection just contains metadata for the particular element? (in this case it's a test method).

It kind of feels like we should expand this a little bit.

What if we added a new class... (the name doesn't matter to me)

```
public enum FeatureType {
    TestMethod
}

public class FeatureItem
{
    public FeatureType Type { get; set; }
    public string Name { get; set; }
}
```

Here's my reasoning.  We could expand this out to include additional items such as has the test passed already, the last reference count (code lens).  This way we leave this object open for extension, should we choose to do so.

src/OmniSharp.Abstractions/Services/IMemberSyntaxFeaturesDiscover.cs
```diff
@@ -0,0 +1,12 @@
+using System.Collections.Generic;
+using Microsoft.CodeAnalysis;
+
+namespace OmniSharp.Abstractions.Services
+{
+    public interface ISyntaxFeaturesDiscover
+    {
+        bool NeedSemanticModel { get; }
+        
+        IEnumerable<string> Discover(SyntaxNode node, SemanticModel semanticModel);
```
david-driscoll: Thoughts if we made this method `Task<IEnumerable<?>>`?

david-driscoll: Actually looking at this again, we might not be able to do that due to the to the SyntaxWalker...

src/OmniSharp.Abstractions/Endpoints.cs
```diff
@@ -64,5 +59,9 @@ public static class V2
             public const string GetCodeActions = "/v2/getcodeactions";
             public const string RunCodeAction = "/v2/runcodeaction";
         }
+        
+        public const string GetTestStartInfo = "/getteststartinfo";
```
david-driscoll: So this replacing the old functionality we have for running tests, so it seems like this command replaces `/gettestcontext`.

For that reason, maybe we should put it under the V2 banner?  `/v2/gettestcontext`.

Thoughts?

src/OmniSharp.Abstractions/Endpoints.cs
```diff
@@ -64,5 +59,9 @@ public static class V2
...
+        
+        public const string RunDotNetTest = "/rundotnettest";
```
david-driscoll: I'm thinking this should go under V2 as well, because it is new.

Also the name, seems very specific, couldn't we name it simply `/v2/runtest`?

src/OmniSharp.DotNetTest/Helpers/TestFeaturesDiscover.cs
```diff
@@ -0,0 +1,56 @@
...
+        {
+            if (node is MethodDeclarationSyntax)
+            {
+                var method = node as MethodDeclarationSyntax;
+                if (IsTestMethod(method, semanticModel))
+                {
+                    var methodName = semanticModel.GetDeclaredSymbol(node).ToDisplayString();
+                    methodName = methodName.Substring(0, methodName.IndexOf('('));
+
+                    yield return $"XunitTestMethod:{methodName}";
```
david-driscoll: The underlying model is still `dotnet test`, so how do we support other frameworks like NUnit / MSTest (god forbid, lol).  And would having to support that make us come back here?

Maybe something as simple as a config setting in `omnisharp.json`?

david-driscoll:
I think that's all I've got for comments so far.

It looks awesome though.

src/OmniSharp.DotNetTest/Helpers/TestFeaturesDiscover.cs
troydai: That's why the endpoint is named `rundotnettest` to eliminate the ambiguity. Moving forward, we can:
1. add new endpoint for other types of test
2. make run test endpoint general to support extensible test provider.

but for now, I'd like to keep the scope narrow.

src/OmniSharp.Abstractions/Endpoints.cs
troydai: will change.

src/OmniSharp.Abstractions/Endpoints.cs
troydai: will change to `/v2/rundotnettest` for the reason in comments i left below.

src/OmniSharp.Abstractions/Services/IMemberSyntaxFeaturesDiscover.cs
troydai: yeah, that walker is synchronized.

src/OmniSharp.DotNetTest/Helpers/TestFeaturesDiscover.cs
david-driscoll: I guess my point is, `dotnet test` is inherently abstract, it just expects a tool to be available that will run tests.

My this is with adding this endpoint is once it's in, we can't drop it very easily.  Where it's easier to add a general one in, and extend that through `MEF` a little easier.

What makes `/rundotnettest` different than `/runtest`, besides the implementation.  I just don't want to see a world where we have `/rundotnettest` `/runmsbuildtest`, `/runfsharptest`, etc.

src/OmniSharp.Abstractions/Models/v1/FileMemberElement.cs
troydai: I get the benefit of a structure here. But the interface now is more extensible. Xunit finder can put whatever data in the collection so as other implementation of a `Feature Finder`. The problem with an enum for the type actually makes it hard to expand. For example, the Enum has to be defined in `OmniSharp.Abstractions`. Then every time a new feature is introduced the `OmhniSharp.Abstractions` will have to be updated.

src/OmniSharp.Abstractions/Endpoints.cs
troydai: will call it `/v2/getteststartinfo` since the function of the endpoint is narrowly defined.

src/OmniSharp.DotNetTest/Helpers/TestFeaturesDiscover.cs
troydai: make sense. I'll update the endpoint name. but I have not idea how other frameworks will be supported. guess we will find out later.

src/OmniSharp.Abstractions/Models/v1/FileMemberElement.cs
david-driscoll: It doesn't have to be an enum, we could simply make it a string instead, and that would work fine as well.

troydai:
@david-driscoll updated.

david-driscoll:
LGTM

Issue #542, opened 2016-04-28
'Unnecessary using directive' is falsely flagged when required for XML cref comments.


jrieken:
_From @lake-fall on April 28, 2016 9:55_

- VSCode Version: 1.0.0
- OS Version: WIndows 10 Version 1511

Steps to Reproduce:

1. Create a C# class.
2. Add an XML comment with `/// <see cref="SqlCommand.Parameters" />`
3. Add the using using `System.Data.SqlClient;` to the namespace.
4. The warning `Unnecessary using directive` appears.
5. This is in conflict to Visual Studio 2015 Community, where removing the using directive generates the warning `CS1574 XML comment has cref attribute 'SqlConnection' that could not be resolved`.

_Copied from original issue: Microsoft/vscode#5938_

jrieken:
_From @bpasero on April 28, 2016 12:45_

@lake-fall please report C# issues against the omnisharp repository.

hal-ler:
After reading about this part in Roslyn, it seems that the issue is caused because OmniSharp does request Roslyn to fully analyze the XML documentation. It can be solved by adding
`.WithXmlReferenceResolver(XmlFileResolver.Default)` to the CSharpCompilationOptions and `documentationMode: DocumentationMode.Diagnose` to the CSharpParseOptions, but it is not clear-cut if this behavior should be on by default because it also triggers Warning 1591 `Missing XML comment for publicly visible type or member` (and other potential side-effects I might not be aware of). Unfortunately it looks like `DocumnentationMode.Parse` is not enough to solve the issue. @david-driscoll Could this be triggered by some flags in `project.json` / `csproj`?

hal-ler:
I will be looking into this as soon as I have a bit of free time.

david-driscoll:
It's probably the analyzer doesn't "see" the reference in the xml comment, wonder if there is something roslyn can do, or if we have to parse the crefs manually.

hal-ler:
The changes posted above by myself make Roslyn track Xml comments nicely, but it is also aggressive in triggering warnings for missing comments. I plan to track the need for Xml documentation in the project system and trigger the additional analysis on demand.

hal-ler:
@lake-fall could you please try the code in the PR to see if it works without unexpected warnings and such on your project?

PR #564, opened 2016-05-29
Added tracking of XML comments via Roslyn (#542)

hal-ler:
Solves #542.

Only enabled when XML document generation is requested to avoid FxCop warnings. Should work for both `csproj` and `project.json` (tested locally on `project.json`).

Still needs unit-tests if you guys like the methodology.

### DIFF
--- a/src/OmniSharp.DotNet/DotNetProjectSystem.cs
+++ b/src/OmniSharp.DotNet/DotNetProjectSystem.cs
@@ -340,6 +340,12 @@ namespace OmniSharp.DotNet
             var parseOptions = new CSharpParseOptions(languageVersion: ParseLanguageVersion(option.LanguageVersion),
                                                       preprocessorSymbols: option.Defines);
 
+            if (option.GenerateXmlDocumentation ?? false)
+            {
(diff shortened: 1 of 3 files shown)


david-driscoll:
Changes look good to me.  If it's possible to get a decent Unit Test in there, that would be great.

cc @troydai @DustinCampbell

hal-ler:
I need to learn a bit about XUnit and using different project files with Roslyn, but it should be doable and I definitely want it with the code.

hal-ler:
After a lot of reading I figured out that the compilation options are read deep within the `ProjectSystem` initialization code. As such I did not find any simple way to test that code. I could look into refactoring the code to make it easier to test, but only if you guys are ok with the idea. The new design would have a static method in the `ProjectSystem` to infer the compilation and parsing option from the project file with a minimum number of dependencies.

david-driscoll:
If it's going to be too complicated, I wouldn't worry to much about it.  Most of the changes are just settings for Roslyn, so we should be comfortable that they have some of these scenarios tested.

david-driscoll:
I'm good with these changes, lets get them in.  :+1:

PR #565, opened 2016-05-29
Parameterize listening interface

jtbm37:
Sometimes the server is not on the same machine as the client. In order to reach it we need the ability to specify the listening interface.

For more details see https://github.com/OmniSharp/omnisharp-emacs/pull/223.


david-driscoll:
LGTM

PR #569, opened 2016-05-30
Fix deployment

david-driscoll:


david-driscoll:
This should fix the deployment issue we're seeing with releases :sheep:

hal-ler:
Looks good, but I would also remove the win7-x86 for now, until that issue is fixed to avoid confusing people. Also does anyone have access to a RedHat box by any chance to double-check that the cross-compiled Linux binaries work properly?

hal-ler:
LGTM

david-driscoll:
:+1:

PR #570, opened 2016-05-30
Fixed archive name with dots in name.

hal-ler:
`ChangeExtension` was interacting poorly with the dot in `netcoreapp1.5`. The resulting archive name looked weird in the release. Sorry for the bug.


hal-ler:
@david-driscoll Possible to take this into the release to get cleaner archive names?

david-driscoll:
LGTM

PR #580, opened 2016-06-06
do not apply changes for code actions through the workspace.  Instead…

david-driscoll:
… simply return the changes for the editor to handle.

cc @DustinCampbell @troydai 

This should fix one of the issues with using incremental changes in the text editor.  Appears to be working well in atom so far.

### DIFF
--- a/src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/RunCodeActionService.cs
+++ b/src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/RunCodeActionService.cs
@@ -44,7 +44,7 @@ namespace OmniSharp.Roslyn.CSharp.Services.Refactoring.V2
             var operations = await action.GetOperationsAsync(CancellationToken.None);
 
             var solution = _workspace.CurrentSolution;
-            IEnumerable<OmniSharp.Models.ModifiedFileResponse> changes = Enumerable.Empty<OmniSharp.Models.ModifiedFileResponse>();
+            var changes = Enumerable.Empty<OmniSharp.Models.ModifiedFileResponse>();
             var directoryName = Path.GetDirectoryName(request.FileName);
             foreach (var o in operations)
             {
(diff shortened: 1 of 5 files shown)


src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/RunCodeActionService.cs
```diff
@@ -43,16 +44,19 @@ public RunCodeActionService(OmnisharpWorkspace workspace, [ImportMany] IEnumerab
             var operations = await action.GetOperationsAsync(CancellationToken.None);
 
             var solution = _workspace.CurrentSolution;
+            IEnumerable<OmniSharp.Models.ModifiedFileResponse> changes = Enumerable.Empty<OmniSharp.Models.ModifiedFileResponse>();
+            var directoryName = Path.GetDirectoryName(request.FileName);
             foreach (var o in operations)
             {
-                o.Apply(_workspace, CancellationToken.None);
+                var applyChangesOperation = o as ApplyChangesOperation;
```
DustinCampbell: Does this mean that we will no longer run certain code actions that we still offer? I think that's the right thing since there are likely code actions that we're running with custom operations for displaying dialogs that can't be run. However, should there be a change elsewhere so that such code actions aren't offered?

david-driscoll: It's possible, I didn't dig too much into the code action code themselves.  I think all the ones that required additional dialog inputs have already been suppressed.  At the very least I've not seen one popup.

david-driscoll: To answer the original question, now that I've looked there a bunch of refactorings that are disabled in `CodeActionHelper`.   It appears they are filtered out from the initial list.

DustinCampbell: If that's the case, LGTM!

DustinCampbell:
:+1:

PR #598, opened 2016-06-29
config encoding For Stdio

nabychan:
vs code need utf-8 encoding for stdio,and other maybe not.

### DIFF
--- a/src/OmniSharp.Host/Program.cs
+++ b/src/OmniSharp.Host/Program.cs
@@ -107,7 +107,7 @@ namespace OmniSharp
 
             if (transportType == TransportType.Stdio)
             {
-                if(encoding != null)
+                if (encoding != null)
                 {
                     Console.InputEncoding = encoding;
                     Console.OutputEncoding = encoding;
--- a/src/OmniSharp.Host/Program.cs
+++ b/src/OmniSharp.Host/Program.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
+using System.Text;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
@@ -30,7 +31,7 @@ namespace OmniSharp
             var otherArgs = new List<string>();
             var plugins = new List<string>();
             var serverInterface = "localhost";
-            var encoding = System.Text.Encoding.Default;
+            Encoding encoding = null;
 
             var enumerator = args.GetEnumerator();
 
@@ -74,10 +75,10 @@ namespace OmniSharp
                     enumerator.MoveNext();
                     serverInterface = (string)enumerator.Current;
                 }
-                else if(arg == "--encoding")
+                else if (arg == "--encoding")
                 {
                     enumerator.MoveNext();
-                    encoding=System.Text.Encoding.GetEncoding((string)enumerator.Current);
+                    encoding = Encoding.GetEncoding((string)enumerator.Current);
                 }
                 else
                 {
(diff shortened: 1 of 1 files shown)


DustinCampbell:
Awesome -- this was on my list to go look at.

david-driscoll:
This is a good change!

Though it looks like `System.Text.Encoding.Default` doesn't exist for `netcoreapp1.0`, I'm not sure how to get the right default for the environment.  I remember seeing something about it at one point, but it's been a while.

DustinCampbell:
FWIW, getting the Default encoding is kind of a nightmare. Here's what we do in the Roslyn compiler: http://source.roslyn.io/#Microsoft.CodeAnalysis/EncodedStringText.cs,ff93fabe53148f02.

DustinCampbell:
I talked to "reluctant encoding guru" @jaredpar, and his advice was to not use Default at all. It's entirely possible for two processes to have different values for Default and these should be the same for client and server. IMO, we should just pick one to be our enforce it. I vote for UTF8.

DustinCampbell:
The other option is to just not set anything if the --encoding flag is not set.

src/OmniSharp.Host/Program.cs
```diff
@@ -30,6 +30,7 @@ public static void Main(string[] args)
             var otherArgs = new List<string>();
             var plugins = new List<string>();
             var serverInterface = "localhost";
+            var encoding = System.Text.Encoding.Default;
```
DustinCampbell: Add using for System.Text.

DustinCampbell: Since Encoding.Default is not available on .NET Core, it's probably best to just not set the encoding if the `--encoding` is not specified. So, set `encoding` to null here and then check for null later. E.g.

```C#
Encoding encoding = null;
```

src/OmniSharp.Host/Program.cs
```diff
@@ -73,6 +74,11 @@ public static void Main(string[] args)
                     enumerator.MoveNext();
                     serverInterface = (string)enumerator.Current;
                 }
+                else if(arg == "--encoding")
+                {
+                    enumerator.MoveNext();
+                    encoding=System.Text.Encoding.GetEncoding((string)enumerator.Current);
```
DustinCampbell: Add using for System.Text.

DustinCampbell: Please add spaces around the `=` to match the formatting in the rest of the file.

src/OmniSharp.Host/Program.cs
```diff
@@ -100,6 +106,8 @@ public static void Main(string[] args)
 
             if (transportType == TransportType.Stdio)
             {
+                Console.InputEncoding = encoding;
+                Console.OutputEncoding = encoding;
```
DustinCampbell: Check that encoding is not null before setting InputEncoding and OutputEncoding:

```C#
if (encoding != null)
{
    Console.InputEncoding = encoding;
    Console.OutputEncoding = encoding;
}
```

src/OmniSharp.Host/Program.cs
```diff
@@ -73,6 +74,11 @@ public static void Main(string[] args)
                     enumerator.MoveNext();
                     serverInterface = (string)enumerator.Current;
                 }
+                else if(arg == "--encoding")
```
DustinCampbell: Add space after `if` to match the formatting in the rest of the file.

DustinCampbell:
@nabychan: I'm *very* happy to merge this once you make the suggested changes and the CI build passes. Thanks a ton for your help!

jaredpar:
I think "reluctant encoding guru" roughly translates to "been burned by this before" 😄

nabychan:
@DustinCampbell OK, I'll have try to make it better

DustinCampbell:
:+1:

DustinCampbell:
Thanks again for your work on this!

src/OmniSharp.Host/Program.cs
```diff
@@ -100,6 +107,11 @@ public static void Main(string[] args)
 
             if (transportType == TransportType.Stdio)
             {
+                if(encoding != null)
```
jaredpar: nit: space between `if` and `(`

DustinCampbell:
Looks great @nabychan

nabychan:
Thanks @DustinCampbell . My first time to participate in projects from GitGub . Not that fluently .

PR #607, opened 2016-07-07
Pass -NoProfile when executing powershell scripts

DustinCampbell:
This avoids loading the default powershell profile when executing powershell scripts during Windows builds, since the user can define a default powershell profile that might do things like change the CWD.


david-driscoll:
LGTM

PR #622, opened 2016-08-16
Added support for text changes to codeformat endpoint.  

david-driscoll:
To opt in pass WantChanges=true.

This is a change to better be able to support the language server protocol.


DustinCampbell:
LGTM

PR #623, opened 2016-08-16
Fix IProjectSystem.GetProjectModel(...) API

DustinCampbell:
This method almost always returns null because all implementations assume that the path passed in is a document path. However, every caller within OmniSharp always pass a project file path. Now, if a document isn't found, each implementation attempts the path as a project path. This fixes the '/projects' endpoint when a file path is specified and the Roslyn workspace events emitted from the OmniSharp server, such as 'ProjectChanged'.

cc @david-driscoll


david-driscoll:
openssl is a know problem.

LGTM

PR #628, opened 2016-08-27
Fix OmniSharp --encoding switch to apply to both stdin and stdout

DustinCampbell:
When the --encoding switch was specified, it unintentionally only applied to stdin, but not stdout. This was due to the fact that the `SharedConsoleWriter` was instantiated before `Console.OutputEncoding` was set. Because `SharedConsoleWriter` initializes itself with the current value of `Console.Out`, it had the default encoding rather than the intended encoding.

A bit subtle.

cc @david-driscoll and @nabychan


david-driscoll:
:+1:

PR #637, opened 2016-09-17
Get unit tests running (and passing) again

DustinCampbell:
This change unskips all tests (except one -- see below) and fixes OmniSharp to make them pass:

* FormattingFacts: Two tests were skipped here because of a bug that had been fixed in Roslyn (https://github.com/dotnet/roslyn/issues/202). In addition, this test is impacted by the recent change to support zero-based indeces in the OmniSharp API. Since the tests don't go through the JSON serialization layer, they have to test against zero-based results. These tests have been updated with regard to their expectations and the formatting services and worker have been cleaned up.

* FixUsingFacts and SnippetFacts: Three tests were skipped between these because of missing references in the test OmnisharpWorkspace. In .NET Core, additional references are needed because many types are forwarded. If anything from the chain of type forwarders is not included, the compiler can't resolve the types.

In addition, I went ahead and pretty much rewrote the FixUsingsService and its associated Worker. It was doing *way* more work than was necessary by calling into every single CodeFixProvider, even those that didn't expect to be called with the diagnostics we're looking for fixes for! After updating to the latest Roslyn binaries locally, I found that it crashed on several CodeFixProviders that it shouldn't have been calling anyway. So, rewrite. It looks much cleaner now.

After this change, the only unit test that is still skipped is `Can_create_a_class_with_a_new_method_in_adjacent_file` which relies on NRefactory. We'll need to update this to use the appropriate Roslyn code fix in the future.


DustinCampbell:
cc @david-driscoll

src/OmniSharp.Roslyn.CSharp/Services/Formatting/FormatAfterKeystrokeService.cs
```diff
@@ -21,7 +20,7 @@ public class FormatAfterKeystrokeService : RequestHandler<FormatAfterKeystrokeRe
         public FormatAfterKeystrokeService(OmnisharpWorkspace workspace, FormattingOptions formattingOptions)
         {
             _workspace = workspace;
-            _options = OmniSharp.Roslyn.CSharp.Workers.Format.Formatting.GetOptions(_workspace, formattingOptions);
+            _options = FormattingWorker.GetOptions(_workspace, formattingOptions);
```
david-driscoll: Pesky namespaces!

david-driscoll:
Reviews + LGTM not mixing!

DustinCampbell:
thanks!

PR #641, opened 2016-09-21
Add configuration to make OmniSharp easier to test with MSBuild-based .NET Core projects

DustinCampbell:
Two settings have been added:

* dotnet.enabled: If set to false, the DotNetProjectSystem will not attempt to initialize and load any projects.
* msbuild.msbuildextensionpath: If set, the value will be passed into MSBuild as the MSBuildExtensionPath property.

### DIFF
--- a/src/OmniSharp.DotNet/DotNetProjectSystem.cs
+++ b/src/OmniSharp.DotNet/DotNetProjectSystem.cs
@@ -92,6 +92,18 @@ namespace OmniSharp.DotNet
 
         public void Initalize(IConfiguration configuration)
         {
+            bool enabled;
+            if (!bool.TryParse(configuration["enabled"], out enabled))
+            {
+                enabled = true;
+            }
+
+            if (!enabled)
+            {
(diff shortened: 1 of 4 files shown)


DustinCampbell:
@david-driscoll: it looks like the lgtm hook is still there.

david-driscoll:
I know I saw that last night.

david-driscoll:
I think I fixed it.  I'll look at this tonight.

DustinCampbell:
@david-driscoll, any chance you could sign off on this?

DustinCampbell:
FWIW, the DotNet project system tweak doesn't really quite work. The very presence of the DotNet project system causes a different version of Microsoft.NuGet.Core to get pulled in than the one used by the targets in the .NET Core SDK. So, I have to comment out the dependency on OmniSharp.DotNet in [OmniSharp\project.json](https://github.com/OmniSharp/omnisharp-roslyn/blob/dev/src/OmniSharp/project.json) to actually get it to work.

PR #646, opened 2016-09-26
Clean up!

DustinCampbell:
This PR addresses several problems:

1. Factors test helper code out of OmniSharp.Tests and into TestUtility
2. Adds OmniSharp.Tests to build.json so that it runs on builds
3. Fixes the "coding guidelines" tests
4. Cleans up all project.json files
5. Tweaks to only ask for code fixes when the provider supports one of the diagnostics under the caret. (Note: except for a special hacky case with Remove Unnecessary Usings.) This should improve the performance of the GetCodeActions end point.

cc @david-driscoll

### DIFF
--- a/OmniSharp.sln
+++ b/OmniSharp.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 14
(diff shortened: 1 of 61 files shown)


DustinCampbell:
Oops. Looks like I forgot to push my change to disable those tests. :smile:

PR #652, opened 2016-10-26
Move OmniSharp to net46

DustinCampbell:
This is necessary to take advantage of newer MSBuild bits in the near future.


DustinCampbell:
Sure, it'll definitely fail to build on machines that don't have a newer framework installed. However, 4.6.1 has extremely wide adoption, which is why [netstandard2.0](https://blogs.msdn.microsoft.com/dotnet/2016/09/26/introducing-net-standard/) is using 4.6.1 as its default surface area.

david-driscoll:
I figured as much, just wanted to make a note about it.

PR #659, opened 2016-11-04
Scripting revamp

filipw:
This PR is an attempt to revamp/normalize the C# scripting support in Omnisharp.

Here is what is done:
 - removed all dependencies on scriptcs - based on Roslyn directly. Omnisharp makes no assumptions about "scripting dialect" any more. It tries to be as close to "standard" as possible. Also renamed things like `ScriptCsProjectContext` to `ScriptProjectContext` etc etc
 - scripting intellisense is now compatible with `InteractiveScriptGlobals` - same as used by `csi.exe`. This gives us same set of global properties as there, i.e. `Args`.
 - scripting is (for now) supported only as .NET Core scripting
 - dependencies for scripts are no longer defined by the old `packages.config`, instead they are pulled from `project.json`
 - no implicit using statements or assembly references are made - except for `System` namespace, `mscorlib` and `System.Runtime` (if needed). 
 - suppress a couple of code check warnings which are not applicable to scripting code - `CS8099` & `CS1701`

This PR is in line with my proposal here https://github.com/OmniSharp/omnisharp-vscode/issues/23#issuecomment-200201767

This is a breaking change of course for scripting story in Omnisharp, but I believe it was never really officially supported but more of an experimental stuff. I believe trying to normalize to "pure" Roslyn scripting is a step in a right direction here, especially as we can easily align with runners like `csi.exe` then or [dotnet-script](https://github.com/filipw/dotnet-script)


filipw:
This PR also - from the scripting perspective - removes Omnisharp's dependency on full desktop framework

david-driscoll:
I am in favor of this change!  Will see if I have time to look at it before summit... if not we'll sit down at summit and talk about it :)

src/OmniSharp.Script/ScriptProjectSystem.cs
```diff
@@ -6,50 +6,40 @@
...
-                MetadataReference.CreateFromFile(typeof(object).GetTypeInfo().Assembly.Location),        // mscorlib
-                MetadataReference.CreateFromFile(typeof(Enumerable).GetTypeInfo().Assembly.Location),    // systemCore
-                MetadataReference.CreateFromFile(typeof(IScriptHost).Assembly.Location)                  // scriptcsContracts
-            };
-
-        OmnisharpWorkspace Workspace { get; }
-        IOmnisharpEnvironment Env { get; }
-        ScriptCsContext Context { get; }
-        ILogger Logger { get; }
+        private CSharpParseOptions CsxParseOptions { get; } = new CSharpParseOptions(LanguageVersion.CSharp6, DocumentationMode.Parse, SourceCodeKind.Script);
```
DustinCampbell: Use `LanguageVersion.Default` to ensure that the latest version of the language is used?

DustinCampbell:
Thanks for making this change @filipw! I definitely agree it's a step in the right direction.

src/OmniSharp.Script/ScriptProjectSystem.cs
```diff
@@ -137,29 +114,28 @@ private ProjectInfo CreateCsxProject(string csxPath, IScriptServicesBuilder base
                 return Context.CsxFileProjects[csxPath];
             }
 
-            // Process the file with ScriptCS first
             Logger.LogInformation($"Processing script {csxPath}...");
             Context.CsxFilesBeingProcessed.Add(csxPath);
-            var localScriptServices = baseBuilder.ScriptName(csxPath).Build();
-            var processResult = localScriptServices.FilePreProcessor.ProcessFile(csxPath);
+
+            var fileParser = new FileParser(Context.RootPath);
```
DustinCampbell: Any reason not to just parse the file with Roslyn?

filipw: to be honest I would love to, and I tried but couldn't find the right APIs. In fact - unless I am mistaken - the necessary APIs ([this?](https://github.com/dotnet/roslyn/blob/0e67586920f6dc2f27474d0d7201134909420c51/src/Compilers/CSharp/Portable/Compilation/SyntaxAndDeclarationManager.cs#L13)) are not available publicly. What we need here is to walk the dependency tree of `#load` directives and gather all the files and their using statements and so on so that we can construct `ProjectInfo` correctly. It seems the only way to publicly get there with Roslyn scripting APIs at the moment is to manually create a script compilation which I wanted to avoid

filipw:
> Thanks for making this change @filipw! I definitely agree it's a step in the right direction.

great! I also had a chat with Tomas yesterday and he suggested that it might be a better idea to treat all CSX files as single project, rather than each of them as a separate 1-file project (which is how it is implemented at the moment). I will try to incorporate this change too.

src/OmniSharp.Script/ScriptProjectSystem.cs
DustinCampbell: This can be done syntactically, correct? If so, SyntaxTree.ParseText(...) is what you're looking for.

filipw: yeah. The only "hurdle" is if I encounter a `#load foo.csx` directive, I need to follow it and pick up the `foo.csx` and that one might have it's own `#loads` and so on.

Scripting APIs would use [this](https://github.com/dotnet/roslyn/blob/master/src/Scripting/Core/ScriptSourceResolver.cs#L11) to resolve that whole dependency tree so maybe that would work, to parse the entry file and then follow it's `#loads` using `ScriptSourceResolver` recursively.

filipw: but that would be better than manual parsing, definitely

DustinCampbell: I guess I'm not seeing that as much of a hurdle. You could still have the file parser but use the SyntaxTree to find the using's, #load's, and #r's. You could keep the logic where it continues processing #load's. The real issue is that the simple text "parsing" isn't going to do all that good of a job. For example, it looks like it would return a weird malformed using if it countered a using static.

filipw: Right I see what you mean now. Of course, let me pluck the naive syntax parsing :+1:

DustinCampbell:
How is this going @filipw?

filipw:
It's coming!

filipw:
I've rebased this and introduced better parsing of files (no more naive parser, instead use `SyntaxTrees`) as suggested. Also improved some of the overall logic.

Tested in the following scenarios:
 - .net core single-file CSX script with nuget references in `project.json`
 - .net core multi-file CSX script  with nuget references in `project.json`
 - accessing script args using globals from `InteractiveScriptGlobals` (same global object as in `csi.exe`)
 - .net desktop CSX script with nuget references in `project.json` and extra GAC references in `project.json`
 - .net desktop CSX script with no nuget references (everything self contained via `#r`)

This is still a simple approach to the problem, but it's definitely an improvement over what's there at the moment. My suggestion would be to go ahead with this merge and I can progressively make this better as we go forward (for example I'd like to include [this](https://github.com/OmniSharp/omnisharp-roslyn/pull/659#issuecomment-259746769) in the future, though behaviorally things are OK already).

Finally, this solution works best on `net46` build of OmniSharp because when we are in `netcoreapp1.0` it's impossible to i.e. infer some of the assemblies without running into a lot of problems with `object has not been imported`. However my understanding is that `net46` will be the default mode of distributing OmniSharp

src/OmniSharp.Script/FileParser.cs
```diff
@@ -0,0 +1,60 @@
...
+            {
+                return;
+            }
+
+            result.LoadedScripts.Add(fullPath);
+
+            var scriptCode = File.ReadAllText(fullPath);
+
+            var syntaxTree = CSharpSyntaxTree.ParseText(scriptCode, CSharpParseOptions.Default.
+                WithPreprocessorSymbols("load", "r").
```
DustinCampbell: Is this line necessary? This would have the effect of implicitly adding the following code at that top of the tree:

``` C#
#define load
#define r
```

Is that what you meant?

src/OmniSharp.Script/FileParser.cs
```diff
@@ -0,0 +1,60 @@
...
+            var refs = syntaxTree.GetCompilationUnitRoot().GetReferenceDirectives().Select(x => x.File.ToString());
+            foreach (var reference in refs)
+            {
+                result.References.Add(reference.Replace("\"", string.Empty));
+            }
+
+            var loads = syntaxTree.GetCompilationUnitRoot().GetLoadDirectives().Select(x => x.File.ToString());
+            foreach (var load in loads)
+            {
+                var filePath = load.Replace("\"", string.Empty);
```
DustinCampbell: Does this work on OSX/Linux?

src/OmniSharp.Script/FileParserResult.cs
```diff
@@ -0,0 +1,20 @@
...
+    public class FileParserResult
+    {
+        public FileParserResult()
+        {
+            Namespaces = new HashSet<string>();
+            References = new HashSet<string>();
+            LoadedScripts = new HashSet<string>();
+        }
+
+        public HashSet<string> Namespaces { get; }
```
DustinCampbell: FYI that read-only auto-props can have initializers. So you could just have this with no constructor:

```C#
public HashSet<string> Namespaces { get; } = new HashSet<string>();
public HashSet<string> References { get; } = new HashSet<string>();
public HashSet<string> LoadedScripts { get; } = new HashSet<string>();
```

Your choice.

src/OmniSharp.Script/ScriptProjectSystem.cs
```diff
@@ -6,50 +6,40 @@
...
 using OmniSharp.Services;
-using ScriptCs;
-using ScriptCs.Contracts;
-using ScriptCs.Hosting;
-using LogLevel = ScriptCs.Contracts.LogLevel;
 
-namespace OmniSharp.ScriptCs
+namespace OmniSharp.Script
 {
     [Export(typeof(IProjectSystem))]
```
DustinCampbell: Add `Shared` attribute so that MEF creates this as a singleton.

src/OmniSharp.Script/ScriptProjectSystem.cs
```diff
@@ -67,48 +57,66 @@ public void Initalize(IConfiguration configuration)
...
-                    .Cache(false)
-                    .Repl(false)
-                    .ScriptName(Env.Path)
-                    .ScriptEngine<NullScriptEngine>();
+            // explicitly inherit scripting library references to all global script object (InteractiveScriptGlobals) to be recognized
+            var inheritedCompileLibraries = DependencyContext.Default.CompileLibraries.Where(x =>
+                    x.Name.ToLowerInvariant().StartsWith("microsoft.codeanalysis")).ToList();
 
-            var scriptServices = baseScriptServicesBuilder.Build();
+            var runtimeContexts = File.Exists(Path.Combine(Env.Path, "project.json")) ? ProjectContext.CreateContextForEachTarget(Env.Path) : null;
```
DustinCampbell: If I understand what this is doing, it'll probably need to change in the future for csproj .NET Core. I might not understand though. :smile:

src/OmniSharp.Script/ScriptProjectSystem.cs
```diff
@@ -67,48 +57,66 @@ public void Initalize(IConfiguration configuration)
...
 
-            var scriptPacks = scriptServices.ScriptPackResolver.GetPacks().ToList();
-            var assemblyPaths = scriptServices.AssemblyResolver.GetAssemblyPaths(Env.Path);
-
-            // Common usings and references
-            Context.CommonUsings.UnionWith(ScriptExecutor.DefaultNamespaces);
+            // if we have no context, then we also have no dependencies
+            // we can assume desktop framework
+            // and add mscorlib
+            if (runtimeContexts == null || !runtimeContexts.Any())
```
DustinCampbell: `if (runtimeContexts?.Any() == false)`

src/OmniSharp.Script/ScriptProjectSystem.cs
```diff
@@ -67,48 +57,66 @@ public void Initalize(IConfiguration configuration)
...
+                            x.Name.ToLowerInvariant().StartsWith("system.runtime")));
+                }
 
-                //script pack usings
-                Context.CommonUsings.UnionWith(scriptPackSession.Namespaces);
+            }
 
-                Context.ScriptPacks.UnionWith(scriptPackSession.Contexts.Select(pack => pack.GetType().ToString()));
+            // inject all inherited assemblies
+//#if NET46
```
DustinCampbell: Can this be deleted?

src/OmniSharp.Script/ScriptProjectSystem.cs
filipw: yes correct I will want to change it in the future. There is no real official project model for scripting so at the moment the idea is that if you have a `project.json` it will be used to determine your dependencies and we can load them in for you to provide language services. We will in the future switch to csproj so that you can have `<packagereference>` there. 

In all other cases, you get a no-frills service with `mscorlib`/`system.runtime` only and all your needed assemblies must be `#r`-ed manually by you

src/OmniSharp.Script/ScriptProjectSystem.cs
filipw: oops

src/OmniSharp.Script/ScriptProjectSystem.cs
DustinCampbell: Sounds good.

src/OmniSharp.Script/FileParser.cs
filipw: yes, this just strips away the quotes, tested on Mac too. when I have i.e. `#load "bar.csx"` in my CSX and I use `GetXXXDirectives()` method on the compilation root, the result string comes back with quotes `"bar.csx"` and I want to strip it away to be able to concat the relative path and traverse these resources

DustinCampbell: Oh sorry. I misread the slash in there. :-) I glanced at it and read the slash, not the double-quote. :smile:

filipw:
@DustinCampbell everything should be OK now, all the review feedback applied - thanks!

DustinCampbell:
Merging! :smile:

filipw:
yay - thanks 😄

gpuido:
Sounds great ! 

what should I put in project.json to get started with something simple like the project in your gif ?

https://github.com/OmniSharp/omnisharp-roslyn/pull/659#issuecomment-262826992

filipw:
> what should I put in project.json to get started with something simple like the project in your gif ?

If you use CSI to execute your scripts, the `project.json` can be empty, just `{}` (it's only needed to "light up" OmniSharp). You will be responsible for bringing in any assembly references you need using `#r` directives. This is the regular CSI experience.

If you use a .NET Core script runner (CSI is not .NET Core compatible at the moment) like https://github.com/filipw/dotnet-script, then you can put any package references you need into the `project.json` too.

gpuido:
Thanks a lot filipw, 

I got some of the frustation out with dotnet-script running .NET Core
but I'm on win10 and I'd like to target dotnet4.6, is this possible ?

> CSI is not .NET Core compatible at the moment

So the 'regular CSI experience' is running on regular .net46 ?
because when I hit F5 vscode ask me to select an environment but only offer to select .NET Core.

DustinCampbell:
If I'm understanding you correctly, the platforms that the debugger in VS Code supports (which is just CoreCLR) is completely orthogonal to what environment your scripts run in.

filipw:
> I got some of the frustation out with dotnet-script running .NET Core
> but I'm on win10 and I'd like to target dotnet4.6, is this possible ?

no, at the moment that runner only exists to run .NET Core scripts. 
In fact, it was built only to bridge the current gap, because CSI doesn't support .NET Core

> So the 'regular CSI experience' is running on regular .net46 ?

correct, CSI is for running scripts as net46 only. It runs on mono (with some hiccups) too.

> because when I hit F5 vscode ask me to select an environment but only offer to select .NET Core.

yep, as Dustin said, the F5 debugging in VS Code only works with CoreCLR, so you can use it to work with scripts running with dotnet-script, but not running with CSI.

gpuido:
Got it now, Thanks guys.

This attempt to normalize the C# scripting support in Omnisharp is much welcomed.

Issue #658, opened 2016-11-03
Unable to Build on Ubuntu


dalealleshouse:
I am unable to build. The output is below. I'm running Ubuntu 16.10 and Mono 4.6.1.

Can someone help me out?

```bash
$ ./build.sh
Downloading NuGet...
Feeds used:
  /home/dalealleshouse/.local/share/NuGet/Cache
  /home/dalealleshouse/.nuget/packages/
  https://dotnet.myget.org/F/dotnet-buildtools/api/v3/index.json
  https://api.nuget.org/v3/index.json

Restoring NuGet package Cake.0.10.1.
WARNING: Unable to find version '0.10.1' of package 'Cake'.
Restoring NuGet package Newtonsoft.Json.8.0.3.
WARNING: Unable to find version '8.0.3' of package 'Newtonsoft.Json'.
Could not find Cake.exe at '/home/dalealleshouse/.vim/bundle/omnisharp-vim/omnisharp-roslyn/scripts/../.tools/Cake/Cake.exe'.
```

willl:
Seems like it's possibly similar to issues #650 and possibly #651,

dalealleshouse:
Thanks @nmschulte , as the other issue suggested, I updated the nuget.exe. This got me a little bit further, but I had to also add some certs as outlined [here](http://stackoverflow.com/questions/15181888/nuget-on-linux-error-getting-response-stream)

Now, I have a new error:
```bash
========================================
Restore
========================================
Executing task: Restore
Restoring packages in /home/dalealleshouse/.vimpkg/bundle/omnisharp-vim/omnisharp-roslyn....
Failed to initialize CoreCLR, HRESULT: 0x80131500

An error occured when executing task 'Restore'.
Error: Failed to restore projects under working directory.
```

dalealleshouse:
I was able to get a bit further by reinstall dotnet core

```bash
Can not find runtime target for framework '.NETCoreApp,Version=v1.0' compatible with one of the target runtimes: 'ubuntu.16.10-x64'. Possible causes:
1. The project has not been restored or restore failed - run `dotnet restore`
2. The project does not list one of 'ubuntu.16.10-x64' in the 'runtimes' section.
3. You may be trying to publish a library, which is not supported. Use `dotnet pack` to distribute libraries.
An error occured when executing task 'OnlyPublish'.
Error: Failed to publish OmniSharp / netcoreapp1.0
```

nmschulte:
[...]

PR #660, opened 2016-11-04
Issue #658 - Unable to Build on Ubuntu 

dalealleshouse:
Added ubuntu 16.10 as a target and changed the target framework from netcoreapp1.0 to netcoreapp1.1 - it's now building on ubuntu 16.10.


david-driscoll:
@DustinCampbell this is a wonderful rid problem, I know you love these!

DustinCampbell:
I was waiting until we released .NET Core 1.1 before looking at this. Now that it's here, @dalealleshouse, could you update the PR to the final release of .NET Core 1.1?

DustinCampbell:
I took a look at this today, and I think this PR actually ends up breaking OmniSharp. The problem is that there is an API breaking change introduced [here](https://github.com/dotnet/corefx/pull/10937/commits/a06e6057511ca074a1f525a02bbf1955cd57897c) between System.Reflection.Metadata, 1.4.1 and System.Reflection.Metadata, 1.4.1-beta-24430-01. Unfortunately, Microsoft.NetCore.App, 1.1 references the former version and Roslyn 2.0.0-rc references the latter version.

The CI passes for this change because OmniSharp.exe is moved to netcoreapp1.1, but all of the tests are still on netcoreapp1.0. So, the tests pass but OmniSharp.exe doesn't work.

build.json
```diff
@@ -1,35 +1,35 @@
...
-  "DotNetFolder": ".dotnet",
-  "DotNetInstallScriptURL": "https://raw.githubusercontent.com/dotnet/cli/rel/1.0.0-preview2/scripts/obtain",
-  "DotNetChannel": "preview",
-  "DotNetVersion": "1.0.0-preview2-003131",
-  "BuildToolsFolder": ".tools",
-  "ArtifactsFolder": "artifacts",
-  "TestProjects": {
-    "OmniSharp.MSBuild.Tests": [
-      "netcoreapp1.0"
+    "UseSystemDotNetPath": "false",
```
DustinCampbell: FWIW, it'd be nice if this change didn't update the whitespace in the files, which messes up the diff.

build.json
```diff
@@ -1,35 +1,35 @@
...
+    "DotNetChannel": "preview",
+    "DotNetVersion": "1.0.0-preview2-003131",
+    "BuildToolsFolder": ".tools",
+    "ArtifactsFolder": "artifacts",
+    "TestProjects": {
+        "OmniSharp.MSBuild.Tests": [
+            "netcoreapp1.0"
+        ],
+        "OmniSharp.Roslyn.CSharp.Tests": [
+            "netcoreapp1.0"
```
JosephWoodward: Does that missing type you're referring to happen to be `System.Reflection.Metadata.ISignatureTypeProvider`?

DustinCampbell: That's the one. We're still working on getting Roslyn updated to .NET Core 1.1, which will address the problem.

DustinCampbell:
Closing this in favor of https://github.com/OmniSharp/omnisharp-roslyn/pull/835.

PR #664, opened 2016-11-10
Don't hard-code CSharp6 in DotNetProjectSystem

DustinCampbell:


DustinCampbell:
Thanks for the quick reviews @willl!

willl:
np @DustinCampbell. It's easy when they're this tiny!

PR #665, opened 2016-11-11
use SymbolDisplayFormat in TypeLookup

filipw:
Instead of manually composing type information, use `SymbolDisplayFormat` to format the response according to the predefined rules:

 - default `SymbolDisplayFormat.FullyQualifiedFormat` but without `global::` for regular C# syntax
 - `SymbolDisplayFormat.FullyQualifiedFormat`but without `global::` and without namespace for script/interactive syntax

Fixes https://github.com/OmniSharp/omnisharp-vscode/issues/394 (currently OmniSharp doesn't show type lookup information for nested types correctly)

### DIFF
--- a/src/OmniSharp.Roslyn.CSharp/Services/Types/TypeLookup.cs
+++ b/src/OmniSharp.Roslyn.CSharp/Services/Types/TypeLookup.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Composition;
 using System.Threading.Tasks;
 using Microsoft.CodeAnalysis;
@@ -16,7 +17,10 @@ namespace OmniSharp.Roslyn.CSharp.Services.Types
     {
         private readonly FormattingOptions _formattingOptions;
         private readonly OmnisharpWorkspace _workspace;
-        private static readonly SymbolDisplayFormat DefaultFormat = SymbolDisplayFormat.FullyQualifiedFormat.WithGlobalNamespaceStyle(SymbolDisplayGlobalNamespaceStyle.Omitted);
+        private static readonly SymbolDisplayFormat DefaultFormat = SymbolDisplayFormat.FullyQualifiedFormat.
+            WithGlobalNamespaceStyle(SymbolDisplayGlobalNamespaceStyle.Omitted).
+            WithMiscellaneousOptions(SymbolDisplayMiscellaneousOptions.None).
+            WithMiscellaneousOptions(SymbolDisplayMiscellaneousOptions.EscapeKeywordIdentifiers);
 
         [ImportingConstructor]
         public TypeLookupService(OmnisharpWorkspace workspace, FormattingOptions formattingOptions)
@@ -37,22 +41,9 @@ namespace OmniSharp.Roslyn.CSharp.Services.Types
                 var symbol = await SymbolFinder.FindSymbolAtPositionAsync(semanticModel, position, _workspace);
                 if (symbol != null)
                 {
-                    //non regular C# code semantics (interactive, script) don't allow namespaces
-                    if (symbol.Kind == SymbolKind.NamedType)
-                    {
-                        if (document.SourceCodeKind == SourceCodeKind.Regular && !symbol.ContainingNamespace.IsGlobalNamespace)
-                        {
-                            response.Type = symbol.ToDisplayString(DefaultFormat);
-                        }
-                        else
-                        {
-                            response.Type = symbol.ToDisplayString(SymbolDisplayFormat.MinimallyQualifiedFormat);
-                        }
-                    }
-                    else
-                    {
-                        response.Type = symbol.ToMinimalDisplayString(semanticModel, position);
-                    }
+                    response.Type = symbol.Kind == SymbolKind.NamedType ? 
+                        symbol.ToDisplayString(DefaultFormat) : 
+                        symbol.ToMinimalDisplayString(semanticModel, position);
                 }
 
                 if (request.IncludeDocumentation)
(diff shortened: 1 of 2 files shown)


DustinCampbell:
Eventually, we'll just use this API: https://github.com/dotnet/roslyn/pull/13623.

src/OmniSharp.Roslyn.CSharp/Services/Types/TypeLookup.cs
```diff
@@ -15,6 +15,14 @@ public class TypeLookupService : RequestHandler<TypeLookupRequest, TypeLookupRes
...
         private readonly FormattingOptions _formattingOptions;
         private readonly OmnisharpWorkspace _workspace;
+        private static readonly SymbolDisplayFormat DefaultFormat = SymbolDisplayFormat.FullyQualifiedFormat.WithGlobalNamespaceStyle(SymbolDisplayGlobalNamespaceStyle.Omitted);
+        private static readonly SymbolDisplayFormat SimpleFormat = new SymbolDisplayFormat(
+                globalNamespaceStyle: SymbolDisplayGlobalNamespaceStyle.Omitted,
+                typeQualificationStyle: SymbolDisplayTypeQualificationStyle.NameAndContainingTypes,
+                genericsOptions: SymbolDisplayGenericsOptions.IncludeTypeParameters,
+                miscellaneousOptions:
+                    SymbolDisplayMiscellaneousOptions.EscapeKeywordIdentifiers |
+                    SymbolDisplayMiscellaneousOptions.UseSpecialTypes);
```
filipw: However `MinimallyQualifiedFormat` skips namespace / containing type information. For example calling `symbol.ToMinimalDisplay...` on `Foo.Bar.Xyz` returns just `Xyz`, is that the intention? Because OmniSharp has shown fully qualified info

filipw:
i've rebased this onto the tip of dev.

Can we take this change in? It doesn't change any existing OmniSharp behavior or format (still shows fully qualified tooltips as always) - except it correctly shows the names of nested classes now. 

The verbosity around `SymbolDisplayFormat` is related to just that.

DustinCampbell:
@filipw: Could you at least add some unit tests that ensure that OmniSharp behavior didn't change for other kinds of symbols? Also, it really might be worth a change like so to ensure that your change is *only* for named types and doesn't affect any other kinds of symbols (like methods, events, properties, parameters, local variables, etc., etc.):

```C#
if (symbol.Kind == SymbolKind.NamedType)
{
    if (document.SourceCodeKind == SourceCodeKind.Regular && !symbol.ContainingNamespace.IsGlobalNamespace)
    {
        response.Type = symbol.ToDisplayString(DefaultFormat);
    }
    else
    {
        response.Type = symbol.ToDisplayString(SimpleFormat);
    }
}
else
{
    response.Type = symbol.ToDisplayString(SymbolDisplayFormat.SymbolDisplayFormat.MinimallyQualifiedFormat);
}
```

src/OmniSharp.Roslyn.CSharp/Services/Types/TypeLookup.cs
```diff
@@ -38,11 +46,11 @@ public TypeLookupService(OmnisharpWorkspace workspace, FormattingOptions formatt
                     //non regular C# code semantics (interactive, script) don't allow namespaces
                     if(document.SourceCodeKind == SourceCodeKind.Regular && symbol.Kind == SymbolKind.NamedType && !symbol.ContainingNamespace.IsGlobalNamespace)
                     {
-                        response.Type = $"{symbol.ContainingNamespace.ToDisplayString()}.{symbol.ToDisplayString(SymbolDisplayFormat.MinimallyQualifiedFormat)}";
+                        response.Type = symbol.ToDisplayString(DefaultFormat);
                     }
                     else
                     {
-                        response.Type = symbol.ToDisplayString(SymbolDisplayFormat.MinimallyQualifiedFormat);
+                        response.Type = symbol.ToDisplayString(SimpleFormat);
```
filipw: thanks for the feedback!

I think it's a great suggestion to add more comprehensive unit tests - regardless of whether we merge anything proposed here or not, the tests will at least be useful for any potential changes in the future to avoid regressions. 

I'll do that.

DustinCampbell: I was looking at doing that too -- in case you were busy. :smile:

filipw:
I have simplified the logic and push a total of 15 tests for different variants of hints.
As a bonus, this PR now fixes https://github.com/OmniSharp/omnisharp-vscode/issues/250 as well (OmniSharp will now have the same behavior as Visual Studio).

I have run the tests against old implementation of `TypeLookup` and the only only difference in output is related to nested types (the original purpose of this PR) and https://github.com/OmniSharp/omnisharp-vscode/issues/250.

Actually, there is one more "byproduct" of this change and I wanted to highlight it to make sure this makes sense. 
When looking at a type hint for something like `IDictionary<string, IEnumerable<int>>`, at the moment OmniSharp would respond with `System.Collections.Generic.IDictionary<string, IEnumerable<int>>`. This is a bit halfway IMHO and inconsistent with other type hint OmniSharp shows. 

This change would make this type hint `System.Collections.Generic.IDictionary<System.String, System.Collections.Generic.IEnumerable<System.Int32>>` which to me makes more sense.

filipw:
I think down the road it would be nicer to have different outpput for generic types, like VS does:
screen shot 2016-11-20 at 20 35 11

But I assume https://github.com/dotnet/roslyn/pull/13623 will take care of that, so no point in rolling out anything by hand?

DustinCampbell:
@filipw: Yes, we'll get that with the Roslyn API. It's several months off but it will pretty remove any code we have in this endpoint and add support for generic type maps (as you showed), proper XML documentation display (with minimally-qualified crefs), and all of the other subtleties in C#'s Quick Info tool tips.

It'll even give us the anonymous type map displays, which are very tricky to get right.

image

PR #673, opened 2016-11-22
Fix project.json search in DotNetProjectSystem

DustinCampbell:
This change fixes several issues with locating project.json projects:

1. The top-level folders defined in global.json were not searched. So, if a user tried to add `"projects": ["MyProject1"]` where `MyProject1` contained a project.json file, the project wouldn't be found. Instead, only child folders of the top-level folders would be search. So, `src/MyProject1` would work. Now, we'll search the top-level folders and their immediate children, which should match how `dotnet restore` behaves a bit more closely. A test has been added for this scenario.

2. We had infrastructure for falling back to a recursive directory search when a project.json or global.json doesn't exist in the current directory. However, we weren't using it! There was an issue with the recursive search (it wouldn't search a directory's sub-directories if the directory contained a project.json file), but that is now fixed.

3. The test for recursively searching directories wasn't correct. This is now fixed.

This fixes https://github.com/OmniSharp/omnisharp-vscode/issues/962 and https://github.com/OmniSharp/omnisharp-vscode/issues/904


src/OmniSharp.DotNet/DotNetWorkspace.cs
```diff
@@ -156,7 +112,7 @@ private static IEnumerable<ProjectDescription> GetProjectReferences(ProjectConte
                     continue;
                 }
 
-                // if this is an assembly reference then don't threat it as project reference
+                // if this is an assembly reference then don't treat it as project reference
```
filipw: lol

src/OmniSharp.DotNet/Projects/ProjectSearcher.cs
```diff
@@ -1,66 +1,86 @@
...
     {
-        public static IEnumerable<string> Search(string solutionRoot)
+        public static IEnumerable<string> Search(string directory)
         {
-            return Search(solutionRoot, maxDepth: 5);
+            return Search(directory, maxDepth: 5);
         }
 
-        public static IEnumerable<string> Search(string solutionRoot, int maxDepth)
+        public static IEnumerable<string> Search(string directory, int maxDepth)
```
DustinCampbell: Note that this function was inconsistent before my changes. Sometimes it would return a directory path to a project.json files, and sometimes it would return multiple _file_ paths to project.json files. Now, it just returns file paths to project.json files.

src/OmniSharp.DotNet/Projects/ProjectSearcher.cs
```diff
@@ -73,19 +93,38 @@ private static IEnumerable<string> FindProjectsThroughGlobalJson(string root)
...
-                                     .Distinct(StringComparer.OrdinalIgnoreCase)
-                                     .ToList();
-            }
-            else
-            {
-                return Enumerable.Empty<string>();
+                var projectPaths = new SortedSet<string>(StringComparer.OrdinalIgnoreCase);
+                var searchDirectories = new Queue<string>();
+
+                // Look in global.json 'projects' search paths and their immediate children
```
david-driscoll: For my sanity (because project.json is starting to fade with csproj out soon :stuck_out_tongue_winking_eye:).

`global.json` never supported wild cards, but instead just looked to see if it a any `project.json` in the folder or it's children.  ie the common case was `projects: ["src", "test"]`.

DustinCampbell: Were you looking for a change here?

PR #674, opened 2016-11-22
Don't throw null-ref when Roslyn CompletionService doesn't return anything

DustinCampbell:
Fixes https://github.com/OmniSharp/omnisharp-vscode/issues/980


src/OmniSharp.Roslyn.CSharp/Services/Intellisense/IntellisenseService.cs
```diff
@@ -44,24 +44,27 @@ public IntellisenseService(OmnisharpWorkspace workspace, FormattingOptions forma
                 // Add keywords from the completion list. We'll use the recommender service to get symbols
                 // to create snippets from.
 
-                foreach (var item in completionList.Items)
+                if (completionList != null)
```
david-driscoll: Usually `git diff` is much better than this! lol

DustinCampbell: totally

PR #676, opened 2016-11-23
Update to roslyn 2.0.0-rc

DustinCampbell:


src/OmniSharp.Roslyn/Services/MetadataFileReferenceCache.cs
```diff
@@ -50,7 +50,7 @@ public MetadataReference GetMetadataReference(string path)
             var documentationFile = Path.ChangeExtension(path, ".xml");
             if (File.Exists(documentationFile))
             {
-                return metadata.GetReference(new XmlDocumentationProvider(documentationFile));
+                return metadata.GetReference(XmlDocumentationProvider.CreateFromFile(documentationFile));
```
DustinCampbell: Hey look! A new public API so we can delete some old code.

src/OmniSharp.Roslyn/XmlDocumentationProvider.cs
```diff
@@ -1,55 +0,0 @@
-﻿using System;
```
DustinCampbell: Note: This file is almost identical (at least, the "meat" of it is identical) to https://github.com/dotnet/roslyn/blob/version-2.0.0-rc/src/Workspaces/Core/Portable/Utilities/Documentation/XmlDocumentationProvider.cs, which we can now use.

src/OmniSharp.Roslyn/Services/MetadataFileReferenceCache.cs
david-driscoll: Yay!

PR #679, opened 2016-11-24
Improve tests

DustinCampbell:
Loads of refactoring and clean up in unit tests. New concepts:

* `TestContent`: This unifies the process of marking up test code with a caret position or spans. It is similar to [`MarkupTestFile`](https://github.com/dotnet/roslyn/blob/master/src/Test/Utilities/Shared/MarkedSource/MarkupTestFile.cs) from Roslyn. The idea is that we should have a single way to specify this information in tests.
* `TestFile`: Couples `TestContent` with a file name. Rather than passing around dictionaries of string -> string, tests now pass around arrays of TestFile.
* `TextPoint` and `TextRange`: These are simple structs for handle line/offset pairs. Note that these do not use the term "Column", which is usually reserved in editors to represent a position within a line after tabs are expanded.

Also, found a few bugs and fixed 'em.


filipw:
> Rather than passing around dictionaries of string -> string, tests now pass around arrays of TestFile.

I swear, it somehow felt like a good idea at a time 😊

DustinCampbell:
@david-driscoll, how does this look to you?

filipw:
:shipit:

DustinCampbell:
Ping @david-driscoll

DustinCampbell:
Thanks!

PR #683, opened 2016-11-26
Add support for running/debugging NUnit tests

HexWrench:
Detect NUnit tests and return an appropriate feature name. When a test
is executed, the test framework name must be set (xunit or nunit). If no
test framework is set, xunit is used (for backwards compatibility).


filipw:
This will also need VS Code change as it currently explicitly searches for `XunitTestMethod` only.

Generally I think any 3rd party framework specific things shouldn't be embedded into OmniSharp, but at the same xunit/nunit are de facto standard for .NET testing so it's not a problem here. 👍

DustinCampbell:
The VS Code side change is here: https://github.com/OmniSharp/omnisharp-vscode/pull/996.

filipw:
great!

src/OmniSharp.DotNetTest/Helpers/TestFeaturesDiscover.cs
```diff
@@ -20,27 +22,41 @@ public IEnumerable<SyntaxFeature> Discover(SyntaxNode node, SemanticModel semant
...
                         Data = methodName
                     };
                 }
             }
         }
 
-        private static bool IsTestMethod(MethodDeclarationSyntax node,
-                                         SemanticModel sematicModel)
+        private bool IsTestMethod(MethodDeclarationSyntax node,
+                                         SemanticModel sematicModel, Func<ITypeSymbol, bool> predicate)
```
DustinCampbell: formatting

src/OmniSharp.DotNetTest/Helpers/TestFeaturesDiscover.cs
```diff
@@ -20,27 +22,41 @@ public IEnumerable<SyntaxFeature> Discover(SyntaxNode node, SemanticModel semant
             if (node is MethodDeclarationSyntax)
             {
                 var method = node as MethodDeclarationSyntax;
-                if (IsTestMethod(method, semanticModel))
+                bool isTestMethod = false;
+                string featureName = null;
+
+                if(IsTestMethod(method, semanticModel, IsDerivedFromFactAttribute))
```
DustinCampbell: spacing between 'if' and '('

src/OmniSharp.DotNetTest/Helpers/TestFeaturesDiscover.cs
```diff
@@ -59,5 +75,24 @@ private static bool IsDerivedFromFactAttribute(ITypeSymbol symbol)
...
+                fullName = $"{symbol.ContainingNamespace}.{symbol.Name}";
+                if (fullName == "NUnit.Framework.TestAttribute"
+                    || fullName == "NUnit.Framework.TestCaseAttribute"
+                    || fullName == "NUnit.Framework.TestCaseSourceAttribute")
+                {
+                    return true;
+                }
+
+                symbol = symbol.BaseType;
+            } while (symbol.Name != "Object");
```
DustinCampbell: Formatting nit: while should go on separate line.

DustinCampbell: More accurate way of doing this:

```C#
while (symbol.SpecialType != SpecialType.System_Object);
```

HexWrench: The changes in the VS Code plugin and in Omnisharp-Roslyn work independently. The new Omnisharp-Roslyn will work with the old plugin, and the new plugin will work with the "old" Omnisharp-Roslyn. In both cases xUnit testing still works.

HexWrench:
The changes in the VS Code plugin and in Omnisharp-Roslyn work independently. The new Omnisharp-Roslyn will work with the old plugin, and the new plugin will work with the "old" Omnisharp-Roslyn. In both cases xUnit testing still works.

DustinCampbell:
I hadn't looked carefully, but yes, it does appear that it'll still work. I had assumed that the change to the protocol would break things, but it looks OK. That said, I'll wait to pull the change into VS Code's master until I've got an OmniSharp build to support it. I want to be sure the feature lights up when I take the PR.

david-driscoll:
Looks good to me, @DustinCampbell we still waiting on the minor nits?

DustinCampbell:
Sorry for taking so long to get back to this. Nope, I wasn't waiting on the nits.

DustinCampbell:
Updated the base branch. I'll go ahead and merge this once the CI build finishes.

Issue #685, opened 2016-11-27
VSCode Intellisense for CSX not showing doc for types other than CSX references


ZurielCorrea:
Create a CSX file in VSCode.

**Scenario 1:**
- Type "System.IO.", 
Expected result: intellisense shows the xml doc descriptions for the types and member types.  
Actual result: intellisense won't show the xml docs.

**Scenario 2:**
- Create a simple assembly with a type that has xml docs in the classes and members. 
- Add a #r ".dll"  to the CSX file, (may require restarting omnisharp or even restarting vscode to make  intellisense recognize the new reference)
Expected result: intellisense shows the xml doc descriptions for the types and member types.  
Actual result: intellisense won't show the xml docs.

**Scenario 3:**
- Create a CSX file called file1.csx and add a public method with a summary xml doc.
- Reference "file1.csx" in another csx file using #load.
- Type the name of the method from file1.csx
Expected result: intellisense shows the xml doc description  
**This one works as expected.**

filipw:
@Zuriel-C thanks, I created https://github.com/OmniSharp/omnisharp-roslyn/pull/686 to (mostly) fix it 

> Add a #r ".dll" to the CSX file, (may require restarting omnisharp or even restarting vscode to make intellisense recognize the new reference)

Correct, these references are not processed in real time yet. `ctrl+shift+p`/`cmd+shift+p` and then `restart omnisharp` will be enough to get it picked up. The plan is to have it react in real time in the future though.

filipw:
Scenario 2 has been fixed.
Scenario 1 has been fixed for .NET Core but not for desktop CLR where references to `mscorlib` and `System.Core` are implicit.

filipw:
> Correct, these references are not processed in real time yet. ctrl+shift+p/cmd+shift+p and then restart omnisharp will be enough to get it picked up. **The plan is to have it react in real time in the future though.**

This is now part of OmniSharp.

PR #686, opened 2016-11-28
load documentation along scripting metadata references

filipw:
This PR ensures we go over the same `IMetadataFileReferenceCache` when adding assembly references to scripting compilation, as MSBuild and DotNet project systems do.

Because the new logic is going through metadata cache, the performance should be better too.
This addresses #685 - although not entirely.

 - It will fix the scenario 2 from that bug report (missing docs from `#r` directive).

screenshot 2016-11-28 10 27 19

 
 - It will also fix scenario 1 from that bug report (missing docs for system types), but only for situations where system assemblies are imported using `project.json` via nuget packages

screenshot 2016-11-28 10 24 18

The outstanding issue is that for situations when scripting is used without a `project.json`, we [implicitly add reference](https://github.com/OmniSharp/omnisharp-roslyn/blob/dev/src/OmniSharp.Script/ScriptProjectSystem.cs#L72) to `mscorlib` and that scenario has no documentation because the docs for that are located in "Reference Assemblies" path somewhere in Program Files. @DustinCampbell do you have any suggestion how to resolve that at runtime?


DustinCampbell:
Finding reference assemblies is super tricky and differs completely between Windows (for .NET Framework) and OSX/Linux (Mono). The most complex MSBuild target is "ResolveAssemblyReferences". We might consider adding some basic support within OmniSharp, but it'll be hard to get right. There's really no "official" way to do it.

filipw:
👍

PR #690, opened 2016-11-30
avoid null reference in TypeLookup

filipw:
At the moment if there is a symbol miss in `TypeLookup` (which can happen if we have missing unresolved references and so on) OmniSharp will always respond with 500 - due a null reference exception in `TypeLookup`.

This fixes the issue.


filipw:
can we merge this? at the moment the experience is not very good with the error 500

PR #694, opened 2016-12-05
MSBuild unit test refactoring

DustinCampbell:
This change adds the first MSBuild project unit test and unifies the "sample projects" into a top-level folder called "test-assets"


DustinCampbell:
It's likely that the CI build will fail when running tests for MSBuild on net46. I'm guessing I'll need to disable that for now.

DustinCampbell:
Ugh. The new MSBuild unit test fails on OSX/Linux because it needs to be restored first to get the targets and packages. However, because we build with an older version of the .NET CLI, restore won't work for the csproj. 😞

PR #705, opened 2016-12-13
Update MSBuild packaging

DustinCampbell:
This is a fairly significant change to rationalize how MSBuild is packaged with OmniSharp.

- [x] Remove Microsoft.Build.Runtime from OmniSharp's project.json (since it won't work for Mono).
- [x] Update build.cake to install Microsoft.Build.Runtime into packages folder
- [x] Update build.cake to copy msbuild to '.msbuild-net46' and '.msbuild-netcoreapp1.0' folders (these will be used for debug and test)
- [x] Update build.cake to copy the appropriate msbuild to the artifacts folders during publish. These are copied to an 'msbuild' subdirectory rather than directly in the OmniSharp folder.
- [x] Add various .NET SDK packages which are used by newer MSBuild builds, which allow the "Sdk" to be specified in the .csproj as an attribute on the Project node.
- [x] Copy Microsoft.CSharp.Core.targets to the 'msbuild' folder from the Microsoft.Net.Compilers package rather than including the file directly in OmniSharp.
- [x] Update MSBuildProjectSystem to properly set MSBUILD_EXE_PATH and MSBuildSDKsPath environment variables so that MSBuild can locate itself.
- [x] Update ProjectFileInfo to set the MSBuildExtensionsPath to the 'msbuild' subdirectory.
- [x] Add net46 MSBuild for Mono
- [x] Various changes to build.cake to package the MSBuild for Mono on net46 builds for OSX/Linux
- [x] Delete a handful of binaries that are incorrectly included in the publish folder for net46 builds on OSX/Linux
- [x] Update MSBuild tests to not deploy Microsoft.Build.Runtime and use the '.msbuild-net46' and '.msbuild-netcoreapp1.0' folders instead.
- [x] Ensure that MSBuild environment variables are set properly at debug time.


scripts/nuget.cake
```diff
@@ -0,0 +1,76 @@
+#load "runhelpers.cake"
+
+using System.Net;
+
+/// <summary>
+/// Downloads and unzips a NuGet package directly without any dependencies.
+/// </summary>
+void DownloadNuGetPackage(string packageID, string version, string outputDirectory, string feedUrl)
```
david-driscoll: I think the latest cake build handles these cases, but I'll circle back and look at that this weekend (god I hope)

DustinCampbell:
Dropping netcoreapp1.0 and moving to the embedded Mono will need to be a separate thing.

DustinCampbell:
Note that the artifacts have increased in size now that we have to include several .NET SDKs. :-(

david-driscoll:
15 and 30 megs isn't too bad

DustinCampbell:
It's a smaller increase on net46, which is good since that's what I want to focus on.

src/OmniSharp.MSBuild/MSBuildProjectSystem.cs
```diff
@@ -67,15 +68,100 @@ public class MSBuildProjectSystem : IProjectSystem
...
+            return Directory.Exists(result)
+                ? result
+                : null;
+        }
+
+        public static void SetUpMSBuildEnvironment(ILogger logger)
+        {
+            var msbuildFolder = Path.Combine(AppContext.BaseDirectory, "msbuild");
+
+            if (!Directory.Exists(msbuildFolder))
```
david-driscoll: This is strict to support the "I'm developing OmniSharp" scenario?

DustinCampbell: Yes, `FindMSBuildFolder(...)` supports debug-time and unit tests.

DustinCampbell:
Thanks for the review!

PR #707, opened 2016-12-15
Update to MSBuild project model and fix for cross-targeting projects

DustinCampbell:
Lots of changes here, but the key pieces are:

1. The MSBuild packages are updated to the latest.
2. We now include the right set of NuGet packages that we'll need to perform restore.
3. We include the NuGet.Build.Tasks package in the msbuild folder.
4. MSBuild tests are enabled (hope they pass this time!)
5. We check to see if there are multiple TFMs and pick one prior to running the "ResolveReferences" task. (Otherwise, we can't process cross-targeting MSBuild projects.)
6. The MSBuild project model includes the list of TFMs specified in the project file.
7. The MSBuild project model includes the output path of the project file.


DustinCampbell:
cc @david-driscoll

DustinCampbell:
I wonder if this is ready to merge into dev?

DustinCampbell:
That ran locally *grumble*

src/OmniSharp.MSBuild/ProjectFile/ProjectFileInfo.cs
```diff
@@ -142,13 +144,30 @@ public void SetProjectId(ProjectId projectId)
...
+            var targetFrameworks = PropertyConverter.ToList(project.GetPropertyValue(PropertyNames.TargetFrameworks), ';');
+
+            // If the project supports multiple target frameworks and specific framework isn't
+            // selected, we must pick one before execution. Otherwise, the ResolveReferences
+            // target might not be available to us.
+            if (string.IsNullOrWhiteSpace(targetFramework) && targetFrameworks.Count > 0)
+            {
+                // For now, we'll just pick the first target framework. Eventually, we'll need to
+                // do better and potentially allow OmniSharp hosts to select a target framework.
+                targetFramework = targetFrameworks[0];
```
david-driscoll: In the past we've exposed this on each request, but a dedicated endpoint would be ideal.

What about support in VScode?  Any UI that might be setup?

DustinCampbell: Eventually, yeah. Providing this on each request might prove _very_ expensive.

Issue #666, opened 2016-11-11
Integrate embedded Mono runtime and MSBuild into OmniSharp


DustinCampbell:
For the latest release of omnisharp-vscode, we have made significant changes to how OmniSharp runs that need to be integrated back into omnisharp-roslyn. The following are the key components the need to be integrated.

* **MSBuild**: In order to properly target .NET Core, .NET Framework and Mono .csproj projects, OmniSharp must include MSBuild itself along with relevant targets and task. On Windows, this can be achieved with the [Microsoft.Build.Runtime](https://www.nuget.org/packages/Microsoft.Build.Runtime/15.1.0-preview-000370-00) package. On macOS/Linux, we need a similar but different set of content. In omnisharp-vscode, this achieved by using a custom build from Xamarin's [fork of MSBuild](https://github.com/mono/msbuild/tree/xplat-c9). Because both of these MSBuild variants target net46 but are completely different, we'll need custom build logic to pull them into OmniSharp.

* **Embedded Mono Runtime**: omnisharp-vscode no longer uses .NET Core builds of OmniSharp. Instead, it runs OmniSharp as a net46 app. On Windows, it just runs on the installed .NET Framework. On macOS/Linux, it runs on an embedded Mono runtime with a set of Mono framework assemblies. There are three variants of the Mono runtime: one for macOS, one for Linux/x86, and one for Linux/x64.

Once the integration is complete, there should only be 5 different builds of OmniSharp:

* Windows (x86 and x64)
* OSX
* Linux (x86 and x64)

In addition, OmniSharp will be able to support the following project types:

* project.json (legacy)
* .csproj targeting .NET Core (new)
* .csproj targeting .NET Framework
* .csproj targeting Mono

david-driscoll:
👍  I need to setup a meeting or time to discuss the changes again... summit was good, but we need to make a plan to consolidate things.

I think one part of that is to update the cake build script, something I was already thinking we should do.

DustinCampbell:
Awesome! I've been working on some bash scripts to construct the Embedded Mono framework that I'd cobbled together by hand.

DustinCampbell:
#705 gets much closer to this being integrated into OmniSharp.

david-driscoll:
And I kind of had half a thought in that last comment.

The historical complaint was the file watchers in mono were fairly buggy, if we were to switch to them now would the behavior be better (now that a lot of the code is shared from .NET Desktop).

[...]

PR #709, opened 2016-12-15
Making MSBuild great again

DustinCampbell:
This PR integrates the new-msbuild branch back into dev, which gets us one large step of the way to #666. Essentially, this integrates MSBuild in a much better way with OmniSharp, using official MSBuild packages and a custom-build for Mono.

Adding a couple of smart people to review, but others are welcome to chime in.


src/OmniSharp.Host/project.json
```diff
@@ -5,7 +5,6 @@
   },
   "dependencies": {
     "OmniSharp.Abstractions": "1.0.0",
-    "OmniSharp.Nuget": "1.0.0",
```
DustinCampbell: @david-driscoll: Please note this change. I think we need to invest in some effort to make this endpoint great again.

src/OmniSharp.MSBuild/MSBuildProjectSystem.cs
```diff
@@ -211,6 +217,24 @@ private static CSharpCompilationOptions CreateCompilationOptions(ProjectFileInfo
                 result = result.WithAllowUnsafe(true);
             }
 
+            var specificDiagnosticOptions = new Dictionary<string, ReportDiagnostic>(projectFileInfo.SuppressedDiagnosticIds.Count);
+
+            // Always suppress CS1701 (Assuming assembly reference 'x' used by 'y' matches identity 'z'. you may need to supply runtime policy)
+            specificDiagnosticOptions.Add("CS1701", ReportDiagnostic.Suppress);
```
filipw: .net cli project system disables also `CS1702` and `CS1705` - should these be aligned?

filipw:
this is a hugely welcome change! 🎉 🎈 🍰 
also, at this point I think there is no point in keeping OmniSharp .NET Core build anymore?

src/OmniSharp.MSBuild/MSBuildEnvironment.cs
```diff
@@ -0,0 +1,126 @@
...
+                msbuildExePath = Path.Combine(msbuildFolder, "MSBuild.dll");
+            }
+
+            if (!File.Exists(msbuildExePath))
+            {
+                logger.LogError("Could not locate MSBuild to set MSBUILD_EXE_PATH");
+                return;
+            }
+
+            if (File.Exists(msbuildExePath))
```
filipw: this check is not necessary anymore, as we will have exited already if the file was not there
similarly, the following `else` is unreachable anymore and can be removed

src/OmniSharp.MSBuild/MSBuildEnvironment.cs
```diff
@@ -0,0 +1,126 @@
...
+
+            if (!Directory.Exists(msbuildFolder))
+            {
+                // If the 'msbuild' folder does not exist beneath OmniSharp.exe, this is likely a development scenario,
+                // such as debugging or running unit tests. In that case, we use one of the .msbuild-* folders at the
+                // solution level.
+                msbuildFolder = FindMSBuildFolderFromSolution();
+            }
+
+            if (msbuildFolder == null || !Directory.Exists(msbuildFolder))
```
filipw: `FindMSBuildFolderFromSolution()` already checks for directory internally so no need to recheck again, just `null` check would be enough

src/OmniSharp.MSBuild/MSBuildProjectSystem.cs
DustinCampbell: Actually, this problem isn't needed anymore. Now that we properly support the `<NoWarn>` attribute in the project file, we actually inherit `CS1702` and `CS1705` from the targets.

src/OmniSharp.MSBuild/MSBuildEnvironment.cs
DustinCampbell: Good catch! I'd forgotten to remove that after some refactoring.

src/OmniSharp.MSBuild/MSBuildEnvironment.cs
DustinCampbell: I really should read my code after refactoring :smile:

Fixing...

DustinCampbell:
@filipw: .NET Core builds are still a good thing so that we have standalone OmniSharp builds. However, once I can get the *other* part of #666 finished (moving onto a local, embedded Mono), we can probably drop them if we want since we'll have standalone net46 builds of OmniSharp.

filipw:
> once I can get the other part of #666 finished

makes sense 👍

src/OmniSharp.Host/project.json
david-driscoll: oh god, no more trump references... I only hope he doesn't follow through repealing NAFTA (though looking at all his other promises... it might disappear!)

DustinCampbell:
OK. In it goes! :smile:

PR #714, opened 2016-12-20
allow loglevel to be passed in as an argument 

filipw:
This PR introduces a new startup argument `-loglevel {value}`, which allows OmniSharp logging level to be configured from outside.

At the moment there is only a `-v` switch which toggles between `information` and `debug` levels, but there is no way to set up more granular logging levels. For now I have not removed the `-v` switch yet, so as to not break any existing use case.

The PR will allow adding this feature https://github.com/OmniSharp/omnisharp-vscode/issues/993 to the editors.

### DIFF
--- a/src/OmniSharp.Host/Program.cs
+++ b/src/OmniSharp.Host/Program.cs
@@ -48,6 +48,15 @@ namespace OmniSharp
                     enumerator.MoveNext();
                     serverPort = int.Parse((string)enumerator.Current);
                 }
+                else if (arg.ToLowerInvariant() == "-loglevel")
+                {
+                    enumerator.MoveNext();
+                    LogLevel level;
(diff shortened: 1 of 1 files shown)


src/OmniSharp.Host/Program.cs
```diff
@@ -48,6 +48,15 @@ public static void Main(string[] args)
                     enumerator.MoveNext();
                     serverPort = int.Parse((string)enumerator.Current);
                 }
+                else if (arg.ToLowerInvariant() == "-loglevel")
```
filipw: 👍

DustinCampbell:
Nope. No complaints. Although, we should consolidate these arguments at some point.

filipw:
great. let's punch this in

Issue #695, opened 2016-12-05
Move off of project.json and back to .csproj


DustinCampbell:
It's going to have to happen sometime. :smile:

One consideration is whether we stick with .NET CLI. Once the embedded Mono work is done (#666), OmniSharp won't need to run on .NET Core. Additionally, if OmniSharp moves off of .NET Core during the move to .csproj, support won't implicitly be dropped for versions of Visual Studio earlier than 2017.

filipw:
yes, I vote for dropping .NET Core as it will not add any value anymore after #666 / #709 and less complexity is always better

david-driscoll:
most definitely, we only jumped on board because mono was a painful dependency.

DustinCampbell:
I'll try to get us moved over to csproj next week in that case.

DustinCampbell:
Looks like I actually got this done. :smile:

PR #720, opened 2017-01-05
added global.json to enforce an older .net sdk version

filipw:
At the moment, looks like a lot of people are already on preview3 or preview4 which causes a failure when loading OmniSharp in VS.


filipw:
this is just a quick fix until https://github.com/OmniSharp/omnisharp-roslyn/issues/695

PR #721, opened 2017-01-05
use the default ScriptMetadataResolver for scripts

### LINKED ISSUE #702: Support GacFileResolver when resolving MetadataReferences for scripts
CSI supports adding GAC references on platforms where GAC is supported.
We need to allow that in OmniSharp scripts too, for example:

```csharp
#r "System.Web"
using System.Web;
```

At the moment you need to provide full path to the reference assembly to make OmniSharp recognize the reference.

filipw:
Fixes #702.

This allows OmniSharp to support intellisense when GAC reference is used. i.e. instead of:

```
#r "C:\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETFramework\v4.6\System.Web.dll"
```

we can use:

```
#r "System.Web"
```

This is in line with the default CSI behavior, which supports GAC references.

### DIFF
--- a/src/OmniSharp.Script/ScriptProjectSystem.cs
+++ b/src/OmniSharp.Script/ScriptProjectSystem.cs
@@ -8,6 +8,7 @@ using System.Reflection;
 using System.Threading.Tasks;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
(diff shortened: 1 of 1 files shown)


filipw:
in action:

filipw:
thanks - let's punch this in then!

PR #722, opened 2017-01-10
include System.ValueTuple in C# Scripts

filipw:
This PR adds `System.ValueTuple` automatically into the compilation references.

This is required because it's already part of "standard scripting" - CSI also automatically adds it https://github.com/dotnet/roslyn/commit/4d6c590f42fc5558aea5ed45f0d1698719533793, allowing users to use tuples straight away without any extra manual set up.

### Before the change

screenshot 2017-01-10 08 43 43

### After the change

screenshot 2017-01-10 08 40 09


filipw:
small step for OmniSharp, large step for CSX script authors :)

PR #725, opened 2017-01-12
Version updates

DustinCampbell:
This moves OmniSharp forward with regard to the versions of its many dependencies:

1. OmniSharp and all tests are now moved to .NET Core 1.1
2. The .NET Core SDK used for build is moved forward to the latest version still supporting project.json, and also targeting Current (as opposed to LTS).
3. Update to latest Roslyn binaries. Now that we've moved to .NET Core 1.1, this can be done safely. There were a few updates here to reflection code in OmniSharp due to types moving around in Roslyn. I've also added coded to provide better messages when types and members can't be found via reflection.
4. Update to latest MSBuild packages
5. Update to latest .NET SDK packages
6. Add new test for the new slimmer .NET Core .csproj project style.


src/OmniSharp.Host/project.json
```diff
@@ -8,36 +8,35 @@
     "OmniSharp.Stdio": "1.0.0",
     "OmniSharp.Plugins": "1.0.0",
     "OmniSharp.Roslyn": "1.0.0",
-    "Microsoft.AspNetCore.Diagnostics": "1.0.0",
-    "Microsoft.AspNetCore.Server.Kestrel": "1.0.0",
+    "Microsoft.AspNetCore.Diagnostics": "1.1.0",
```
filipw: we should review these ASP.NET Core dependencies at some point, I don't think all of them are used anymore

src/OmniSharp.Nuget/project.json
```diff
@@ -16,17 +16,6 @@
         "NuGet.Packaging": "3.5.0-beta2-1484",
         "NuGet.Packaging.Core": "3.5.0-beta2-1484",
         "Microsoft.Web.Xdt": "2.1.1"
```
filipw: is this project still used by anything? it feels like it could be removed

src/OmniSharp.Host/project.json
DustinCampbell: I *really* want to invert OmniSharp at some point and build the HTTP endpoints on top of the stdio. Make Http a plugin so that core OmniSharp is lean and mean.

src/OmniSharp.Nuget/project.json
DustinCampbell: I'm leaving it so @david-driscoll can have a go at it. At the moment, it isn't actually built as part of OmniSharp.

DustinCampbell:
>great - is Roslyn 2.0.0 RTM tied to VS 2017 release?

Yes

src/OmniSharp.Nuget/project.json
filipw: 🔥

david-driscoll: You can 💣 it, we'll come back to it again in the future with something better!

DustinCampbell:
Excellent! Merging. Next up: OmniSharp moves to .NET Core .csproj.

PR #727, opened 2017-01-23
do not implicity use UTF-8 for SourceText

filipw:
`SourceText` will attempt to detect encoding from BOM (and fallback to UTF-8 anyway) so we shouldn't impose it on the loaded files.


filipw:
done - https://github.com/OmniSharp/omnisharp-roslyn/issues/728

PR #729, opened 2017-01-24
Improve tests

DustinCampbell:
This change refactors the tests to use a the xUnit ITestOutputHelper for logging to ensure that we don't just dump stuff to the console during tests. In addition, we can now run *all* tests on net46 rather than just a select few.


build.json
```diff
@@ -16,20 +16,24 @@
       "net46"
     ],
     "OmniSharp.Roslyn.CSharp.Tests": [
-      "netcoreapp1.1"
+      "netcoreapp1.1",
+      "net46"
```
filipw: whoa I can't believe this was not there before 😄

PR #734, opened 2017-01-27
Respect "nowarn" setting in dotnet project system

filipw:
Fixes https://github.com/OmniSharp/omnisharp-vscode/issues/542

At the moment we completely ignored the user defined suppressions.


kirkone:
I have no clue why but this is not working with *.csproj based projects anymore.

when I run "dotnet build" I get no warnings but VSCode goes crazy with warnings at the same time.
Any idea what I am doing wrong?

Greetings,
KirK

PR #735, opened 2017-01-27
Use Application.BaseDirectory when searching for OmniSharp.sln in MSBuildEnvironment

DustinCampbell:
This allows the MSBuild tools to be found when attaching a debugger to OmniSharp (as opposed to starting it under the debugger).

This also makes things work when setting the "omnisharp.path" in VS Code to bin/Debug/net46/OmniSharp.exe.


DustinCampbell:
@david-driscoll, @filipw: Look OK to you?

filipw:
:shipit:

PR #737, opened 2017-01-27
Clean up logging somewhat

DustinCampbell:


filipw:
This is a very welcome change - thank you - especially as logging has been in quite a disarray.

This reminds me of two things for the future:
 - since we will now have the capability of [externally configuring the logging level](https://github.com/OmniSharp/omnisharp-roslyn/pull/714) it will be a good idea to generally revise the logging usage throughout the application. For example a lot of current "Information" logs could be downgraded and so on
 - one of the bigger problems IMHO has been that in many cases, on server errors, the log would only show "error 500, ErrorMiddleware" and that's it (unless you enabled verbose level). This should also be addressed as it will help a lot

DustinCampbell:
Agreed on both points. I'd like to get both done so the logging is sensible. I also have a change for VS Code logging once this goes in to make it look much more like the Console logger, reducing a lot of noise.

PR #738, opened 2017-01-30
Update to VS 2017 RC .NET Core SDK packages

DustinCampbell:
This PR includes the following changes

* Update OmniSharp to the latest .NET Core SDK packages
* Include "ImportAfter" targets which will should allow it to perform package restores via MSBuild in the near feature.
* Clean up build.cake somewhat (at least, reducing some noise).
* Add support for setting MSBuildSDKsPath in omnisharp.json. This can be used to redirect OmniSharp to a different set of SDK packages if desired.


DustinCampbell:
whoa! Mac CI finishes before AppVeyor kicks off? What strange alternate universe is this?

PR #739, opened 2017-01-30
Fix running unit tests defined in nested classes

DustinCampbell:
Fixes https://github.com/OmniSharp/omnisharp-vscode/issues/743 and fixes https://github.com/OmniSharp/omnisharp-vscode/issues/1151

The issue here is that the metadata name for nested classes was not correct. OmniSharp would report "A.B.C" for the following code, but it should be "A.B+C".

```C#
namespace A
{
    class B
    {
        class C
        {
        }
    }
}
```

### DIFF
--- a/src/OmniSharp.Abstractions/Extensions/ISymbolExtensions.cs
+++ b/src/OmniSharp.Abstractions/Extensions/ISymbolExtensions.cs
@@ -1,0 +1,71 @@
+﻿using System;
+using System.Collections.Generic;
+using Microsoft.CodeAnalysis;
+namespace OmniSharp.Extensions
+{
+    public static class ISymbolExtensions
+    {
+        private readonly static CachedStringBuilder s_cachedBuilder;
+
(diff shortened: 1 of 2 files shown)


src/OmniSharp.Abstractions/Extensions/ISymbolExtensions.cs
```diff
@@ -0,0 +1,71 @@
...
+                {
+                    break;
+                }
+
+                symbols.Push(symbol);
+                symbol = symbol.ContainingSymbol;
+            }
+
+            var builder = s_cachedBuilder.Acquire();
+            try
```
david-driscoll: I'm suprised this behavior is not part of the Roslyn Api

DustinCampbell: Yeah, Roslyn didn't want to be an API for metadata.

PR #741, opened 2017-01-31
Support project type GUID added when the .NET CLI manipulates an SLN file

DustinCampbell:
The .NET CLI offers commands to add and remove projects from an SLN file (e.g. `dotnet sln Solution1 add Project1.csproj`). When it does this, it uses a different project type GUID which we need to support in OmniSharp.

cc @blackdwarf


blackdwarf:
@DustinCampbell nice, I wanted to do this tonight. :)

DustinCampbell:
I have harder deadlines than that. :smile:

blackdwarf:
Got it. That is the GUID we use, so looks fine to me. Can't wait for the VSIX to try it out. 😄

PR #743, opened 2017-02-01
loglevel should be a "--" argument

filipw:
See https://github.com/OmniSharp/omnisharp-vscode/pull/1169#discussion_r98970819


filipw:
Actually I would like to get rid of all this manual parsing and move to use https://github.com/aspnet/Common/tree/dev/src/Microsoft.Extensions.CommandLineUtils instead

DustinCampbell:
> Actually later on I would like to get rid of all this manual parsing and move to https://www.nuget.org/packages/Microsoft.Extensions.CommandLineUtils/ instead

Agree with that. We should be using a library for command-line args.

PR #746, opened 2017-02-03
Preparing for package auto-restore for MSBuild projects

DustinCampbell:
Fixes https://github.com/OmniSharp/omnisharp-vscode/issues/1114

This PR represents a lot of clean up to get ready to introduce package auto-restore for MSBuild projects. Here are the highlights:

* Unified some process launching logic into a ProcessHelper class.
* Moved a few types around into various namespaces.
* Unified code between DotNet and MSBuild project systems that create documents.
* Introduced OmniSharpFileTextLoader class for use with OmniSharpWorkspace.
* Introduced DotNetCliService as a central service that can run .NET CLI operations.
* Renamed some types that had misspelled "OmniSharp" as "Omnisharp".
* Simplified IAssemblyLogger and added trace logging.
* Added code to MSBuildProjectSystem to watch for changes to "project.assets.json" file and update the project in response. This fixes issues where a `dotnet restore` would not refresh the project without restarting OmniSharp.


src/OmniSharp.Roslyn/OmniSharpTextLoader.cs
```diff
@@ -0,0 +1,51 @@
+﻿using System;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Text;
+
+namespace OmniSharp
+{
+    public class OmniSharpTextLoader : TextLoader
```
filipw: 👍

src/OmniSharp.Abstractions/Services/DotNetCliService.cs
```diff
@@ -0,0 +1,66 @@
...
+                lock (restoreLock)
+                {
+                    var exitStatus = new ProcessExitStatus(-1);
+                    _eventEmitter.RestoreStarted(projectPath);
+                    _semaphore.Wait();
+                    try
+                    {
+                        // A successful restore will update the project lock file which is monitored
+                        // by the dotnet project system which eventually update the Roslyn model
+                        exitStatus = ProcessHelper.Run("dotnet", "restore", projectPath);
```
filipw: it's not necessary at the moment, but it might be useful to allow to set omnisharp-level command parameters to be included, for example, I'd like to include `--ignore-failed-sources` in the command etc.

DustinCampbell: Makes sense!

src/OmniSharp.Roslyn/OmniSharpTextLoader.cs
filipw: we can remove this code now as well https://github.com/OmniSharp/omnisharp-roslyn/blob/dev/src/OmniSharp.Script/ScriptProjectSystem.cs#L236
I can do that in a separate PR though, no need to hold up here

DustinCampbell:
> I admire your patience with cleaning up some of that scattered code that has accumulated

Often, I refactor to better understand the code.

PR #747, opened 2017-02-03
Implement package auto-restore for .csproj projects

DustinCampbell:
Fixes https://github.com/OmniSharp/omnisharp-vscode/issues/770


src/OmniSharp.MSBuild/MSBuildProjectSystem.cs
```diff
@@ -483,6 +492,104 @@ private void UpdateReferences(Project project, IList<string> references)
...
+            else
+            {
+                // Note: This is a bit of misnmomer. It's entirely possible that a package reference has been removed
+                // and a restore needs to happen in order to update project.assets.json file. Otherwise, the MSBuild targets
+                // will still resolve the removed reference as a reference in the user's project. In that case, the package
+                // reference isn't so much "unresolved" as "incorrectly resolved".
+                IEnumerable<PackageReference> unresolvedPackageReferences;
+
+                // Did the project file change? Diff the package references and see if there are unresolved dependencies.
+                if (previousProjectFileInfo != null)
```
filipw: would be really good if NuGet.ProjectModel provided some API for these types of checks

DustinCampbell: Which checks? These are the package references defined in the .csproj. NuGet provides no APIs for the .csproj file itself.

src/OmniSharp.MSBuild/ProjectFile/PackageReference.cs
```diff
@@ -0,0 +1,40 @@
+﻿using System;
+using NuGet.Packaging.Core;
+
+namespace OmniSharp.MSBuild.ProjectFile
+{
+    public class PackageReference : IEquatable<PackageReference>
```
filipw: nuget has this https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.PackageManagement/Utility/PackageReferenceComparer.cs, maybe there is no need to introduce a class into omnisharp?

DustinCampbell: Possibly. I *believe* these are used by NuGet to handle VS package restore, which is a much bigger thing than what we're doing. I just need to keep a list of the PackageReferences items computed by MSBuild so that I can tell if changes occurred to the project file. This is before NuGet is really involved.

Note that this class is not exactly the same. It has at least one piece of additional metadata that is provided via MSBuild targets.

DustinCampbell: Note: NuGet's PackageReference is way different: https://github.com/NuGet/NuGet.Client/blob/4cccb13833ad29d6a0bcff055460d964f1b49cfe/src/NuGet.Core/NuGet.Packaging/PackageReference.cs. These aren't the same thing.

src/OmniSharp.MSBuild/MSBuildProjectSystem.cs
filipw: oh, there is no API in nuget for reading package references from csproj? I didn't realize that
(there used to be APIs for reading references from `packages.config` and to read references from `project.json`)

src/OmniSharp.MSBuild/ProjectFile/PackageReference.cs
filipw: right now I remember, in the past I used to define my own `PackageReference` on the [old nuget apis](https://github.com/scriptcs/scriptcs/blob/dev/src/ScriptCs.Core/PackageReference.cs) too, for the very same reason you just mentioned

src/OmniSharp.MSBuild/MSBuildProjectSystem.cs
DustinCampbell: Nope. The API for reading from .csproj is MSBuild. :smile:

PR #750, opened 2017-02-05
Added Microsoft.Extensions.CommandLineUtils

### LINKED ISSUE #744: Use command line parsing library
At the moment the command line arguments are all over the place, and they are parsed manually too.
This should be standardized, not to mention that there is no need to reinvent the wheel - we could just use a parser lib such as https://www.nuget.org/packages/Microsoft.Extensions.CommandLineUtils/ or https://www.nuget.org/packages/Mono.Options/

filipw:
Removed manual arg parsing in favor of https://www.nuget.org/packages/Microsoft.Extensions.CommandLineUtils/
Fixes #744 

This has a couple of benefits:
 - obviously no need to maintain the arg parsing code anymore
 - automatic help generation
 - better/automatic support for short/long names of options
 - automatic support for no value / single value / multi value options
 - in the future we could easily introduce **commands** that could do interesting things (at the moment we have only **options**) i.e. `omnisharp plugin foo`

screenshot 2017-02-05 12 57 41

Note that we have a feature that supports passing `omnisharp.json` values via command line. This is great but is a bit of mess since "real/fixed" OmniSharp options are mixed up with `omnisharp.json` args which are then integrated into the runtime configuration. Ideally the `omnisharp.json`stuff would come at the end or even better be delimited with `--` as it's a general practice for building CLI tools. I have not done that as part of this part (though it would be very easy) as I didn't want to introduce a big breaking change straight away.

Also clean up some code around DI and changed OmniSharp to have a result code and better error message in case it crashes at startup.

### DIFF
--- a/src/OmniSharp.Host/Internal/CommandOptionExtensions.cs
+++ b/src/OmniSharp.Host/Internal/CommandOptionExtensions.cs
@@ -3,9 +3,9 @@ using Microsoft.Extensions.CommandLineUtils;
 
 namespace OmniSharp.Host.Internal
 {
-    public static class CommandOptionExtensions
+    internal static class CommandOptionExtensions
     {
-        public static T GetValueOrDefault<T>(this CommandOption opt, T defaultValue)
+        internal static T GetValueOrDefault<T>(this CommandOption opt, T defaultValue)
         {
             if (opt.HasValue())
             {
--- a/src/OmniSharp.Host/Internal/CommandOptionExtensions.cs
+++ b/src/OmniSharp.Host/Internal/CommandOptionExtensions.cs
@@ -1,0 +1,18 @@
+using System.ComponentModel;
+using Microsoft.Extensions.CommandLineUtils;
+
+namespace OmniSharp.Host.Internal
+{
+    public static class CommandOptionExtensions
+    {
+        public static T GetValueOrDefault<T>(this CommandOption opt, T defaultValue)
+        {
+            if (opt.HasValue())
+            {
+                return (T)TypeDescriptor.GetConverter(typeof(T)).ConvertFrom(opt.Value());
+            }
+
+            return defaultValue;
+        }
+    }
+}
--- a/src/OmniSharp.Host/Program.cs
+++ b/src/OmniSharp.Host/Program.cs
@@ -1,6 +1,5 @@
 using System;
 using System.Collections.Generic;
-using System.ComponentModel;
 using System.Diagnostics;
 using System.IO;
 using System.Text;
@@ -9,6 +8,7 @@ using Microsoft.Extensions.CommandLineUtils;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
+using OmniSharp.Host.Internal;
 using OmniSharp.Host.Loader;
 using OmniSharp.Plugins;
 using OmniSharp.Services;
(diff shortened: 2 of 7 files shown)


src/OmniSharp.Host/Internal/CommandOptionExtensions.cs
```diff
@@ -0,0 +1,18 @@
+using System.ComponentModel;
+using Microsoft.Extensions.CommandLineUtils;
+
+namespace OmniSharp.Host.Internal
+{
+    public static class CommandOptionExtensions
```
DustinCampbell: Public class inside a namespace called "Internal"? Is that intentional?

src/OmniSharp.Host/Program.cs
```diff
@@ -18,168 +21,144 @@ namespace OmniSharp
...
-
-            if (transportType == TransportType.Stdio)
-            {
-                builder.UseServer(new StdioServer(Console.In, writer));
-            }
-            else
+            catch (Exception e)
             {
-                builder.UseKestrel();
+                Console.Error.WriteLine(e.GetBaseException().Message);
```
DustinCampbell: Should we print the stack too? (Just asking)

src/OmniSharp.Host/Program.cs
```diff
@@ -18,168 +21,144 @@ namespace OmniSharp
 {
     public class Program
     {
-        public static OmniSharpEnvironment Environment { get; set; }
```
DustinCampbell: Thank you! This felt really dirty to me.

src/OmniSharp.Host/Program.cs
```diff
@@ -18,168 +21,144 @@ namespace OmniSharp
...
-            else
+            catch (Exception e)
             {
-                builder.UseKestrel();
+                Console.Error.WriteLine(e.GetBaseException().Message);
+                return 0xbad;
             }
+        }
+
+        public static int Run(string[] args)
```
DustinCampbell: Does this need to public?

src/OmniSharp.Host/Program.cs
```diff
@@ -18,168 +21,144 @@ namespace OmniSharp
...
+            var verboseOption = omnisharpApp.Option("-v | --verbose", "Explicitly set 'Debug' log level.", CommandOptionType.NoValue);
+            var hostPidOption = omnisharpApp.Option("-hpid | --hostPID", "Host process ID.", CommandOptionType.SingleValue);
+            var stdioOption = omnisharpApp.Option("-stdio | --stdio", "Use STDIO over HTTP as OmniSharp commincation protocol.", CommandOptionType.NoValue);
+            var zeroBasedIndicesOption = omnisharpApp.Option("-z | --zero-based-indices", "Use zero based indices in request/responses (defaults to 'false').", CommandOptionType.NoValue);
+            var serverInterfaceOption = omnisharpApp.Option("-i | --interface", "Server interface address (defaults to 'localhost').", CommandOptionType.SingleValue);
+            var encodingOption = omnisharpApp.Option("-e | --encoding", "Input / output encoding for STDIO protocol.", CommandOptionType.SingleValue);
+            var pluginOption = omnisharpApp.Option("-pl | --plugin", "Plugin name(s).", CommandOptionType.MultipleValue);
```
DustinCampbell: Hmmm... should this be "pl" or "pi"? @david-driscoll?

src/OmniSharp.Host/Program.cs
```diff
@@ -18,168 +21,144 @@ namespace OmniSharp
...
 
-            using (var app = builder.Build())
+            // omnisharp.json arguments should not be parsed by the CLI args parser
+            // they will contain "=" so we should filter them out
+            var omnisharpJsonArgs = args.Where(x => x.Contains("="));
+
+            var omnisharpApp = new CommandLineApplication(throwOnUnexpectedArg: false);
+            omnisharpApp.HelpOption("-? | -h | --help");
+
+            var applicationRootOption = omnisharpApp.Option("-s | --solution", "Solution, project file or directory for OmniSharp to point at (defaults to current directory).", CommandOptionType.SingleValue);
```
DustinCampbell: Since we have an opportunity to name the long name, how about something more general. E.g. `--source`

filipw:
@DustinCampbell thanks for looking at this - pushed the changes you mentioned. 
I also changed `--solution` to `--source`, sounds much better indeed.

src/OmniSharp.Host/Program.cs
david-driscoll: I think pl is fine

PR #751, opened 2017-02-05
Fix code actions that add files to the workspace

DustinCampbell:
Fixes https://github.com/OmniSharp/omnisharp-vscode/issues/975

This PR cleans up a lot of the code for running code actions. It also tries to add some defensive code around a strange interaction in the '/v2/runcodeaction' end point. Essentially, this end point can either go ahead apply the code action to the workspace or supply the file edits needed to apply the code action. However, in the cases of code actions that add files, this is weird.

When file edits are requested for a code action that adds files, we add an empty file with that name to the workspace. This is very weird and could potentially leave the workspace in a bad state where an empty file was added if the editor host requesting the file edits never actually applies them. This needs revisiting in the future, but doing so now would potentially break editors.

History aside, the actual bug is that the endpoint tried to remove the document from the workspace before adding it. This throws an exception if the document doesn't already exist, which causes the end point fail. This code has been in OmniSharp for a long time, so this either didn't already work or the exception was added to Roslyn at some point. Now, we no longer try to remove it as that doesn't really make sense since a document is being added.


DustinCampbell:
Thanks!

filipw:
> History aside, the actual bug is that the endpoint tried to remove the document from the workspace before adding it. This throws an exception if the document doesn't already exist, which causes the end point fail. This code has been in OmniSharp for a long time, so this either didn't already work or the exception was added to Roslyn at some point.

looks like it has never worked in OmniSharp properly 😄  https://github.com/dotnet/roslyn/blob/version-1.0.0/src/Workspaces/Core/Portable/Workspace/Workspace.cs#L648

PR #753, opened 2017-02-05
Handle nested code actions

DustinCampbell:
Fixes https://github.com/OmniSharp/omnisharp-vscode/issues/302

This change digs into the first level of an action's "nested" code actions. This is a feature added more recently to Roslyn which is used to support code actions being hidden under submenus in VS. We can't assume submenu support in OmniSharp, so we expand the nested actions in place.

image


src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/BaseCodeActionService.cs
```diff
@@ -23,41 +25,63 @@ public abstract class BaseCodeActionService<TRequest, TResponse> : RequestHandle
...
             }
 
+            var refactoringContext = await GetRefactoringContext(originalDocument, request, actions);
+            if (refactoringContext != null)
+            {
+                await CollectRefactoringActions(refactoringContext.Value);
+            }
+
+            // TODO: Determine good way to order code actions.
             actions.Reverse();
```
david-driscoll: Any thoughts on this (besides just reversing the order like we do today?)

src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/BaseCodeActionService.cs
```diff
@@ -23,41 +25,63 @@ public abstract class BaseCodeActionService<TRequest, TResponse> : RequestHandle
...
+        private readonly MethodInfo _getNestedCodeActions;
+
         protected BaseCodeActionService(OmniSharpWorkspace workspace, CodeActionHelper helper, IEnumerable<ICodeActionProvider> providers, ILogger logger)
         {
             this.Workspace = workspace;
             this.Providers = providers;
             this.Logger = logger;
             this._helper = helper;
+
+            // Sadly, the CodeAction.NestedCodeActions property is still internal.
```
david-driscoll: 👎 (at it still being internal)

src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/BaseCodeActionService.cs
DustinCampbell: We have some notion of ordering in Roslyn using attributes but [it's messy](https://github.com/dotnet/roslyn/blob/master/src/Features/CSharp/Portable/CodeFixes/GenerateEnumMember/GenerateEnumMemberCodeFixProvider.cs#L18). At some point, I intend to order them similarly in OmniSharp, but it's down the priority list a bit. For now, I just wanted to add a TODO to indicate that it wasn't ideal. :smile:

src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/BaseCodeActionService.cs
DustinCampbell: We discussed making it public awhile back, but we tabled the issue because folks didn't like the word "Nested" or "Composite" because it's really about grouping related code actions together rather than making up a larger code action out of little ones. I need to bring this back up with the team again.

DustinCampbell:
Thanks for the approval! Burning down some bugs. :smile:

PR #754, opened 2017-02-05
Only show the remove unnecessary usings code fix for the appropriate diagnostic

### LINKED ISSUE #742: Suggests "Remove Unnecessary Usings" in a file with no using directives
image

DustinCampbell:
fixes #742

### DIFF
--- a/src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/BaseCodeActionService.cs
+++ b/src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/BaseCodeActionService.cs
@@ -145,15 +145,23 @@ namespace OmniSharp.Roslyn.CSharp.Services.Refactoring.V2
                         continue;
                     }
 
+                    var typeName = codeFix.GetType().FullName;
+
                     // TODO: This is a horrible hack! However, remove unnecessary usings only
                     // responds for diagnostics that are produced by its diagnostic analyzer.
                     // We need to provide a *real* diagnostic engine to address this.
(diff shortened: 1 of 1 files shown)


src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/BaseCodeActionService.cs
```diff
@@ -121,24 +121,32 @@ private TextSpan GetTextSpan(ICodeActionRequest request, SourceText sourceText)
...
+                    if (typeName != CodeActionHelper.RemoveUnnecessaryUsingsProviderName)
                     {
                         if (!diagnosticIds.Any(id => codeFix.FixableDiagnosticIds.Contains(id)))
                         {
                             continue;
                         }
                     }
 
+                    if (typeName == CodeActionHelper.RemoveUnnecessaryUsingsProviderName &&
+                        !diagnosticIds.Contains("CS8019")) // ErrorCode.HDN_UnusedUsingDirective
```
david-driscoll: Still valid (I assume) if you're inside an unused using?

DustinCampbell: Right. This is the "unused using directive" that comes from the compiler. The problem is that the CodeFixProvider doesn't listen for that diagnostic. Instead, there's *another* diagnostic analyzer that does some extra work and introduces a different diagnostic that the provider responds to. Because we don't have a diagnostic analyzer engine yet for running other analyzers, we need to "force" remove unnecessary usings to operate on this diagnostic. Not ideal, but we'll get it fixed eventually.

Issue #227, opened 2015-05-20
CSX project system should listen file added events


jrieken:
I would be cool if the ScriptCS project system listens to additions and deletions of CSX files so that OmniSharp doesn't need to be restarted.

nosami:
Yes!

filipw:
does the client currently notify omnisharp at all about new document creation/deletion?

jrieken:
they should! there is a route ```/filesChanged```

jrieken:
also on windows we could use the native file watcher, see #193

nosami:
We add new cs files to a project without a watcher.

jrieken:
yeah, but only when their path matches the path of another directory. So, as a first step we could add the logic here: https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp/Roslyn/BufferManager.cs. That would make sure new files sort even without file events

Issue #689, opened 2016-11-30
Unable to use multiple #load in CSX file


TzeWey:
Using the latest `dev` build, create a folder, 3 CSX files and an empty project.json (to get OmniSharp to load).

File 1 - `A.CSX`
```
class ClassA{}
```

File 2 - `B.CSX`
```
class ClassB{}
```

File 3 - `C.CSX`
```
#load "A.CSX"
#load "B.CSX"
ClassA A = new ClassA();
ClassB B = new ClassB();
```

Intellisense shows that ClassA could not be found.
![image](https://cloud.githubusercontent.com/assets/3044158/20759569/36e027ec-b758-11e6-8af0-0ad157ce6901.png)

After swapping the `#load` around so that `C.CSX` looks like this (and restarting OmniSharp):
```
#load "B.CSX"
#load "A.CSX"
ClassA A = new ClassA();
ClassB B = new ClassB();
```

Intellisense would show that ClassB could not be found instead.
![image](https://cloud.githubusercontent.com/assets/3044158/20759524/1ad00fae-b758-11e6-823a-dfe9aa5c6998.png)

I am not able to figure out why this is happening. Is this behavior expected? Or am I doing something wrong?

Thanks!

filipw:
thanks - will look into what's happening

filipw:
when this is implemented https://github.com/OmniSharp/omnisharp-roslyn/pull/659#issuecomment-259746769, it should be fixed

filipw:
fixed by github.com/OmniSharp/omnisharp-roslyn/pull/760

PR #760, opened 2017-02-08
Further improvements to CSX support

filipw:
This is a follow up PR to the large work we did in https://github.com/OmniSharp/omnisharp-roslyn/pull/659
In that PR we removed custom scripting code and focused on "standard" Roslyn scripting.

In this PR, we take it a step further and introduce suggestions from Tomas (https://github.com/OmniSharp/omnisharp-roslyn/pull/659#issuecomment-259746769) on how to approach CSX projects (which have no project system after all). So what we do here, is we move away from the current model of "project-per-file" to a "project-per-entry-file" model (a project is created only for a top-level CSX file and all its #load dependencies are delegated to Roslyn's `ScriptSourceResolver` - we never manually specify project-to-project relationships anymore).

This has some terrific results for us:
 - we can **delete a ton** of custom parsing code, and let Roslyn resolvers do the heavy lifting for us
 - we get support for workspace level changes out of the box - user can change `#r` references or `#load` directives and we pick them up in real time
 - multiple `#load` per project are now supported (in the past, due to our  "project-per-file" model, only 1 was allowed - this is because a submission compilation can only have one project reference)

Fixes https://github.com/OmniSharp/omnisharp-roslyn/issues/227 and https://github.com/OmniSharp/omnisharp-roslyn/issues/689 - two most important issues we have for CSX at the moment.

Here is the behavior demo for "real time" `#load`:

And "real time" `#r`:


src/OmniSharp.Roslyn.CSharp/Services/Navigation/GotoDefinitionService.cs
```diff
@@ -55,25 +55,32 @@ public GotoDefinitionService(OmniSharpWorkspace workspace, MetadataHelper metada
                     }
                     else if (location.IsInMetadata && request.WantMetadata)
                     {
-                        var cancellationSource = new CancellationTokenSource(TimeSpan.FromMilliseconds(request.Timeout));
-                        var metadataDocument = await _metadataHelper.GetDocumentFromMetadata(document.Project, symbol, cancellationSource.Token);
-                        if (metadataDocument != null)
+                        try
```
filipw: sorry this change got in by accident, will revert

src/OmniSharp.Roslyn/MetadataHelper.cs
```diff
@@ -62,8 +62,7 @@ public Task<Document> GetDocumentFromMetadata(Project project, ISymbol symbol, C
 
             var service = _csharpMetadataAsSourceService.CreateInstance(temporaryDocument.Project.LanguageServices);
             var method = _csharpMetadataAsSourceService.GetMethod(AddSourceToAsync);
-
-            return method.Invoke<Task<Document>>(service, new object[] { temporaryDocument, topLevelSymbol, cancellationToken });
+                return method.Invoke<Task<Document>>(service, new object[] {temporaryDocument, topLevelSymbol, cancellationToken});
```
DustinCampbell: Did you mean to indent here?

src/OmniSharp.Script/ScriptProjectSystem.cs
```diff
@@ -107,28 +131,43 @@ public void Initalize(IConfiguration configuration)
...
+            // Each .CSX file becomes an entry point for it's own project
+            // Every #loaded file will be part of the project too
             foreach (var csxPath in allCsxFiles)
             {
                 try
                 {
-                    CreateCsxProject(csxPath);
+                    var csxFileName = Path.GetFileName(csxPath);
+                    var project = ProjectInfo.Create(
+                        id: ProjectId.CreateNewId(Guid.NewGuid().ToString()),
```
DustinCampbell: I don't think you need to create a new Guid for this. `ProjectId.Create()` already does that. The `debugName` parameter is to give it something human-readable for debugging. ([ref](http://source.roslyn.io/#Microsoft.CodeAnalysis.Workspaces/Workspace/Solution/ProjectId.cs,bac27363e2d822dc))

src/OmniSharp.Script/ScriptProjectSystem.cs
```diff
@@ -107,28 +131,43 @@ public void Initalize(IConfiguration configuration)
...
+                    _workspace.AddDocument(project.Id, csxPath, SourceCodeKind.Script);
+                    _projects[csxPath] = project;
+                    _logger.LogInformation($"Added CSX project '{csxPath}' to the workspace.");
                 }
                 catch (Exception ex)
                 {
-                    Logger.LogError($"{csxPath} will be ignored due to the following error:", ex.ToString());
-                    Logger.LogError(ex.ToString());
-                    Logger.LogError(ex.InnerException?.ToString() ?? "No inner exception.");
+                    _logger.LogError(0, ex, $"{csxPath} will be ignored due to an following error");
```
DustinCampbell: We've got an extension method that allows you to call LogError without needing to pass an event ID now.

src/OmniSharp.Roslyn/MetadataHelper.cs
filipw: no 😧

adamralph:
All kinds of awesome

filipw:
thanks, pushed the mentioned changes - should be good to go now :shipit:

DustinCampbell:
:shipit:

PR #770, opened 2017-02-12
Correctly set formatting options on the workspace

filipw:
fixes https://github.com/OmniSharp/omnisharp-roslyn/issues/759

This PR changes our approach to CSharp formatting options. So far, we'd use the OmniSharp workspace without paying any attention to formatting settings - those were always applied on demand via each of the formatting endpoints. As a result, any formatting change that happened differently than through those endpoints (i.e. running code actions) would not respect the settings.

The change here is to introduce a discoverable provider, `IWorkspaceOptionsProvider` (maybe anyone can suggest a better name?) which is discovered at app startup and provides the workspace with formatting options. In our case, the only provider at the moment would be `CSharpWorkspaceOptionsProvider` but in the future we could easily add more, or have different ones loaded via the plugin infrastructure, so from that perspective we are also introducing a new extensibility point into OmniSharp.


david-driscoll:
At least once CI passes... :)

src/OmniSharp.Host/Startup.cs
```diff
@@ -189,15 +189,28 @@ public void ConfigureServices(IServiceCollection services)
...
+                    logger.LogError(e, message);
+                }
+            }
+
+            // run all workspace options providers
+            foreach (var workspaceOptionsProvider in PluginHost.GetExports<IWorkspaceOptionsProvider>())
+            {
+                try
+                {
+                    Workspace.Options = workspaceOptionsProvider.Process(Workspace.Options);
```
DustinCampbell: This implies an order to the `IWorkspaceOptionsProviders`. Is there a way we can define that?

src/OmniSharp.Host/Startup.cs
```diff
@@ -189,15 +189,28 @@ public void ConfigureServices(IServiceCollection services)
...
+                }
+                catch (Exception e)
+                {
+                    var message = $"The workspace optiosn provider '{workspaceOptionsProvider.GetType().Name}' threw exception during initialization.";
```
DustinCampbell: Might be helpful to see the `workspaceOptionsProvider.GetType().FullName` here.

tests/TestUtility/AbstractTestFixture.cs
```diff
@@ -69,6 +71,10 @@ protected Task<OmniSharpWorkspace> CreateWorkspaceAsync(params TestFile[] testFi
 
             var workspace = plugInHost.GetExport<OmniSharpWorkspace>();
 
+            // OmniSharp ships only one provider, CSharpWorkspaceOptionsProvider
+            var formattingProvider = plugInHost.GetExports<IWorkspaceOptionsProvider>().FirstOrDefault() as CSharpWorkspaceOptionsProvider;
```
DustinCampbell: Given the comment, would `Single()` be a better choice?

tests/TestUtility/AbstractTestFixture.cs
```diff
@@ -69,6 +71,10 @@ protected Task<OmniSharpWorkspace> CreateWorkspaceAsync(params TestFile[] testFi
...
+            workspace.Options = formattingProvider?.Process(workspace.Options);
```
DustinCampbell: As written, if the CSharpWorkspaceOptionsProvider isn't included in the MEF composition, this will set workspace.Options to null. Intentional?

src/OmniSharp.Abstractions/Services/IWorkspaceOptionsProvider.cs
```diff
@@ -0,0 +1,9 @@
+using Microsoft.CodeAnalysis.Options;
+
+namespace OmniSharp.Services
+{
+    public interface IWorkspaceOptionsProvider
+    {
+        OptionSet Process(OptionSet workOptionSet);
```
DustinCampbell: I did a bit of a double-take when I read `workOptionSet`. I wasn't sure if this meant "working option set" or "workspace option set". Maybe just `optionSet` to avoid confusion?

DustinCampbell:
All of the CI failed with `Error: /Users/travis/build/OmniSharp/omnisharp-roslyn/.tools/xunit.runner.console/tools/xunit.runner.utility.desktop.dll does not exist`.

It looks like there was an update to xunit.runner.console. I bet that broke it.

DustinCampbell:
Yup. performing a `git clean -xdf` and then building fails. I'll get a fix out and then we can rebase this PR.

DustinCampbell:
PR is out to fix build: https://github.com/OmniSharp/omnisharp-roslyn/pull/772.

DustinCampbell:
There we go. :smile:

filipw:
> PR is out to fix build: #772.

thanks! I didn't notice it - it worked when I ran the tests from VS 👍

src/OmniSharp.Host/Startup.cs
filipw: I wanted to avoid getting into that yet, because that would need a more generic solution anyway - for the plugins.
I believe the best approach would be to use MEF metadata something like: 

```
    public interface IPlugin
    {
        string Name { get; }
        int Order { get; }
    }

    [MetadataAttribute]
    public class PluginMetadataAttribute : ExportAttribute, IPluginMetadata
    {
        public string Name { get; set; }
        public int Order { get; set; }

        public PluginMetadataAttribute()
            : base(typeof(IPlugin))
        {
        }
    }

[PluginMetadata(Name = "DefaultCSharpFormatting", Order = -1)]
```

And then use this metadata when plugins are discovered; this would apply to all plugins in the future. However, I believe we would need to design it a bit more carefully - I'll open a separate issue to track that

filipw:
> None of my comments are blocking, though some of them are things we should think about (especially the ordering thing).

thanks for the feedback - I pushed the smaller things into this branch and spun off https://github.com/OmniSharp/omnisharp-roslyn/issues/773 specifically as part of the plugin infrastructure

DustinCampbell:
:shipit:

DustinCampbell:
> thanks! I didn't notice it - it worked when I ran the tests from VS 👍 

Tests run for you in VS? I haven't been able to get tests to run in VS since moving OmniSharp to netcoreapp1.1.

filipw:
They do, but only when `netcoreapp1.1` is defined before `net46` in the frameworks.

screenshot 2017-02-13 21 22 24

DustinCampbell:
Ahhhh.... We should change that then.

PR #772, opened 2017-02-13
Fix the build to use updated binary name from xunit.runner.console package

DustinCampbell:


DustinCampbell:
FYI: @david-driscoll, if you don't see this and CI passes, I'll just go ahead and merge to unblock @filipw.

PR #775, opened 2017-02-13
Fix test discovery in VS 2015 by swapping the TFMs

DustinCampbell:
@filipw [mentioned](https://github.com/OmniSharp/omnisharp-roslyn/pull/770#issuecomment-279510689) that the TFMs can be swapped in the project.json files to make the VS 2015 Test Window properly discover tests. This change does exactly that.


DustinCampbell:
I don't think this needs a code review. So long as CI passes, this should be fine.

PR #640, opened 2016-09-21
initial run at cli based plugins

david-driscoll:
Latest go at plugins... this is loading refactoring essentials (the pcl)

PR #780, opened 2017-02-21
Find all Todo's and add as quick fixes (to problems tray)

eByte23:
This commit finds all TODOs based on single line comments with a regex.
image

### DIFF
--- a/src/OmniSharp.Roslyn.CSharp/Services/Diagnostics/CodeCheckService.cs
+++ b/src/OmniSharp.Roslyn.CSharp/Services/Diagnostics/CodeCheckService.cs
@@ -2,10 +2,13 @@
 using System.Collections.Generic;
 using System.Composition;
 using System.Linq;
+using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
 using OmniSharp.Mef;
 using OmniSharp.Models;
+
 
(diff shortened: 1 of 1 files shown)


david-driscoll:
@eByte23 It would be great if you could give a little more insight into what the end goal you see for this feature (todo's, and other related lines?)

While I like the idea (personally) I'm hesitant to bring it into the full project as not everyone will want this behavior.  In addition editors like vscode or atom could offer similar functionality outside of OmniSharp itself by simple file parsing.

This is definitely a good candidate for a plugin, which means I need to revive #640 and get that finished off soon.

eByte23:
@david-driscoll I reason this came about was the need/want to see all todo's in all files in an easy many all current extension in no display it this way. And this way you can also search for todo's only and search on the to-do text.

I believe that visual studio has a to-do/tasks list by default so I believe it should be a similar feature in omnisharp.

Yes I did consider a plugin however that is obviously not complete yet and this would only be for to-do so I'd feel nearly point less in requiring the same plugin in every solution or project

filipw:
> In VS, we have an engine for crawling the entire solution looking for TODOs.

this one? https://github.com/dotnet/roslyn/issues/11110

DustinCampbell:
yes

mholo65:
Discussed this with @david-driscoll and @filipw.
This could be implemented in the future as a plugin, thus should not be a core part of OmniSharp.

We should soon get some documentation on how to create plugins. Stay tuned, thanks!

PR #781, opened 2017-02-21
allow unsafe in C# scripts

filipw:
At the moment C# script compiler APIs in Roslyn implicitly always allows unsafe blocks.
OmniSharp should match that because at the moment we break when defining i.e.

```
unsafe void Print(string txt) {
  // do stuff
}
```


DustinCampbell:
Totally.

filipw:
👍

PR #782, opened 2017-02-22
Merge latest changes from my migrate-to-csproj fork

DustinCampbell:
I'm not expecting CI to work because I expect Travis doesn't support Mono 4.8.0 yet. However, it works on my machine! :smile:


DustinCampbell:
Getting closer! The OSX build succeeds now but the Linux build fails. It looks like the issue is that MSBuild isn't available in Mono 4.8.0 on Ubuntu.

DustinCampbell:
OK. Sounds like there's an "msbuild" package that needs to be installed for Linux.

DustinCampbell:
And... it looks like the msbuild package isn't recent enough to support the latest .NET Core projects. :-(

DustinCampbell:
Xamarin has updated the msbuild Debian package. I think it might work now!

DustinCampbell:
OK. I'm going to go ahead and pull this into the migrate-to-csproj branch, now that it's working.

PR #785, opened 2017-02-28
use IgnoreCorLibraryDuplicatedTypes binder flag in scripts

### LINKED ISSUE #784: Set IgnoreCorLibraryDuplicatedTypes for scripting
At the moment, in some edge cases, it's possible that we would incorrectly report the following type of issue in scripting

```
error CS0433: The type 'Console' exists in both 'System.Console, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' 
and 'mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089
```

This has been fixed in Roslyn in this PR https://github.com/dotnet/roslyn/commit/43a72d9a7ae6cdee6053253c184f6472ec632c9a however, to enable this we'll need to use reflection as the API that solves the problem (`.WithTopLevelBinderFlags(BinderFlags.IgnoreCorLibraryDuplicatedTypes))`) is internal.

filipw:
This is what the Roslyn Script Compiler API does internally (see https://github.com/dotnet/roslyn/blob/version-2.0.0-rc4/src/Scripting/CSharp/CSharpScriptCompiler.cs#L67). We should do the same in order to avoid invalid error reporting in some cases (unfortunately the API required is internal).

Fixes #784

### DIFF
--- a/src/OmniSharp.Script/ScriptProjectSystem.cs
+++ b/src/OmniSharp.Script/ScriptProjectSystem.cs
@@ -65,6 +65,11 @@ namespace OmniSharp.Script
             _env = env;
             _logger = loggerFactory.CreateLogger<ScriptProjectSystem>();
             _projects = new Dictionary<string, ProjectInfo>();
+
+            var topLevelBinderFlagsProperty = typeof(CSharpCompilationOptions).GetProperty("TopLevelBinderFlags", BindingFlags.Instance | BindingFlags.NonPublic);
+            var binderFlagsType = typeof(CSharpCompilationOptions).GetTypeInfo().Assembly.GetType("Microsoft.CodeAnalysis.CSharp.BinderFlags");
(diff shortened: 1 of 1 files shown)


DustinCampbell:
+ @tmat as FYI about needing a public API to avoid using reflection.

tmat:
The underlying issue is being addressed in Core CLR: https://github.com/dotnet/corefx/issues/5540
I'd suggest creating a tracking issue to remove the reflection once the Core CLR fix ships.

DustinCampbell:
Good idea -- thanks Tomas!

filipw:
Thanks guys 👍 

I opened a tracking issues https://github.com/OmniSharp/omnisharp-roslyn/issues/786.
BTW, looks like the exact same reflection based trick is implemented in PowerShell https://github.com/PowerShell/PowerShell/pull/2704 and in Orleans https://github.com/dotnet/orleans/pull/2466.

tmat:
Yes, I'm aware of these two.

PR #790, opened 2017-03-09
Script code check improvements

filipw:
- instead of excluding some diagnostics in the `CodeCheckService` (which was a hack), moved the suppression to the compilation options of the `ScriptProjectSystem`
 - we do not need to suppress `CS8099` anymore, because we are setting `SourceFileResolver` on the compilation (beforehand, it was missing, which generated the diagnostic).

### DIFF
--- a/src/OmniSharp.Roslyn.CSharp/Services/Diagnostics/CodeCheckService.cs
+++ b/src/OmniSharp.Roslyn.CSharp/Services/Diagnostics/CodeCheckService.cs
@@ -33,13 +33,6 @@ namespace OmniSharp.Roslyn.CSharp.Services.Diagnostics
                 var semanticModel = await document.GetSemanticModelAsync();
                 IEnumerable<Diagnostic> diagnostics = semanticModel.GetDiagnostics();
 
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -31,6 +31,7 @@ buildlog
 /omnisharp*.tar.gz
 scripts/Omnisharp*
 .msbuild-*/
+packages/
```
DustinCampbell: Was this intentional?

filipw: ugh, I somehow had that folder in the root. after git clean and rebuild, it's gone, must have been a leftover from a long ago; I will revert this

filipw: removed

david-driscoll:
Mac build fell over, restarted.

DustinCampbell:
Yeah, I restarted the Mac build as well this morning. I guess Travis is having trouble?

DustinCampbell:
OK. I'm going to assume this is good. It's passed lots of times at this point. However, it looks like Travis is dropping PR events and not triggering.

filipw:
thanks

PR #791, opened 2017-03-09
Update Roslyn, NuGet and MEF packages

DustinCampbell:


filipw:
✨

tools/packages.config
```diff
@@ -2,12 +2,12 @@
...
     <package id="Microsoft.Build.Runtime" version="15.1.548" />
-    <package id="Microsoft.NET.Sdk" version="1.0.0-alpha-20170217-2" />
+    <package id="Microsoft.NET.Sdk" version="1.0.0-alpha-20170224-6" />
     <package id="Microsoft.NET.Sdk.Web" version="1.0.0-alpha-20170130-3-281" />
     <package id="Microsoft.NET.Sdk.Publish" version="1.0.0-alpha-20170130-3-281" />
     <package id="Microsoft.NET.Sdk.Web.ProjectSystem" version="1.0.0-alpha-20170130-3-281" />
-    <package id="Microsoft.Net.Compilers" version="2.0.0-rc4-61325-08" />
-    <package id="NuGet.Build.Tasks" version="4.0.0-rtm-2283" />
-    <package id="NuGet.Build.Tasks.Pack" version="4.0.0-rtm-2283" />
+    <package id="Microsoft.Net.Compilers" version="2.0.1" />
```
filipw: what happened here 😄 

screenshot 2017-03-09 22 27 20

DustinCampbell: I asked the same question in my team room. Somebody failed about five months ago I guess. :smile:

DustinCampbell: That was according to @jasonmalinowski and @pilchie.

Pilchie: Yep, every rolling build of Roslyn produces 3 sets of nuget packages:  Next build-numbered version for myget, next major pre-release (beta/rc/etc), next release.

This means that we don't need to know ahead of time if a given build is going to be our last.

On the other hand, it means that the wrong packages are laying around and the final release packages can be uploaded by mistake instead of the rc ones for example...

filipw:
All green now, let's move it!

PR #794, opened 2017-03-10
cache ScriptMetadataResolver results

filipw:
This fixes https://github.com/OmniSharp/omnisharp-vscode/issues/1306

Basically, instead of letting `ScriptMetadataResolver` run entire reference resolution every time the semantic model is needed (which is how it is right now, and can sky rocket into GBs of memory usage, because it happens pretty much on every request into OmniSharp), we cache the results and reuse the same references over and over. This alleviates the memory usage.

When a new `#r` is added by user, it will still cause a jump in RAM but it would be much much smaller than what can be observed now.


filipw:
thanks 🍺

veremenko-y:
Awesome! Thanks @filipw!

PR #796, opened 2017-03-10
Try installing Mono via apt-get and Homebrew

DustinCampbell:


DustinCampbell:
This doesn't appear to be any better.

PR #797, opened 2017-03-13
Fix null reference in DotNetProjectSystem when project reference is invalid

DustinCampbell:
This keeps the DotNetProjectSystem continuing to load in the face of missing project references. It still reports the failure, but continues to load.


filipw:
:shipit:

DustinCampbell:
Thanks!

PR #800, opened 2017-03-20
Re-work test infrastructure

DustinCampbell:
This PR introduces a number of new features for the test infrastructure, simplifying how tests are written.

* Test assets can be shadow-copied to a temporary location and cleaned up after the test run. This allows for tests to perform operations on test assets that might change them. E.g. running 'dotnet restore' during a test.
* An `OmniSharpTestHost` has been added which replicates an OmniSharp host in test cases. When using this, all OmniSharp services should be present in the host's MEF composition. So, in many cases, tests no longer have to create the service they're testing and a dummy `OmniSharpWorkspace`. Instead, they create an `OmniSharpTestHost` in a using statement and access the service directly.
* Base classes for various test fixtures have helpers to make it easy to create an `OmniSharpTestHost`
* A new test fixture that makes it easier to test a single request handler.

There's some additional clean up in here as I'm preparing for VSTest support. I apologize for including that in the PR if you find it confusing.


DustinCampbell:
Seems like Travis is failing to acquire the Mono 4.8.0 MDK on OSX. I tried the same `curl` on my Mac and it worked fine. 😞  The build also works fine on my machine and tests pass.

DustinCampbell:
Hmmm... Travis's status for OSX builds looks fishy too. :smile:

image

DustinCampbell:
Looks like it just started moving again. :smile:

image

DustinCampbell:
There we go. Clean build!

src/OmniSharp.Host/Startup.cs
```diff
@@ -27,44 +25,40 @@ namespace OmniSharp
...
-        public Startup(IOmniSharpEnvironment omniSharpEnvironment)
+        public Startup(IOmniSharpEnvironment env)
         {
-            _omniSharpEnvironment = omniSharpEnvironment;
-            var appEnv = PlatformServices.Default.Application;
+            _env = env;
 
             var configBuilder = new ConfigurationBuilder()
-                .SetBasePath(appEnv.ApplicationBasePath)
+                .SetBasePath(AppContext.BaseDirectory)
```
filipw: 👍

src/OmniSharp.Host/Startup.cs
```diff
@@ -130,88 +159,84 @@ public void ConfigureServices(IServiceCollection services)
...
-            // ProjectEventForwarder register event to OmnisharpWorkspace during instantiation
-            PluginHost.GetExport<ProjectEventForwarder>();
+            InitializeWorkspace(Workspace, PluginHost, Configuration, logger);
 
-            // Initialize all the project systems
-            foreach (var projectSystem in PluginHost.GetExports<IProjectSystem>())
+            logger.LogInformation("Configuration finished.");
+        }
+
+        private static List<Assembly> DiscoverOmniSharpAssemblies(IAssemblyLoader loader, ILogger logger)
```
filipw: thanks, this is much more readable than the previous `shouldLoad` delegate 😄

DustinCampbell: Sometimes I need to unwind LINQ for my addled brain to understand it. :smile:

david-driscoll: You hate my linq don't you? 😭

src/OmniSharp.Roslyn.CSharp/Services/Buffer/UpdateBufferService.cs
```diff
@@ -20,7 +20,7 @@ public UpdateBufferService(OmniSharpWorkspace workspace)
 
         public async Task<object> Handle(UpdateBufferRequest request)
         {
-            await _workspace.BufferManager.UpdateBuffer(request);
+            await _workspace.BufferManager.UpdateBufferAsync(request);
```
david-driscoll: Really Async naming???? 😲 

(However since we leverage Roslyn it makes sense to follow similar patterns here 👍 )

DustinCampbell: /shrug - It's the recommended pattern for .NET in general.

david-driscoll: I prefer without, but that's a personal preference. 🍶

DustinCampbell: I do whatever Stephen Toub tells me to do. I have for years. :smile:

DustinCampbell: http://stackoverflow.com/questions/15951774/does-the-use-of-the-async-suffix-in-a-method-name-depend-on-whether-the-async

DustinCampbell: Note that I'm not make the Handle method async. I think that's overkill. However, I ran into an issue where I'd failed to await a Task returned from a method because it does not state it was Async, which caused me an hour of debugging to find the source of a bizarre re-entrancy problem. When I realized my fail, I went ahead and renamed. :smile:

PR #803, opened 2017-03-23
Add first actual unit test for legacy 'dotnet test' support

DustinCampbell:
This unit test copies a project.json test asset to a temporary location, restores it, and runs 'dotnet test' through the appropriate service.


filipw:
this is implicitly testing against `1.0.0-preview2-1-003177` cause that's what will be inside `.dotnet-legacy`, right?

DustinCampbell:
> this is implicitly testing against 1.0.0-preview2-1-003177 cause that's what will be inside .dotnet-legacy, right?

That's right. This is a test to make sure project.json dotnet test keeps working. I'm going to make a bunch of changes to it in order to support csproj dotnet test, so I'm adding test infrastructure to ensure I don't break anything.

filipw:
Let's get this in!

Issue #366, opened 2015-12-15
Improve formatting options experience


jrieken:
_From @anfomin on December 14, 2015 17:34_

I've configured VS Code to always use tabs:

``` json
// Place your settings in this file to overwrite the default settings
{
    "editor.insertSpaces": false
}
```

But C# **Format Code** command (Shift+Alt+F) always replaces tabs with spaces. I'm using OS X El Capitan.

_Copied from original issue: Microsoft/vscode#1300_

jrieken:
This is because OmniSharp doesn't support per request formatting options (https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp.Abstractions/Models/v1/FormatRangeRequest.cs#L6). Moving to OmniSharp for that.

jrieken:
@anfomin - Do workaround this, you can create an `omnisharp.json` with the config

anfomin:
@jrieken I tried to create `omnisharp.json` in my project room directory with content:
```json
{
	"TextEditorOptions": {
		"tabsToSpaces": false
	}
}
```

But `Format Code` command still uses spaces.

anfomin:
I have found file I needed:
```json
{
	"FormattingOptions": {
		"useTabs": true
	}
}
```

Now it's OK.

0xdeafcafe:
Is there a way to affect omnisharp globally with the method above?

filipw:
here is a new wiki article I just posted today, hopefully this helps 
https://github.com/OmniSharp/omnisharp-roslyn/wiki/Configuration%20Options

When the new C# Extension for VS Code is released, the VS Code intendation/tab settings will be inherited by OmniSharp automatically.

DustinCampbell:
There are two things that we need to support to improve the experience.

1. The omnisharp.json file should be reloaded when it changes so that users don't have to restart OmniSharp after changing this file.
2. We need a new end point for specifying formatting options. This will allow hosts to pass along options that aren't stored in omnisharp.json file. (E.g. omnisharp-vscode passes the "editor.insertSpaces" and "editor.tabSize" VS Code options along as "formattingOptions:useTabs", "formattingOptions:tabSize" and "formattingOptions:indentationSize" to OmniSharp.

filipw:
I can go ahead and implement this

Thaina:
I'm not so sure that omnisharp itself was normally listening to files changed in workspace, or is it?
[...]

PR #804, opened 2017-03-24
Watch for changes in omnisharp.json

filipw:
This PR introduces live change tracking in `omnisharp.json`. There are a couple of issues open for that, will attach them to this PR when I locate them 😄 
At the moment changes in formatting settings in `omnisharp.json` are only read by OmniSharp at startup - so if you change anything you need to restart the server to have them take effect which is not an ideal experience.

Note that we do not roll out a file watcher by hand, but rather rely on change tracking capabilities of `Microsoft.Extensions.Configuration`. On OmniSharp side, technically, the way it's implemented, is that we let each of our `IWorkspaceOptionsProvider` run as soon as the configuration change is reported.

Here is the change in action:


filipw:
related issue: (see comment 1) https://github.com/OmniSharp/omnisharp-roslyn/issues/366#issuecomment-268984041

src/OmniSharp.Host/Startup.cs
```diff
@@ -20,6 +20,7 @@
 using OmniSharp.Services.FileWatching;
 using OmniSharp.Stdio.Logging;
 using OmniSharp.Stdio.Services;
+using System.IO;
```
DustinCampbell: sort order? Shouldn't System namespaces be on top?

filipw:
> What options does this work for?

it deals with the strongly typed [OmnisharpOptions](https://github.com/OmniSharp/omnisharp-roslyn/blob/dev/src/OmniSharp.Abstractions/Options/OmniSharpOptions.cs) so at the moment that's formatting only

src/OmniSharp.Host/Startup.cs
filipw: yes, I deleted it now as it wasn't even used; it crept in here because of some code that I ended up ultimately removing

filipw:
related issue: https://github.com/OmniSharp/omnisharp-vscode/issues/1290#issuecomment-284827335

filipw:
one more related issue: https://github.com/OmniSharp/omnisharp-vscode/issues/1080

src/OmniSharp.Host/Startup.cs
```diff
@@ -49,7 +49,7 @@ public Startup(IOmniSharpEnvironment env)
                 new PhysicalFileProvider(env.Path),
                 "omnisharp.json",
                 optional: true,
-                reloadOnChange: false);
+                reloadOnChange: true);
```
davidfowl: 👏

Issue #806, opened 2017-03-28
unable to build tests on windows 10 x64: missing xunit.runner.utility.desktop.dll


jmzagorski:
I tried running build.ps1 on windows 10 x64 after getting the net451 developer kit, but it fails `Error: Could not find file ...\omnisharp-roslyn\.tools\xunit.runner.console\tools\xunit.runner.utility.desktop.dll`  (i trimmed the full path in that example) and I can verify the file is not there. I am unsure where to look to try and debug this error or if it is specified to windows 10. I recently migrated from windows 7 where I did not have build errors.

```
omnisharp-roslyn> .\build.ps1
Preparing to run build script...
Running build script...
Analyzing build script...
Processing build script...
Installing addins...
Compiling build script...

========================================
Cleanup
========================================
Executing task: Cleanup
Finished executing task: Cleanup

========================================
BuildEnvironment
========================================
Executing task: BuildEnvironment
dotnet-install: .NET SDK version 1.0.0-preview2-003121 is already installed.
.NET Command Line Tools (1.0.0-preview2-003121)

Product Information:
 Version:            1.0.0-preview2-003121
 Commit SHA-1 hash:  1e9d529bc5

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.14393
 OS Platform: Windows
 RID:         win10-x64
Feeds used:
  https://dotnet.myget.org/F/dotnet-buildtools/api/v3/index.json
  https://api.nuget.org/v3/index.json

Package "xunit.runner.console.2.2.0" is already installed.
Finished executing task: BuildEnvironment

========================================
PopulateRuntimes
========================================
Executing task: PopulateRuntimes
Finished executing task: PopulateRuntimes

========================================
Setup
========================================
Executing task: Setup
Finished executing task: Setup

========================================
Restore
========================================
Executing task: Restore
Restoring packages....
Package restore successful!
Restoring packages....
Package restore successful!
Finished executing task: Restore

========================================
BuildTest
========================================
Executing task: BuildTest
Finished executing task: BuildTest

========================================
Test
========================================
Executing task: Test
An error occured when executing task 'Test'.
Error: Could not find file ...\omnisharp-roslyn\.tools\xunit.runner.console\tools\xunit.runner.utility.desktop.dll'.
```

DustinCampbell:
[...]

PR #807, opened 2017-03-28
Add retry logic for test clean up

DustinCampbell:
Avoid `UnauthorizedAccessException` when cleaning up test directories if any of the contents are still locked by another process (such as 'dotnet').

Should address #806.

### DIFF
--- a/tests/TestUtility/TestAssets.TestProject.cs
+++ b/tests/TestUtility/TestAssets.TestProject.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Threading;
 
 namespace TestUtility
 {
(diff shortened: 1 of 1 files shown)


DustinCampbell:
Hmmm... Travis's Linux build seems to hang when running OmniSharp.DotNetTest.Tests, which is the set of tests that require clean up. However, the OSX build passes and I just tried on my own Ubuntu 16.04 machine and it passes there too.

😕

DustinCampbell:
Hmmm... retried a couple of times and Ubuntu carried on and finished successfully. Travis blew up a couple of those times setting up the instance, so I wonder if this is just some intermittent issue?

david-driscoll:
It's entirely possible 😞

PR #809, opened 2017-03-30
Added support for a global omnisharp.json

### LINKED ISSUE #717: Global omnisharp.json
_From @zeroskyx on December 30, 2016 17:12_

Is there any way to use omnisharp.json globally instead of having to copy it to every project manually?
Placing one in C:\Users\<User>\AppData\Roaming\Code\User does not seem to work.

Help is greatly appreciated. Thanks in advance
-Simon

_Copied from original issue: OmniSharp/omnisharp-vscode#1101_

filipw:
Fixes #717 

The global `omnisharp.json` file location is:
 - On Windows: `%APPDATA%\OmniSharp\omnisharp.json`
 - On Mac/Linux: `~/.omnisharp/omnisharp.json`

 The folder and an empty `{}` file are created automatically if they are not there.
 The live reload logic introduced here https://github.com/OmniSharp/omnisharp-roslyn/pull/804 also supports this global file, meaning any changes to it are automatically reflected.
 The global `omnisharp.json` file is given a lower precedence than a local one, so global settings can be overridden locally. It is however given a higher precedence than environment variables and command line arguments.


src/OmniSharp.Host/Services/OmniSharpEnvironment.cs
```diff
@@ -38,6 +40,20 @@ public class OmniSharpEnvironment : IOmniSharpEnvironment
             TraceType = traceType;
             TransportType = transportType;
             OtherArgs = otherArgs;
+
+            // On Windows: %APPDATA%\OmniSharp\omnisharp.json
```
DustinCampbell: Should this be "%USERPROFILE%\.omnisharp\omnisharp.json" to match more closely to the Unix location? Usually, users aren't expected to edit or place files under %APPDATA%.

src/OmniSharp.Host/Startup.cs
```diff
@@ -20,6 +20,8 @@
 using OmniSharp.Services.FileWatching;
 using OmniSharp.Stdio.Logging;
 using OmniSharp.Stdio.Services;
+using System.IO;
+using OmniSharp.Host.Internal;
```
DustinCampbell: Sort usings?

src/OmniSharp.Host/Startup.cs
```diff
@@ -36,18 +38,21 @@ public Startup(IOmniSharpEnvironment env)
...
+                .AddJsonFile(OmniSharpConstants.OmnisharpConfigFile, optional: true)
                 .AddEnvironmentVariables();
 
             if (env.OtherArgs?.Length > 0)
             {
                 configBuilder.AddCommandLine(env.OtherArgs);
             }
 
+            // Use the global omnisharp config if there's any in the shared path
+            configBuilder.CreateAndAddGlobalOptionsFile(env);
```
DustinCampbell: It strikes me a little weird that we'd start creating files on the user's machine without them knowing, even if it's an empty file for convenience.

src/OmniSharp.Host/Internal/OmniSharpConstants.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace OmniSharp.Host.Internal
+{
+    internal static class OmniSharpConstants
+    {
+        internal const string OmnisharpConfigFile = "config.json";
+        internal const string OmnisharpOptionsFile = "omnisharp.json";
+    }
+}
```
DustinCampbell: Multiple references to "OmniSharp" seem redundant. Maybe remove it from the member names? Also, the member names have different casing of "OmniSharp". :smile:

src/OmniSharp.Host/Internal/ConfigurationBuilderExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+                var omnisharpGlobalFilePath = Path.Combine(env.SharedPath, OmniSharpConstants.OmnisharpOptionsFile);
+                if (!File.Exists(omnisharpGlobalFilePath))
+                {
+                    File.WriteAllText(omnisharpGlobalFilePath, "{}");
+                }
+
+                configBuilder.AddJsonFile(
+                    new PhysicalFileProvider(env.SharedPath),
+                    OmniSharpConstants.OmnisharpOptionsFile,
+                    optional: true,
```
DustinCampbell: We make the file optional but always create it anyway?

src/OmniSharp.Host/Services/OmniSharpEnvironment.cs
```diff
@@ -1,12 +1,14 @@
 ﻿using System;
 using Microsoft.Extensions.Logging;
+using System.IO;
```
DustinCampbell: Sort usings

src/OmniSharp.Host/Services/OmniSharpEnvironment.cs
```diff
@@ -1,12 +1,14 @@
...
 
 namespace OmniSharp.Services
 {
     public class OmniSharpEnvironment : IOmniSharpEnvironment
     {
         public string Path { get; }
         public string SolutionFilePath { get; }
+        public string SharedPath { get; }
```
DustinCampbell: Should this be `SharedDirectoryPath` for clarity?

src/OmniSharp.Host/Services/OmniSharpEnvironment.cs
```diff
@@ -38,6 +40,20 @@ public class OmniSharpEnvironment : IOmniSharpEnvironment
             TraceType = traceType;
             TransportType = transportType;
             OtherArgs = otherArgs;
+
+            // On Windows: %APPDATA%\OmniSharp\omnisharp.json
+            // On Mac/Linux: ~/.omnisharp/omnisharp.json
+            var root = Environment.GetEnvironmentVariable("APPDATA") ??
+                Environment.GetEnvironmentVariable("HOME");
+
+            if (!string.IsNullOrEmpty(Environment.GetEnvironmentVariable("APPDATA")))
```
DustinCampbell: Nit: It seems a little funny that we get the environment variable for APPDATA and then get it again immediately.

src/OmniSharp.Host/Startup.cs
filipw: I agree, I was a bit on the fence with this one. The idea was to be able to say: "got here, and edit the `omnisharp.json` file to have global shared settings", rather than having to say "please create it".

However, you are right, an empty file is a bit silly. 

I think we should still create the folder though, it would be somewhat awkward to have to manually create folders that OmniSharp needs.

filipw:
I moved it from `APPDATA` to `USERPROFILE` as suggested - the paths are now unified for both OS families (`{HOME}|{USERPROFILE}/.omnisharp/omnisharp.json`). There is no more automatic file creation, but we still have directory creation there. 

Aside from convenience for the user, having the directory there allows us to use the file watcher under the hood of the Options library - as it can watch for a non-existing file, and light up as soon as it's created, which is a nice user experience. This wouldn't be possible if the folder wasn't there.

Also addressed all the other code review feedback.

src/OmniSharp.Host/Services/OmniSharpEnvironment.cs
david-driscoll: Wasn't there a huge fight about that when `dotnet` was still `dnx`?  And they moved things to `APPDATA` because that was the "windows way"?  Some people wanted `.dotnet` and others wanted it elsewhere?

Personally I like using `USERPROFILE` myself, but I don't to make things any more confusing than they have to be.

src/OmniSharp.Host/Startup.cs
david-driscoll: Other things (not necessarily language servers specificly) do this kind of thing.  If the files don't exist create them.  Atom or VSCode both on start.

From a UX point of view it would be easier to point the user to a file and say "Edit your global settings here".  However since we (omnisharp) aren't really an editor, we could leave that for the editors to implement themselves.

src/OmniSharp.Host/Services/OmniSharpEnvironment.cs
DustinCampbell: No idea. On my machine, I see the following under `USERPROFILE`:

* .dnx
* .node-gyp
* .nuget
* .templateengine
* .vscode
* .vscode-insiders
* .vscode-oss-dev

I think we're in good company in picking `USERPROFILE`. :smile:

src/OmniSharp.Host/Startup.cs
DustinCampbell: Yup. I gated my review on APPDATA vs. USERPROFILE and not whether we create the file or not. I just commented that it struck me as a little weird. :smile:

Shall we keep it as is and change it if we need to?

filipw: the local file `{working_dir}/omnisharp.json` is not implicitly created, and users have to manually add it, so at least in that sense the behavior for the global one is consistent

DustinCampbell:
I think we're in good shape here. @david-driscoll, holler if you disagree.

DustinCampbell:
@filipw: I think this change warrants an update to https://github.com/OmniSharp/omnisharp-roslyn/wiki/Configuration-Options.

filipw:
indeed 📚

PR #812, opened 2017-04-03
Add PhysicalFileProvider.WrapForPolling() extension method to avoid using FileSystemWatcher

DustinCampbell:
Recently, we've added support for reloading the user's local or global OmniSharp options when the relevant "omnisharp.json" file changes on disk. This causes the Microsoft.Extensions.Configuration API to start using a FileSystemWatcher. Unfortunately, this watcher looks for changes recursively, which can easily hit the [ulimit](https://ss64.com/osx/ulimit.html) for open files OSX just to watch for a couple of files.

This change adds a wrapper for PhysicalFileProvider that polls the file system every 2 seconds for just the files we care about.

This change also rationalizes some of the property names in OmniSharpEnvironment and fixes an issue where "Path" could be invalid if it was specified relative to the current working directory. E.g., if the OmniSharp is started with `-s OmniSharp.sln` from the directory where that file exists, Path would be an empty string.

### DIFF
--- a/src/OmniSharp.Host/Program.cs
+++ b/src/OmniSharp.Host/Program.cs
@@ -61,7 +61,7 @@ namespace OmniSharp
             var omnisharpApp = new CommandLineApplication(throwOnUnexpectedArg: false);
             omnisharpApp.HelpOption("-? | -h | --help");
 
-            var applicationRootOption = omnisharpApp.Option("-s | --source", "Solution, project file or directory for OmniSharp to point at (defaults to current directory).", CommandOptionType.SingleValue);
+            var applicationRootOption = omnisharpApp.Option("-s | --source", "Solution or directory for OmniSharp to point at (defaults to current directory).", CommandOptionType.SingleValue);
             var portOption = omnisharpApp.Option("-p | --port", "OmniSharp port (defaults to 2000).", CommandOptionType.SingleValue);
             var logLevelOption = omnisharpApp.Option("-l | --loglevel", "Level of logging (defaults to 'Information').", CommandOptionType.SingleValue);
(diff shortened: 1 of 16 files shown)


src/OmniSharp.Host/Program.cs
```diff
@@ -61,7 +61,7 @@ private static int Run(string[] args)
             var omnisharpApp = new CommandLineApplication(throwOnUnexpectedArg: false);
             omnisharpApp.HelpOption("-? | -h | --help");
 
-            var applicationRootOption = omnisharpApp.Option("-s | --source", "Solution, project file or directory for OmniSharp to point at (defaults to current directory).", CommandOptionType.SingleValue);
```
DustinCampbell: From what I could tell, all of our project systems assume that `OmniSharpEnvironment.Path` is a directory. If OmniSharp had been launched with anything other than a path to a directory or a solution file, it wouldn't have worked. Let me know if I've read this incorrectly.

src/OmniSharp.Host/Program.cs
```diff
@@ -61,7 +61,7 @@ private static int Run(string[] args)
...
+            var applicationRootOption = omnisharpApp.Option("-s | --source", "Solution or directory for OmniSharp to point at (defaults to current directory).", CommandOptionType.SingleValue);
```
DustinCampbell: From what I could tell, all of our project systems assume that `OmniSharpEnvironment.Path` is a directory. If OmniSharp had been launched with anything other than a path to a directory or a solution file, it wouldn't have worked. Please let me know if I've read this incorrectly.

filipw:
Sorry for adding you more work 😥  :1st_place_medal:

src/OmniSharp.Host/Program.cs
filipw: correct

DustinCampbell:
Nah. This wasn't your fault at all.

PR #813, opened 2017-04-04
Added support for referencing NuGet packages in C# scripts

seesharper:
This PR adds support for resolving NuGet packages in C# scripts 

Example 
```
#r "nuget:AutoMapper/6.0.0"
```
The feature is implemented by decorating the default MetadataReferenceResolver with a resolver that will know how to handle references to NuGet packages. The code for the NuGetMetadataReferenceResolver is currently maintained at this repo 

https://github.com/seesharper/Dotnet.Script.NuGetMetadataResolver

The reason for keeping the NuGetMetadataReferenceResolver outside of this project is so that it can be used by other script runners such as https://github.com/filipw/dotnet-script


src/OmniSharp.Script/ScriptProjectSystem.cs
```diff
@@ -79,16 +89,29 @@ public class ScriptProjectSystem : IProjectSystem
         private readonly Dictionary<string, ProjectInfo> _projects;
         private readonly OmniSharpWorkspace _workspace;
         private readonly IOmniSharpEnvironment _env;
+        private readonly ILoggerFactory loggerFactory;
```
DustinCampbell: Please add an underscore to match the other fields in this file.

src/OmniSharp.Script/ScriptProjectSystem.cs
```diff
@@ -39,38 +42,45 @@ public class ScriptProjectSystem : IProjectSystem
...
         private static readonly CSharpParseOptions ParseOptions = new CSharpParseOptions(LanguageVersion.Default, DocumentationMode.Parse, SourceCodeKind.Script);
-
-        private static readonly Lazy<CSharpCompilationOptions> CompilationOptions = new Lazy<CSharpCompilationOptions>(() =>
+        private readonly Lazy<CSharpCompilationOptions> _compilationOptions;
+        private CSharpCompilationOptions CreateCompilation()
         {
+            var resolver = NuGetMetadataReferenceResolver.Create(ScriptMetadataResolver.Default,
+                NugetFrameworkProvider.GetFrameworkNameFromAssembly(),loggerFactory, _env.Path);
+           
+
```
DustinCampbell: remove extra blank line.

src/OmniSharp.Script/ScriptProjectSystem.cs
```diff
@@ -18,6 +18,9 @@
 
 namespace OmniSharp.Script
 {
+    using Dotnet.Script.NuGetMetadataResolver;    
+    using Microsoft.DotNet.InternalAbstractions;
```
DustinCampbell: Please don't nest usings inside of namespaces. Put them with the rest of the namespaces at the top of the file.

src/OmniSharp.Script/CachingScriptMetadataResolver.cs
```diff
@@ -1,27 +1,32 @@
...
 namespace OmniSharp.Script
 {
     public class CachingScriptMetadataResolver : MetadataReferenceResolver
     {
+        private readonly MetadataReferenceResolver defaultReferenceResolver;
         private static Dictionary<string, ImmutableArray<PortableExecutableReference>> DirectReferenceCache = new Dictionary<string, ImmutableArray<PortableExecutableReference>>();
         private static Dictionary<string, PortableExecutableReference> MissingReferenceCache = new Dictionary<string, PortableExecutableReference>();
-        private static MetadataReferenceResolver _defaultRuntimeResolver = ScriptMetadataResolver.Default;
+
+
```
DustinCampbell: remove extra blank line

src/OmniSharp.Script/ScriptProjectSystem.cs
```diff
@@ -39,38 +42,45 @@ public class ScriptProjectSystem : IProjectSystem
 
         private const string CsxExtension = ".csx";
         private static readonly CSharpParseOptions ParseOptions = new CSharpParseOptions(LanguageVersion.Default, DocumentationMode.Parse, SourceCodeKind.Script);
-
-        private static readonly Lazy<CSharpCompilationOptions> CompilationOptions = new Lazy<CSharpCompilationOptions>(() =>
+        private readonly Lazy<CSharpCompilationOptions> _compilationOptions;
+        private CSharpCompilationOptions CreateCompilation()
         {
+            var resolver = NuGetMetadataReferenceResolver.Create(ScriptMetadataResolver.Default,
+                NugetFrameworkProvider.GetFrameworkNameFromAssembly(),loggerFactory, _env.Path);
```
DustinCampbell: missing space between `,` and `loggerFactory`

src/OmniSharp.Script/CachingScriptMetadataResolver.cs
```diff
@@ -1,27 +1,32 @@
...
+        public CachingScriptMetadataResolver(MetadataReferenceResolver defaultReferenceResolver)
+        {
+            this.defaultReferenceResolver = defaultReferenceResolver;
```
DustinCampbell: Please use underscore for private field name.

src/OmniSharp.Script/ScriptProjectSystem.cs
```diff
@@ -39,38 +42,45 @@ public class ScriptProjectSystem : IProjectSystem
...
-                metadataReferenceResolver: new CachingScriptMetadataResolver(),
-                sourceReferenceResolver: ScriptSourceResolver.Default,
-                assemblyIdentityComparer: DesktopAssemblyIdentityComparer.Default).
+                    OutputKind.DynamicallyLinkedLibrary,
+                    usings: DefaultNamespaces,
+                    allowUnsafe: true,
+                    metadataReferenceResolver:
+                    new CachingScriptMetadataResolver(resolver),
+                    sourceReferenceResolver: ScriptSourceResolver.Default,
+                    assemblyIdentityComparer: DesktopAssemblyIdentityComparer.Default).
```
DustinCampbell: If you like, maybe move the `.` to the next line like you've done elsewhere. That always looks better to me.

src/OmniSharp.Script/ScriptProjectSystem.cs
```diff
@@ -121,18 +144,38 @@ public void Initalize(IConfiguration configuration)
...
+                {                    
+                    var runtimeId = RuntimeEnvironment.GetRuntimeIdentifier();
+                    var inheritedAssemblyNames = DependencyContext.Default.GetRuntimeAssemblyNames(runtimeId).Where(x =>
+                        x.FullName.StartsWith("system.", StringComparison.OrdinalIgnoreCase) ||
+                        x.FullName.StartsWith("microsoft.codeanalysis", StringComparison.OrdinalIgnoreCase) ||
+                        x.FullName.StartsWith("mscorlib", StringComparison.OrdinalIgnoreCase));
+
+                    foreach (var inheritedAssemblyName in inheritedAssemblyNames)
+                    {
+                        var assembly = Assembly.Load(inheritedAssemblyName);
```
DustinCampbell: We generally use an [`IAssemblyLoader`](https://github.com/OmniSharp/omnisharp-roslyn/blob/dev/src/OmniSharp.Abstractions/Services/IAssemblyLoader.cs) so that we can track assemblies that are loaded at runtime. Just import `IAssemblyLoader` in the constructor and use it here.

src/OmniSharp.Script/ScriptProjectSystem.cs
DustinCampbell: `_env.Path` changed to `_env.TargetDirectory` in dev. That's why the build is failing.

DustinCampbell:
Thanks for the PR! My feedback is mostly nits to try and match the code style of the files you're changing. Also, the CI build is failing because `IOmniSharpEnvironment.Path` changed to `IOmniSharpEnvironment.TargetDirectory`.

seesharper:
Thanks for your feedback. I think I got them covered now :)

filipw:
@seesharper thanks a lot for the PR, always great to see people investing their time in CSX tooling 😀

I'm gonna loop in @tmat for his comments on this. Tomas, does this align with the Roslyn vision of supporting nuget references ([seems to me it does](https://github.com/dotnet/roslyn/issues/6900#issuecomment-230858010))? In general we want OmniSharp to stay aligned with CSI.exe scripting. In this case, we can get ahead of the curve a little bit, which I think would be fine, as long as syntax-wise this is what you have thought of for Roslyn down the road?

I also like that it's a done as only a custom `MetadataReferenceResolver` rather than a ton of machinery for injecting stuff into `ScriptOptions` and things like that.

src/OmniSharp.Script/ScriptProjectSystem.cs
```diff
@@ -79,16 +86,31 @@ public class ScriptProjectSystem : IProjectSystem
...
         private readonly OmniSharpWorkspace _workspace;
         private readonly IOmniSharpEnvironment _env;
+        private readonly ILoggerFactory _loggerFactory;
         private readonly ILogger _logger;
+        private readonly IAssemblyLoader _assemblyLoader;
+        private static readonly Lazy<string> _targetFrameWork = new Lazy<string>(ResolveTargetFramework);
+
+        private static string ResolveTargetFramework()
+        {
+            return Assembly.GetEntryAssembly().GetCustomAttributes()
```
filipw: what problem did you run into that you added this code?

seesharper: No problem. I was just making sure that I did not change the way metadata references are being added when running on the full desktop framework.

src/OmniSharp.Script/CachingScriptMetadataResolver.cs
```diff
@@ -47,7 +51,7 @@ public override ImmutableArray<PortableExecutableReference> ResolveReference(str
                 return DirectReferenceCache[key];
             }
 
-            var result = _defaultRuntimeResolver.ResolveReference(reference, baseFilePath, properties);
+            var result = _defaultReferenceResolver.ResolveReference(reference, baseFilePath, properties);
```
filipw: it's not visible what happens here for nuget packages, can you explain a little? do they get installed? what if there is no network? what if there is no `nuget` in the user's path? what's the lock file?

src/OmniSharp.Script/ScriptProjectSystem.cs
```diff
@@ -39,38 +41,43 @@ public class ScriptProjectSystem : IProjectSystem
 
         private const string CsxExtension = ".csx";
         private static readonly CSharpParseOptions ParseOptions = new CSharpParseOptions(LanguageVersion.Default, DocumentationMode.Parse, SourceCodeKind.Script);
-
-        private static readonly Lazy<CSharpCompilationOptions> CompilationOptions = new Lazy<CSharpCompilationOptions>(() =>
+        private readonly Lazy<CSharpCompilationOptions> _compilationOptions;
+        private CSharpCompilationOptions CreateCompilation()
         {
+            var resolver = NuGetMetadataReferenceResolver.Create(ScriptMetadataResolver.Default,
+                NugetFrameworkProvider.GetFrameworkNameFromAssembly(), _loggerFactory, _env.TargetDirectory);
```
filipw: Is this where the nuget framework to be used is determined? if so, I don't think it's correct to assume the version of nuget framework from the version of OmniSharp. A desktop OmniSharp (in fact, in VS Code only that version is used) can be used to provide language services for a .NET Core script and vice versa, a .NET Core OmniSharp can be used to provide language services for a destkop framework script

filipw:
Also, let's wait for input from @tmat but I think it would be better to delimit version with a comma. This is the design used in `#r` implementation in F# https://github.com/fsharp/fslang-design/blob/master/RFCs/FS-1027-fsi-references.md#summary

seesharper:
Yes, that was the general idea:) I will get back to you tomorrow with more details around how I ended up with this approach. 11pm here now :)

tmat:
>             if (DirectReferenceCache.ContainsKey(key))

It would be better to use a tuple here.

---
Refers to: src/OmniSharp.Script/CachingScriptMetadataResolver.cs:49 in f1d6dd7. [](commit_id = f1d6dd7442d60d49dd0f3da5dcced5bad236b5f2, deletion_comment = False)

tmat:
There is a related discussion in F# design repo: https://github.com/fsharp/fslang-design/issues/167 

Currently the syntax F# is considering is:
 
```#r "nuget: <package-name>, <package-version>"```

It would be good if C# and F# used the same syntax.

To make this work for all nupkgs (specifically with CoreFX nupkgs) we'll need to be able to deal with both reference assemblies and implementation assemblies in the resolver. Also someone pointed out that batching the resolution (i.e. have an API that allows the compiler to resolve multiple #r's in one go) would be necessary to avoid perf issues with discovering dependencies.

**Comment by tmat** [comment]
It would be useful to write up some specification, how the resolver works. For example, does it check ```NuGet.config``` files for feeds, how is the lookup cached across script executions -- is there a ```script.assets.json``` file that ```csi /restore script.csx``` would create, etc.?

File: src/OmniSharp.Script/ScriptProjectSystem.cs (hunk 11 above)
Comment (seesharper) [inline src/OmniSharp.Script/ScriptProjectSystem.cs]: How can we within OmniSharp determine if the we a running the script in the context of a .Net Core app when there is no project context? The whole idea here is to support "self-contained" scripts that has no additional context other than the script itself.

**Comment by seesharper** [comment]
The challenge , revisited 

* Being able provide language services for scripts even if they provide no context in the form of a project.json/csproj file. 

* Being able to resolve references to NuGet packages in C# scripts.

So how do OmniSharp know if it should provide metadata for the full framework or .Net Core?

The PR uses information from the host process (OmniSharp) to determine if we are running .Net Core or the full framework. This is probably not good enough as @filipw pointed out as OmniSharp can provide language services for both runtimes regardless of the runtime that OmniSharp itself runs in. 

So how can we provide OmniSharp with enough information about script target runtime?

What if we did something similar to SheBang on nix systems?

```
#!/bin/sh
``` 

we could do 

```
#r "runtime:NetCoreApp1.1"
```

In my head, it sort of makes sense that we "reference" the runtime.

If we find multiple csx files with different "runtimes" we would have to handle this somehow though.

File: src/OmniSharp.Script/ScriptProjectSystem.cs (hunk 11 above)
Comment (filipw) [inline src/OmniSharp.Script/ScriptProjectSystem.cs]: so far - because CSI is desktop FX only and there was no formal specification for how .NET Core scripting should behave, OmniSharp would assume desktop FX mode for scripts in case no `project.json` was there, which would be the typical majority use case. 

if `project.json` was found, then the first runtime FX defined there would be used.
in other words, the only way to force .NET Core scripting was through an external file, and I think for now this is how it should stay because it is non-standard scripting anyway.

Comment (seesharper) [inline src/OmniSharp.Script/ScriptProjectSystem.cs]: I understand that this is a problem and that is why .Net Core metadata references are only added when OmniSharp itself is running as a .Net Core app. It's not ideal, but at least it would provide a way for people wanting to try out project-less .Net Core scripting by pointing to the .Net Core OmniSharp binary from VsCode. Also commented on an alternative approach here .

https://github.com/OmniSharp/omnisharp-roslyn/pull/813#issuecomment-291850551

**Comment by filipw** [comment]
> So how do OmniSharp know if it should provide metadata for the full framework or .Net Core?

I think for the sake of baby steps and incremental progress we could consider for now that for no known runtime context (such as CSI scripts) we just assume latest desktop FX, but maybe I'm being overly simplistic.
The natural next step would then be to graduate the discussion to how the CSX could manifest it's runtime environment.

I fear that if we dive into this now we will never get anything out as the scope would just be too large; not to mention such discussion really belongs in the Roslyn repo as that's the codebase that will drive this, and OmniSharp would merely consume it.

File: src/OmniSharp.Script/ScriptProjectSystem.cs (hunk 11 above)
Comment (DustinCampbell) [inline src/OmniSharp.Script/ScriptProjectSystem.cs]: C# for VS Code does not ship a .NET Core version of OmniSharp. It runs OmniSharp with an embedded Mono runtime. So, this approach will not work unless the user downloads or builds a .NET Core-based OmniSharp and sets VS Code to use that. And, if they do that, they may have other problems since .NET Core OmniSharp can't handle csproj projects properly.

Note, this will also be a problem long term as the .NET Core versions of OmniSharp will be going away in favor of a Mono-based approach. See https://github.com/OmniSharp/omnisharp-roslyn/issues/666 for explanation.

Comment (seesharper) [inline src/OmniSharp.Script/ScriptProjectSystem.cs]: I sort of understood #666 wouldn't help me much :)
Jokes aside, is there anything we can do to hint OmniSharp about the scripting runtime?

What if we did this

```
#r "runtime:NetCoreApp1.1"
```
In the case of a non-existing project.json, we could create one at runtime based on the runtime reference and feed that into the existing logic that resolves metadata references based on project context,

In other words we create a project context based on script runtime references.

The really nice thing about this approach is that we could also add the NuGet references found in the script to the dynamically created project.json and basically skip the NuGetMetadataReferenceResolver all together.

**Comment by tmat** [comment]
@seesharper Great write up. Could you add it to the readme.md of https://github.com/seesharper/Dotnet.Script.NuGetMetadataResolver?

Re the resolver implementation: One thing I'm worried about is specifying all transitive dependencies explicitly. That's not very script-friendly. Note that ```#r "assembly.dll"``` currently automatically references all transitive dependencies. So it would be odd if ```#r "nuget"``` didn't do that. Also inconvenient for the user, especially in the REPL. I'd prefer if the resolver found all dependencies and stored them in a .json file next to the script. If such file already exists it would just use it and skip the dependency resolution unless ```#r nuget``` directives in the script changed. 

Re runtime: We already have support for ```#!``` -- it's just a comment for the compiler. But I imagine we can come up with some convention for specifying the host environment. I don't think we should be specifying runtime (Core CLR vs Desktop CLR) though but rather the target API set. For this purpose using #r on a (NETStandard.Library)[https://www.nuget.org/packages/NETStandard.Library/]" meta-package seems most fitting.

For example, 
```#r "nuget:NETStandard.Library, 1.6.1"```

I think there are two categories of scripts:

1) csi runnable scripts
These are universal scripts that might use host variables and default imports defined by csi (such as Args) but require no other host-specific types/variables/imports.

2) custom host scripts
These are host-specific scripts that assume an environment created by some application that runs them. This is where we could use ```#!```. Perhaps we could develop some convention that allows the app author to specify what the script environment looks like statically, to enable editors to provide intellisense.

**Comment by seesharper** [comment]
Before discussing further I would just like to show you what kind of script developer happiness I'm aiming 
for here. No project file or any other files besides the actual script file.

nuget

**Comment by seesharper** [comment]
I am not giving up on this yet :)
I have created a new version of DotNet.Script.NuGetMetadataResolver that takes a new approach to resolving NuGet packages. Updated readme at https://github.com/seesharper/Dotnet.Script.NuGetMetadataResolver

Question is, should I abandon this PR and start a new PR based on the latest sources?

**Comment by seesharper** [comment]
@tmat @DustinCampbell Could you take another look at this in the light of the changes I've made to 
https://github.com/seesharper/Dotnet.Script.NuGetMetadataResolver and the simplified ScriptProjectSystem?

**Comment by seesharper** [comment]
@tmat @DustinCampbell @filipw What do you say, guys. Is this happening or not?

File: NuGet.Config [hunk 12]
```diff
@@ -1,9 +1,16 @@
...
+﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <packageSources>
     <clear />
     <add key="NuGet" value="https://api.nuget.org/v3/index.json" />
     <add key="dotnet-core" value="https://dotnet.myget.org/F/dotnet-core/api/v3/index.json" />
-    <add key="cli-deps" value="https://dotnet.myget.org/F/cli-deps/api/v3/index.json" />
+    <add key="cli-deps" value="https://dotnet.myget.org/F/cli-deps/api/v3/index.json" />    
   </packageSources>
+  <disabledPackageSources>
```
david-driscoll: These probably shouldn't be committed (looks like custom sources you have locally)

src/OmniSharp.Script/CachingScriptMetadataResolver.cs
david-driscoll: Any other feedback to @filipw's question above?

src/OmniSharp.Script/ScriptProjectSystem.cs
```diff
@@ -120,47 +143,37 @@ public void Initalize(IConfiguration configuration)
...
+            {
 
-                    inheritedCompileLibraries.AddRange(DependencyContext.Default.CompileLibraries.Where(x =>
-                            x.Name.ToLowerInvariant().StartsWith("system.runtime")));
-                }
+                inheritedCompileLibraries.AddRange(DependencyContext.Default.CompileLibraries.Where(x =>
+                        x.Name.ToLowerInvariant().StartsWith("system.runtime")));
             }
-
+
```
david-driscoll: Minor nit, there is extra whitespace here, that isn't needed.

david-driscoll:
Also looks like the branch needs to be updated with the latest changes.

seesharper:
Got it. I'll bring it up date and fix the issues you mentioned 👍

seesharper:
@david-driscoll @DustinCampbell PR is now up to date and hopefully most of the requested changes has been resolved

seesharper:
I have spent many hours on this PR and I think that I would be entitled to a response that either accepts or rejects this PR. Don't leave me hanging, guys

filipw:
@seesharper sorry for the delay! 😇

Overall I think this is a very valuable addition, and because:
 - it's implemented as a resolver which is pluggable into any script host 👍
 - syntax-wise it aligns with future CSI functionality

there should be no issue having it in OmniSharp. 
In fact, especially in the CSX area, OmniSharp has been really spearheading how the language tooling should behave, so I think it would be great to have this too.

That said, I have the following concerns:
 - it brings in a dependency on a 3rd party package. In the past we have had some unfortunate experience with depending on scriptcs binaries for script preprocessing so I am a bit worried what happens if we have issues on this functionality? Then we can't even fix it straight up, we will have to wait for you to release a new package
 - a feature like this should be possible to be toggled from the `omnisharp.json` similarly to how you can switch on/off the package restore here https://github.com/OmniSharp/omnisharp-roslyn/blob/cdd99aef194660835e61c58f19fb51ab3dabc52e/src/OmniSharp/config.json#L5
 - how are the errors reported? is everything sent into the OmniSharp log? what happens if `nuget` is not in my PATH (for example, at the moment it's not there on neither my OS X or my Windows). Shouldn't we ship `nuget.exe` with OmniSharp to facilitate this? or at least have an option to provide a path to it

seesharper:
Thanks for your response, @filipw 👍 

I do share your concerns and I would be happy to hand over the code currently resided in DotNet.Script.NuGEtMetadataReferenceResolver.  The whole reason I made it a separate package was that script runners can use the exact same code when actually executing scripts.  Question is where this project should live? We could put it here. No problem for me. 

If an error  occurs, the exception is logged and we fall back to the the way it used to work. 
There is also logging going on inside the library, but because of the "OmniSharp" filter in the logger, this does not make it out to the console. 

Shipping NuGet.Exe is probably not a bad idea since we then also can have a certain control of the actual NuGet version. Alternatively we could just document that we need Nuget >= "some version" and that it has to be in your path. 

Feature toggling sounds great. Then we can ship this as an opt-in feature.

Now that I think about it, for version "1", we create a feature toggle and document minimum NuGet version and the "in you path" requirement

tmat:
@seesharper I think we should built DotNet.Script.NuGetMetadataReferenceResolver into Roslyn at some point. Not sure when that would be at this moment, but I think that's the right place for this functionality long term.

seesharper:
@tmat I agree. Would that be a separate PR then? I just want to get the wheels moving on this one.

tmat:
@seesharper Yes, that would be completely separate PR. We will need to do some design work in Roslyn in order to prepare for this change.

david-driscoll:
@filipw @DustinCampbell thoughts?  Should we perhaps use a source copy of `DotNet.Script.NuGEtMetadataReferenceResolver` to allow us to control it, or go with the package approach for now?

filipw:
I think let's do the following - this feature should be updated to be opt in (as discussed earlier today), and can stay as external dependency for now. 
if there are issues and we need flexibility in addressing them, we will pull it in (it's just a couple of classes)

seesharper:
What if we did this?
Put the code from DotNet.Script.MetadataReferenceResolver into a new project in the roslyn solution.
We can name this OmniSharp.Script.Nuget and publish a package from this project.
In this package we can also ship nuget.exe so that we can control the nuget version used,

tmat:
I don't think using nuget.exe is a good idea. Do we need nuget.exe? Are there not nuget libraries providing APIs we can use?

NuGet.Config
```diff
@@ -1,9 +1,9 @@
...
-    <add key="cli-deps" value="https://dotnet.myget.org/F/cli-deps/api/v3/index.json" />
-  </packageSources>
-</configuration>
+    <packageSources>
+        <clear />
+        <add key="NuGet" value="https://api.nuget.org/v3/index.json" />
+        <add key="dotnet-core" value="https://dotnet.myget.org/F/dotnet-core/api/v3/index.json" />
+        <add key="cli-deps" value="https://dotnet.myget.org/F/cli-deps/api/v3/index.json" />
+    </packageSources>
+</configuration>
```
DustinCampbell: This contains just unnecessary whitespace changes that you probably didn't mean to include.

src/OmniSharp.Script/ScriptHelper.cs
```diff
@@ -8,6 +8,8 @@
 
 namespace OmniSharp.Script
 {
+    using Dotnet.Script.NuGetMetadataResolver;
```
DustinCampbell: Please move using to be in the same block as the rest of the usings. Note that we don't nest them in namespaces in the OmniSharp source.

src/OmniSharp.Script/ScriptProjectSystem.cs
```diff
@@ -16,6 +16,8 @@
 
 namespace OmniSharp.Script
 {
+    using Dotnet.Script.NuGetMetadataResolver;
```
DustinCampbell: Please move using to be in the same block as the rest of the usings. Note that we don't nest them in namespaces in the OmniSharp source.

src/OmniSharp.Script/ScriptProjectSystem.cs
```diff
@@ -60,28 +65,33 @@ public void Initalize(IConfiguration configuration)
...
+            {
+                runtimeContexts = TryCreateRuntimeContextsFromScriptFiles();
+            }
+
             // if we have no context, then we also have no dependencies
             // we can assume desktop framework
             // and add mscorlib
             if (runtimeContexts == null || runtimeContexts.Any() == false)
-            {
+            {
```
DustinCampbell: Unnecessary whitespace

src/OmniSharp.Script/ScriptProjectSystem.cs
```diff
@@ -29,6 +31,8 @@ public class ScriptProjectSystem : IProjectSystem
         private readonly OmniSharpWorkspace _workspace;
         private readonly IOmniSharpEnvironment _env;
         private readonly ILogger _logger;
+        private IScriptProjectProvider _scriptProjectProvider;
```
DustinCampbell: This can be readonly

src/OmniSharp.Script/ScriptProjectSystem.cs
```diff
@@ -201,5 +226,13 @@ Task<object> IProjectSystem.GetWorkspaceModelAsync(WorkspaceInformationRequest r
...
             return Task.FromResult<object>(new ScriptContextModelCollection(scriptContextModels));
         }
+
+        private static string ResolveTargetFramework()
+        {
+            return Assembly.GetEntryAssembly().GetCustomAttributes()
+                .OfType<System.Runtime.Versioning.TargetFrameworkAttribute>()
+                .Select(x => x.FrameworkName)
+                .FirstOrDefault();
+        }
```
DustinCampbell: Note that, in VS Code, this will always be net46.

DustinCampbell:
> I don't think using nuget.exe is a good idea. Do we need nuget.exe? Are there no nuget libraries providing APIs we can use?

We shouldn't do this. How would it run cross-platform if Mono isn't present?

There are definitely NuGet packages we can use.

src/OmniSharp.Script/ScriptProjectSystem.cs
```diff
@@ -29,6 +31,8 @@ public class ScriptProjectSystem : IProjectSystem
         private readonly OmniSharpWorkspace _workspace;
         private readonly IOmniSharpEnvironment _env;
         private readonly ILogger _logger;
+        private IScriptProjectProvider _scriptProjectProvider;
+        private static readonly Lazy<string> _targetFrameWork = new Lazy<string>(ResolveTargetFramework);
```
david-driscoll: Rename to `_targetFramework`

seesharper:
The reason for depending on NuGet.exe is that the resolver was intended to be a NetStandard library so that it also can be used from .Net Core script runners. The https://www.nuget.org/packages/NuGet.PackageManagement package is only available for desktop and the only real option for creating the project.lock.json file we need for resolving metadata references is to feed the generated project.json to NuGet.exe. NuGet.exe still understands project.json as opposed to the DotNet cli that requires an old preview sdk to handle project.json. 

"How would it run cross-platform if Mono isn't present?"
Will Omnisharp even run on *nix platform without Mono? I know there is an option to run OmniSharp as a .Net Core app, but I got the impression that this is something OmniSharp will move away from.

DustinCampbell:
> I know there is an option to run OmniSharp as a .Net Core app, but I got the impression that this is something OmniSharp will move away from.

Yes, that's the true. However, it's still a little weird that we would use nuget.exe. If you don't want to use the NuGet API, why not just shell out to "dotnet nuget" and require that users have the .NET CLI installed? That way, they could get a different version simply by have a global.json in their directory try. Note that there's a DotNetCliService in OmniSharp that this functionality could be added to.

mholo65:
:+1: on generic NuGet support via `DotNetCliService`. It might be useful once Cake support is added in order to dowload addins/tools cia NuGet...

seesharper:
@DustinCampbell 
> Yes, that's the true. However, it's still a little weird that we would use nuget.exe. If you don't want to use the NuGet API, why not just shell out to "dotnet nuget" and require that users have the .NET CLI installed? That way, they could get a different version simply by have a global.json in their directory try. Note that there's a DotNetCliService in OmniSharp that this functionality could be added to.

It is not that I don't want to use the NuGet API, it is just that it is not possible to create a resolver that will work with .Net Core script runners since we cannot make it a NetStandard library. Relying on "dotnet restore" is IMHO a more fragile solution as the user would need another file (global.json) to make this work unless the current sdk is the latest preview that actually supported project.json. NuGet.CommandLine is a single package containing a single file (NuGet.Exe) and is from what I can understand more maintainable than requiring an old sdk preview . We can ship this thing with whatever version of NuGet.CommandLine we desire and have full control of the version. It will require Mono on *nix platforms, but that is already a requirement to make OmniSharp work? Or at least it will be in the future if OmniSharp is moving away from a .Net Core host

DustinCampbell:
> since we cannot make it a NetStandard library

I'm not understanding this bit. Could you explain?

> Relying on "dotnet restore" is IMHO a more fragile solution as the user would need another file (global.json) to make this work unless the current sdk is the latest preview that actually supported project.json

Ah yes, I'd forgotten that we're still using project.json for our scripting solution. @filipw, what's the lifetime of that? What would it take to move on?

seesharper:
This package https://www.nuget.org/packages/NuGet.PackageManagement which we would need to restore packages is a desktop only library and quite frankly I think the code would be more complex as the NuGet API is rather fine grained and the chance for doing something wrong is certainly possible. NuGet 4.0 (the one that ships with vs 2017) continues support for project.json and they have even improved the performance for project.json based projects. 
> "nuget update for UWP, WPF and other project.json based projects is now much faster. In our sample solution with 20 UWP projects, updating packages is 6 times faster than before. For packages.config based projects, nuget update is now about 20% faster."

Even if they decide to discontinue support for project.json some time in the future we can still ship a stable NuGet.CommandLine that support project.json. 

What about csproj. Could we do csproj instead of project.json?

The problem here is that there is no easy way to get a list of metadata references like we do with project.json based projects.

DustinCampbell:
I'm not sure I get why NuGet.PackageManagement being desktop-only is a problem. Note that we use the NuGet API  already within OmniSharp and we'll be using it more.

DustinCampbell:
Also, I'm not sure we *want* to continue supporting project.json forever. I thought that was just a stop-gap until we had something better. @filipw?

seesharper:
It is not a problem if you only look at it in the perspective of OmniSharp. But on the execution side (script runners), I wanted this to be available as a NetStandard library. DotNet.Script for instance is a .net core script runner that could take advantage of this.

DustinCampbell:
Could a .NET Script runner work if that script is .NET Framework? The changes here will always assume .NET Framework at design-time in VS Code. I get the sense that were mixing a bunch of runtime semantics here that might ultimately cause a ton of problems.

seesharper:
Absolutely, .net core is not a requirement, it's an opt-in option

tmat:
If NuGet APIs are not available as a netstandard library we should follow up with NuGet team to fix that - would you mind filing an issue on https://github.com/NuGet/NuGet.Client to raise awareness about the issue?
BTW, not having good NuGet APIs was the main reason why we didn't ship nuget support with the first version of scripting. Since then our team haven't had the necessary resources to follow up on it. My hope is we can get back on it soon(ish) and work with NuGet team on APIs that we need for scripting.

seesharper:
@tmat Maybe we can get away with using https://www.nuget.org/packages/NuGet.Commands/4.0.0 which is NetStandard1.3, but the problem is that the NuGet API documentation is very scarce and I believe that trying to figure out how to put the various pieces together just to do a simple restore, is way more error prone than simply shell out to NuGet.exe which has all the batteries included. At least we should start there and move into using the API directly when proper documentation becomes available.

> The NuGet v3 Client and NuGetGallery APIs are constantly evolving and we are working on having a stable surface area that we can document soon.

I really just want the get the ball rolling on this one and create an awesome scripting experience. 
If we opt-in, we should not be stepping on anybody's toes either :)

filipw:
> Also, I'm not sure we want to continue supporting project.json forever. I thought that was just a stop-gap until we had something better. @filipw?

Let me explain the background/context a little bit.

OmniSharp supports two scripting modes: desktop-CLR CSX files and .NET Core CSX files. The 1st one is the primary use case and affects the vast majority of users.

In the desktop-CLR mode, CSI scripting and all of its features are supported straight up, as that's the default Roslyn scripting. This is the key functionality we want to have in OmniSharp. However, CSI is desktop CLR only. In fact it's not just CSI - the same applied to other alternative script runners like scriptcs and others, which are also desktop only. In other words, no one offered .NET Core-based C# scripting.

As a measure to fill that gap, I created a https://github.com/filipw/dotnet-script which plugs into .NET CLI and supports 1:1 the syntax of standard-CSI based scripting (no scriptcs or Cake-like customizations), and let's the users use .NET Core runtime for executing C# scripts. The extra thing - beyond the CSI standards - that this project does, it allows you to supply dependencies via `project.json` (similarly to the scripting experience offered by Azure Functions). 
Since this is the only command line runner out there that supports .NET Core CSX, we also support it in OmniSharp at the moment, because syntax-wise it matches perfectly with CSI, it's just the runtime that's different. (Note that I will happily retire the whole project once https://github.com/dotnet/roslyn/issues/17666 is implemented and CSI can be used for .NET Core scripts). Ultimately, this is why we have the `project.json` bits in OmniSharp for CSX - they have nothing to do with CSI (which is our primary use case) but they just support your ability to do something with .NET Core scripting via the only .NET Core script runner out there. The fact that this solution uses `project.json` is not so much for "legacy" reasons, but rather because no one has ever defined how an external dependency specification should look like for CSX files. And since Azure Functions is the only product that offered such a feature, and it already exposed dependencies as `project.json` (and it still does - never moved to `csproj`), the .NET Core runner simply mimicked that.

Now to the point of @seesharper solution discussed here - it's implemented as a `MetadataReferenceResolver` which is the proper Roslyn extensibility point for this. It means it can be plugged in into a script runner (like CSI, or the aforementioned dotnet-script or Azure Functions or anything else), but it can also plugged into a tooling provider like OmniSharp, and guarantee uniform experience for dependency resolution both at runtime and design time. As such, I personally don't think it shouldn't take dependencies on anything from inside OmniSharp (like `DotNetCliService`) because that would break the symmetry between the tooling and the runtime.

While this feature will ultimately come into CSI, at the moment, the only realistic use case for the whole functionality in the immediate future is the .NET Core runner where I could just plug it in and let people experience something they will in the future in CSI. Because of that, it's a very much a niche scenario and personally - since this is a very much beta feature anyway, and only affecting users on .NET Core - I am fine with tying it to `nuget.exe` for now.

seesharper:
How would I go about creating a feature toggle for this?
Does it go into config.json?
How do I access this config inside the ScriptProjectSystem class?
What should be the name of the setting? EnableScriptNuGetReferences (default false)

filipw:
OmniSharp already supports project system specific configuration, as long as it was added under the key matching the project system key (in this case `Script`). `config.json` is the "parent" of `omnisharp.json` and all the defaults go there, so yeah the `false` default should go in there and then users can override it via their `omnisharp.json`

It's passed in here https://github.com/OmniSharp/omnisharp-roslyn/blob/dev/src/OmniSharp.Script/ScriptProjectSystem.cs#L47 and can be read from there. See corresponding examples in MsBuild (https://github.com/OmniSharp/omnisharp-roslyn/blob/v1.17.0/src/OmniSharp.MSBuild/MSBuildProjectSystem.cs#L75) and .NET CLI (https://github.com/OmniSharp/omnisharp-roslyn/blob/v1.17.0/src/OmniSharp.DotNet/DotNetProjectSystem.cs#L105-L121).

seesharper:
@filipw @DustinCampbell @david-driscoll 
Could you please review the latest changes. 
I have added an option in config.json that makes all this opt-in.

The requested changes were a little outdated, but I think I've addressed most of them.
```
{
    "dotnet": {
        "alias": "default",
        "projects": "**/project.json",
        "enablePackageRestore": true,
        "packageRestoreTimeout": "180"
    },
    "formattingOptions": {
        "newLine": "\n",
        "useTabs": false,
        "tabSize": 4,
        "indentationSize": 4
    },
    "script": {
        "enableScriptNuGetReferences" : false
    }
}

```

seesharper:
FYI: Just tested this on a Mac and it works equally well there.

filipw:
@seesharper thanks - looks got to me. There are still some small things @DustinCampbell pointed out but after that, as far as I'm concerned, we could take this in as an experimental scripting feature 👍

@DustinCampbell @david-driscoll what do you guys think?

filipw:
actually one more thing - looks like the configuration is not being consumed at the moment. Did you forget a push?

seesharper:
@filipw @DustinCampbell @david-driscoll 

The configuration is being consumed here 

https://github.com/OmniSharp/omnisharp-roslyn/pull/813/files#diff-ed5a3c116a19423c18b1e630f80455b9R77

I have also updated to the latest version of Dotnet.Script.NuGetMetadataResolver (2.0.3) and this PR should be able to be merged.

filipw:
I meant that `NuGetMetadataReferenceResolver` shouldn't be used at all if the setting is `false`

seesharper:
Ah, I see. The ScriptHelper class is a static class. Is there any way that I can access the configuration from within the ScriptHelper class?

seesharper:
@filipw Probably need to make the ScriptHelper class an instance class in order to pass in "IConfiguration" ?

filipw:
that's OK - it could be a singleton similar to i.e. https://github.com/OmniSharp/omnisharp-roslyn/blob/dev/src/OmniSharp.Roslyn/MetadataHelper.cs#L32 + https://github.com/OmniSharp/omnisharp-roslyn/blob/89fa23cc1d3572795dceb612be7b1784630d7a8a/src/OmniSharp.Host/Startup.cs#L97

seesharper:
@filipw Now the NuGetMetadataResolver is only used when the "enableScriptNuGetReferences" option is "true"

seesharper:
@filipw @DustinCampbell @david-driscoll 
Would appreciate you taking the time to review the (hopefully) final changes.

seesharper:
@filipw @DustinCampbell @david-driscoll 
Is this happening, guys? It's getting a little tedious to constantly have to keep this branch up-to-date.

src/OmniSharp.Script/CachingScriptMetadataResolver.cs
DustinCampbell: @seesharper: did this ever get answered?

seesharper:
Thanks @DustinCampbell . Build passing. @david-driscoll , @filipw . Are we merging this now?

filipw:
The merge button is unavailable until @david-driscoll approves

seesharper:
@david-driscoll Would really appreciate your final approval on this PR

david-driscoll:
Sorry missed the final approvals here... LGTM

seesharper:
Thanks a lot, @david-driscoll. Appreciate it 👍

adamralph:
This is great! Is there an issue anywhere tracking the addition of this to csi.exe?

Issue #795, opened 2017-03-10
Lots of errors when opening aspnet/SignalR


davidfowl:
I'm running on OSX.

I didn't want to open an uber issue but it seems like there are multiple things happening here. The repro steps are to clone https://github.com/aspnet/SignalR and run build.sh. After the build finishes, open vs code at the root (`code .`).

1. I get warnings telling me there are unresolved dependencies even though restore ran.
2. I get 100s of errors talking about ambiguous references. It seems as though the sample projects don't see the src projects within the same solution so packages are being resolved...

![image](https://cloud.githubusercontent.com/assets/95136/23789882/971b6578-0531-11e7-9321-cc9a813570d0.png)

I'll try to pick it apart into specific issues but I was caught by surprise because I thought things would just work.

davidfowl:
@DustinCampbell am assuming nobody is looking into this issue. There's a fundamental problem with project references in msbuild within omnisharp. If nobody gets to this issue I'll try to take a look this weekend. I can't use VS code anymore on any project in ASP.NET because of it.

DustinCampbell:
Sorry @davidfowl. I forgot to post back here after trying to repro a few weeks ago. Interestingly, I can repro if I run `./build.sh` but I can't repro if I just `dotnet restore SignalR.sln` with .NET Core SDK 1.0.1. I do get several messages about unresolved dependencies which I dismiss with <kbd>ESC</kbd>. However, I don't get any errors.

Here's what I did in the Terminal:

```
git clone https://github.com/aspnet/SignalR.git
cd SignalR/
dotnet restore SignalR.sln
code .
```

No errors anywhere. 😞 

![image](https://cloud.githubusercontent.com/assets/116161/24745387/6da8f608-1a6a-11e7-9c13-b961dbb882ab.png)

Hovering over the methods where your screenshot shows errors or pressing <kbd>F12</kbd> on them shows that they're indeed binding properly.

I _suspect_ this has to do with the fact that the SignalR is using .NET Core SDK 2.0.0-preview1-005418, which is beyond RTM. Today OmniSharp ships RTM versions of the various MSBuild SDKs and points to those at designtime. So, at designtime, an older version of, say, Microsoft.Net.Web.Sdk is being used than what `./build.sh` just ran against. That's just a guess though. I haven't dug in deeply enough to know for sure.

DustinCampbell:
[...]

PR #814, opened 2017-04-06
Handle package reference version ranges in .csproj files

DustinCampbell:
When I did the initial punch through to get package references working in .csproj, I failed to account for version ranges. This should address the unresolved dependencies warnings reported in #795.

### DIFF
--- a/src/OmniSharp.MSBuild/ProjectFile/PackageReference.cs
+++ b/src/OmniSharp.MSBuild/ProjectFile/PackageReference.cs
@@ -5,12 +5,12 @@ namespace OmniSharp.MSBuild.ProjectFile
 {
     public class PackageReference : IEquatable<PackageReference>
     {
-        public PackageDependency Identity { get; }
+        public PackageDependency Dependency { get; }
         public bool IsImplicitlyDefined { get; }
 
         public PackageReference(PackageDependency dependency, bool isImplicitlyDefined)
         {
-            this.Identity = dependency;
+            this.Dependency = dependency;
             this.IsImplicitlyDefined = isImplicitlyDefined;
         }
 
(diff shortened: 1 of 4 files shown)


src/OmniSharp.MSBuild/ProjectFile/PackageReference.cs
```diff
@@ -5,21 +5,20 @@ namespace OmniSharp.MSBuild.ProjectFile
 {
     public class PackageReference : IEquatable<PackageReference>
     {
-        public PackageIdentity Identity { get; }
+        public PackageDependency Identity { get; }
```
filipw: maybe you'd want to go from `public PackageIdentity Identity { get; }` to `public PackageDependency Dependency { get; }` instead?

DustinCampbell: Yes -- thanks!

src/OmniSharp.MSBuild/ProjectFile/PropertyConverter.cs
```diff
@@ -165,10 +165,9 @@ public static OutputKind ToOutputKind(string propertyValue)
             }
         }
 
-        public static NuGetVersion ToNuGetVersion(string propertyValue)
+        public static VersionRange ToVersionRange(string propertyValue)
         {
-            NuGetVersion version;
-            if (NuGetVersion.TryParse(propertyValue.Trim(), out version))
+            if (VersionRange.TryParse(propertyValue.Trim(), out var version))
```
filipw: maybe I'm weird but doesn't it strike you as a bit inconsistent that in the same namespace they'd have two different naming styles - `NuGetVersion` and `VersionRange` (used for, well, `NuGetVersions`)?

DustinCampbell: There's a bunch of them: `NuGetVersion`, `SemanticVersion`, `VersionRange`, `FloatRange`. They're all in NuGet.Versioning. `NuGetVersion` is a specific version, used for package identity. A range wouldn't be suitable for that.

PR #815, opened 2017-04-06
Ensure that MSBuildProjectSystem doesn't add project references twice

### LINKED ISSUE #795: Lots of errors when opening aspnet/SignalR

DustinCampbell:
Fixes #795

It's possible for project references to appear twice within an MSBuild project after execution:

* As a ProjectReference item
* As a ReferencePath item with ReferenceSourceTarget="ProjectReference"

This change makes sure that ReferencePath items that were sourced from project references aren't added. We just use the project references, which is a better experience at designtime when the project is built.

Note: There was code that used to do this, but it was unintentionally removed in a refactoring a few months ago (by me -- oops!): https://github.com/OmniSharp/omnisharp-roslyn/commit/ef4b3838d39666ed5ca9780ea133bfdc08ff5844.

### DIFF
--- a/src/OmniSharp.MSBuild/ProjectFile/ProjectFileInfo.cs
+++ b/src/OmniSharp.MSBuild/ProjectFile/ProjectFileInfo.cs
@@ -218,7 +218,8 @@ namespace OmniSharp.MSBuild.ProjectFile
             var projectAssetsFile = projectInstance.GetPropertyValue(PropertyNames.ProjectAssetsFile);
 
             var sourceFiles = GetFullPaths(projectInstance.GetItems(ItemNames.Compile));
-            var references = GetFullPaths(projectInstance.GetItems(ItemNames.ReferencePath));
+            var references = GetFullPaths(
+                projectInstance.GetItems(ItemNames.ReferencePath).Where(ReferenceSourceTargetIsNotProjectReference));
             var projectReferences = GetFullPaths(projectInstance.GetItems(ItemNames.ProjectReference));
             var analyzers = GetFullPaths(projectInstance.GetItems(ItemNames.Analyzer));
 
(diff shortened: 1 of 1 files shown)


filipw:
have you thought of publishing all this MSBuild-related code as a nuget package? 

For example I would like to use something like that in other projects - being able to go from `csproj` to everything needed to create a `CSharpCompilation`/describing the project (some kind of a replacement for https://www.nuget.org/packages/Microsoft.DotNet.ProjectModel)

DustinCampbell:
> have you thought of publishing all this MSBuild-related code as a nuget package?

Probably not. We already have that in Roslyn with the [MSBuildProjectLoader](https://github.com/dotnet/roslyn/blob/master/src/Workspaces/Core/Desktop/Workspace/MSBuild/MSBuildProjectLoader.cs). It just doesn't work with the latest MSBuild and isn't portable. However, it's looking likely that I'll be called upon to fix that.

filipw:
oh that's what I've been missing. Is it part of any of the Roslyn nuget packages?

DustinCampbell:
Sure, but it's only available on desktop: https://www.nuget.org/packages/Microsoft.CodeAnalysis.Workspaces.Common/.

The API sits next to MSBuildWorkspace.

filipw:
thanks!

PR #816, opened 2017-04-06
More improvements for 'dotnet test' support

DustinCampbell:
* Clean up endpoints to provide more data
* Fire TestMessage event with messages from the test runner
* Get debugging working for 'dotnet vstest'.

The debugging requires a bit of explanation. In order to debug, we query the VS Test protocol for the command-line args to launch a test host. We then start the test host and return the process ID back to the editor host. Once the editor host has attached a debugger, it sends another request to OmniSharp letting it know that it's ready to start. OmniSharp then sends a message to 'dotnet test' to confirm that the debugger is attached. This is a bit of a dance and I'm still looking at ways to improve it.


src/OmniSharp.DotNetTest/Legacy/LegacyTestManager.cs
```diff
@@ -63,22 +71,42 @@ public override RunDotNetTestResponse RunTest(string methodName, string testFram
...
+                RedirectStandardError = true
             };
 
             var testProcess = Process.Start(startInfo);
 
-            var results = new List<TestResult>();
+            var output = new StringBuilder();
+            var error = new StringBuilder();
+
+            testProcess.OutputDataReceived += (_, e) =>
```
david-driscoll: Do we know if this can return a partial line? (or line and a half?)

I need to go looking at the .net process model again... :face_with_head_bandage:

DustinCampbell: It's a line. At least, that's what I've always observed and what the [docs](https://msdn.microsoft.com/en-us/library/system.diagnostics.process.outputdatareceived(v=vs.110).aspx) say:

> The event is enabled during asynchronous read operations on StandardOutput. To start asynchronous read operations, you must redirect the StandardOutput stream of a Process, add your event handler to the OutputDataReceived event, and call BeginOutputReadLine. Thereafter, **the OutputDataReceived event signals each time the process writes a line to the redirected StandardOutput stream**, until the process exits or calls CancelOutputRead.

david-driscoll: 👍 

I hadn't jumped to the docs yet.  I just remember my experience with node that it wasn't always the case (of course we're talking two completely different tech stacks 😄)

PR #818, opened 2017-04-07
Clean up MSBuild environment initialization and allow it to use Visual Studio 2017 if it's present

DustinCampbell:
Fixes https://github.com/OmniSharp/omnisharp-vscode/issues/1368

This change ensures that OmniSharp will use the MSBuild tools installed with Visual Studio 2017 if they're present on the machine. This allows OmniSharp to properly handle VS 2017 projects where the targets/tasks aren't include with OmniSharp's local MSBuild, such as WebApplication projects as reported in the bug listed above. If VS 2017 is not on the machine, OmniSharp will continue to use its local MSBuild which has a fallback to the Microsoft Build Tools for targets/tasks that it can't find.

I've also taken the opportunity to clean up a lot of the MSBuild environment initialized code and project file processing.

### DIFF
--- a/src/OmniSharp.MSBuild/MSBuildEnvironment.cs
+++ b/src/OmniSharp.MSBuild/MSBuildEnvironment.cs
@@ -11,18 +11,30 @@ namespace OmniSharp.MSBuild
         public const string MSBuildSDKsPathName = "MSBuildSDKsPath";
 
         private static bool s_isInitialized;
-        private static string s_msbuildFolder;
+
+        private static bool s_usingVisualStudio;
+
+        private static string s_msbuildExePath;
         private static string s_msbuildExtensionsPath;
         private static string s_msbuildSDKsPath;
 
         public static bool IsInitialized => s_isInitialized;
 
-        public static string MSBuildFolder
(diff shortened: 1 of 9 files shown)


DustinCampbell:
Thanks!

src/OmniSharp.MSBuild/MSBuildEnvironment.cs
```diff
@@ -39,82 +51,110 @@ public static string MSBuildSDKsPath
...
             {
-                throw new InvalidOperationException("MSBuildEnvironment is already initialized");
+                throw new InvalidOperationException("MSBuild environment is already initialized.");
             }
 
-            // If OmniSharp is running normally, MSBuild is located in an 'msbuild' folder beneath OmniSharp.exe.
-            var msbuildFolder = Path.Combine(AppContext.BaseDirectory, "msbuild");
+            // If MSBuild can locate VS 2017 and set up a build environment, we don't need to do anything.
+            // MSBuild will take care of itself.
+            if (MSBuildHelpers.TryGetVisualStudioBuildEnvironment())
```
filipw: nice

PR #821, opened 2017-04-10
Clean up vstest protocol again and simplify debug session

DustinCampbell:
I'm continuing here to work on VS Test support for C# for VS Code. This change largely removes the need to launch a test process for the debugger to attach to. Instead, it just returns the process start info for the debugger to launch. This still requires C# for VS Code to notify OmniSharp (via the `/v2/debugtest/run` end point) to start a test run once the debugger is connected.


filipw:
> This change largely removes the need to launch a test process for the debugger to attach to. Instead, it just returns the process start info for the debugger to launch.

very interesting!

PR #824, opened 2017-04-17
Update 'dotnet test' support

DustinCampbell:
This change improves the 'dotnet test' support:

* Rename `/v2/debugtest/run` endpoint to `/v2/debugtest/launch`
* Add a `TargetProcessId` to the `/v2/debugtest/launch` endpoint to allow OmniSharp to subscribe to be notified with the launched process exists.
* Made reading from the underlying socket asynchronous for the debugging endpoints (since the debugger interaction is asynchronous). This change includes threading through cancellation.
* Made everything but the models internal. These classes are not intended to be public API surface area.


filipw:
> Made everything but the models internal. These classes are not intended to be public API surface area.

👍 

I think now that the plugin effort back on the table, this is really important conversation to have. We could start by making most things internal to begin with, and only go with a handful of the most obvious extensibility points like `IProjectSystem` and the models, and then include more things (like events) once someone actually needs them.

src/OmniSharp.DotNetTest/VSTestManager.cs
```diff
@@ -157,7 +122,11 @@ public override void DebugRun()
 
             while (!done)
             {
-                var message = ReadMessage();
+                var (success, message) = await TryReadMessageAsync(cancellationToken);
```
filipw: nice!

PR #827, opened 2017-04-18
In MSBuildProjectSystem, add NoWarn values that do not parse as ushorts

### LINKED ISSUE #826: MSBuild NoWarn doesn't work for error codes with "CS" prefix
Currently, the MSBuildProjectSystem doesn't include values from NoWarn that don't parse as numbers. This incorrect, it should prefix the value with "CS" and it if its a number. If it's not a number, it should just add it.

DustinCampbell:
Fixes #826


DustinCampbell:
Thanks!

PR #829, opened 2017-04-19
Fixed GoToMetadata in CSX

### LINKED ISSUE #755: GoToMetadata doesn't work in CSX files
This is caused due to this check https://github.com/dotnet/roslyn/blob/9502c50d4a09457fae78f2f621484c59501993e3/src/Compilers/CSharp/Portable/Compilation/CSharpCompilation.cs#L1193 in our reflection call into `AbstractMetadataAsSourceService`. In order to use metadata we need to add a document with a namespace and things break down there.

The solution could be to not flag our CSX projects as submissions however it would break the scripting implementation in many places.

filipw:
Fixes #755 

At the moment Go To Metadata doesn't work in CSX files. The reason is, in the current implementation of metadata handling on the server, a metadata-document was added to current project. This is not allowed in `submission` type projects (which are used for CSX files), as they can only have one root syntax tree, cannot have namespaces etc. - and an ugly unhandled exception was thrown instead. To fix this, we actually create a temporary project (which is thrown away) for CSX metadata navigation.

The fix doesn't affect the original code path so there is no risk merging this. However similar implementation (a temporary project) could also be considered there, because at the moment we flood the current project with new documents every time metadata is accessed. I left it intact because perhaps there was some reason to implement it like that?

I also added GoToMetadata tests for CSX to validate this fix.


src/OmniSharp.Roslyn/MetadataHelper.cs
```diff
@@ -58,8 +58,14 @@ public Task<Document> GetDocumentFromMetadata(Project project, ISymbol symbol, C
         {
             var filePath = GetFilePathForSymbol(project, symbol);
             var topLevelSymbol = GetTopLevelContainingNamedType(symbol);
-            var temporaryDocument = project.AddDocument(filePath, string.Empty);
 
+            // since submission projects cannot have new documents added to it
+            // we will use temporary project to hold metadata documents
+            var metadataProject = project.IsSubmission ? project.Solution.AddProject("metadataTemp", "metadataTemp.dll", LanguageNames.CSharp).
+                WithCompilationOptions(project.CompilationOptions).
+                WithMetadataReferences(project.MetadataReferences) : project;
```
DustinCampbell: Tiny, tiny, tiny nit: This might be more readable if formatted like so:

C#
```
var metadataProject = project.IsSubmission
    ? project.Solution.AddProject("metadataTemp", "metadataTemp.dll", LanguageNames.CSharp)
        .WithCompilationOptions(project.CompilationOptions)
        .WithMetadataReferences(project.MetadataReferences)
    : project;
```

tests/TestUtility/OmniSharpTestHost.cs
```diff
@@ -136,7 +136,12 @@ public void AddFilesToWorkspace(params TestFile[] testFiles)
                 this.Workspace,
                 "project.json",
                 new[] { "dnx451", "dnxcore50" },
-                testFiles);
+                testFiles.Where(f => f.FileName.EndsWith(".cs", StringComparison.CurrentCultureIgnoreCase)).ToArray());
```
DustinCampbell: Do we want `CurrentCulture`? Should that be `OrdinalIgnoreCase`

filipw: ah sure, I just picked "that long thing" from intellisense without thinking twice

PR #831, opened 2017-04-25
Enable lots of tests to run for CSX

filipw:
In the last PR I introduced a possibility to run our regular `AbstractTestFixture`-based tests against CSX files. 
In this PR I enabled a bunch of existing tests to run both against `.cs` files and against `.csx` files.

As a result there are 36 more tests.

### DIFF
--- a/tests/TestUtility/TestHelpers.cs
+++ b/tests/TestUtility/TestHelpers.cs
@@ -5,6 +5,8 @@ using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using OmniSharp;
 using OmniSharp.Services;
+using System.IO;
+using System.Reflection;
 
 namespace TestUtility
 {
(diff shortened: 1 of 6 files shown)


DustinCampbell:
looks like honest test failures.

filipw:
you are right, it is. Before submitting, I ran the tests from VS and they all passed, but Test Explorer only runs tests for `net46`, not for the `netcoreapp1.1` - and the 2 test failures here are actually on the .NET Core version only. 

I'll make the necessary tweaks, the root cause is at the moment `System.Linq` namespace is not seen in the .NET Core version of CSX project and that's why it fails.

filipw:
I believe it's all fine now (pending the App Veyor network connectivity issues)

tests/TestUtility/TestHelpers.cs
```diff
@@ -5,6 +5,8 @@
 using Microsoft.CodeAnalysis.CSharp;
 using OmniSharp;
 using OmniSharp.Services;
+using System.IO;
+using System.Reflection;
```
DustinCampbell: Shouldn't these be sorted to the top?

filipw: actually they are not needed at all, the snuck in there due to some code that ended up being removed

PR #836, opened 2017-04-27
Increase the timeout specified when running published bits during build

DustinCampbell:
Our build script finishes up by running the published OmniSharp bits with a timeout. This timeout is no longer long enough and the bits continue trying to run after the build is finished. I *believe* the reason for this is because the MSBuildProjectSystem now loads referenced projects outside of the directory OmniSharp is started with. More projects means more time processing, which causes the bits run longer. I'm increasing the timeout from 10 seconds to 30 seconds.


DustinCampbell:
Yeah, I figured it was extra projects getting loaded because I saw a *lot* more output. Plus, I think we run these in the folder of OmniSharp.csproj, which would reference all of the OmniSharp projects *outside* of the folder.

PR #837, opened 2017-04-27
Add back code to handle project paths within a solultion file correctly

DustinCampbell:
This was causing solution files not to load properly on *Nix.


DustinCampbell:
Going to go ahead and push this one so I can kick off a new release.

PR #838, opened 2017-04-28
When finding partial methods, pick the one with body 

filipw:
At the moment, when we have the following code:

```
public partial class MyClass 
{
     public MyClass()
    {
          Method();
     } 

      partial void Method()
      {
         Console.WriteLine("foo");
      }
}

  public partial class MyClass  
  {
      partial void Method();
  }
```

 - using `ctrl+T` in VS Code, and typing `Method()` will always send the user to the one without body
 - using "go to definition" in VS Code over the method invocation, will also always send the user to the one without body

This PR ensures that we send the caret to the version with body instead.

### DIFF
--- a/tests/OmniSharp.Roslyn.CSharp.Tests/FindSymbolsFacts.cs
+++ b/tests/OmniSharp.Roslyn.CSharp.Tests/FindSymbolsFacts.cs
@@ -3,7 +3,8 @@ using System.Threading.Tasks;
 using OmniSharp.Models;
 using OmniSharp.Models.FindSymbols;
 using OmniSharp.Roslyn.CSharp.Services.Navigation;
-using Microsoft.CodeAnalysis;using TestUtility;
+using Microsoft.CodeAnalysis;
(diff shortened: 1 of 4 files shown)


tests/OmniSharp.Roslyn.CSharp.Tests/FindSymbolsFacts.cs
```diff
@@ -3,7 +3,7 @@
 using OmniSharp.Models;
 using OmniSharp.Models.FindSymbols;
 using OmniSharp.Roslyn.CSharp.Services.Navigation;
-using TestUtility;
+using Microsoft.CodeAnalysis;using TestUtility;
```
DustinCampbell: line break missing here?

filipw: argh, conflict resolution gone sideways 😃

filipw:
thanks 👍

PR #840, opened 2017-04-30
Use CompletionService in IntellisenseService

filipw:
At the moment, the `CompletionService` is only used for keywords, everything else is coming from `Recommender`.

This PR introduces relying on `CompletionService` for all completions. 
The symbols are still processed the same way as before, so none of the existing features are broken, but extra features available via  `CompletionService` also show up in intellisense now (as we process all of its providers). 
This will close a lot of open bugs via a relatively risk-free change.

I will push some tests into this soon but here are some of the examples of new features that are introduced via this PR:

### Override support

Note: because of [this Regex](https://github.com/OmniSharp/omnisharp-vscode/blob/634261f9911c29148347325f2e528bfcfc62999a/src/features/completionItemProvider.ts#L43) in omnisharp-vscode, where brackets are removed, it appears a little broken. This change would need to be done in omnisharp-vscode.


src/OmniSharp.Roslyn.CSharp/Services/Intellisense/CompletionItemExtensions.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        static CompletionItemExtensions()
+        {
+            var symbolCompletionItemType = typeof(CompletionItem).GetTypeInfo().Assembly.GetType("Microsoft.CodeAnalysis.Completion.Providers.SymbolCompletionItem");
+            _getSymbolsAsync = symbolCompletionItemType.GetMethod("GetSymbolsAsync", BindingFlags.Public | BindingFlags.Static);
+        }
+
+        public static async Task<IEnumerable<ISymbol>> GetCompletionSymbols(this CompletionItem completionItem, IEnumerable<ISymbol> recommendedSymbols, Document document)
+        {
+            // for SymbolCompletionProvider, use the logic of extracting information from recommended symbols
```
DustinCampbell: I kind of hate that we pre-compute recommendedSymbols (which is essentially like calling the CompletionService _twice_) just to handle this special case.

src/OmniSharp.Roslyn.CSharp/Services/Intellisense/CompletionItemExtensions.cs
```diff
@@ -0,0 +1,47 @@
...
+            if (completionItem.Properties.ContainsKey("Provider") && completionItem.Properties["Provider"] == "Microsoft.CodeAnalysis.CSharp.Completion.Providers.SymbolCompletionProvider")
```
DustinCampbell: It might be good to throw an exception if the type `Microsoft.CodeAnalysis.CSharp.Completion.Providers.SymbolCompletionProvider` doesn't exist. That way, we aren't searching around for the reason why completion starts failing if this internal type's name changes.

src/OmniSharp.Roslyn.CSharp/Services/Intellisense/CompletionItemExtensions.cs
```diff
@@ -0,0 +1,47 @@
...
+namespace OmniSharp.Roslyn.CSharp.Services.Intellisense
+{
+    internal static class CompletionItemExtensions
+    {
+        private static MethodInfo _getSymbolsAsync;
+
+        static CompletionItemExtensions()
+        {
+            var symbolCompletionItemType = typeof(CompletionItem).GetTypeInfo().Assembly.GetType("Microsoft.CodeAnalysis.Completion.Providers.SymbolCompletionItem");
+            _getSymbolsAsync = symbolCompletionItemType.GetMethod("GetSymbolsAsync", BindingFlags.Public | BindingFlags.Static);
```
DustinCampbell: We've got some extension methods for this sort of thing (used by [`MetadataHelper`](https://github.com/OmniSharp/omnisharp-roslyn/blob/dev/src/OmniSharp.Roslyn/MetadataHelper.cs)) that throw exceptions nicely when they fail. I'd recommend using those to ensure we can easily spot failures if internal API names change.

src/OmniSharp.Roslyn.CSharp/Services/Intellisense/CompletionItemExtensions.cs
filipw: unfortunately since this change https://github.com/dotnet/roslyn/pull/15688, the `SymbolCompletionProvider` doesn't store encoded symbols in the properties of the completion item anymore, so just calling `GetSymbols()` results in an empty array. Instead, the only thing I could think of was that they are fetched using `Recommender`. I am sure you'd know more background behind this change and if there is a better way to do it?

src/OmniSharp.Roslyn.CSharp/Services/Intellisense/CompletionItemExtensions.cs
filipw: yes good point 👍

filipw:
thanks, I agree that in order to *properly* benefit from all the power of `CompletionService` we'll need a brand new endpoint.
What I tried to do here was to improve the current behavior as much as possible, with relatively small change to the current way the server operates, to solve a few annoyances 😃

src/OmniSharp.Roslyn.CSharp/Services/Intellisense/IntellisenseService.cs
```diff
@@ -37,54 +37,60 @@ public IntellisenseService(OmniSharpWorkspace workspace, FormattingOptions forma
...
-                }
 
-                var model = await document.GetSemanticModelAsync();
-                var symbols = await Recommender.GetRecommendedSymbolsAtPositionAsync(model, position, _workspace);
+                            // for other completions, i.e. keywords, create a simple AutoCompleteResponse
+                            // we'll just assume that the completion text is the same
+                            // as the display text.
+                            var response = new AutoCompleteResponse()
+                            {
+                                CompletionText = item.DisplayText,
```
david-driscoll: To support completion properly we need a new api similar to what LSP provides with `additionalTextEdits`?

DustinCampbell: Right, but we'll want to be more of a "resolve" API, so we don't have to go computing this for every completion item.

src/OmniSharp.Roslyn.CSharp/Services/Intellisense/CompletionItemExtensions.cs
DustinCampbell: Ah yes, I'd forgotten about that. Computing all symbol IDs was simply too expensive.

src/OmniSharp.Roslyn.CSharp/Services/Intellisense/IntellisenseService.cs
filipw: actually overrides and partials won't even reach that specific line because they'd branch out [here](https://github.com/OmniSharp/omnisharp-roslyn/pull/840/files#diff-d2949d6205937df4d863d016cb0be1f7R55) since these providers encode symbols. So the formatting/presentation of i.e. symbol information will be consistent with the rest of the experience, but yeah ultimately the `DisplayText` will be used for completion.

From what I experimented, even the simple `DisplayText` at the moment isn't that bad. It actually, in case of overrides, has the following format https://github.com/dotnet/roslyn/blob/614299ff83da9959fa07131c6d0ffbc58873b6ae/src/Features/Core/Portable/Completion/Providers/AbstractOverrideCompletionProvider.ItemGetter.cs#L21. So you get to see the full signature, and that's what can be inserted into the code even using the current object/endpoint model we have (of course things like missing usings would have to be resolved manually separately by the user)

It's definitely sub par experience compared to VS but at the same time maybe it's better than not having anything?
But of course, we could also just disable the complex providers like override/partial from participating in the old intellisense endpoint and only add them once we have a new one

DustinCampbell:
The failures on *Nix look real.

filipw:
> The failures on *Nix look real.

Yes it was actually caused by this https://github.com/dotnet/corefx/issues/15825#issuecomment-277722247 "bug" ("behavior"?) of CoreFX - default US culture-based sort behavior is different on Windows and *nix. It should be fine now.

src/OmniSharp.Roslyn.CSharp/Services/Intellisense/CompletionItemExtensions.cs
```diff
@@ -0,0 +1,48 @@
...
+        }
+
+        public static async Task<IEnumerable<ISymbol>> GetCompletionSymbolsAsync(this CompletionItem completionItem, IEnumerable<ISymbol> recommendedSymbols, Document document)
+        {
+            // for SymbolCompletionProvider, use the logic of extracting information from recommended symbols
+            if (completionItem.Properties.ContainsKey("Provider") && completionItem.Properties["Provider"] == "Microsoft.CodeAnalysis.CSharp.Completion.Providers.SymbolCompletionProvider")
+            {
+                var symbols = recommendedSymbols.Where(x => x.Name == completionItem.Properties["SymbolName"] && (int)x.Kind == int.Parse(completionItem.Properties["SymbolKind"])).Distinct();
+                return symbols ?? Enumerable.Empty<ISymbol>();
```
DustinCampbell: I don't think `Where(...)` or `Distinct(...)` will ever return null.

filipw: you are right 😊

DustinCampbell:
> Note: because of this Regex in omnisharp-vscode, where brackets are removed, it appears a little broken. This change would need to be done in omnisharp-vscode.

FWIW, I've never noticed that regex. It existed before I was on the project, so it's presence is a mystery.

DustinCampbell:
All right. Merging... :smile:

filipw:
awesome! 🐣☀️

lukepuplett:
Hello - sorry to bother but a cluster of issues all point to this. I don't get intellisense for C# object initializers in VSCode and it seems this has been a feature for years, right? I'm fully up to date, I think; I don't know how to check the version of OmniSharp.

My question is: where do I go for help on this?

darkguy2008:
I have to +1 @lukepuplett as it seems I'm experiencing the same thing. Trying to get intellisense for class properties in its initializer is sending me straight to the Process namespace (halfway through the full .NET namespace list) instead of sending me to the top of the list (where the properties are).

I opened a bug report in VSCode here but maybe it's related to this instead? https://github.com/microsoft/vscode/issues/128963

PR #846, opened 2017-05-02
replaced `InteractiveScriptGlobals` with `CommandLineScriptGlobals ` as CSX host object type

filipw:
Replaced `InteractiveScriptGlobals` with `CommandLineScriptGlobals ` as CSX host object type. This is required, because CSI uses `CommandLineScriptGlobals` as host object for CSX script execution. `InteractiveScriptGlobals` is **only** used for REPL mode ([reference](https://github.com/dotnet/roslyn/blob/master/src/Scripting/Core/Hosting/CommandLine/CommandLineRunner.cs#L176-L196)), so naturally it makes sense to offer `CommandLineScriptGlobals` in OmniSharp intellisense. 

As a bonus, I extracted some of the project/compilation set up logic from `ScriptProjectSystem` into `ScriptHelper`. This way, it can be used in the tests as well, to ensure that the test set up is using the same logic as the project system would. This makes tests more reliable than before, where the code to create CSX project/compilation was maintained in tests separately and needed to always stay in sync with project system to be trustworthy.

### DIFF
--- a/tests/OmniSharp.Roslyn.CSharp.Tests/IntellisenseFacts.cs
+++ b/tests/OmniSharp.Roslyn.CSharp.Tests/IntellisenseFacts.cs
@@ -310,7 +310,7 @@ namespace OmniSharp.Roslyn.CSharp.Tests
         public async Task Returns_host_object_members_in_csx()
         {
             const string source =
-                "Pri$$";
+                "Prin$$";
 
             var completions = await FindCompletionsAsync("dummy.csx", source);
             ContainsCompletions(completions.Select(c => c.CompletionText), new[] { "Print", "PrintOptions" }); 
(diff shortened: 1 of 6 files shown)


tests/OmniSharp.Roslyn.CSharp.Tests/IntellisenseFacts.cs
```diff
@@ -202,6 +202,16 @@ public MyClass1()
...
         }
 
+        [Fact]
+        public async Task Returns_host_object_members_in_csx()
+        {
+            const string source =
+                "Prin$$";
+
+            var completions = await FindCompletionsAsync("dummy.csx", source);
+            ContainsCompletions(completions.Select(c => c.CompletionText), new[] { "Print", "PrintOptions" });
```
filipw: note that we are ahead of VS 😃
https://github.com/dotnet/roslyn/issues/13886

DustinCampbell: Yeah, VS Code scripting support is well ahead of VS. You're doing great work! 👍

filipw:
looks like I got into conflicts with myself 😄

tests/OmniSharp.Roslyn.CSharp.Tests/IntellisenseFacts.cs
filipw: thanks! 😄

filipw:
I rebased this and resolved the conflicts - should be good to go

PR #847, opened 2017-05-03
Stop deploying SDKs with OmniSharp

### LINKED ISSUE #765: Don't deploy .NET Core SDKs
Today, we deploy the .NET Core SDKs with OmniSharp to make everything work with MSBuild and .csproj. However, we really shouldn't. Instead, we should set the "MSBuildSDKsPath" environment variable to the SDKs that are deployed with the .NET Core SDK (e.g. the CLI). However, there's a bit of extra work here to discover where the .NET Core SDK is installed and which .NET Core SDK is specified in the user's global.json.

DustinCampbell:
Fixes #765

This change stops deploying the .NET Core MSBuild SDKs with OmniSharp. That was a stop gap for .NET Core 1.0. However, with .NET Core 2.0 around the corner, it's time to stop deploying them and rely on the .NET Core SDK installed on the user's machine.

To locate the correct .NET Core SDK, we simply launch `dotnet --info` in the working directory of the project. That should properly handle the case where the SDK is specified in a global.json file. Then, we process the text returned by `dotnet --info` and locate the "Base Path" value. Combining this with "Sdks" should be the correct value.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -179,59 +179,6 @@ Task("SetupMSBuild")
 
     CopyDirectory(msbuildNetCoreAppInstallFolder, msbuildNetCoreAppFolder);
 
-    var sdks = new []
-    {
-        "Microsoft.NET.Sdk",
-        "Microsoft.NET.Sdk.Publish",
-        "Microsoft.NET.Sdk.Web",
-        "Microsoft.NET.Sdk.Web.ProjectSystem",
-        "NuGet.Build.Tasks.Pack"
-    };
-
-    var net46SdkFolder = CombinePaths(msbuildNet46Folder, "Sdks");
-    var netCoreAppSdkFolder = CombinePaths(msbuildNetCoreAppFolder, "Sdks");
-
-    foreach (var sdk in sdks)
-    {
-        var sdkInstallFolder = CombinePaths(env.Folders.Tools, sdk);
-        var net46SdkTargetFolder = CombinePaths(net46SdkFolder, sdk);
-        var netCoreAppSdkTargetFolder = CombinePaths(netCoreAppSdkFolder, sdk);
-
-        CopyDirectory(sdkInstallFolder, net46SdkTargetFolder);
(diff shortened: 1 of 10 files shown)


DustinCampbell:
Note, this also shrinks the drops by a good bit.

filipw:
very nice! also the build script should be faster 

On a side note, I think we should also review what is the role of `OmniSharp.Abstractions`. At the moment it's a mix of a little of everything i.e. config class with hardcoded Roslyn versions or endpoint paths, all the request/response models, a bunch of interfaces, a bunch of extension methods and utilities, CLI wrapper like the`DotNetCliService` etc.

DustinCampbell:
Yes. I've been going through Abstractions and moving things out. For example, a lot of implementations have moved to OmniSharp.Host and OmniSharp.Roslyn.

DustinCampbell:
> also the build script should be faster

Not by much. Restoring these packages and copying their contents was a fairly small amount of time compared to other steps, like publish.

juergenhoetzel:
I use _-use-global-dotnet-sdk_ but the **non-existing** local *SDK* path is still referenced in the *msbuild.sh* script:
`
msbuild.sh: line 3: cd: ./.dotnet/sdk/1.0.1/: No such file or directory`

juergenhoetzel:
There is an __Enviroment__ property in __RunOptions__:

```csharp
    public IDictionary<string, string> Environment { get; }
```

using this property, the actual __SDK_DIR__ could be passed to __msbuild.sh__?

DustinCampbell:
Yes, this should be done. I'm not sure what your comment has to do with this particular PR though. `msbuild.sh` was added awhile ago.

juergenhoetzel:
> Yes, this should be done. I'm not sure what your comment has to do with this particular PR though. msbuild.sh was added awhile ago.

I thought the functionality of passing the correct ***SDK_DIR***  to **msbuild.sh** should have been part of this PR, because building on GNU/Linux systems fails if you use this new feature  (dont deploy *SDK*)

PR #848, opened 2017-05-05
Added an option to configure a code actions folder

filipw:
**Use case**: I'd like to load my own or 3rd party code actions into Omnisharp.

In this PR I added a possibility to add the following `CodeActions` node to the `omnisharp.json`.

```
{
    "FormattingOptions": {
        "IndentationSize": 2
    },
    "CodeActions": {
        "LocationPaths": [
            "C:/codeactions"
        ]
    }
}
```

This will be picked up by `RoslynFeaturesHostServicesProvider` as a source location for loading assemblies containing refactorings and code fixes.
Because `omnisharp.json` is hierarchical (global/local) we can specify code actions on a global and local level if needed.

No I could download something like https://marketplace.visualstudio.com/items?itemName=josefpihrt.Roslynator2017 and extract its DLLs into `C:/codeactions` and profit. Relative and absolute paths are supported.

### DIFF
--- a/src/OmniSharp.Abstractions/Options/CodeActionOptions.cs
+++ b/src/OmniSharp.Abstractions/Options/CodeActionOptions.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 
@@ -8,21 +9,23 @@ namespace OmniSharp.Options
     {
         public string[] LocationPaths { get; set; }
 
-        public IEnumerable<string> GetLocations(IOmniSharpEnvironment env)
+        public IEnumerable<string> GetNormalizedLocationPaths(IOmniSharpEnvironment env)
         {
-            if (LocationPaths == null) return Enumerable.Empty<string>();
+            if (LocationPaths == null || LocationPaths.Length == 0) return Enumerable.Empty<string>();
 
-            var normalizePaths = new HashSet<string>();
+            var normalizePaths = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
             foreach (var locationPath in LocationPaths)
             {
                 if (Path.IsPathRooted(locationPath))
                 {
                     normalizePaths.Add(locationPath);
                 }
-
-                normalizePaths.Add(Path.Combine(env.TargetDirectory, locationPath));
+                else
+                {
+                    normalizePaths.Add(Path.Combine(env.TargetDirectory, locationPath));
+                }
             }
-
+            
             return normalizePaths;
         }
     }
--- a/src/OmniSharp.Abstractions/Options/CodeActionOptions.cs
+++ b/src/OmniSharp.Abstractions/Options/CodeActionOptions.cs
@@ -1,0 +1,9 @@
+﻿using System.IO;
+
+namespace OmniSharp.Options
+{
+    public class CodeActionOptions
+    {
+        public string LocationPath { get; set; }
+    }
+}
--- a/src/OmniSharp.Abstractions/Options/CodeActionOptions.cs
+++ b/src/OmniSharp.Abstractions/Options/CodeActionOptions.cs
@@ -1,9 +1,29 @@
-﻿using System.IO;
+﻿using System.Collections.Generic;
+using System.IO;
+using System.Linq;
 
 namespace OmniSharp.Options
 {
     public class CodeActionOptions
     {
-        public string LocationPath { get; set; }
+        public string[] LocationPaths { get; set; }
+
+        public IEnumerable<string> GetLocations(IOmniSharpEnvironment env)
+        {
+            if (LocationPaths == null) return Enumerable.Empty<string>();
+
+            var normalizePaths = new HashSet<string>();
+            foreach (var locationPath in LocationPaths)
+            {
+                if (Path.IsPathRooted(locationPath))
+                {
+                    normalizePaths.Add(locationPath);
+                }
+
+                normalizePaths.Add(Path.Combine(env.TargetDirectory, locationPath));
+            }
+
+            return normalizePaths;
+        }
     }
 }
--- a/src/OmniSharp.Abstractions/Options/OmniSharpOptions.cs
+++ b/src/OmniSharp.Abstractions/Options/OmniSharpOptions.cs
@@ -4,15 +4,8 @@ namespace OmniSharp.Options
 {
     public class OmniSharpOptions
     {
-        public CodeActionOptions CodeActions { get; set; } = new CodeActionOptions();
+        public CodeActionOptions CodeActions { get; } = new CodeActionOptions();
 
-        public FormattingOptions FormattingOptions { get; }
-
-        public OmniSharpOptions() : this(new FormattingOptions()) { }
-
-        public OmniSharpOptions(FormattingOptions options)
-        {
-            FormattingOptions = options ?? throw new ArgumentNullException(nameof(options));
-        }
+        public FormattingOptions FormattingOptions { get; } = new FormattingOptions();
     }
 }
--- a/src/OmniSharp.Abstractions/Options/OmniSharpOptions.cs
+++ b/src/OmniSharp.Abstractions/Options/OmniSharpOptions.cs
@@ -4,7 +4,7 @@ namespace OmniSharp.Options
 {
     public class OmniSharpOptions
     {
-        public CodeActionOptions CodeActions { get; } = new CodeActionOptions();
+        public RoslynExtensionsOptions RoslynExtensionsOptions { get; } = new RoslynExtensionsOptions();
 
         public FormattingOptions FormattingOptions { get; } = new FormattingOptions();
     }
--- a/src/OmniSharp.Abstractions/Options/OmniSharpOptions.cs
+++ b/src/OmniSharp.Abstractions/Options/OmniSharpOptions.cs
@@ -4,6 +4,8 @@ namespace OmniSharp.Options
 {
     public class OmniSharpOptions
     {
+        public CodeActionOptions CodeActions { get; set; }
+
         public FormattingOptions FormattingOptions { get; }
 
         public OmniSharpOptions() : this(new FormattingOptions()) { }
--- a/src/OmniSharp.Abstractions/Options/OmniSharpOptions.cs
+++ b/src/OmniSharp.Abstractions/Options/OmniSharpOptions.cs
@@ -4,7 +4,7 @@ namespace OmniSharp.Options
 {
     public class OmniSharpOptions
     {
-        public CodeActionOptions CodeActions { get; set; }
+        public CodeActionOptions CodeActions { get; set; } = new CodeActionOptions();
 
         public FormattingOptions FormattingOptions { get; }
 
--- a/src/OmniSharp.Abstractions/Services/IAssemblyLoader.cs
+++ b/src/OmniSharp.Abstractions/Services/IAssemblyLoader.cs
@@ -8,7 +8,7 @@ namespace OmniSharp.Services
     {
         Assembly Load(AssemblyName name);
 
-        IEnumerable<Assembly> LoadAll(string folderPath);
+        IReadOnlyList<Assembly> LoadAllFrom(string folderPath);
     }
 
     public static class IAssemblyLoaderExtensions
--- a/src/OmniSharp.Abstractions/Services/IAssemblyLoader.cs
+++ b/src/OmniSharp.Abstractions/Services/IAssemblyLoader.cs
@@ -7,6 +7,8 @@ namespace OmniSharp.Services
     public interface IAssemblyLoader
     {
         Assembly Load(AssemblyName name);
+
+        IEnumerable<Assembly> LoadAll(string folderPath);
     }
 
     public static class IAssemblyLoaderExtensions
--- a/src/OmniSharp.Host/Services/AssemblyLoader.cs
+++ b/src/OmniSharp.Host/Services/AssemblyLoader.cs
@@ -34,9 +34,9 @@ namespace OmniSharp.Host.Loader
             return result;
         }
 
-        public IEnumerable<Assembly> LoadAll(string folderPath)
+        public IReadOnlyList<Assembly> LoadAllFrom(string folderPath)
         {
-            if (folderPath == null) return Enumerable.Empty<Assembly>();
+            if (string.IsNullOrWhiteSpace(folderPath)) return new Assembly[0];
 
             var assemblies = new List<Assembly>();
             foreach (var filePath in Directory.EnumerateFiles(folderPath, "*.dll"))
--- a/src/OmniSharp.Host/Services/AssemblyLoader.cs
+++ b/src/OmniSharp.Host/Services/AssemblyLoader.cs
@@ -1,4 +1,7 @@
 ﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.IO;
 using System.Reflection;
 using Microsoft.Extensions.Logging;
 using OmniSharp.Services;
@@ -11,7 +14,7 @@ namespace OmniSharp.Host.Loader
 
         public AssemblyLoader(ILoggerFactory loggerFactory)
         {
-            this._logger = loggerFactory.CreateLogger<AssemblyLoader>();
+            _logger = loggerFactory.CreateLogger<AssemblyLoader>();
         }
 
         public Assembly Load(AssemblyName name)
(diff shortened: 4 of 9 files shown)


filipw:
The reason why I thought it might be a good idea is it could give a chance for user to opt-in into features not yet available as part of Roslyn, but already built in the community or self-built. (i.e. https://github.com/dotnet/roslyn/issues/8925 which has been requested already in Omnisharp repos and so on).

See: http://recordit.co/dEAKIBAvbl

src/OmniSharp/app.config
```diff
@@ -3,4 +3,20 @@
   <startup>
     <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.6" />
   </startup>
+  <runtime>
```
filipw: this was added so that a refactoring written against Roslyn 1.0.0 (which is often the case) can load too

src/OmniSharp.Abstractions/Options/CodeActionOptions.cs
```diff
@@ -0,0 +1,29 @@
...
+        public string[] LocationPaths { get; set; }
+
+        public IEnumerable<string> GetLocations(IOmniSharpEnvironment env)
+        {
+            if (LocationPaths == null) return Enumerable.Empty<string>();
+
+            var normalizePaths = new HashSet<string>();
+            foreach (var locationPath in LocationPaths)
+            {
+                if (Path.IsPathRooted(locationPath))
```
david-driscoll: If the paths are relative shouldn't they be rooted in the directory where the configuration file lives?  However this may be super painful to pull off (would require a custom options provider)

src/OmniSharp/app.config
DustinCampbell: This will need to be updated everytime we update Roslyn to a new version (or, at least, everytime the assembly version changes). Correct?

filipw: In the current format, yes, correct. While this is technically not necessary, most of the code actions out there (i.e. packaged as VSIX-es) reference Roslyn assemblies from the host (VS). So it has to be low version if you want to avoid having a dependency on something like let's say Update 3 of Visual Studio to be installed by the user. 
I believe even the official Roslyn SDK templates default to Roslyn 1.0.0.

Otherwise, we'd only be able to load into the current process code actions built specifically against the version of Roslyn OmniSharp is built against.

I only added the three most common ones, maybe there are other DLLs that would be worthy of redirecting?

src/OmniSharp.Abstractions/Options/CodeActionOptions.cs
filipw: yeah so at the moment, relative always means relative to the folder in which OmniSharp is running. 
I think this is reasonable for most cases, but you are right, when you are editing the global file like `c:\users\filip\.omnisharp\omnisharp.json` it might appear a bit counter intuitive.

DustinCampbell:
Is this change all that interesting if we add a proper diagnostics engine that loads analyzers referenced by a project?

filipw:
> Is this change all that interesting if we add a proper diagnostics engine that loads analyzers referenced by a project?

however stand alone refactorings without analyzer (i.e. initalize a field from constructor, LINQ into foreach and so on) are installed as VSIX - they are not referenced by projects. So this PR attempts to cover only that use case.

Setting a code actions folder in a global `omnisharp.json` would be to me roughly equivalent to installing a VSIX into my Visual Studio.

DustinCampbell:
The refactorings you mentioned will very likely be added to Roslyn itself.

filipw:
yeah absolutely, then when they ship in the box, there is no need to opt into anything anymore 😃 
But it's exactly the same case as with installing these types of add ons into Visual Studio.

but until then, if you wish, you could tap into anything from here https://github.com/JosefPihrt/Roslynator/tree/master/source/Refactorings/Refactorings or here https://github.com/code-cracker/code-cracker/tree/master/src/CSharp/CodeCracker/Refactoring or few other places that people have been building.

This also has the reverse effect - if I'm an author of a refactoring, I'm loving the fact that I can build it, and it works in VS and I there is a way that I can also load it through OmniSharp into VS Code and other places. In fact, for example, at work, I have some domain specific refactorings for internal purposes only, that developers in our organization install into their VS - I think it would be great to be able to load them into OmniSharp.

I think ultimately the discussion should be what are the potential extensibility point in OmniSharp, and I think this is an excellent candidate one, especially as, in a way, it aligns with Visual Studio so experience-wise should be intuitive.

DustinCampbell:
Note: I'm not against this change. I'm just pushing on whether it's the right change, so we don't have to support it later if we decide that we should have done something else.

david-driscoll:
Would anyone complain if we called this a beta / alpha feature?  Perhaps annotate the code (and docs when we get there) to indicate that.

Ideally I'd love some sort of nuget based process going forward, but this is a nice segway into that.

filipw:
An alternative solution to the same problem could be something like this:

 - someone creates a hypothetical `ThirdPartyCodeActionProvider` implementation of OmniSharp's `ICodeActionProvider` which subclasses our `RoslynCodeActionProvider` and introduces the type of functionality contained in this current PR (ability to load code actions from some external paths)
 - package is published into Nuget
 - user installs this Nuget package into, say,`%USERPROFILE%\.omnisharp\plugins
 - at startup we scan that plugins folder and pick up the plugin and swap the default `RoslynCodeActionProvider` with `ThirdPartyCodeActionProvider` and everything else remains unchanged.

The difference would be `omnisharp.json` cannot participate anymore, so the plugin would need to have its own config file, but I think that's fine.

This could also have a different format, and the plugin could just bundle the code actions straight up and offer them as direct dependency - the downside here is that you'd need such dedicated loader for everything, rather than a having a more generic solution.

src/OmniSharp.Abstractions/Options/OmniSharpOptions.cs
```diff
@@ -4,6 +4,8 @@ namespace OmniSharp.Options
 {
     public class OmniSharpOptions
     {
+        public CodeActionOptions CodeActions { get; set; } = new CodeActionOptions();
```
DustinCampbell: Why is this settable and instantiated here? Shouldn't it follow the same pattern as FormattingOptions below?

src/OmniSharp.Abstractions/Services/IAssemblyLoader.cs
```diff
@@ -7,6 +7,8 @@ namespace OmniSharp.Services
     public interface IAssemblyLoader
     {
         Assembly Load(AssemblyName name);
+
+        IEnumerable<Assembly> LoadAll(string folderPath);
```
DustinCampbell: I'd prefer this be called `LoadAllFrom` to call out that it uses the LoadFrom context rather than the Load context.

src/OmniSharp.Host/Services/AssemblyLoader.cs
```diff
@@ -30,5 +33,43 @@ public Assembly Load(AssemblyName name)
             _logger.LogTrace($"Assembly loaded: {name}");
             return result;
         }
+
+        public IEnumerable<Assembly> LoadAll(string folderPath)
```
DustinCampbell: Seems like we could return a better interface than `IEnumerable<Assembly>` since the instance is a `List<Assembly>`. Maybe `IReadOnlyList<Assembly>` so that callers could index into it if they want.

src/OmniSharp.Host/Services/AssemblyLoader.cs
```diff
@@ -30,5 +33,43 @@ public Assembly Load(AssemblyName name)
...
+        {
+            if (folderPath == null) return Enumerable.Empty<Assembly>();
```
DustinCampbell: `string.IsNullOrWhitespace`?

src/OmniSharp.Roslyn.CSharp/Services/RoslynFeaturesHostServicesProvider.cs
```diff
@@ -12,7 +14,7 @@ public class RoslynFeaturesHostServicesProvider : IHostServicesProvider
         public ImmutableArray<Assembly> Assemblies { get; }
 
         [ImportingConstructor]
-        public RoslynFeaturesHostServicesProvider(IAssemblyLoader loader)
+        public RoslynFeaturesHostServicesProvider(IAssemblyLoader loader, OmniSharpOptions options, IOmniSharpEnvironment env)
```
DustinCampbell: To separate concerns, this should probably be added to another `IHostServicesProvider` rather than tacked onto the `RoslynFeaturesHostServicesProvider`, since this is specifically about loading the Roslyn features assemblies. Note that `OmniSharpWorkspace` aggregates multiple `IHostServicesProviders`.

src/OmniSharp/app.config
DustinCampbell: There are others that will come up eventually, but this is good for now.

src/OmniSharp.Abstractions/Options/CodeActionOptions.cs
```diff
@@ -0,0 +1,29 @@
...
+
+namespace OmniSharp.Options
+{
+    public class CodeActionOptions
+    {
+        public string[] LocationPaths { get; set; }
+
+        public IEnumerable<string> GetLocations(IOmniSharpEnvironment env)
+        {
+            if (LocationPaths == null) return Enumerable.Empty<string>();
```
DustinCampbell: `LocationPaths == null || LocationPaths.Length == 0`?

src/OmniSharp.Abstractions/Options/CodeActionOptions.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace OmniSharp.Options
+{
+    public class CodeActionOptions
+    {
+        public string[] LocationPaths { get; set; }
```
DustinCampbell: It's kind of odd that there's a public, settable `LocationPaths` property that you're really not supposed to call and there's a `GetLocations(..)` method below that's the one you're really supposed to call.

src/OmniSharp.Abstractions/Options/CodeActionOptions.cs
```diff
@@ -0,0 +1,29 @@
...
+
+        public IEnumerable<string> GetLocations(IOmniSharpEnvironment env)
```
DustinCampbell: Perhaps this should be called something like `GetNormalizedLocationPaths(...)` or something like that to indicate what it does differently over just accessing the `LocationPaths` property?

src/OmniSharp.Abstractions/Options/CodeActionOptions.cs
DustinCampbell: Up to you

src/OmniSharp.Abstractions/Options/CodeActionOptions.cs
```diff
@@ -0,0 +1,29 @@
...
+        {
+            if (LocationPaths == null) return Enumerable.Empty<string>();
+
+            var normalizePaths = new HashSet<string>();
```
DustinCampbell: Do we want this to be case-sensitive? Or, should you add `StringComparer.OrdinalIgnoreCase`? (We do that in other places)

src/OmniSharp.Roslyn.CSharp/Services/RoslynFeaturesHostServicesProvider.cs
```diff
@@ -21,6 +23,15 @@ public RoslynFeaturesHostServicesProvider(IAssemblyLoader loader)
 
             builder.AddRange(loader.Load(Features, CSharpFeatures));
 
+            var codeActionLocations = options.CodeActions.GetLocations(env);
+            if (codeActionLocations != null && codeActionLocations.Any())
```
DustinCampbell: Optional: `codeActionLocations?.Any() == true`

src/OmniSharp.Abstractions/Options/CodeActionOptions.cs
```diff
@@ -0,0 +1,29 @@
...
+            {
+                if (Path.IsPathRooted(locationPath))
+                {
+                    normalizePaths.Add(locationPath);
+                }
+
+                normalizePaths.Add(Path.Combine(env.TargetDirectory, locationPath));
+            }
+
+            return normalizePaths;
```
DustinCampbell: Note that the paths will now be returned in a different order than they were specified, which could result in subtle bugs. (e.g. Imagine two assemblies that reference the same dependency but different versions. Their load order could be important.)

src/OmniSharp.Abstractions/Options/CodeActionOptions.cs
DustinCampbell: Yeah, you're right that will be counter intuitive for the global config file case. Maybe a doc comment on `LocationPaths` to help describe the expected values?

src/OmniSharp.Roslyn.CSharp/Services/RoslynFeaturesHostServicesProvider.cs
DustinCampbell: Also, I just realized that `RoslynFeaturesHostServicesProvider` is not marked as `[Shared]`. It probably doesn't matter in practice (since `OmniSharpWorkspace` is a singleton), but I just noticed.

src/OmniSharp.Abstractions/Options/CodeActionOptions.cs
filipw: it's a bit unfortunate, yes. The reason for this is that this class (`CodeActionOptions`) is used to bind to the ASP.NET Configuration which mandates the presence of a public setter. So I thought leaving it as is and adding a public method that would normalize the paths would be a lesser evil than having another abstraction layer wrapping this or (sic!) a property with setter only to which the config would bind

src/OmniSharp.Abstractions/Options/OmniSharpOptions.cs
filipw: that's a good observation. Let's change it to `public CodeActionOptions CodeActions { get; } = new CodeActionOptions();` I think this should still bind fine to the ASP.NET Configuration and be more elegant. The other property (`FormattingOptions`) should use the same approach, it would be cleaner. Currently it can be set through the constructor too, but that constructor is redundant - I don't believe we use it anywhere, only the default constructor is in use

src/OmniSharp.Roslyn.CSharp/Services/RoslynFeaturesHostServicesProvider.cs
filipw: I didn't know there can be many of them - makes sense, thanks!

src/OmniSharp.Abstractions/Options/CodeActionOptions.cs
filipw: 👍

DustinCampbell:
I'm wondering if the folder should have a more general name. Maybe "RoslynExtensions" instead of "CodeActions". After all, there's nothing restricting this to just Code Actions. It would also load, say, CompletionProviders.

filipw:
> I'm wondering if the folder should have a more general name. Maybe "RoslynExtensions" instead of "CodeActions". After all, there's nothing restricting this to just Code Actions. It would also load, say, CompletionProviders.

yes makes sense to me. We still need relevant exports to be there (i.e. `ICodeActionProvider`) to understand which types should be discovered, but as far as loading assemblies, indeed, it can bring in all kinds of extensions/add-ons.

src/OmniSharp.Host/Services/AssemblyLoader.cs
```diff
@@ -30,5 +33,43 @@ public Assembly Load(AssemblyName name)
             _logger.LogTrace($"Assembly loaded: {name}");
             return result;
         }
+
+        public IReadOnlyList<Assembly> LoadAllFrom(string folderPath)
+        {
+            if (string.IsNullOrWhiteSpace(folderPath)) return new Assembly[0];
```
DustinCampbell: You could use `Array.Empty<Assembly>()` and avoid allocating here if you like.

src/OmniSharp.Roslyn.CSharp/Services/RoslynFeaturesHostServicesProvider.cs
```diff
@@ -1,12 +1,15 @@
 ﻿using System.Collections.Immutable;
 using System.Composition;
 using System.Reflection;
+using OmniSharp.Options;
 using OmniSharp.Services;
+using System.Linq;
```
DustinCampbell: Sort usings? It looks like `System.Linq` wound up on the bottom.

src/OmniSharp.Roslyn.CSharp/Services/ExternalFeaturesHostServicesProvider.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System.Collections.Immutable;
+using System.Composition;
+using System.Reflection;
+using OmniSharp.Options;
+using OmniSharp.Services;
+using System.Linq;
```
DustinCampbell: Sort usings? It looks like `System.Linq` wound up on the bottom.

DustinCampbell:
@filipw: CI passed. I'm good with this change (only found a couple of tiny nits). Are there more changes you wanted to make?

filipw:
Thanks guys, should be good to go.

DustinCampbell:
I won the "merge first contest". :smile: Kicking this again.

PR #854, opened 2017-05-13
Drop Kestrel as default server

david-driscoll:
Work in progress... will be working on tonight.


src/OmniSharp.Host/OmniSharp.Host.csproj
```diff
@@ -15,21 +16,19 @@
...
   <ItemGroup>
     <ProjectReference Include="..\OmniSharp.Abstractions\OmniSharp.Abstractions.csproj" />
-    <ProjectReference Include="..\OmniSharp.Stdio\OmniSharp.Stdio.csproj" />
     <ProjectReference Include="..\OmniSharp.Plugins\OmniSharp.Plugins.csproj" />
     <ProjectReference Include="..\OmniSharp.Roslyn\OmniSharp.Roslyn.csproj" />
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.AspNetCore.Diagnostics" Version="1.1.0" />
-    <PackageReference Include="Microsoft.AspNetCore.Server.Kestrel" Version="1.1.0" />
```
DustinCampbell: 😮

david-driscoll:
So this builds, but I have not tested it yet for http or stdio... in theory they should work, but not with any level of confidence! 😄

src/OmniSharp.Http/OmniSharp.Http.csproj
```diff
@@ -0,0 +1,24 @@
...
+        <!-- Needed for Microsoft.Composition -->
+        <PackageTargetFallback Condition=" '$(TargetFramework)' == 'netstandard1.6' ">$(PackageTargetFallback);dotnet5.4;portable-net45+win8</PackageTargetFallback>
+    </PropertyGroup>
+
+    <ItemGroup>
+        <PackageReference Include="Microsoft.AspNetCore.Diagnostics" Version="1.1.0" />
+        <PackageReference Include="Microsoft.AspNetCore.Server.Kestrel" Version="1.1.0" />
+        <PackageReference Include="Microsoft.AspNetCore.Hosting" Version="1.1.0" />
+        <PackageReference Include="Microsoft.AspNetCore.Http.Features" Version="1.1.0" />
+        <ProjectReference Include="..\OmniSharp.Host\OmniSharp.Host.csproj" />
```
DustinCampbell: This feels like a layering violation. Currently, nothing references `OmniSharp.Host` and it really shouldn't provide public API -- just implementations. Could API be moved into `OmniSharp.Abstractions` instead of referencing `OmniSharp.Host`?

src/OmniSharp.Host/OmniSharpCommandLineApplication.cs
```diff
@@ -0,0 +1,77 @@
...
+        {
+            Application = new CommandLineApplication(throwOnUnexpectedArg: false);
+            Application.HelpOption("-? | -h | --help");
+
+            _applicationRoot = Application.Option("-s | --source", "Solution or directory for OmniSharp to point at (defaults to current directory).", CommandOptionType.SingleValue);
+            _logLevel = Application.Option("-l | --loglevel", "Level of logging (defaults to 'Information').", CommandOptionType.SingleValue);
+            _verbose = Application.Option("-v | --verbose", "Explicitly set 'Debug' log level.", CommandOptionType.NoValue);
+            _hostPid = Application.Option("-hpid | --hostPID", "Host process ID.", CommandOptionType.SingleValue);
+            _stdio = Application.Option("-stdio | --stdio", "Use STDIO over HTTP as OmniSharp commincation protocol.", CommandOptionType.NoValue);
+            _stdio = Application.Option("-lsp | --lsp", "Use Language Server Protocol.", CommandOptionType.NoValue);
```
DustinCampbell: Setting the same variable twice.

src/OmniSharp.Http/OmniSharp.Http.csproj
```diff
@@ -0,0 +1,24 @@
...
+        <ProjectReference Include="..\OmniSharp.Roslyn.CSharp\OmniSharp.Roslyn.CSharp.csproj" />
+        <ProjectReference Include="..\OmniSharp.DotNet\OmniSharp.DotNet.csproj" />
+        <ProjectReference Include="..\OmniSharp.DotNetTest\OmniSharp.DotNetTest.csproj" />
+        <ProjectReference Include="..\OmniSharp.MSBuild\OmniSharp.MSBuild.csproj" />
+        <ProjectReference Include="..\OmniSharp.Script\OmniSharp.Script.csproj" />
```
DustinCampbell: Why does this need to reference all of the leaf projects?

OmniSharp.sln
```diff
@@ -60,88 +60,258 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "OmniSharp.DotNetTest.Tests"
...
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "OmniSharp.Http", "src\OmniSharp.Http\OmniSharp.Http.csproj", "{296B8800-007B-41D5-9001-50A371B8B5F6}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|Any CPU = Debug|Any CPU
+		Debug|x64 = Debug|x64
 		Debug|x86 = Debug|x86
+		Release|Any CPU = Release|Any CPU
+		Release|x64 = Release|x64
```
DustinCampbell: Are these additional configurations needed for anything?

src/OmniSharp.Host/OmniSharpCommandLineApplication.cs
DustinCampbell: Oh, I guess setting `-lsp` implicitly sets `_stdio`? Might be worth a source comment.

src/OmniSharp.Http/OmniSharp.Http.csproj
DustinCampbell: Or, are we planning to put hosting APIs here now?

src/OmniSharp.Http/Program.cs
```diff
@@ -0,0 +1,114 @@
...
+using OmniSharp.Services;
+using OmniSharp.Stdio;
+using OmniSharp.Stdio.Services;
+using OmniSharp.Utilities;
+
+namespace OmniSharp.Http
+{
+    public class Program
+    {
+        public static int Main(string[] args) => OmniSharp.Start(() =>
```
DustinCampbell: How do I start this application? :smile:

src/OmniSharp.Stdio/Program.cs
```diff
@@ -0,0 +1,266 @@
...
+using OmniSharp.Stdio.Logging;
+using OmniSharp.Stdio.Protocol;
+using OmniSharp.Stdio.Services;
+using OmniSharp.Utilities;
+
+namespace OmniSharp.Stdio
+{
+    public class Program : IDisposable
+    {
+        public static int Main(string[] args) => OmniSharp.Start(() =>
```
DustinCampbell: How do I start this application?

src/OmniSharp.Http/OmniSharp.Http.csproj
david-driscoll: It doesn't, that's copy and paste problem I think.

src/OmniSharp.Http/OmniSharp.Http.csproj
david-driscoll: My initial thinking is to have the hosting Api's here.  Which will allow us to have different startup projects.  Which will then be packaged with their specific dependencies.

* `OmniSharp.Http` - Starts an http based server (kestrel)
* `OmniSharp.Stdio` / `OmniSharp` (for legacy reasons) - Starts an stdio based server
* `OmniSharp.Lsp`* - Starts an LSP based server (could be integrated into `Stdio` potentially.
* `OmniSharp.Sockets`* - Starts up a server that runs over a traditional socket
* `OmniSharp.WebSockets`* - Starts up a server that runs over a WebSocket.

Now these are cool endgame ideas, which may or may not require large additional assemblies / libraries, ie Kestrel.  The three big things we need to support are Http, Stdio, and in the future Language Server Protocol.

src/OmniSharp.Host/OmniSharpCommandLineApplication.cs
david-driscoll: I think it's just a copy and paste fail 😢

OmniSharp.sln
david-driscoll: Nope... just visual studio being mean I think.

david-driscoll:
Updated to be in sync with the latest code base, I think we discussed some of the parts on slack, but I expect more comments. 🕺 💃

src/OmniSharp.Host/Internal/CommandOptionExtensions.cs
```diff
@@ -1,11 +1,11 @@
...
 using Microsoft.Extensions.CommandLineUtils;
 
-namespace OmniSharp.Host.Internal
+namespace OmniSharp.Internal
 {
-    internal static class CommandOptionExtensions
+    public static class CommandOptionExtensions
     {
-        internal static T GetValueOrDefault<T>(this CommandOption opt, T defaultValue)
+        public static T GetValueOrDefault<T>(this CommandOption opt, T defaultValue)
```
filipw: we ended up with public class in an internal namespace 😃

src/OmniSharp.Stdio/Program.cs
```diff
@@ -0,0 +1,264 @@
...
+                var writer = new SharedConsoleWriter();
+                var plugins = application.CreatePluginAssemblies();
+                var configuration = new OmniSharpConfigurationBuilder(environment).Build();
+                var serviceProvider = OmniSharpMefBuilder.CreateDefaultServiceProvider(configuration);
+                var mefBuilder = new OmniSharpMefBuilder(serviceProvider, environment, writer, new StdioEventEmitter(writer));
+                var compositionHost = mefBuilder.Build();
+                var loggerFactory = serviceProvider.GetRequiredService<ILoggerFactory>();
+                var cancellation = new CancellationTokenSource();
+
+                using (var program = new Program(Console.In, writer, environment, configuration, serviceProvider, compositionHost, loggerFactory, cancellation))
```
filipw: this feels a bit weird IMHO, a `Main` in a `Program`, that then instantiates `Program` and calls another "Main" (`Start`) on it. I think for sanity reasons it should be split into dedicated classes, and let `Program` just act as an entry point

src/OmniSharp/Program.cs
```diff
@@ -7,17 +7,6 @@ public class Program
...
-            argsList.Remove("--debug");
-            Console.WriteLine($"Attach debugger to process {Process.GetCurrentProcess().Id} to continue...");
-            while (!Debugger.IsAttached)
-            {
-                Thread.Sleep(100);
-            }
-        }
-
-        return OmniSharp.Program.Main(argsList.ToArray());
+        return OmniSharp.Stdio.Program.Main(args);
```
filipw: is there even a point of having this project anymore? Could this not simply be merged with OmniSharp.Stdio? 

One DLL less..

filipw: Note that for compatibility purposes (omnisharp.exe name) we can have a project named `OmniSharp.Stdio.proj` but build its output assembly name as `OmniSharp`

david-driscoll: Compatibility was the main reason, that can be shifted around.

src/OmniSharp.Stdio/Program.cs
david-driscoll: Sure, I'll change it tonight.

src/OmniSharp.Host/Internal/CommandOptionExtensions.cs
david-driscoll: That's how the aspnet team does things. ;)  https://github.com/aspnet/Mvc/blob/b3fe473e2971da62208729d792be162a6ba60cc7/src/Microsoft.AspNetCore.Mvc.Formatters.Json/Internal/JsonArrayPool.cs

filipw: 😱😱😱

david-driscoll:
Updated... the strangest thing is failing to work at the moment.  For some reason kestrel won't load properly, and I can't figure out why.

david-driscoll:
@DustinCampbell any insight into the build failures on linux / mac?  Seems like since `dotnet publish` isn't used that `libuv.dll` isn't added.

DustinCampbell:
Yeah, we're not bundling libuv and `dotnet publish` requires a RID, which isn't relevant in our world of Mono builds. I had this as a discussion item in https://github.com/OmniSharp/omnisharp-roslyn/pull/915.

I think we should just `apt-get install libuv1`. On OSX, `brew install libuv`

david-driscoll:
okay, I'll look to add those to the build.

DustinCampbell:
No worries! I've got it.

build.cake
```diff
@@ -349,9 +352,14 @@ Task("BuildMain")
...
     .IsDependentOn("Restore")
-    .IsDependentOn("BuildMain")
     .Does(() =>
 {
+    foreach (var project in buildPlan.HostProjects)
+    {
+        var projectName = project + ".csproj";
+        var projectFilePath = CombinePaths(env.Folders.Source, project, projectName);
+        BuildProject(env, projectName, projectFilePath, configuration, "test");
+    }
```
DustinCampbell: This is kind of weird. The `BuildTest` task shouldn't build the host projects. IMO, it should just depend on `BuildMain` (which should be renamed to `BuildHosts`).

build.cake
```diff
@@ -337,10 +337,13 @@ Task("BuildMain")
     .IsDependentOn("Restore")
     .Does(() =>
 {
-    var projectName = buildPlan.MainProject + ".csproj";
-    var projectFilePath = CombinePaths(env.Folders.Source, buildPlan.MainProject, projectName);
+    foreach (var project in buildPlan.HostProjects)
```
DustinCampbell: Should this task be renamed to `BuildHosts`?

build.cake
```diff
@@ -349,9 +352,14 @@ Task("BuildMain")
 Task("BuildTest")
     .IsDependentOn("Setup")
     .IsDependentOn("Restore")
-    .IsDependentOn("BuildMain")
```
DustinCampbell: Why?

DustinCampbell: Actually, it turns out that this creates a small problem in that the host projects get built twice -- once during `BuildTest` and then again during `Publish`, since that task depends on `BuildMain`. Best just to have this depend on `BuildMain` and not build the host projects in `BuildTest.

build.json
```diff
@@ -11,10 +11,14 @@
   "MonoFramework": "framework-5.2.0.213.zip",
   "MonoMSBuildRuntime": "Microsoft.Build.Runtime.Mono-alpha5.zip",
   "MonoMSBuildLib": "Microsoft.Build.Lib.Mono-alpha5.zip",
-  "MainProject": "OmniSharp",
+  "HostProjects": [
+    "OmniSharp.Stdio",
```
DustinCampbell: I'm wondering if we should rename `OmniSharp.Stdio` to `OmniSharp`. Thoughts?

msbuild.ps1
```diff
@@ -0,0 +1,12 @@
+echo test
+$PSScriptRoot = Split-Path -Parent -Path $MyInvocation.MyCommand.Definition
+$SDK_DIR = "$PSScriptRoot/.dotnet/sdk/1.0.1/"
```
DustinCampbell: I would expect this to be broken. We install the 1.0.4 version of SDK.

.vscode/tasks.json
```diff
@@ -1,22 +1,41 @@
...
             "taskName": "build",
+            "echoCommand": true,
             // Show the output window only if unrecognized errors occur.
-            "showOutput": "silent",
+            "showOutput": "always",
             // Use the standard MS compiler pattern to detect errors, warnings and infos
-            "problemMatcher": "$msCompile"
+            "problemMatcher": "$msCompile",
+            "isBuildCommand": true,
+            "command": "${workspaceRoot}/msbuild.sh",
```
DustinCampbell: This should be broken. `msbuild.sh` doesn't exist anymore.

DustinCampbell: You should be able to just call `msbuild`. Our build script expects Mono 5.2 to be present.

.vscode/tasks.json
```diff
@@ -1,22 +1,41 @@
...
+            "windows": {
+                "command": "powershell",
+                "args": [
+                    "${workspaceRoot}/msbuild.ps1",
```
DustinCampbell: What's the need for this indirection? Why not just set the command to `"${workspaceRoot}/.dotnet/dotnet"` on Windows? That's how we build.

msbuild.ps1
DustinCampbell: On second thought (after looking at the tasks.json), we *really* don't need this script.

scripts/runhelpers.cake
```diff
@@ -134,13 +135,21 @@ ExitStatus Run(string command, string arguments, RunOptions runOptions)
...
     Context.Log.Write(Verbosity.Diagnostic, LogLevel.Debug, "  CWD: {0}", workingDirectory);
 
-    var process = System.Diagnostics.Process.Start(
-            new ProcessStartInfo(command, arguments)
-            {
-                WorkingDirectory = workingDirectory,
-                UseShellExecute = false,
-                RedirectStandardOutput = runOptions.Output != null
-            });
+    var pInfo = new ProcessStartInfo(command, arguments)
```
DustinCampbell: What's this? Hungarian notation for a pointer to some `Info` class? :smile:

How about `startInfo`?

src/OmniSharp.Host/Endpoint/EndpointHandler.cs
```diff
@@ -5,31 +5,26 @@
...
 using OmniSharp.Models;
 using OmniSharp.Models.UpdateBuffer;
 using OmniSharp.Plugins;
+using OmniSharp.Stdio.Protocol;
 
-namespace OmniSharp.Middleware.Endpoint
+namespace OmniSharp.Endpoint
 {
-    class LanguageModel
+    public abstract class EndpointHandler
```
DustinCampbell: Since you're making a public API, you might consider renaming from `Endpoint` to `EndPoint`.

src/OmniSharp.Host/Endpoint/EndpointHandler.cs
```diff
@@ -5,31 +5,26 @@
...
     {
-        public string Language { get; set; }
-        public string FileName { get; set; }
-    }
-
-    abstract class EndpointHandler
-    {
-        public abstract Task<object> Handle(HttpContext context);
+        public abstract Task<object> Handle(RequestPacket context);
```
DustinCampbell: `packet`

src/OmniSharp.Host/Endpoint/EndpointHandler.cs
```diff
@@ -5,31 +5,26 @@
...
 
         public static EndpointHandler Create<TRequest, TResponse>(IPredicateHandler languagePredicateHandler, CompositionHost host,
-            ILogger logger, OmniSharpEndpointMetadata item,
+            ILogger logger,
+            OmniSharpEndpointMetadata item,
```
DustinCampbell: `metadata`?

src/OmniSharp.Host/Endpoint/EndpointHandler.cs
```diff
@@ -38,7 +33,8 @@ abstract class EndpointHandler
         }
 
         public static EndpointHandler Factory(IPredicateHandler languagePredicateHandler, CompositionHost host,
-            ILogger logger, OmniSharpEndpointMetadata item,
+            ILogger logger,
+            OmniSharpEndpointMetadata item,
```
DustinCampbell: `metadata`?

src/OmniSharp.Host/OmniSharp.Host.csproj
```diff
@@ -4,6 +4,7 @@
     <TargetFramework>net46</TargetFramework>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <PlatformTarget>AnyCPU</PlatformTarget>
+    <RootNamespace>OmniSharp</RootNamespace>
```
DustinCampbell: `OmniSharp.Host`

src/OmniSharp.Host/OmniSharpWorkspaceInitializer.cs
```diff
@@ -0,0 +1,100 @@
...
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using OmniSharp.Options;
+using OmniSharp.Roslyn;
+using OmniSharp.Roslyn.Options;
+using OmniSharp.Services;
+
+namespace OmniSharp
+{
+    public class OmniSharpWorkspaceInitializer
```
DustinCampbell: Why does everything need to be prefixed with `OmniSharp`? We're in the `OmniSharp` namespace!

src/OmniSharp.Host/OmniSharpMefBuilder.cs
```diff
@@ -0,0 +1,141 @@
...
+using OmniSharp.FileWatching;
+using OmniSharp.Mef;
+using OmniSharp.Options;
+using OmniSharp.Roslyn;
+using OmniSharp.Services;
+using OmniSharp.Stdio.Services;
+
+namespace OmniSharp
+{
+    public class OmniSharpMefBuilder
```
DustinCampbell: Please don't prefix class names with `OmniSharp`. That's what namespaces are for.

src/OmniSharp.Host/OmniSharpConfigurationBuilder.cs
```diff
@@ -0,0 +1,57 @@
+using System;
+using System.Collections.Generic;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.FileProviders;
+using OmniSharp.Internal;
+using OmniSharp.Utilities;
+
+namespace OmniSharp
+{
+    public class OmniSharpConfigurationBuilder : IConfigurationBuilder
```
DustinCampbell: I don't think we need the `OmniSharp` prefix.

src/OmniSharp.Host/OmniSharpCommandLineApplicationExtensions.cs
```diff
@@ -0,0 +1,23 @@
+using System.Linq;
+using OmniSharp.Plugins;
+using OmniSharp.Services;
+
+namespace OmniSharp
+{
+    public static class OmniSharpCommandLineApplicationExtensions
```
DustinCampbell: The `OmniSharp` class name prefix looks silly here.

src/OmniSharp.Host/OmniSharpCommandLineApplication.cs
```diff
@@ -0,0 +1,95 @@
...
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.CommandLineUtils;
+using Microsoft.Extensions.Logging;
+using OmniSharp.Internal;
+
+namespace OmniSharp
+{
+    public class OmniSharpCommandLineApplication
```
DustinCampbell: What's the point of prefixing the class name with `OmniSharp`?

src/OmniSharp.Host/Program.cs
```diff
@@ -3,24 +3,46 @@
...
+using OmniSharp.Roslyn;
 using OmniSharp.Services;
 using OmniSharp.Stdio;
 using OmniSharp.Stdio.Services;
 using OmniSharp.Utilities;
 
 namespace OmniSharp
 {
-    public class Program
+    public class OmniSharp
```
DustinCampbell: `OmniSharp.OmniSharp`? Please, no. :smile:

src/OmniSharp.Http/OmniSharp.Http.csproj
DustinCampbell: Perhaps we should rename `OmniSharp.Host` to `OmniSharp.Hosting`?

src/OmniSharp.Http/OmniSharp.Http.csproj
DustinCampbell: well, fix it!

src/OmniSharp.Http/OmniSharpHttpCommandLineApplication.cs
```diff
@@ -0,0 +1,21 @@
+using Microsoft.Extensions.CommandLineUtils;
+using OmniSharp.Internal;
+
+namespace OmniSharp.Http
+{
+    public class OmniSharpHttpCommandLineApplication : OmniSharpCommandLineApplication
```
DustinCampbell: I'm going to keep complaining -- don't unnecessarily prefix class names with `OmniSharp`. Also, why is this public?

src/OmniSharp.Http/OmniSharpHttpEnvironment.cs
```diff
@@ -0,0 +1,7 @@
+namespace OmniSharp.Http
+{
+    public class OmniSharpHttpEnvironment
```
DustinCampbell: Still hating the prefix thing. :smile:

src/OmniSharp.Http/Program.cs
DustinCampbell: I don't think this needs to be public.

src/OmniSharp.Http/Program.cs
```diff
@@ -0,0 +1,33 @@
...
+using Microsoft.AspNetCore.Server.Kestrel;
+using Microsoft.AspNetCore.Server.Kestrel.Internal;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using OmniSharp.Services;
+using OmniSharp.Stdio;
+
+namespace OmniSharp.Http
+{
+    public class Program
```
DustinCampbell: I don't think this needs to be public.

src/OmniSharp.Http/Program.cs
```diff
@@ -0,0 +1,33 @@
...
+        public static int Main(string[] args) => OmniSharp.Start(() =>
+        {
+            var application = new OmniSharpHttpCommandLineApplication();
+            application.OnExecute(() =>
+            {
+                var environment = application.CreateEnvironment();
+                var writer = new SharedConsoleWriter();
+                var plugins = application.CreatePluginAssemblies();
+
+                var program = new Host(environment, writer, plugins, application.Port, application.Interface);
```
DustinCampbell: `host`

src/OmniSharp.Http/Host.cs
```diff
@@ -0,0 +1,86 @@
...
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using OmniSharp.Eventing;
+using OmniSharp.Plugins;
+using OmniSharp.Stdio.Services;
+using OmniSharp.Utilities;
+
+namespace OmniSharp.Http
+{
+    public class Host
```
DustinCampbell: This doesn't need to be public.

src/OmniSharp.Http/Startup.cs
```diff
@@ -0,0 +1,78 @@
...
+using OmniSharp.Eventing;
+using OmniSharp.Http.Middleware;
+using OmniSharp.Options;
+using OmniSharp.Roslyn;
+using OmniSharp.Services;
+using OmniSharp.Stdio.Services;
+
+namespace OmniSharp.Http
+{
+    public class Startup
```
DustinCampbell: I don't think this needs to be public.

src/OmniSharp.Http/Startup.cs
```diff
@@ -0,0 +1,78 @@
...
+    {
+        private readonly IOmniSharpEnvironment _environment;
+        private readonly IEventEmitter _eventEmitter;
+        private readonly ISharedTextWriter _writer;
+        private readonly IConfigurationRoot _configuration;
+        // private OmniSharpWorkspace _workspace;
```
DustinCampbell: Delete commented field.

src/OmniSharp.Http/Startup.cs
DustinCampbell: I'm not sure though. This is ASP.NET stuff.

src/OmniSharp.Host/OmniSharpWorkspaceInitializer.cs
```diff
@@ -0,0 +1,100 @@
...
+    {
+        private readonly OmniSharpWorkspace _workspace;
+        private readonly CompositionHost _compositionHost;
+        private readonly IConfiguration _configuration;
+        private readonly ILogger _logger;
+        private readonly IOptionsMonitor<OmniSharpOptions> _options;
+
+        public OmniSharpWorkspaceInitializer(
```
DustinCampbell: This class is kind of weird. It looks like it can be reduced to a single static method.

src/OmniSharp.Http/Startup.cs
```diff
@@ -0,0 +1,78 @@
...
+                return false;
+            });
+
+            app.UseRequestLogging();
+            app.UseExceptionHandler("/error");
+            app.UseMiddleware<EndpointMiddleware>(_compositionHost);
+            app.UseMiddleware<StatusMiddleware>(workspace);
+            app.UseMiddleware<StopServerMiddleware>();
+
+            new OmniSharpWorkspaceInitializer(serviceProvider, _compositionHost, _configuration, logger).Initialize();
```
DustinCampbell: That's a pretty smelly pattern: allocate an object just to call a method and throw it away immediately. If the goal is to apply side effects to the arguments, just turn `OmniSharpWorkspaceInitializer` (or better, *`WorkspaceInitializer`* :smile:) into a single helper method.

src/OmniSharp.Stdio/Host.cs
```diff
@@ -0,0 +1,226 @@
...
+using OmniSharp.Plugins;
+using OmniSharp.Services;
+using OmniSharp.Stdio.Logging;
+using OmniSharp.Stdio.Protocol;
+using OmniSharp.Stdio.Services;
+using OmniSharp.Utilities;
+
+namespace OmniSharp.Stdio
+{
+    public class Host : IDisposable
```
DustinCampbell: Does this need to be public?

src/OmniSharp.Stdio/Host.cs
```diff
@@ -0,0 +1,226 @@
...
+            _loggerFactory?.Dispose();
+            _cancellation?.Dispose();
+        }
+
+        public void Start()
+        {
+            var logger = _loggerFactory.CreateLogger<Program>();
+            _loggerFactory.AddStdio(_writer, (category, level) => OmniSharp.LogFilter(category, level, _environment));
+
+            new OmniSharpWorkspaceInitializer(_serviceProvider, _compositionHost, _configuration, logger).Initialize();
```
DustinCampbell: Looks weird here too. :smile:

src/OmniSharp.Stdio/OmniSharp.Stdio.csproj
```diff
@@ -1,19 +1,22 @@
...
-  <PropertyGroup>
-    <TargetFramework>net46</TargetFramework>
-    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-    <PlatformTarget>AnyCPU</PlatformTarget>
-  </PropertyGroup>
+    <PropertyGroup>
+        <TargetFramework>net46</TargetFramework>
+        <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+        <PlatformTarget>AnyCPU</PlatformTarget>
+        <AssemblyName>OmniSharp</AssemblyName>
```
DustinCampbell: It looks like you *are* changing this to `OmniSharp`. In that case, could we rename the project too?

src/OmniSharp.Stdio/OmniSharp.Stdio.csproj
```diff
@@ -1,19 +1,22 @@
...
-    <ProjectReference Include="..\OmniSharp.Abstractions\OmniSharp.Abstractions.csproj" />
-  </ItemGroup>
+    <ItemGroup>
+        <ProjectReference Include="..\OmniSharp.Host\OmniSharp.Host.csproj" />
+        <ProjectReference Include="..\OmniSharp.Roslyn.CSharp\OmniSharp.Roslyn.CSharp.csproj" />
+        <ProjectReference Include="..\OmniSharp.DotNet\OmniSharp.DotNet.csproj" />
+        <ProjectReference Include="..\OmniSharp.DotNetTest\OmniSharp.DotNetTest.csproj" />
+        <ProjectReference Include="..\OmniSharp.MSBuild\OmniSharp.MSBuild.csproj" />
+        <ProjectReference Include="..\OmniSharp.Script\OmniSharp.Script.csproj" />
+    </ItemGroup>
```
DustinCampbell: Note: whitespace is changing here. I'm not sure that was intentional.

src/OmniSharp.Stdio/OmniSharp.Stdio.csproj
```diff
@@ -1,19 +1,22 @@
...
-    <TargetFramework>net46</TargetFramework>
-    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-    <PlatformTarget>AnyCPU</PlatformTarget>
-  </PropertyGroup>
+    <PropertyGroup>
+        <TargetFramework>net46</TargetFramework>
+        <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+        <PlatformTarget>AnyCPU</PlatformTarget>
+        <AssemblyName>OmniSharp</AssemblyName>
+        <PreserveCompilationContext>true</PreserveCompilationContext>
```
DustinCampbell: Is this needed since we're net46?

src/OmniSharp.Stdio/OmniSharpStdioCommandLineApplication.cs
```diff
@@ -0,0 +1,21 @@
+using Microsoft.Extensions.CommandLineUtils;
+using OmniSharp.Internal;
+
+namespace OmniSharp.Stdio
+{
+    public class OmniSharpStdioCommandLineApplication : OmniSharpCommandLineApplication
```
DustinCampbell: Boo on the prefix

src/OmniSharp/app.config
```diff
@@ -1,56 +0,0 @@
-﻿<?xml version="1.0" encoding="utf-8" ?>
```
DustinCampbell: Where did this move to? These binding redirects are really important.

tests/OmniSharp.Stdio.Tests/OmniSharp.Stdio.Tests.csproj
```diff
@@ -12,7 +12,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <ProjectReference Include="..\..\src\OmniSharp.Stdio\OmniSharp.Stdio.csproj" />
+      <ProjectReference Include="..\..\src\OmniSharp.Stdio\OmniSharp.Stdio.csproj" />
```
DustinCampbell: unneeded whitespace change.

tests/OmniSharp.Stdio.Tests/StdioServerFacts.cs
```diff
@@ -81,76 +98,83 @@ public void ServerRepliesWithErrorToInvalidRequest()
...
+        //        }
+        //    );
+
+        //    using (BuildTestServerAndStart(new StringReader(JsonConvert.SerializeObject(request) + "\r\n"), writer))
+        //    {
+        //        Assert.True(writer.Completion.WaitOne(TimeSpan.FromSeconds(10)), "Timeout");
+        //        Assert.Null(writer.Exception);
+        //    }
+
+        //}
```
DustinCampbell: What's the plan for these commented tests?

tests/TestUtility/OmniSharpTestHost.cs
```diff
@@ -26,7 +28,7 @@ public class OmniSharpTestHost : DisposableObject
         private static Lazy<Assembly[]> s_lazyAssemblies = new Lazy<Assembly[]>(() => new[]
         {
             typeof(OmniSharpEndpoints).GetTypeInfo().Assembly, // OmniSharp.Abstractions
-            typeof(Startup).GetTypeInfo().Assembly, // OmniSharp.Host
+            typeof(OmniSharp.OmniSharp).GetTypeInfo().Assembly, // OmniSharp.Host
```
DustinCampbell: bleeding eyes.

tests/TestUtility/OmniSharpTestHost.cs
```diff
@@ -97,31 +99,26 @@ public static OmniSharpTestHost Create(string path = null, ITestOutputHelper tes
                 throw new InvalidOperationException($"Local .NET CLI path does not exist. Did you run build.(ps1|sh) from the command line?");
             }
 
-            var builder = new ConfigurationBuilder();
+             var builder = new ConfigurationBuilder();
```
DustinCampbell: Unnecessary single space added

tests/TestUtility/TestUtility.csproj
```diff
@@ -6,14 +6,14 @@
...
-    <ProjectReference Include="..\..\src\OmniSharp.Script\OmniSharp.Script.csproj" />
-  </ItemGroup>
+    <ItemGroup>
+        <ProjectReference Include="..\..\src\OmniSharp.DotNetTest\OmniSharp.DotNetTest.csproj" />
+        <ProjectReference Include="..\..\src\OmniSharp.DotNet\OmniSharp.DotNet.csproj" />
+        <ProjectReference Include="..\..\src\OmniSharp.Host\OmniSharp.Host.csproj" />
+        <ProjectReference Include="..\..\src\OmniSharp.MSBuild\OmniSharp.MSBuild.csproj" />
+        <ProjectReference Include="..\..\src\OmniSharp.Roslyn.CSharp\OmniSharp.Roslyn.CSharp.csproj" />
+        <ProjectReference Include="..\..\src\OmniSharp.Script\OmniSharp.Script.csproj" />
+    </ItemGroup>
```
DustinCampbell: Unintentional whitespace changes.

build.cake
```diff
@@ -438,11 +446,10 @@ void CopyMonoBuild(BuildEnvironment env, string sourceFolder, string outputFolde
     FileHelper.Delete(CombinePaths(outputFolder, "System.Threading.Thread.dll"));
 }
 
-string PublishMonoBuild(BuildEnvironment env, BuildPlan plan, string configuration, bool archive)
+string PublishMonoBuild(string project, BuildEnvironment env, BuildPlan plan, string configuration, bool archive)
 {
     Information("Publishing Mono build...");
```
DustinCampbell: It might be worth printing the project name that's being published so that the text isn't just duplicated in the build log.

build.cake
david-driscoll: I had to do this because building the hosts was emitting an exe for both projects, which then running unit tests would fail, because it could not find the dll.  It was the best solution I could think of at the time anyway.

build.json
david-driscoll: We could, but then it obscures that we have different host projects (I don't have a preference)

src/OmniSharp.Host/Endpoint/EndpointHandler.cs
david-driscoll: where did you get the extra capital `P` from?  I know I'm Canadian and everything... but I've never seen that! https://en.wikipedia.org/wiki/Endpoint

src/OmniSharp.Host/Program.cs
david-driscoll: really?? 😄

src/OmniSharp/app.config
david-driscoll: they didn't move anywhere... I'll bring them back 😱 😨

src/OmniSharp.Host/OmniSharp.Host.csproj
david-driscoll: Today everything is just in the `OmniSharp` namespace, this was just codifying that.  I can change, and update all the namespaces if we want.

src/OmniSharp.Stdio/OmniSharp.Stdio.csproj
david-driscoll: This was based on @filipw's comment at one point about using a different executable name.  So this was intenation. And also happens to OmniSharp.Http, though I'm fine with changing if we want.

build.cake
DustinCampbell: I'm not following. What you're doing here should be functionally the same as depending on `BuildMain`.

build.json
DustinCampbell: I'm betting folks will figure it out. :smile:

src/OmniSharp.Host/Endpoint/EndpointHandler.cs
DustinCampbell: I'm don't feel strongly about it. I just noticed that it's different than what we do in System.Net: https://msdn.microsoft.com/en-us/library/system.net.endpoint(v=vs.110).aspx

src/OmniSharp.Host/Program.cs
DustinCampbell: Dude.

src/OmniSharp/app.config
DustinCampbell: yeah, we'll need em.

src/OmniSharp.Stdio/OmniSharp.Stdio.csproj
DustinCampbell: Well, it kind of defeats your argument about obscuring the fact that we have different hosts. :smile:

That said, I'm OK if we don't rename the project. I'm fine with it.

OmniSharp.sln.GhostDoc.xml
```diff
@@ -0,0 +1,53 @@
+<GhostDoc>
```
filipw: is this file needed?

david-driscoll: Nope, was just an extension I had installed.

src/OmniSharp.Host/Endpoint/EndpointHandler.cs
david-driscoll: 😱

src/OmniSharp.Stdio/OmniSharp.Stdio.csproj
david-driscoll: Not entirely sure, will test.

david-driscoll:
Updated with rest of the feedback

.vscode/tasks.json
```diff
@@ -1,22 +1,41 @@
...
             // Show the output window only if unrecognized errors occur.
-            "showOutput": "silent",
+            "showOutput": "always",
             // Use the standard MS compiler pattern to detect errors, warnings and infos
-            "problemMatcher": "$msCompile"
+            "problemMatcher": "$msCompile",
+            "isBuildCommand": true,
+            "command": "msbuild",
+            "windows": {
+                "command": "powershell",
```
DustinCampbell: Why run "dotnet" with Powershell? Does the following not work?

```
"windows": {
    "command": "${workspaceRoot}/.dotnet/dotnet",
    "args": [
        "OmniSharp.sln",
        // Ask msbuild to generate full paths for file names.
        "/property:GenerateFullPaths=true"
    ]
}
```

OmniSharp.sln
DustinCampbell: it can be mean sometimes :smile:

build.cake
DustinCampbell: OK. I see it now. You're trying to set this "TestOutputType" property as an environment variable. I get why you added this, but it feels pretty messy.

I would argue that OmniSharp.Http and OmniSharp.Stdio should just be super slim exes. The functionality that needs testing should go into libraries. Or, maybe better, we could just have a single OmniSharp.exe that dynamically loads OmniSharp.Http or OmniSharp.Stdio depending on command line args? What do you think?

src/OmniSharp.Host/Internal/CommandOptionExtensions.cs
DustinCampbell: doesn't make it right. That's a smell.

What semantic meaning are you trying to convey by having a public class in an internal namespace. Why does this need to be public at all? Is it part our public API surface area?

scripts/archiving.cake
```diff
@@ -3,14 +3,19 @@
 
 using System.IO.Compression;
 
+string GetPackagePrefix(string project)
+{
+    return project.EndsWith(".Stdio") ? string.Empty : "." + project.Substring(project.IndexOf('.')+1).ToLower();
```
DustinCampbell: Did you mean this?

```C#
return project.EndsWith(".Stdio") ? string.Empty : project.Substring(project.IndexOf('.')).ToLower();
```

david-driscoll: The extra '.' was intentional so that the package comes out like `omnisharp.http-win-x64` vs `omnisharphttp-win-x64`.

DustinCampbell: Sure, I get that. But you already have a "." because you're searching for ".". You should be able to just use `project.IndexOf('.')` rather than `project.IndexOf('.')+1`.

david-driscoll: stop making sense...

src/OmniSharp.Http/OmniSharp.Http.csproj
DustinCampbell: FWIW, I'd be interested in what you think about a design where there's a single startup project that dynamically loads the correct server based on command-line args. Would that be possible?

david-driscoll: It's totally doable now since we're only bound to the desktop framework.  That would be mean that we should ship all the hosts in package, which would mean we're bring kestrel with us going forward.  And then the same thing would happen if/when we support the new sockets api.  I don't have a preference, but there would be extra "baggage" as it were.

DustinCampbell: If we design for this case, one could imagine that the "transport hosts" could be plugins.

DustinCampbell: Note: I'm not asking for further changes here. We should get this PR in and start testing and refactoring.

DustinCampbell:
woo-hoo!

filipw:
epic! 🏄

PR #856, opened 2017-05-13
Support to run MSTest based tests.

AbhitejJohn:
Accompanying PR : https://github.com/OmniSharp/omnisharp-vscode/pull/1478

### DIFF
--- a/src/OmniSharp.DotNetTest/TestFrameworks/MSTestFramework.cs
+++ b/src/OmniSharp.DotNetTest/TestFrameworks/MSTestFramework.cs
@@ -8,8 +8,7 @@
 
         protected override bool IsTestAttributeName(string typeName)
         {
-            return typeName == "Microsoft.VisualStudio.TestTools.UnitTesting.TestMethodAttribute"
-                || typeName == "Microsoft.VisualStudio.TestTools.UnitTesting.DataTestMethodAttribute";
+            return typeName == "Microsoft.VisualStudio.TestTools.UnitTesting.TestMethodAttribute";
         }
     }
 }
--- a/src/OmniSharp.DotNetTest/TestFrameworks/MSTestFramework.cs
+++ b/src/OmniSharp.DotNetTest/TestFrameworks/MSTestFramework.cs
@@ -1,0 +1,15 @@
+﻿namespace OmniSharp.DotNetTest.TestFrameworks
+{
+    internal class MSTestFramework : TestFramework
+    {
+        public override string FeatureName { get; } = "MSTestMethod";
+        public override string Name { get; } = "mstest";
+        public override string MethodArgument { get; } = "--test";
+
+        protected override bool IsTestAttributeName(string typeName)
+        {
+            return typeName == "Microsoft.VisualStudio.TestTools.UnitTesting.TestMethodAttribute"
+                || typeName == "Microsoft.VisualStudio.TestTools.UnitTesting.DataTestMethodAttribute";
+        }
+    }
+}
--- a/build.json
+++ b/build.json
@@ -21,12 +21,14 @@
   ],
   "TestAssets": [
     "NUnitTestProject",
-    "XunitTestProject"
+    "XunitTestProject",
+    "MSTestProject"
   ],
   "LegacyTestAssets": [
     "BasicTestProjectSample01",
     "BasicTestProjectSample02",
     "LegacyNunitTestProject",
-    "LegacyXunitTestProject"
+    "LegacyXunitTestProject",
+    "LegacyMSTestProject"
   ]
 }
--- a/src/OmniSharp.DotNetTest/Legacy/LegacyTestManager.cs
+++ b/src/OmniSharp.DotNetTest/Legacy/LegacyTestManager.cs
@@ -67,7 +67,9 @@ namespace OmniSharp.DotNetTest.Legacy
             var testStartInfo = testStartInfoMessage.DeserializePayload<TestStartInfo>();
 
             var fileName = testStartInfo.FileName;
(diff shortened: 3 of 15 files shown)


AbhitejJohn:
Tagging @DustinCampbell for a review.

willl:
Should unit tests/test projects be added as part of this @david-driscoll?

They exist for [xUnit](https://github.com/OmniSharp/omnisharp-roslyn/tree/dev/test-assets/test-projects/XunitTestProject) and [NUnit](https://github.com/OmniSharp/omnisharp-roslyn/tree/dev/test-assets/test-projects/NUnitTestProject).

src/OmniSharp.DotNetTest/TestFrameworks/MSTestFramework.cs
```diff
@@ -0,0 +1,15 @@
...
+    internal class MSTestFramework : TestFramework
+    {
+        public override string FeatureName { get; } = "MSTestMethod";
+        public override string Name { get; } = "mstest";
+        public override string MethodArgument { get; } = "--test";
+
+        protected override bool IsTestAttributeName(string typeName)
+        {
+            return typeName == "Microsoft.VisualStudio.TestTools.UnitTesting.TestMethodAttribute"
+                || typeName == "Microsoft.VisualStudio.TestTools.UnitTesting.DataTestMethodAttribute";
```
filipw: `DataTestMethodAttribute` is not needed here, because it inherits from `TestMethodAttribute`.
The code in OmniSharp checks the whole inheritance tree for the attributes.

DustinCampbell:
To add tests:

1. Add a test project it test-assets/test-projects similar to https://github.com/OmniSharp/omnisharp-roslyn/tree/dev/test-assets/test-projects/XunitTestProject.
2. Add the test project to our build plan [here](https://github.com/OmniSharp/omnisharp-roslyn/blob/dev/build.json#L22-L25). This will ensure that the build script restores and builds the test project.
3. Add tests to [RunTestFacts.cs](https://github.com/OmniSharp/omnisharp-roslyn/blob/dev/tests/OmniSharp.DotNetTest.Tests/RunTestFacts.cs) and [GetTestStartInfoFacts.cs](https://github.com/OmniSharp/omnisharp-roslyn/blob/dev/tests/OmniSharp.DotNetTest.Tests/GetTestStartInfoFacts.cs).

src/OmniSharp.DotNetTest/TestFrameworks/MSTestFramework.cs
filipw: this is similar to the xunit code in OmniSharp not checking for `TheoryAttribute` -> because it inherits from `FactAttribute`

AbhitejJohn:
Thanks @DustinCampbell . Will update the PR soon. Would we need one for the legacy scenarios as well? I see a LegacyXunitTestProject & LegacyNunitTestProject .

DustinCampbell:
There's no need to add a legacy test project. Those are for older project.json. Those tests are just to ensure we don't break that older support. We never supported MSTest for project.json .NET Core projects (was it even possible?)

AbhitejJohn:
>We never supported MSTest for project.json .NET Core projects (was it even possible?) 

MSTest tests did work with project.json. There was a dotnet-test-mstest that drove those scenarios which leads me to guess that there are more changes required to enable that support in.

DustinCampbell:
If that's the case, this change will likely make MS Test work for project.json projects. It might be worth adding some legacy tests as well. There are `LegacyRunTestFacts` and `LegacyGetStartInfoFacts`.

AbhitejJohn:
@dotnet-bot test this please.

DustinCampbell:
@AbhitejJohn: Note that this repo doesn't use the @dotnet-bot build/test infrastructure.

AbhitejJohn:
@DustinCampbell : Oh ok. It looks like we have a problem with the Legacy workflow though - When we run the test we pass in a commandline similar to `dotnet.exe ..... --wait-command --test Main.Test.MainTest.Test`. dotnet-test-mstest however expects that when a `--wait-command` is passed in, the tests to be run would be sent via a `RunTestsMessage` message as [this](https://github.com/dotnet/cli/blob/master/Documentation/manpages/sdk/dotnet-test.1) doc details. 
>\-\-wait\-command: Indicates that the runner should wait to receive commands through the TCP channel instead of running tests right away. We use this to get around the command line size limit.

This step however seems to be missing in the vs code implementation which expects the adapter to pick up the tests from the commandline arguments itself. That seems to be confusing dotnet-test-mstest which defaults to running all the tests in the source instead. Should we change this workflow to reflect the above instead - either remove `--wait-command` or send out the filtered tests via a message to the adapters? The VS 2015 workflow follows this sequence of steps and things do work out well there.

DustinCampbell:
@AbhitejJohn: Yeah, I don't think the project.json support was ever truly completed. It was gotten to a point of "good enough" and somewhat abandoned when the developer who worked on it moved on. While adding the new vstest support, I tried to just preserve the legacy support since there will be fewer and fewer people using it.

I'd recommend removing `--wait-command`. Note that the legacy support already trims off several other arguments: https://github.com/OmniSharp/omnisharp-roslyn/blob/dev/src/OmniSharp.DotNetTest/Legacy/LegacyTestManager.cs#L160-L164.

AbhitejJohn:
@DustinCampbell : Fixed in the latest update.

DustinCampbell:
Thanks @AbhitejJohn! Looks good! Do you have any other changes? If not, I'll merge as soon as CI completes.

AbhitejJohn:
@DustinCampbell : No, I'm done.

PR #857, opened 2017-05-15
Fix Path.Combine exception in MSBuildProjectSystem.GetSdksPath

nanoant:
This fixes regression introduced in 2fbcc854f, where null info.BasePath
is passed to Path.Combine causing exception.

### DIFF
--- a/src/OmniSharp.MSBuild/MSBuildProjectSystem.cs
+++ b/src/OmniSharp.MSBuild/MSBuildProjectSystem.cs
@@ -308,7 +308,7 @@ namespace OmniSharp.MSBuild
         {
             var info = _dotNetCli.GetInfo(Path.GetDirectoryName(projectFilePath));
 
-            if (info.IsEmpty && !string.IsNullOrWhiteSpace(info.BasePath))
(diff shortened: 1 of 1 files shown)


src/OmniSharp.MSBuild/MSBuildProjectSystem.cs
```diff
@@ -308,7 +308,7 @@ private string GetSdksPath(string projectFilePath)
         {
             var info = _dotNetCli.GetInfo(Path.GetDirectoryName(projectFilePath));
 
-            if (info.IsEmpty && !string.IsNullOrWhiteSpace(info.BasePath))
+            if (info.IsEmpty && string.IsNullOrWhiteSpace(info.BasePath))
```
DustinCampbell: I think you want `||` here. Note that `info.BasePath` will always be null if `info.IsEmpty` is true (see [`DotNetInfo.cs`](https://github.com/OmniSharp/omnisharp-roslyn/blob/dev/src/OmniSharp.Abstractions/Services/DotNetInfo.cs))

nanoant:
@DustinCampbell applied your remark and changed '&&' into '||' in new PR version.

Issue #202, opened 2015-05-01
MSBuild support


davidfowl:
Our msbuild project file support is shaky at best, This is an item to keep track of the pending work and challenges involved.

The problems:
- Differences between xbuild and msbuild (maybe we can get the oss msbuild building and running on mono to replace xbuild)
- Using the "right" version of msbuild
- Modifying the project file in a way that round trips with VS
  - Adding references

Hard/Impossible to fix:
- SDKs that aren't installed
- Customizations made to project files that require context (properties). Maybe we can solve this with configuration?

Brain dump from conversation with msbuild expert:
- VS 2012 and earlier (4.X): Uses the project ToolsVersion unless that ToolsVersion does not exist, in which case it falls back to 4.0
- VS 2013 and later (12.0+): Uses the "current" ToolsVersion (same as the engine, so 12.0 on msbuild 12, 14.0 on msbuild 14)
- The engine and tasks both have to run against the same version of the framework (*Microsoft.Build.Framework.dll, that is')
- The Tasks and Utilities assemblies are versioned in the file name, but that doesn't really matter as long as the version of M.B.Framework matches

AndersNS:
Detecting the correct target platform needs to be done too. Omnisharp-roslyn seems to default to x86 for me. I worked around this by adding `{ "Platform", "AnyCPU"}` the properties dictionary used by omnisharp-roslyn when creating a ProjectCollection. 

Without this I got this error:

> [ERROR:OmniSharp.MSBuild.MSBuildProjectSystem] The OutputPath property is not set for ? project 'NameSpace.CoolProject.csproj'.  Please check to make sure that you have specified a valid combination of Configuration and Platform for this project.  Configuration='Debug'  Platform='x86'.

nanoant:
I was today struggling with getting our MSBuild (VS 2015) projects to work with OmniSharp-roslyn, basically because it somehow defaults to `$(Configuration)=Debug` and `$(Platform)=AnyCPU`, but such platform does not exist neither in solution or project.

I have managed to get rid of the error simply settings the platform in `ProjectFileInfo::Create` to `{ "Platform", "x64" }`, but this is really hackish way.

Is there any other, more elegant, probably via `config.json` way to set the platform?

DustinCampbell:
[...]

PR #858, opened 2017-05-15
Expose MsBuild Configuration and Platform setting

nanoant:
This allows choosing other (than default) MsBuild configuration, e.g.
`MsBuild:Configuration=Release` `MsBuild:Platform=x64`, especially usable
for projects that are not using default (AnyCPU) platforms.

This relates to #202.

P.S. Do we need to add some tests on that one? Or is there any other particular way to override Configuration/Platform planned?

### DIFF
--- a/src/OmniSharp.MSBuild/Options/MSBuildOptions.cs
+++ b/src/OmniSharp.MSBuild/Options/MSBuildOptions.cs
@@ -4,6 +4,8 @@ namespace OmniSharp.Options
     {
         public string ToolsVersion { get; set; }
         public string VisualStudioVersion { get; set; }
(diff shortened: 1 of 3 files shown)


DustinCampbell:
> Do we need to add some tests on that one? Or is there any other particular way to override Configuration/Platform planned?

We don't have great test infrastructure for this yet. It's pretty close, but would require a bit of refactoring to pass in custom options when creating an [OmniSharpTestHost](https://github.com/OmniSharp/omnisharp-roslyn/blob/dev/tests/TestUtility/OmniSharpTestHost.cs).

Eventually, I hope we can add an endpoint to support changing Configuration, Platform, and also TargetFramework.

PR #862, opened 2017-05-17
Clean up unresolved dependencies check

DustinCampbell:
Fixes https://github.com/OmniSharp/omnisharp-vscode/issues/1272

This change cleans up the logic (which had gotten kinda messy) that resolves MSBuild package references against a project's lock file. This also adds logging when package references can't be resolved, which should help diagnose situations where VS Code is getting restore prompts unnecessarily.

### DIFF
--- a/src/OmniSharp.DotNet/DotNetProjectSystem.cs
+++ b/src/OmniSharp.DotNet/DotNetProjectSystem.cs
@@ -302,29 +302,30 @@ namespace OmniSharp.DotNet
         {
             var libraryManager = state.ProjectContext.LibraryManager;
             var allDiagnostics = libraryManager.GetAllDiagnostics();
-            var unresolved = libraryManager.GetLibraries().Where(dep => !dep.Resolved);
-            var needRestore = allDiagnostics.Any(diag => diag.ErrorCode == ErrorCodes.NU1006) || unresolved.Any();
+            var unresolvedLibraries = libraryManager.GetLibraries().Where(dep => !dep.Resolved);
+            var needRestore = allDiagnostics.Any(diag => diag.ErrorCode == ErrorCodes.NU1006) || unresolvedLibraries.Any();
 
             if (needRestore)
             {
+                var unresolvedDependencies = unresolvedLibraries.Select(library =>
+                    new PackageDependency
+                    {
+                        Name = library.Identity.Name,
+                        Version = library.Identity.Version?.ToString()
+                    });
+
+                var projectFile = state.ProjectContext.ProjectFile;
+
                 if (allowRestore && _enableRestorePackages)
                 {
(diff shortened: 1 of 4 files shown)


src/OmniSharp.MSBuild/Resolution/PackageDependencyResolver.cs
```diff
@@ -0,0 +1,108 @@
...
+using System.Collections.Immutable;
+using System.IO;
+using System.Linq;
+using Microsoft.Extensions.Logging;
+using NuGet.ProjectModel;
+using OmniSharp.MSBuild.ProjectFile;
+
+namespace OmniSharp.MSBuild.Resolution
+{
+    internal class PackageDependencyResolver
```
david-driscoll: Is there much value in writing tests for this class?

src/OmniSharp.DotNet/DotNetProjectSystem.cs
```diff
@@ -302,29 +302,30 @@ private void UpdateUnresolvedDependencies(ProjectState state, bool allowRestore)
         {
             var libraryManager = state.ProjectContext.LibraryManager;
             var allDiagnostics = libraryManager.GetAllDiagnostics();
-            var unresolved = libraryManager.GetLibraries().Where(dep => !dep.Resolved);
-            var needRestore = allDiagnostics.Any(diag => diag.ErrorCode == ErrorCodes.NU1006) || unresolved.Any();
+            var unresolvedLibraries = libraryManager.GetLibraries().Where(dep => !dep.Resolved);
+            var needRestore = allDiagnostics.Any(diag => diag.ErrorCode == ErrorCodes.NU1006) || unresolvedLibraries.Any();
```
filipw: when lock file doesn't exist at all I believe the error code would be `NU1009`. 
Should that be taken into account as well? a restore would then generate the lock file

filipw: actually, maybe it doesn't matter anymore - it's legacy

DustinCampbell: Yeah, I wasn't trying to clean up `DotNetProjectSystem` too much. I've been thinking of that as a bit frozen so I don't unintentionally break anything.

src/OmniSharp.MSBuild/Resolution/PackageDependencyResolver.cs
DustinCampbell: I think there's definitely value in having tests, but I don't want to gate this PR on them. I want to write tests for the MSBuildProjectSystem in general because there's really not much today. Now that we have the test infrastructure in place, I'd like to write some broader MSBuildProjectSystem tests.

DustinCampbell: Added https://github.com/OmniSharp/omnisharp-roslyn/issues/863 to track adding tests.

PR #866, opened 2017-05-18
Error handling when loading assemblies from an external folder

filipw:
Follow up to #848 

Added try/catch to the directory enumeration inside `AssemblyLoader`. It can throw due to a number of problems - directory not found, path too long, lack of permissions etc.
We should gracefully exit with no assemblies loaded and not crash OmniSharp.

### DIFF
--- a/src/OmniSharp.Host/Services/AssemblyLoader.cs
+++ b/src/OmniSharp.Host/Services/AssemblyLoader.cs
@@ -38,6 +38,12 @@ namespace OmniSharp.Host.Loader
         {
             if (string.IsNullOrWhiteSpace(folderPath)) return Array.Empty<Assembly>();
 
+            if (!Directory.Exists(folderPath))
+            {
+                _logger.LogWarning($"Attempted to load assemblies from '{folderPath}' but that path doesn't exist.");
+                return Array.Empty<Assembly>();
+            }
+
             var assemblies = new List<Assembly>();
             foreach (var filePath in Directory.EnumerateFiles(folderPath, "*.dll"))
             {
--- a/src/OmniSharp.Host/Services/AssemblyLoader.cs
+++ b/src/OmniSharp.Host/Services/AssemblyLoader.cs
@@ -38,23 +38,26 @@ namespace OmniSharp.Host.Loader
         {
             if (string.IsNullOrWhiteSpace(folderPath)) return Array.Empty<Assembly>();
 
-            if (!Directory.Exists(folderPath))
+            try
             {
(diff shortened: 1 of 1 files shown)


src/OmniSharp.Host/Services/AssemblyLoader.cs
```diff
@@ -38,17 +38,26 @@ public IReadOnlyList<Assembly> LoadAllFrom(string folderPath)
...
+                    }
                 }
-            }
 
-            return assemblies;
+                return assemblies;
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError(ex, $"An error occurred when attempting to access '{folderPath}'.");
```
DustinCampbell: This message could be wrong if something else fails. I can't see what would fail though. :smile:

Is there any reason not to check `Directory.Exists(folderPath)`?

filipw: I had it in the initial commit (https://github.com/OmniSharp/omnisharp-roslyn/pull/866/commits/76906f90f7a2041504a54ffcb8e4f2cd774a59aa) but then I decided to go for try/catch instead as that would cover a lot more potential edge cases i.e. `PathTooLongException`, `SecurityException`, `UnauthorizedAccessException`, `IOException` rather than just `DirectoryNotFoundException`

DustinCampbell: Fair enough.

src/OmniSharp.Host/Services/AssemblyLoader.cs
```diff
@@ -38,17 +38,26 @@ public IReadOnlyList<Assembly> LoadAllFrom(string folderPath)
...
 
-            var assemblies = new List<Assembly>();
-            foreach (var filePath in Directory.EnumerateFiles(folderPath, "*.dll"))
+            try
             {
-                var assembly = LoadFromPath(filePath);
-                if (assembly != null)
+                var assemblies = new List<Assembly>();
+
+                foreach (var filePath in Directory.EnumerateFiles(folderPath, "*.dll"))
```
filipw: maybe this should be `SearchOption.AllDirectories`?

DustinCampbell: I'd rather that be passed in. Could be surprising behavior.

filipw: yeah maybe it's too much at all to even think about it. let's keep it simple for now and when/if more advanced use cases come up then we can decide what to do (i.e. allow this setting to be bound from config too and so on)

PR #868, opened 2017-05-18
Don't set completion text to full signature unless its override or partial method completion

DustinCampbell:
This change fixes a regression where overloads would not be collapsed in the VS Code completion list. In addition, I added a couple of tweaks to ensure that we report the "EnumMember" and "Const" kinds.

### DIFF
--- a/src/OmniSharp.Roslyn.CSharp/Services/Intellisense/CompletionItemExtensions.cs
+++ b/src/OmniSharp.Roslyn.CSharp/Services/Intellisense/CompletionItemExtensions.cs
@@ -12,15 +12,17 @@ namespace OmniSharp.Roslyn.CSharp.Services.Intellisense
 {
     internal static class CompletionItemExtensions
     {
-        private const string GetSymbolsAsync = "GetSymbolsAsync";
+        private const string GetSymbolsAsync = nameof(GetSymbolsAsync);
+        private const string InsertionText = nameof(InsertionText);
+        private const string NamedParameterCompletionProvider = "Microsoft.CodeAnalysis.CSharp.Completion.Providers.NamedParameterCompletionProvider";
         private const string OverrideCompletionProvider = "Microsoft.CodeAnalysis.CSharp.Completion.Providers.OverrideCompletionProvider";
         private const string ParitalMethodCompletionProvider = "Microsoft.CodeAnalysis.CSharp.Completion.Providers.PartialMethodCompletionProvider";
-        private const string Provider = "Provider";
+        private const string Provider = nameof(Provider);
         private const string SymbolCompletionItem = "Microsoft.CodeAnalysis.Completion.Providers.SymbolCompletionItem";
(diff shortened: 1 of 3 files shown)


src/OmniSharp.Roslyn.CSharp/Services/Intellisense/CompletionItemExtensions.cs
```diff
@@ -13,24 +12,36 @@ namespace OmniSharp.Roslyn.CSharp.Services.Intellisense
 {
     internal static class CompletionItemExtensions
     {
+        private const string GetSymbolsAsync = "GetSymbolsAsync";
```
david-driscoll: Use named?

src/OmniSharp.Roslyn.CSharp/Services/Intellisense/CompletionItemExtensions.cs
```diff
@@ -13,24 +12,36 @@ namespace OmniSharp.Roslyn.CSharp.Services.Intellisense
...
+        private const string OverrideCompletionProvider = "Microsoft.CodeAnalysis.CSharp.Completion.Providers.OverrideCompletionProvider";
+        private const string ParitalMethodCompletionProvider = "Microsoft.CodeAnalysis.CSharp.Completion.Providers.PartialMethodCompletionProvider";
+        private const string Provider = "Provider";
+        private const string SymbolCompletionItem = "Microsoft.CodeAnalysis.Completion.Providers.SymbolCompletionItem";
+        private const string SymbolCompletionProvider = "Microsoft.CodeAnalysis.CSharp.Completion.Providers.SymbolCompletionProvider";
+        private const string SymbolKind = "SymbolKind";
```
david-driscoll: Use nameof?

src/OmniSharp.Roslyn.CSharp/Services/Intellisense/CompletionItemExtensions.cs
DustinCampbell: thanks!

src/OmniSharp.Roslyn.CSharp/Services/Intellisense/CompletionItemExtensions.cs
DustinCampbell: thanks!

DustinCampbell:
Note: This isn't quite right. Some tests failing.

filipw:
thanks!

filipw:
aaaaah mystery solved!

I never saw this on the original PR because I tested against the version of C# for VS Code which had the magic completion Regex still present (which we remved here https://github.com/OmniSharp/omnisharp-vscode/commit/01aa52e197d37e6832006376bf8da96eb1bb5591). Turns out the purpose of this was to collapse overloads together on the client side.

It absolutely makes sense to do this on the server side correctly instead - thanks @DustinCampbell for fixing 😅

PR #870, opened 2017-05-21
Use SymbolFinder.FindDerivedClassesAsync inside "find implementations" endpoint

filipw:
Replace the hand-rolled code used to find derived named type symbols with [SymbolFinder.FindDerivedClassesAsync](http://source.roslyn.io/#Microsoft.CodeAnalysis.Workspaces/FindSymbols/SymbolFinder_Hierarchy.cs,dbb07fa6e6e5a08c). That API didn't exist at the time of writing that feature - it was only added, I believe, for VS15.

Also enabled the find implementations code to run on CSX files.

### DIFF
--- a/src/OmniSharp.Roslyn.CSharp/Helpers/QuickFixExtensions.cs
+++ b/src/OmniSharp.Roslyn.CSharp/Helpers/QuickFixExtensions.cs
@@ -7,23 +7,23 @@ namespace OmniSharp.Helpers
 {
     internal static class QuickFixExtensions
     {
-        internal static async Task AddQuickFix(this ICollection<QuickFix> quickFixes, OmniSharpWorkspace workspace, Location location)
-        {
-            if (location.IsInSource)
-            {
-                var quickFix = await QuickFixHelper.GetQuickFix(workspace, location);
-                quickFixes.Add(quickFix);
-            }
-        }
-
-        internal static async Task AddQuickFixes(this ICollection<QuickFix> quickFixes, OmniSharpWorkspace workspace, IEnumerable<ISymbol> symbols)
+        internal static void AddRange(this ICollection<QuickFix> quickFixes, IEnumerable<ISymbol> symbols, OmniSharpWorkspace workspace)
         {
             foreach (var symbol in symbols)
             {
                 foreach (var location in symbol.Locations)
                 {
-                    await AddQuickFix(quickFixes, workspace, location);
+                    quickFixes.Add(location, workspace);
                 }
+            }
+        }
+
+        internal static void Add(this ICollection<QuickFix> quickFixes, Location location, OmniSharpWorkspace workspace)
+        {
+            if (location.IsInSource)
+            {
+                var quickFix = location.GetQuickFix(workspace);
+                quickFixes.Add(quickFix);
             }
         }
     }
--- a/src/OmniSharp.Roslyn.CSharp/Helpers/QuickFixExtensions.cs
+++ b/src/OmniSharp.Roslyn.CSharp/Helpers/QuickFixExtensions.cs
@@ -1,0 +1,30 @@
+﻿using System.Collections.Generic;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
+using OmniSharp.Models;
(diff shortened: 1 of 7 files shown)


src/OmniSharp.Roslyn.CSharp/Helpers/QuickFixExtensions.cs
```diff
@@ -0,0 +1,30 @@
...
+
+namespace OmniSharp.Helpers
+{
+    internal static class QuickFixExtensions
+    {
+        internal static async Task AddQuickFix(this ICollection<QuickFix> quickFixes, OmniSharpWorkspace workspace, Location location)
+        {
+            if (location.IsInSource)
+            {
+                var quickFix = await QuickFixHelper.GetQuickFix(workspace, location);
```
DustinCampbell: I hadn't looked at this code before, so forgive me if I'm creating work for you. :smile: However, none of these helpers should be async. Ultimately, these call into `QuickFixHelper`, which is async just to ask for the tree and then the `SourceText`. This is all remarkably inefficient.

Instead of passing in `OmniSharpWorkspace` could these just take `SourceText`? After all, we already have (or nearly have) a `SourceText` in all of the places that call into `QuickFixHelper.GetQuickFix(...)`.

src/OmniSharp.Roslyn.CSharp/Helpers/QuickFixExtensions.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System.Collections.Generic;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
+using OmniSharp.Models;
+
+namespace OmniSharp.Helpers
+{
+    internal static class QuickFixExtensions
+    {
+        internal static async Task AddQuickFix(this ICollection<QuickFix> quickFixes, OmniSharpWorkspace workspace, Location location)
```
DustinCampbell: You *could* just call this `Add` since it's going onto an `ICollection<QuickFix>`.

src/OmniSharp.Roslyn.CSharp/Helpers/QuickFixExtensions.cs
```diff
@@ -0,0 +1,30 @@
...
+        {
+            if (location.IsInSource)
+            {
+                var quickFix = await QuickFixHelper.GetQuickFix(workspace, location);
+                quickFixes.Add(quickFix);
+            }
+        }
+
+        internal static async Task AddQuickFixes(this ICollection<QuickFix> quickFixes, OmniSharpWorkspace workspace, IEnumerable<ISymbol> symbols)
```
DustinCampbell: If you make the rename above, this could be `AddRange`. Also, does it seem better to swap the last two parameters?

src/OmniSharp.Roslyn.CSharp/Helpers/QuickFixExtensions.cs
DustinCampbell: Also, does it seem better to swap the last two parameters? What do you think?

filipw:
thanks a lot for the comments! 
Would you mind having a look again?

I pushed a new commit now, where I renamed the extension methods to `Add` and `AddRange` (makes total sense), and removed all asynchronicity on all these quick fix related methods. Since we always construct the `QuickFix` from a `Location` instance, that old async call to get the syntax tree can really be simply replaced by the `.SourceTree` property of the `Location`. As such I renamed `QuickFixHelper` to a `LocationExtensions` as it makes more logical sense to me.

I left the workspace in there, because the helper looks up the document based on the location and attaches the project information to the `QuickFix` object. Since we don't have these documents or their `SourceText` earlier (these locations are results of `SymbolFinder` operations and can be scattered all over solution) I wasn't sure what we could replace it with to make it more efficient.

filipw:
@DustinCampbell I pushed one additional change to this PR, to include the original location of a non-abstract symbol in the result set. This is the same behavior as VS currently has and that didn't exist in OmniSharp (for example `go to implementation` on a variable should take you to the declaration of the variable).

PR #873, opened 2017-05-25
Ensure that the MSBuild project system adds projects with the .csproj extension, even if they have a different GUID

DustinCampbell:


src/OmniSharp.MSBuild/MSBuildProjectSystem.cs
```diff
@@ -160,7 +160,8 @@ private IEnumerable<string> GetProjectPathsFromSolution(string solutionFilePath)
                     continue;
                 }
 
-                if (project.Kind == ProjectKind.CSharpProject)
+                if (project.Kind == ProjectKind.CSharpProject ||
+                    Path.GetExtension(projectFilePath) == ".csproj")
```
DustinCampbell: Actually, this should be a case-insensitive match. Let me tweak that.

PR #879, opened 2017-05-31
Harden DotNetInfo a bit to handle situations when 'dotnet' doesn't launch properly or return data

DustinCampbell:
Fixes https://github.com/OmniSharp/omnisharp-vscode/issues/1532

### DIFF
--- a/src/OmniSharp.Abstractions/Services/DotNetInfo.cs
+++ b/src/OmniSharp.Abstractions/Services/DotNetInfo.cs
@@ -26,7 +26,10 @@ namespace OmniSharp.Services
         {
             IsEmpty = false;
 
-            Version = SemanticVersion.Parse(version);
+            Version = SemanticVersion.TryParse(version, out var value)
+                ? value
+                : null;
+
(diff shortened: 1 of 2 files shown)


src/OmniSharp.Abstractions/Services/DotNetInfo.cs
```diff
@@ -26,7 +26,10 @@ private DotNetInfo(string version, string osName, string osVersion, string osPla
         {
             IsEmpty = false;
 
-            Version = SemanticVersion.Parse(version);
+            Version = SemanticVersion.TryParse(version, out var value)
+                ? value
+                : null;
```
david-driscoll: Should `Version` be null or `string.Empty`?

DustinCampbell: Unfortunately, `Version` is a `SemanticVersion`.

david-driscoll: doh I'm slow! lol

PR #882, opened 2017-06-02
Add IsSuggestionMode to AutoCompleteResponse for determining when the completion list should commit aggressively

DustinCampbell:
Related to https://github.com/OmniSharp/omnisharp-vscode/issues/1524


DustinCampbell:
The intention here is to able to determine when the completion list is in suggestion mode and shouldn't have define commit characters.

It's a bit heavy to add the values of `SuggestionModeItem` to every `AutoCompleteResponse` and I'm not sure how I'd even present that in VS Code. I'm working on a new '/completion' end point that we can use for future completion work that should marry a bit better with @david-driscoll's LSP work.

filipw:
👍👍

DustinCampbell:
Thanks!

PR #883, opened 2017-06-02
Support go to definition from metadata source

### LINKED ISSUE #876: Allow Go To Definition to work inside metadata files
At the moment we only support "go to definition" from:
 -  a source file into another source file 
 - a source file into metadata

However, we don't support "go to definition" from metadata to metadata. 
This is something that works fine in Visual Studio and it's not unreasonable to expect similar feature from OmniSharp.

I made a prototype of this yesterday and will PR soon.

filipw:
This is an attempt to fix #876 
I am not sure if this is the best way to do it, but I wanted to fit into the existing endpoint structure to avoid breaking changes.

### The current flow of the client trying to go to metadata is the following

1. Invoke `/gotodefinition`. The server code recognizes the symbol location is in metadata, and uses Roslyn's `MetadataAsSourceService` to generate a temporary document a provide basic metadata information to the caller - such as the position of the symbol in the metadata source. No metadata source is returned though.
2. Invoke `/metadata` and pass in the meta information (i.e. `TypeName`, `AssemblyName`). The server will then once again call into `MetadataAsSourceService`, generate a temporary document again, and return the source text of that document, along with its temporary name.
3. The client stitches these pieces of information together - displays the metadata source text, with its temporary name, and positions the user correctly based on the location obtained from the original `/gotodefinition` call.

It's impossible at the moment to continue the navigation from metadata to other metadata sources, because that would require going through `/gotodefinition` endpoint again, and it doesn't support starting points that are not a document inside the workspace.

### This PR changes the above flow in the following, non-breaking way

1. We still start by invoking `/gotodefinition`. The code recognizes the symbol location is in metadata, and uses Roslyn's `MetadataAsSourceService` to generate a temporary document a provide basic metadata information to the caller - such as the position of the symbol in the metadata source. No metadata source is returned though. The difference is that the temporary document document containing metadata source is stored in memory for future reuse (not added to current workspace, just stored on a side in a dictionary).
2. Invoke `/metadata` and pass in the meta information (i.e. `TypeName`, `AssemblyName`). The server now doesn't call into `MetadataAsSourceService` again (so we have a perf boost), but rather it will use the cached document and return the source text of that document, along with its temporary name.
3. The client stitches these pieces of information together - displays the metadata source text, with its temporary name, and positions the user correctly based on the location obtained from the original `/gotodefinition` call.
4. The client can now call `/gotodefinition` and ask for anything based on the location in the metadata source document - that is because the `/gotodefinition` endpoint has also been modified to not only work with documents that are part of Workspace but also with metadata sources using the aformentioned metadata cache to look up the sources. From there the process goes back to point 1, and we can have a neverending navigation cycle between metadata documents.

Let me know what you think, and whether this approach makes sense - we can absolutely change it to something better. 
I also specifically avoided storing the metadata source documents inside the workspace as that started impacting a lot of other endpoints in an unwelcome way (i.e. go to symbol).


src/OmniSharp.Roslyn.CSharp/Services/Navigation/GotoDefinitionService.cs
```diff
@@ -31,8 +31,12 @@ public GotoDefinitionService(OmniSharpWorkspace workspace, MetadataHelper metada
         {
             var quickFixes = new List<QuickFix>();
 
-            var document = _workspace.GetDocument(request.FileName);
+            var document = request.IsMetadataFile() ? 
+                _metadataHelper.FindDocumentInMetadataCache(request.FileName) : 
+                _workspace.GetDocument(request.FileName);
```
david-driscoll: This could technically work for other endpoints as well right?  Thinking of `/findusages`, `/typelookup` and maybe even `/highlight`.

Would be great if we could simplify this.

david-driscoll: We could probably just make it `_metadataHelper.FindDocumentInMetadataCache(request.FileName) ?? _workspace.GetDocument(request.FileName);` to avoid the first if there.

src/OmniSharp.Roslyn/MetadataHelper.cs
```diff
@@ -43,36 +47,62 @@ public MetadataHelper(IAssemblyLoader loader)
...
+                }
+            }
+            else
+            {
+                // for regular projects we will use current project to store metadata
+                metadataProject = project;
+            }
+
+            Document metadataDocument = null;
+            if (!_metadataDocumentCache.TryGetValue(fileName, out metadataDocument))
```
david-driscoll: could use `out var metadataDocument`

src/OmniSharp.Roslyn/OmniSharpWorkspace.cs
```diff
@@ -127,6 +127,8 @@ public IEnumerable<Document> GetDocuments(string filePath)
 
         public Document GetDocument(string filePath)
         {
+            if (filePath == null) return null;
```
david-driscoll: Should probably use `string.IsNullOrEmpty()`

src/OmniSharp.Roslyn/MetadataHelper.cs
```diff
@@ -43,36 +47,62 @@ public MetadataHelper(IAssemblyLoader loader)
...
+            if (project.IsSubmission)
+            {
+                metadataProject = project.Solution.Projects.FirstOrDefault(x => x.Name == MetadataKey);
+                if (metadataProject == null)
+                {
+                    metadataProject = project.Solution.AddProject(MetadataKey, $"{MetadataKey}.dll", LanguageNames.CSharp)
                     .WithCompilationOptions(project.CompilationOptions)
-                    .WithMetadataReferences(project.MetadataReferences)
-                : project;
+                    .WithMetadataReferences(project.MetadataReferences);
```
DustinCampbell: Should this be indented?

filipw:
thanks for the comments 👍 and sorry for the delay (long weekend) 🌴
I'll make the changes you mentioned

src/OmniSharp.Roslyn.CSharp/Services/Navigation/GotoDefinitionService.cs
filipw: yes this approach should work for other endpoints too

tests/OmniSharp.Roslyn.CSharp.Tests/FindReferencesFacts.cs
```diff
@@ -228,6 +230,62 @@ public class Foo {
             Assert.Equal("a.cs", usages.QuickFixes.ElementAt(0).FileName);
         }
 
+        [Fact]
+        public async Task CanFindReferencesFromMetadata()
```
filipw: that's a good point - I have applied this change to the other test. This one I removed completely cause it was a left over from when I wanted to include other features beyond Go To Definition in this PR (they'll be handled by separate PRs).

filipw:
I addressed the comments, squashed the commits to remove the noise I managed to generate and rebased this on top of the latest `dev`.

DustinCampbell:
Are you good here @filipw? If so, I think you can merge this once CI completes again.

filipw:
I think so!

PR #884, opened 2017-06-02
WIP: New completion service

DustinCampbell:
This change adds new '/completion' and '/completionItem/resolve' end points for better exposing the Roslyn completion service. In order to make this work, I finally had to make the '/open' and '/close' end points work as well. In order for the workspace to properly handle open/closed states for documents, I added a SourceText wrapper for OmniSharp that allows us to control our own SourceTextContainer instances. FWIW, getting a workspace text layer right is one of the more complex parts of the Roslyn workspace design. I have a lot more I want to do here, such as take advantage of the [VS text layer](https://www.nuget.org/packages?q=Microsoft.VisualStudio.Text).

cc @Pilchie

TODO:
- [x] Return commit characters from the '/completion' end point
- [x] Return commit character modification rules for items
- [ ] Return optional caret position
- [x] Tests


filipw:
awesome! 

I have two questions:
 - at the moment the existing autocomplete service does some custom sorting at the end, before responding (by exact match, by ignore case match, by subsequence match etc), how does the completion service sort?
 - is XML documentation included here?

DustinCampbell:
> at the moment the existing autocomplete service does some custom sorting at the end, before responding (by exact match, by ignore case match, by subsequence match etc), how does the completion service sort?

The completion service returns the list already sorted as it appears in VS. Note that the list returned by the `/completion` end point is not filtered by the word that's already typed as the `/autocomplete` end point. Instead, the expectation is that the filtering happens in the client (which is what VS Code does).

> is XML documentation included here?

Yes, XML documentation is returned by this list, though it is already processed. See the `Description_has_minimally_qualified_crefs` test I added.

src/OmniSharp.Roslyn.CSharp/Services/IntelliSense/V2/CompletionService.cs
```diff
@@ -72,11 +76,114 @@ public CompletionService(OmniSharpWorkspace workspace, ILoggerFactory loggerFact
...
+                throw new ArgumentException($"Cannot resolve completion item from '{fileName}' because the last {OmniSharpEndpoints.V2.Completion} request was for '{previousFileName}'");
+            }
+
+            if (itemIndex < 0 || itemIndex >= previousCompletionList.Items.Length)
+            {
+                throw new ArgumentOutOfRangeException($"Invalid item index: {itemIndex}. Should be within range 0 to {previousCompletionList.Items.Length}");
+            }
+
+            var previousItem = previousCompletionList.Items[itemIndex];
+            if (!string.Equals(previousItem.DisplayText, displayText))
```
Pilchie: The doc comment makes it sound like setting `displayText` is optional, but this will always fail

src/OmniSharp.Roslyn.CSharp/Services/IntelliSense/V2/CompletionService.cs
```diff
@@ -2,68 +2,72 @@
...
+                return CompletionModels.CompletionResponse.Empty;
             }
 
             var completionList = await service.GetCompletionsAsync(document, position, trigger);
+            if (completionList == null)
+            {
+                return CompletionModels.CompletionResponse.Empty;
+            }
 
             var isSuggestionMode = completionList.SuggestionModeItem != null;
```
Pilchie: This loses the placeholder text for the item, and I don't think there is a way to resolve the description text for the builder anymore.

src/OmniSharp.Roslyn.CSharp/Services/IntelliSense/V2/CompletionService.cs
```diff
@@ -72,11 +76,114 @@ public CompletionService(OmniSharpWorkspace workspace, ILoggerFactory loggerFact
...
+                Line = startTextLine.LineNumber,
+                Column = change.TextChange.Span.Start - startTextLine.Start
+            };
+
+            var endTextLine = text.Lines.GetLineFromPosition(change.TextChange.Span.End);
+            var endPoint = new Models.Point
+            {
+                Line = endTextLine.LineNumber,
+                Column = change.TextChange.Span.End - endTextLine.Start
+            };
```
Pilchie: A helper to go from a TextChange to a Range seems useful

Pilchie:
This looks awesome!  Also @cyrusnajmabadi might want to look.

src/OmniSharp.Roslyn.CSharp/Services/IntelliSense/V2/CompletionService.cs
DustinCampbell: Will do. We may have one and I just didn't spot it.

src/OmniSharp.Roslyn.CSharp/Services/IntelliSense/V2/CompletionService.cs
DustinCampbell: You're right. I had intended it to be optional. Thanks!

src/OmniSharp.Roslyn.CSharp/Services/IntelliSense/V2/CompletionService.cs
DustinCampbell: Yup. There's no notion of a "builder" in any of the editors that use OmniSharp. That's very much a VS concept. Ultimately, this is just used on the [client side](https://github.com/DustinCampbell/omnisharp-vscode/blob/7d5bb940206baedc90275c4204332e6619190a69/src/features/completionItemProvider.ts#L48) to handle commit characters (if the client has that concept). If OmniSharp supports a client that has a "builder", this end point could easily be augmented by exploding the properties of the 'SuggestionModeItem'.

DustinCampbell:
@ruant : It's not dead yet, but I have some tweaks that I need to make before it can be merged. Unfortunately, this is stacked up behind a few other high priority work items.

filipw:
this is superseded by https://github.com/OmniSharp/omnisharp-roslyn/pull/1877

PR #887, opened 2017-06-07
Added test for loading external code actions

filipw:
This is a follow up to https://github.com/OmniSharp/omnisharp-roslyn/pull/848 which didn't have tests in it.

This PR:
 - contains a unit test for loading an external refactoring (it's an `Add ConfigureAwait(false)` refactoring)
 - introduces a `binaries` folder in the `test-assets`, which is exposed to test infrastructure. The folder contains a DLL with the aforementioned refactoring.
 - extends the test infrastructure to allow passing in custom configuration, which is then translated to `OmnisharpOptions` - this will be helpful in testing different states of the configuration


tests/TestUtility/TestAssets.cs
```diff
@@ -12,13 +12,16 @@ public partial class TestAssets
         public string OmniSharpSolutionPath { get; }
         public string TestAssetsFolder { get; }
         public string TestProjectsFolder { get; }
+        public string TestBinariesFolder { get; }
+
```
DustinCampbell: extra blank line

PR #894, opened 2017-06-15
Minor typo in ProjectFileInfo logging message

DavidMoore:
Minor typo 'reference' misspelled as 'refernce'

### DIFF
--- a/src/OmniSharp.MSBuild/ProjectFile/ProjectFileInfo.cs
+++ b/src/OmniSharp.MSBuild/ProjectFile/ProjectFileInfo.cs
@@ -148,7 +148,7 @@ namespace OmniSharp.MSBuild.ProjectFile
                                 Version version;
                                 if (FrameworkReferenceResolver.Default.TryGetAssembly(referenceName, framework, out path, out version))
                                 {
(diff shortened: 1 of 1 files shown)


DustinCampbell:
I didn't see that you submitted this against master. We actually work out of the dev branch and master is very much out-of-date. The message that you're fixing doesn't even exist in dev any longer. The reason the CI is failing is because master doesn't have an update that changes the name of an xUnit binary that we copy when running tests.

@david-driscoll: Is there any reason not to merge dev into master and be done with it? Working out of dev doesn't seem to be adding much value and it seems that master is actually broken at the moment because of the xUnit issue I mentioned above.

filipw:
If I recall correctly dev was introduced when .NET CLI was extremely unstable as a way to have a "prerelease" OmniSharp.
I agree, at this point, we should really just come back to a single branch again - it would make everything simpler.

david-driscoll:
Yeah, there's no reason for the existence anymore, it was born out of a stability issue, we can move back to one branch safely.

DustinCampbell:
OK. We've merged the dev branch into master. This typo, and the entire message, has gone away. Closing this PR...

Issue #863, opened 2017-05-17
Write tests for MSBuildProjectSystem


DustinCampbell:
Now that we have the `OmniSharpTestHost` infrastructure, it should be much easier to write tests that exercise MSBuildProjectSystem:

* Loading various .NET Core and .NET Framework project types
* Handling unresolved package references
* Etc.

PR #897, opened 2017-06-26
Move (back) over to MSBuild solution parsing API

DustinCampbell:
Fixes https://github.com/OmniSharp/omnisharp-vscode/issues/1580

This also adds initial tests for #863.


filipw:
great - out of curiosity, what was the original reason for the custom parsing code?

DustinCampbell:
There are a couple of things:

1. Before, we were using the "project type GUID" to detect Unity projects and such. But I've now handled that through other means. Checking for ".csproj" is good enough.

2. When I originally tried to use MSBuild's API, I tried to use the "AbsolutePath" property for each project and found that had a bug with non-Windows slashes. Sticking with the "RelativePath" found in the solution file and keeping our existing logic for converting slashes avoids that bug.

filipw:
thanks

DustinCampbell:
Thanks guys!

PR #898, opened 2017-06-27
Improved the logic of adding default assembly references for Desktop CLR CSX scripts

filipw:
This PR improves the way we are adding default assembly references for Desktop CLR CSX scripts. We need to make sure we add references to all assemblies type forwarders might point to when working with a Desktop CLR script (for .NET Core scripts, all references should be provided via the project model).

The main difference was so far we references `mscorlib.dll` and `System.Core.dll` but missed `System.dll`.

Fixes https://github.com/OmniSharp/omnisharp-vscode/issues/1581


src/OmniSharp.Script/ScriptProjectSystem.cs
```diff
@@ -139,6 +127,45 @@ public void Initalize(IConfiguration configuration)
...
+                    .Select(a => a.Location)
+                    .Distinct()
+                    .Select(l => _metadataFileReferenceCache.GetMetadataReference(l));
+
+                foreach (var reference in references)
+                {
+                    commonReferences.Add(reference);
+                }
+
+                Assembly FromName(string assemblyName)
```
DustinCampbell: Why not use an `IAssemblyLoader`?

filipw: and here I was, so happy that I used a local function... 😋
you are right though

I also noticed it needs a rebase

filipw:
> Is there any point of adding any unit tests at all?

Good point.
I'll check - let's hold off on the merge.

I think we should be able to have a test based on using the `Regex` type (https://github.com/OmniSharp/omnisharp-vscode/issues/1581) that would fail on the old OmniSharp code and would work here.

filipw:
Actually the test infrastructure has it's own logic for default references, that's why this issue never showed up in tests before 
https://github.com/OmniSharp/omnisharp-roslyn/blob/dev/tests/TestUtility/TestHelpers.cs#L26

I think this should be good to merge for now then.
In the future we can eliminate that little duplication of default references between the test infrastructure with the project system.

david-driscoll:
Sounds good to me.  @DustinCampbell ?

DustinCampbell:
Fine by me!

filipw:
everything is green. I think this is good to merge - no reason to keep it hanging around anymore

PR #899, opened 2017-06-28
Rework code action service

DustinCampbell:
This change reworks how we compute code fixes and refactorings. Much of this is just basic refactoring and clean up, but there are substantial changes to to code fixes are computed. Now, we'll return code fixes when there are multiple diagnostics with overlapping spans. For example, consider this scenario:

```C#
class C
{
    void M()
    {
        StringBuilder
    }
}
```

There are three diagnostics at the position immediately following `StringBuilder`:

1. CS1001: Identifier expected
2. CS1002: ; expected
3. CS0246: The type or name 'StringBuilder' could not be found

The problem is that the first two diagnostics have a zero-width span immediately following `StringBuilder`, but the last diagnostic has the same span as `StringBuilder`. Previously, if the code action service was called with the line and column after `StringBuilder`, only the first two diagnostics would be considered for fixes because their spans are an exact match. However, it's the last diagnostic that we really want because that's the one for which the 'add import' code fix is triggered by. This change ensures that we'll consider all relevant diagnostics when computing code fixes.

### DIFF
--- a/src/OmniSharp.Abstractions/Utilities/DictionaryExtensions.cs
+++ b/src/OmniSharp.Abstractions/Utilities/DictionaryExtensions.cs
@@ -1,0 +1,19 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace OmniSharp.Utilities
+{
+    internal static class DictionaryExtensions
+    {
+        public static TValue GetOrAdd<TKey, TValue>(this IDictionary<TKey, TValue> dictionary, TKey key, Func<TKey, TValue> valueGetter)
+        {
+            if (!dictionary.TryGetValue(key, out var value))
+            {
+                value = valueGetter(key);
+                dictionary.Add(key, value);
+            }
+
+            return value;
+        }
+    }
+}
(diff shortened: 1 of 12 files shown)


DustinCampbell:
> would also be cool to one day tap into Roslyn's code action ordering

Yup, I'm hoping to look at that soon.

DustinCampbell:
And FWIW, if I add extension ordering, it'd be temporary. That's really an implementation detail that should be handled by a proper service from Roslyn.

src/OmniSharp.Abstractions/Utilities/DictionaryExtensions.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace OmniSharp.Utilities
+{
+    internal static class DictionaryExtensions
+    {
+        public static TValue GetOrAdd<TKey, TValue>(this IDictionary<TKey, TValue> dictionary, TKey key, Func<TKey, TValue> valueGetter)
```
david-driscoll: someone add this to the BCL. 🔮

DustinCampbell: Right? :smile:

DustinCampbell:
Thanks guys!

PR #900, opened 2017-06-29
Update Roslyn dependency to 2.3.0-beta2

DustinCampbell:
This change also tweaks the BufferManager to fix some bugs that the latest Roslyn now exposes (an OmniSharp test started failing without this change).


filipw:
I would like to mention that I have a project running that uses Roslyn 3 nightlies, it might be the first open source project that does that 😀

DustinCampbell:
Very brave.

Issue #896, opened 2017-06-26
Improve the "scan-all" behavior of all project systems


filipw:
As reported on chat today - the [SearchOption.AllDirectories](https://github.com/OmniSharp/omnisharp-roslyn/blob/dev/src/OmniSharp.Script/ScriptProjectSystem.cs#L58) is a bit too much and can cause OmniSharp to load a long time.

filipw:
Actually, this is not just scripting - MSBuild project system does the same thing if it doesn't have SLN https://github.com/OmniSharp/omnisharp-roslyn/blob/beb61b8683c382c498c62556f8fa7ace43631cf7/src/OmniSharp.MSBuild/MSBuildProjectSystem.cs#L134

`project.json` project system tries to work around it by having a depth limit of 5 https://github.com/OmniSharp/omnisharp-roslyn/blob/a4dced28f5f3b38d806134e89a6896c9d8075997/src/OmniSharp.DotNet/Projects/ProjectSearcher.cs#L12

At the very least, in all cases, we could exclude the common nasty folder names like `node_modules`, `bower_components` or `.`-prefixed folders etc.

DustinCampbell:
I think there are a could of things to do here.

1. Allow the user to be able to specify folders in an omnisharp.json to exclude from the project search.
2. Don't wait until projects are synchronously loaded before OmniSharp is finished loading. The project search should continue after OmniSharp has finished its initial configuration.

FWIW, I actually dislike the depth limit because it's hard to tell why a project didn't load if it happens to be lower in the directory tree.

mholo65:
As reported [here](https://github.com/OmniSharp/omnisharp-roslyn/issues/926#issuecomment-380247367), Cake Project System also suffers from this.

Time to think of some default exclude patterns and create a helper for enumerating files? Maybe we can start from default [.gitignore`](https://github.com/github/gitignore/blob/master/VisualStudio.gitignore) that Github uses. Additionally, make the exclude patterns configurable via OmniSharp configuration system.

PR #902, opened 2017-06-30
added support for disabling any project system via configuration

filipw:
At the moment we had [logic in place](https://github.com/OmniSharp/omnisharp-roslyn/blob/dev/src/OmniSharp.DotNet/DotNetProjectSystem.cs#L105-L114) that allowed users to disable the "legacy" `project.json` project system - using a configuration flag from `omnisharp.json` or any other of the several configuration sources we have in place.

For consistency, this PR extends this concept to allow the `"enabled": false` flag to be used against MsBuild and Script project systems too. 
It can help for example performance, as the user can easily (even globally) switch off scanning of the file system done by these project systems.

I also moved out some of the logic from the `Startup` class because, well, SRP.

### DIFF
--- a/src/OmniSharp.DotNet/DotNetProjectSystem.cs
+++ b/src/OmniSharp.DotNet/DotNetProjectSystem.cs
@@ -102,17 +102,6 @@ namespace OmniSharp.DotNet
 
         public void Initalize(IConfiguration configuration)
         {
-            if (!bool.TryParse(configuration["enabled"], out var enabled))
-            {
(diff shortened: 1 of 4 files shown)


filipw:
This indirectly can help against https://github.com/OmniSharp/omnisharp-roslyn/issues/896 too

DustinCampbell:
I looks to me like this one is ready to go. Do you not have merge access @filipw?

filipw:
I do! 😊

Issue #904, opened 2017-07-04
Sdk element in MSBuild project causes InvalidProjectFileException on load


natemcmaster:
OmniSharp cannot parse projects that use the `<Sdk>` element under project. (I believe this is new in MSBuild 15.3)

**Repro**
```xml
<Project>
  <Sdk Name="Microsoft.NET.Sdk" />
  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>netcoreapp2.0</TargetFramework>
  </PropertyGroup>
</Project>
```

**Error**
```
Microsoft.Build.Exceptions.InvalidProjectFileException: The element <Sdk> beneath element <Project> is unrecognized.  /private/tmp/tst/tst.csproj
...

 at Microsoft.Build.Evaluation.ProjectCollection.LoadProject (System.String fileName, System.String toolsVersion) [0x00001] in <bdf083ca20c84af487947992fd9a4a29>:0 
  at Microsoft.Build.Evaluation.ProjectCollection.LoadProject (System.String fileName) [0x00001] in <bdf083ca20c84af487947992fd9a4a29>:0 
  at OmniSharp.MSBuild.ProjectFile.ProjectFileInfo.LoadProject (System.String filePath, System.String solutionDirectory, System.String sdksPath, Microsoft.Extensions.Logging.ILogger logger, OmniSharp.Options.MSBuildOptions options, System.Collections.Generic.ICollection`1[T] diagnostics, System.Collections.Immutable.ImmutableArray`1[System.String]& targetFrameworks) [0x00062] in <4547000d373140b29fc6ad48137693ba>:0 
  at OmniSharp.MSBuild.ProjectFile.ProjectFileInfo.Create (System.String filePath, System.String solutionDirectory, System.String sdksPath, Microsoft.Extensions.Logging.ILogger logger, OmniSharp.Options.MSBuildOptions options, System.Collections.Generic.ICollection`1[T] diagnostics) [0x0000a] in <4547000d373140b29fc6ad48137693ba>:0 
  at OmniSharp.MSBuild.MSBuildProjectSystem.LoadProject (System.String projectFilePath) [0x0003f] in <4547000d373140b29fc6ad48137693ba>:0 
```

**Details:**
macOS Sierra
mono: not installed
dotnet: 2.0.0-preview2-006497
VS Code: 1.13.1
C# extension: 1.12.0-beta1

DustinCampbell:
[...]

Issue #903, opened 2017-06-30
OmniSharp.MSBuild.ProjectFile.ProjectFileInfo - Specified cast is not valid.


DustinCampbell:
_From @igorkovalchuk on June 30, 2017 13:37_

## Environment data

Debian 9 "Stretch"

$ mono --version
Mono JIT compiler version 5.2.0.179 (tarball Fri Jun 16 15:46:58 UTC 2017)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
	TLS:           __thread
	SIGSEGV:       altstack
	Notifications: epoll
	Architecture:  amd64
	Disabled:      none
	Misc:          softdebug 
	LLVM:          supported, not enabled.
	GC:            sgen (concurrent by default)

VS Code version: 1.13.1

C# Extension version: 1.11.0

## Details

I am working with Xamarin.Forms projects in Linux environment, so It is not possible to reproduce the exact configuration easy (it will take a lot of time), so I publish this issue for your information. Maybe you know already the source of this error.

Generally, you may ignore this issue completely, because I am working tightly only with PCL projects and can use the MonoDevelop for the Droid project if it will be necessary, 
but maybe this issue also will appear for other users in the similar specific cases/projects further. So it is FYI.

## Steps to reproduce

- build the xamarin-android project on Linux (or try to use the existing jenkins [build](https://jenkins.mono-project.com/view/Xamarin.Android/job/xamarin-android-linux/lastSuccessfulBuild/Azure/) );
- create a simple Xamarin.Forms Android project with PCL using the Visual Studio 2017;
- try to build it on Linux from the command line (it is possible, I may provide detailed info if you will ask, because it needs some tweaks);
- open the project in Visual Studio Code (it needs some tweaks - I may provide details if you need);
- VS Code: you may work with PCL project(s), intellisense will work normally;
- VS Code: but the Droid project will show an error and intellisense will not work in the Droid project;

After the auto-update from C# extension 1.10.0 to version 1.11.0 I've got these error messages:

The ... task could not be instantiated from the assembly...

Please verify the task assembly has been built using the same version of the Microsoft.Build.Framework assembly as the one installed on your computer... Specified cast is not valid.

## Log

[fail]: OmniSharp.MSBuild.ProjectFile.ProjectFileInfo
[...]

PR #907, opened 2017-07-05
MSBuild Update

DustinCampbell:
This represents several updates for the OmniSharp MSBuild projects system:

1. Use the "Compile" target rather than the "ResolveReferences" target. This allows MSBuild to run more stuff while evaluating the project file, such as generated files (fixes https://github.com/OmniSharp/omnisharp-vscode/issues/1531). Note that this requires copying the C# build tasks into OmniSharp.
2. Update MSBuild to latest bits (fixes https://github.com/OmniSharp/omnisharp-roslyn/issues/904).
3. Binding redirects added for MSBuild (fixes https://github.com/OmniSharp/omnisharp-roslyn/issues/903).
4. ~~Support for testing projects with the latest .NET Core SDK.~~

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -194,9 +194,6 @@ Task("SetupMSBuild")
         DeleteFile(CombinePaths(folder, "Microsoft.VisualBasic.Core.targets"));
         DeleteFile(CombinePaths(folder, "VBCSCompiler.exe"));
         DeleteFile(CombinePaths(folder, "VBCSCompiler.exe.config"));
-        DeleteFile(CombinePaths(folder, "csi.exe"));
-        DeleteFile(CombinePaths(folder, "csi.exe.config"));
-        DeleteFile(CombinePaths(folder, "csi.rsp"));
         DeleteFile(CombinePaths(folder, "vbc.exe"));
         DeleteFile(CombinePaths(folder, "vbc.exe.config"));
         DeleteFile(CombinePaths(folder, "vbc.rsp"));
--- a/build.cake
+++ b/build.cake
@@ -30,6 +30,7 @@ public class BuildPlan
     public string DotNetChannel { get; set; }
     public string DotNetVersion { get; set; }
     public string LegacyDotNetVersion { get; set; }
+    public string FutureDotNetVersion { get; set; }
     public string DownloadURL { get; set; }
     public string MSBuildRuntimeForMono { get; set; }
     public string MSBuildLibForMono { get; set; }
@@ -38,6 +39,7 @@ public class BuildPlan
     public string[] TestProjects { get; set; }
     public string[] TestAssets { get; set; }
     public string[] LegacyTestAssets { get; set; }
+    public string[] FutureTestAssets { get; set; }
 
     private string currentRid;
     private string[] targetRids;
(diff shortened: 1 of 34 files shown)


build.json
```diff
@@ -1,8 +1,9 @@
 {
-  "DotNetInstallScriptURL": "https://raw.githubusercontent.com/dotnet/cli/rel/1.0.0/scripts/obtain",
+  "DotNetInstallScriptURL": "https://raw.githubusercontent.com/dotnet/cli/release/2.0.0/scripts/obtain",
   "DotNetChannel": "preview",
-  "DotNetVersion": "1.0.1",
+  "DotNetVersion": "1.0.4",
   "LegacyDotNetVersion": "1.0.0-preview2-1-003177",
+  "FutureDotNetVersion": "2.0.0-preview2-006497",
```
filipw: the range of supported versions is getting out of control 😀

build.cake
```diff
@@ -176,18 +178,31 @@ Task("SetupMSBuild")
...
-    CopyFile(csharpTargetsPath, CombinePaths(csharpTargetsNet46Folder, csharpTargetsName));
-    CopyFile(csharpTargetsPath, CombinePaths(csharpTargetsNetCoreAppFolder,csharpTargetsName));
+    // Delete unnecessary files
+    foreach (var folder in new[] { roslynNet46Folder, roslynNetCoreAppFolder })
+    {
+        DeleteFile(CombinePaths(folder, "Microsoft.CodeAnalysis.VisualBasic.dll"));
+        DeleteFile(CombinePaths(folder, "Microsoft.VisualBasic.Core.targets"));
+        DeleteFile(CombinePaths(folder, "VBCSCompiler.exe"));
+        DeleteFile(CombinePaths(folder, "VBCSCompiler.exe.config"));
+        DeleteFile(CombinePaths(folder, "csi.exe"));
```
filipw: it's not related to the PR, but it just occurred to me, can we keep these 3 `csi` files? it's only ~50kb I think

if they are there, they open a world of possibilities. we could add similar functionality into VS Code that already exists in VS - you can either execute CSX with a shortcut or highlight a piece of code and execute that. In VS it's done using the "C# Interactive Window" but csi.exe could be a simple replacement

filipw:
the test failures seem legit at the moment?

build.cake
DustinCampbell: You already have APIs to execute scripts though.

DustinCampbell:
Yeah, the test failures were me getting a bit greedy. I tried to add testing of .NET 2.0 SDKs at the very end , but it occurs to me that there's going to be problems with that until some upcoming Mono changes make their way in. I'll roll that bit back for now.

build.cake
filipw: you mean to have such APIs built into OmniSharp? yes that would be possible, I remember discussing this with Jason some time ago.
it could also be a third party extension, I built one like that [for scriptcs](https://marketplace.visualstudio.com/items?itemName=filipw.scriptcsRunner) long time ago

But the simple way would be to just have CSI.exe as part of the bundle. The main reason is there is no distribution channel for CSI.exe at the moment other than installing MSBuild tools on your machine (and no distribution channel for non-Windows platforms at all), which is quite annoying

Again, this should be discussed separately, as it doesn't impact this PR, but it would be great to be able to simply ship CSI.exe with OmniSharp so that as a user you don't need to install anything else if you i.e. just want to start a C# REPL or something like that. The C# extension in VS Code could just facilitate shelling scripts into CSI (again, to simulate a bit the experience you have in VS with C# Interactive Window) and give folks an exploratory playground for C# straight up.

DustinCampbell: No, I meant you have APIs to execute scripts from Roslyn.

I'm OK with leaving these. I'm just a little leery of supporting an execution path through "csi" here. Essentially, it would mean that *we're* supporting it. So long as we're willing to sign up for that... :smile:

DustinCampbell:
OK. CI passes now that I rolled back the 'future' testing stuff. @filipw: I went ahead and kept CSI.exe as well.

PR #911, opened 2017-07-13
Use p/invoke rather than ugly bash commands to locate Mono on Unix

DustinCampbell:


filipw:
all green after restart

PR #918, opened 2017-07-18
Fix solution parsing (again!)

DustinCampbell:
It turns out that my recent change to use the MSBuild solution parsing API (for a second time) has broken Unity solutions. In this particular case, the MSBuild API throws if a solution file contains multiple projects with the same name, which is how Unity solutions are generated. Thanks to @Leopotam for reporting this issue!

The MSBuild API is much more restrictive about the format of a solution file than Visual Studio. I suppose that makes sense given that MSBuild can fail fast if a solution file is invalid. However, an IDE must be more resilient against faults. So, I've written a brand new solution parser from scratch. It is heavily derived from the [parser in MSBuild](https://github.com/Microsoft/msbuild/blob/master/src/Build/Construction/Solution/SolutionFile.cs).

Fixes https://github.com/OmniSharp/omnisharp-vscode/issues/1645

### DIFF
--- a/src/OmniSharp.MSBuild/MSBuildProjectSystem.cs
+++ b/src/OmniSharp.MSBuild/MSBuildProjectSystem.cs
@@ -5,7 +5,6 @@ using System.Composition;
 using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
-using Microsoft.Build.Construction;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.Extensions.Configuration;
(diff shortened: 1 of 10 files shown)


DustinCampbell:
Note: I'm expecting AppVeyor to fail initially due to the outage their experiencing around NuGet (http://status.appveyor.com/incidents/m2vdvw39kdk8). I'm currently tracking the incident there and will re-kick the CI once it's resolved.

Also, I'm betting the OSX build will fail as it seems to time out a lot during US business hours. I'll kick it later if it does time out.

PR #920, opened 2017-07-20
Fix Mono path detection to actually find Mono on the path before calling real_path

DustinCampbell:


DustinCampbell:
Sigh.. kicking OSX CI again.

DustinCampbell:
FWIW, once I get the PR to switch to net46, CI will run faster and timeout less.

PR #924, opened 2017-07-23
Several updates to build

DustinCampbell:
This PR cherry-picks several changes from https://github.com/OmniSharp/omnisharp-roslyn/pull/915.

* We're moving to the very latest Mono. This is necessary to take advantage of the new `--assembly-loader` flag for the Mono runtime, which does a better job of simulating the .NET Framework assembly loading behavior on Mono. In addition, the msbuild that ships with Mono is much more mature, so we can call it directly.
* Install the .NET Core SDK during Travis runs. This allows the sanity runs of OmniSharp that occur at the end of the build to work properly, which has been broken since the Sdks were removed from OmniSharp.
* Platform detection code to determine whether we're on Windows, OSX, Linux, and whether the current platform is 32- or 64-bit. This is needed for pulling the correct Mono runtime.
* Validation step that tests the installed Mono to ensure it's the correct version.

I structured each commit to be digestible.


scripts/platform.cake
```diff
@@ -0,0 +1,73 @@
+#load "runhelpers.cake"
+
+using System;
+
+public class Platform
```
filipw: it's nice that we have robust platform information captured in one place. Maybe for consistency we can replace the calls to the built-in Cake `IsRunningOnWindows()` global methods used in several places with `platform.IsWindows`. This way we have a guarantee the two logics would not diverge

filipw:
it's a shame Cake is on a very old version of Roslyn (1.0.0-rc2). In the newer versions of Roslyn scripting, top level extension methods are supported in scripts which would be perfect for complex build script like this one

DustinCampbell:
> it's a shame Cake is on a very old version of Roslyn (1.0.0-rc2).

Agreed, though it's worse than that. AFAICT, Cake only seems to use Roslyn Scripting on Windows. It seems to use the Mono scripting dialect on OSX/Linux, which has very different semantics.

scripts/platform.cake
DustinCampbell: Good point. I'll do that.

mholo65:
@filipw @DustinCampbell we are updating Roslyn and unifying script engines in next release :wink: 

PR here https://github.com/cake-build/cake/pull/1645

PR #935, opened 2017-08-08
C# 7.1

filipw:
- added C# `7.1` and `Latest` to the MSBuild property converter
 - updated CSX from `LanguageVersion.Default` (which is C# 7) to `LanguageVersion.Latest` because we should be cutting edge there 😃 Jokes aside, there is no way in scripting to specify language version (since there is no project file) so it makes sense to be on the latest-greatest version


filipw:
thanks for quick merge!

PR #937, opened 2017-08-08
Msbuild project fixes

### LINKED ISSUE #936: .NET Core projects are not processed properly on Linux if Mono not installed
I introduced a bug in how the MSBuild 15.0 toolset gets set up when Mono is not installed.

DustinCampbell:
Fixes #936
Fixes https://github.com/OmniSharp/omnisharp-vscode/issues/1676

This PR addresses a few issues with using MSBuild for execution on Mono:

1. There was a bug introduced where we would set `MSBuildExtensionsPath` to `$mono_prefix/xbuild` in situations that caused things to break. OmniSharp will now only do that if `$mono_prefix/msbuild` and `$mono_prefix/xbuild/15.0` exist. If we set it when those paths don't exist, it's likely that we'll lose the targets to properly handle .NET Core projects.
2. The Mono targets have hardcoded a redirect of `CscToolPath` to a different location. Unfortunately, that breaks OmniSharp when we call the `Compile` target, since the `Csc` task requires a correct path to `csc.exe`. So, we override `CscToolsPath` and set it to point to our own copy of `csc.exe` (which is the same value as `RoslynTargetsPath`).
3. I encountered an issue where we would attempt to use an MSBuild Toolset, even if it pointed to a location that didn't exist on disk. We now try to figure out the correct toolset to use, and, if it doesn't exist, use the highest toolset version that *does* exist.

### DIFF
--- a/src/OmniSharp.MSBuild/MSBuildEnvironment.cs
+++ b/src/OmniSharp.MSBuild/MSBuildEnvironment.cs
@@ -25,6 +25,7 @@ namespace OmniSharp.MSBuild
         private static string s_msbuildExtensionsPath;
         private static string s_targetFrameworkRootPath;
         private static string s_roslynTargetsPath;
+        private static string s_cscToolPath;
 
         public static bool IsInitialized => s_isInitialized;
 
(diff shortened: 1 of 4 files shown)


david-driscoll:
Once travis completes. 😄

DustinCampbell:
Just awaiting the OSX build.

DustinCampbell:
All good.

PR #940, opened 2017-08-15
Adds LoadFrom to IAssemblyLoader

mholo65:
- Makes it possible to load single assembly from path.
- Make private method in `AssemblyLoader` public.
- Rename to `LoadFrom` to align with `LoadAllFrom`

### DIFF
--- a/src/OmniSharp.Host/Services/AssemblyLoader.cs
+++ b/src/OmniSharp.Host/Services/AssemblyLoader.cs
@@ -44,7 +44,7 @@ namespace OmniSharp.Host.Loader
 
                 foreach (var filePath in Directory.EnumerateFiles(folderPath, "*.dll"))
                 {
-                    var assembly = LoadFromPath(filePath);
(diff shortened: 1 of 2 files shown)


src/OmniSharp.Host/Services/AssemblyLoader.cs
```diff
@@ -60,7 +60,7 @@ public IReadOnlyList<Assembly> LoadAllFrom(string folderPath)
             }
         }
 
-        private Assembly LoadFromPath(string assemblyPath)
+        public Assembly LoadFrom(string assemblyPath)
         {
             Assembly assembly = null;
```
david-driscoll: I'm with @filipw for consistency lets add a guard `if (string.IsNullOrWhiteSpace(assemblyPath)) return null;`  One less exception to get thrown.

mholo65:
@david-driscoll @filipw Added check for null/whitespace path to avoid trowing exception.

PR #941, opened 2017-08-16
Don't crash when encountering external code action DLL with missing dependencies 

filipw:
Fixed a bug where an external code action DLL with missing dependencies would crash OmniSharp. This can happen when - for example - the refactoring DLL is built in a way that it takes dependency on Visual Studio DLLs or GAC DLLs that don't exist on the machine where OmniSharp runs.

If that happens, we will now just log a warning and continue, rather than crashing OmniSharp.

### DIFF
--- a/src/OmniSharp.Roslyn/HostServicesAggregator.cs
+++ b/src/OmniSharp.Roslyn/HostServicesAggregator.cs
@@ -43,11 +43,11 @@ namespace OmniSharp
                         builder.Add(assembly);
                         logger.LogTrace("Successfully added {assembly} to host service assemblies.", assembly.FullName);
                     }
-                    catch (Exception)
+                    catch (Exception ex)
                     {
                         // if we can't see exported types, it means that the assembly cannot participate
-                        // in MefHostServices as one or more of its dependencies (typically a Visual Studio or GACed DLL) is missing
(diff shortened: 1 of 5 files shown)


src/OmniSharp.Roslyn/HostServicesAggregator.cs
```diff
@@ -34,7 +37,18 @@ public class HostServicesAggregator
...
+                    {
+                        var exportedTypes = assembly.ExportedTypes;
+                        builder.Add(assembly);
+                        logger.LogTrace("Successfully added {assembly} to host service assemblies.", assembly.FullName);
+                    }
+                    catch (Exception)
+                    {
+                        // if we can't see exported types, it means that the assembly cannot participate
+                        // in MefHostServices as one or more of its dependencies (typically a Visual Studio or GACed DLL) is missing
+                        logger.LogWarning("Expected to use {assembly} in host services but the assembly cannot be used due to missing dependencies.", assembly.FullName);
```
DustinCampbell: This message could be misleading since you're just testing for Exception. I can imagine lots of other errors.

filipw: you are right - I have updated message, thanks

filipw:
Merging this, thanks

PR #949, opened 2017-08-22
Do not throw NullReferenceException when trying to "go to definition" on a namespace

filipw:
At the moment, when you try to go to definition on a namespace, we throw a NullReferenceException. 
OmniSharp doesn't crash but a nasty exception shows up in the logs.


DustinCampbell:
@filipw: Are there other changes you wanted to make, or should this be merged?

filipw:
no - this is ready to be merged, let's get it in

PR #962, opened 2017-09-18
fixed LanguageVersion assignment

### LINKED ISSUE #961: How to use C# 7.1 in omnisharp?
Things like default literal are already supported by compilers but omnisharp doesn't pick up the LangVersion csproj attribute and so roslyn gives errors for these features.
Unless I'm missing something and there is another way this should be fixed.

filipw:
we assigned the property to itself instead of the parameter 😄
Fixes #961

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # Changelog
 All changes to the project will be documented in this file.
 
+## [1.25.0] - not released yet
+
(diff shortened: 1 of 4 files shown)


DustinCampbell:
Travis keeps crapping out on the OSX build. I'm going to go ahead and merge this since the last commit passed CI.

NinoFloris:
Thanks for the quick turnaround! Hope to see it soon in vscode :)

PR #964, opened 2017-09-18
Fix typo in ProjectFileInfo that caused the sign key file not to propagate to the OmniSharpWorkspace

DustinCampbell:


filipw:
looks good - let's merge it!

PR #971, opened 2017-10-02
Added UseUrls method

Brumawen:
Configured Urls as per https://docs.microsoft.com/en-us/aspnet/core/fundamentals/hosting?tabs=aspnetcore2x

Fixed issue where the host URL was defaulting to 5000 regardless of the port number being passed in the parameters.


DustinCampbell:
I fixed the xUnit issues that caused the CI to break on this PR.

DustinCampbell:
Thanks for the contribution @Brumawen!

PR #972, opened 2017-10-02
Update to latest xUnit and fix up tests affected by analyzers

DustinCampbell:
This should fix the failed build in #971


DustinCampbell:
This should fix the failed build in #971

DustinCampbell:
This change is larger mechanical, since I only updated the xUnit version and used its code fixes to address diagnostics that the xUnit analyzers flagged. So, I'm going to go ahead and merge this so I can get #971 updated. Let me know if you have any concerns.

filipw:
looks good!

PR #974, opened 2017-10-03
Use MSBuild SDK Resolvers

DustinCampbell:
Currently, we launch `dotnet --info` for every MSBuild project. This allows us to detect the .NET Core SDK that is used for that project without having to replicate a lot of code to search for a `global.json`, or look in various places. Once we've have the output of `dotnet --info`, we look for the "Base Path" property and use that to decide where the location of the "Sdks" are within the relevant .NET Core SDK's directory. Once we've got that, we set the `MSBuildSDKsPath` environment variable to tell MSBuild where the SDKs are before we load the project.

This approach has several problems:

1. Setting an environment variable is fragile! It means that we can't load MSBuild projects in parallel because the projects might target a different .NET Core SDK. Also, another process (even another OmniSharp process) could change or unset it.
2. Parsing `dotnet --info` is fragile! The .NET Core SDK's informational output is subject to change, and there are proposals to do it.

To avoid this problem, OmniSharp will deploy the same "SDK Resolvers" that are deployed with Visual Studio. This is a bit of code that knows how to resolve a project to the correct .NET Core SDK. In fact, it uses the same native binary that "dotnet" does to find the SDK. Once present, MSBuild uses the resolver to locate the .NET Core SDK for any project that contains an `Sdk` attribute.

For tests, we still set the `MSBuildSDKsPath` environment variable so that we can use the .NET Core SDKs that the OmniSharp build scripts install locally during our test runs. We *could* write our own SDK Resolver to resolve our local .NET Core SDKs during testing, but that's probably overkill. If it becomes an issue, we can look into that.

### DIFF
--- a/NuGet.Config
+++ b/NuGet.Config
@@ -4,6 +4,6 @@
         <clear />
         <add key="NuGet" value="https://api.nuget.org/v3/index.json" />
         <add key="dotnet-core" value="https://dotnet.myget.org/F/dotnet-core/api/v3/index.json" />
-        <add key="cli-deps" value="https://dotnet.myget.org/F/cli-deps/api/v3/index.json" />
+        <add key="dotnet-cli" value="https://dotnet.myget.org/F/cli-deps/api/v3/index.json" />
     </packageSources>
 </configuration>
--- a/build.cake
+++ b/build.cake
@@ -218,7 +218,7 @@ Task("InstallMonoAssets")
     Run("chmod", $"+x '{CombinePaths(env.Folders.Mono, "run")}'");
 });
 
-void CopyDotNetHostResolver(string os, string arch, string hostFileName, string targetFolderBase)
+void CopyDotNetHostResolver(BuildEnvironment env, string os, string arch, string hostFileName, string targetFolderBase)
 {
     var source = CombinePaths(
         env.Folders.Tools,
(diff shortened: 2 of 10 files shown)


DustinCampbell:
I just tested on OSX when Mono is installed and when it isn't, and it's working after a small bug fix. I also updated the Mono runtime and MSBuild to newer builds.

This is good for review now.

DustinCampbell:
+ @livarcocc and @nguerrera in case they're interested.

build.cake
```diff
@@ -227,17 +248,43 @@ Task("CreateMSBuildFolder")
...
+        CopyDotNetHostResolver(env, "win", "x86", "hostfxr.dll", msbuildSdkResolverFolder, copyToArchSpecificFolder: true);
+        CopyDotNetHostResolver(env, "win", "x64", "hostfxr.dll", msbuildSdkResolverFolder, copyToArchSpecificFolder: true);
+    }
+    else if (Platform.Current.IsMacOS)
+    {
+        CopyDotNetHostResolver(env, "osx", "x64", "libhostfxr.dylib", msbuildSdkResolverFolder, copyToArchSpecificFolder: false);
+    }
+    else if (Platform.Current.IsLinux)
+    {
+        CopyDotNetHostResolver(env, "linux", "x64", "libhostfxr.so", msbuildSdkResolverFolder, copyToArchSpecificFolder: false);
```
nguerrera: Is the lack of linux-x86 no longer an issue for you?

DustinCampbell: I realized that it's kind of a non-issue since .NET Core isn't supported on linux-x86 anyway. Since I can't install a .NET Core SDK on linux-x86, I can't see a need to resolve to it on linux-x86. :smile:

DustinCampbell:
Thanks!

NuGet.Config
```diff
@@ -4,6 +4,6 @@
         <clear />
         <add key="NuGet" value="https://api.nuget.org/v3/index.json" />
         <add key="dotnet-core" value="https://dotnet.myget.org/F/dotnet-core/api/v3/index.json" />
-        <add key="cli-deps" value="https://dotnet.myget.org/F/cli-deps/api/v3/index.json" />
```
livarcocc: Awesome, we actually want to terminate the cli-deps feed, since it was actually built to contain all the packages the CLI depends on, but we have moved away from that approach since then.

DustinCampbell: Yeah, this was an old feed from long ago.

PR #978, opened 2017-10-09
Pin .NET Core SDK and fix Travis OSX build

DustinCampbell:


build.cake
```diff
@@ -163,7 +168,8 @@ Task("InstallDotNetCoreSdk")
         // Install legacy .NET Core SDK (used to 'dotnet restore' project.json test projects)
         InstallDotNetSdk(env, buildPlan,
             version: buildPlan.LegacyDotNetVersion,
-            installFolder: env.Folders.LegacyDotNetSdk);
+            installFolder: env.Folders.LegacyDotNetSdk,
+            noPath: true);
```
DustinCampbell: ℹ️ Our build script should be adding the .NET Core SDK that we built with to the path. However, we don't want the legacy .NET Core SDK we're using on the path.

DustinCampbell:
Once CI passes. I'm going to go ahead and merge this and update the other PRs that have failed because of this issue.

PR #982, opened 2017-10-18
Fix build on CentOS 7

omajid:
This fixes two issues building Omnisharp on CentOS 7:

System.Version needs a major and minor version or it refuses to parse it. For Linux distributions with a single major-only version in /etc/os-release, add a ".0" suffix to the version.

Add a documentation note about `msbuild` being a dependency as well since it is sometimes not included with the default mono installation.


omajid:
I am not sure if my patch has anything to do with the CI failures. Please let me know if I need to fix something.

DustinCampbell:
Nope. It's a different issue.

PR #987, opened 2017-10-24
Listen for file renames

rchande:
Use the improved file watching in https://github.com/OmniSharp/omnisharp-vscode/pull/1805 to better handle file renames.
When adding files to the workspace, we now set up file watching for each directory for which we add a document. If a file deletion comes in from such a directory, we'll remove any applicable documents from the workspace. If a .cs file is created, we kick off a task for the buffer change notification service to use its logic to add the file to applicable projects.

These changes allow users to rename files from the VS Code file tree without getting errors about duplicate type declarations. For projects with file globbing, the workspace will automatically wire up the new file and the user won't see any errors about missing types.


rchande:
Tag @DustinCampbell @filipw

rchande:
Fixes https://github.com/OmniSharp/omnisharp-vscode/issues/785

src/OmniSharp.Abstractions/Models/v1/FilesChanged/ChangeTypeConverter.cs
```diff
@@ -0,0 +1,25 @@
...
+namespace OmniSharp.Models.FilesChanged
+{
+    class FileChangeTypeConverter : JsonConverter
+    {
+        public override bool CanConvert(Type objectType) => objectType == typeof(FileChangeType);
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
+        {
+            switch (reader.Value as string)
+            {
+                case "Change":
```
DustinCampbell: These should probably be case-insensitive string matches.

src/OmniSharp.Abstractions/Models/v1/FilesChanged/FilesChangedRequest.cs
```diff
@@ -1,8 +1,13 @@
...
 using OmniSharp.Mef;
 
 namespace OmniSharp.Models.FilesChanged
 {
-    [OmniSharpEndpoint(OmniSharpEndpoints.FilesChanged, typeof(IEnumerable<Request>), typeof(FilesChangedResponse))]
-    public class FilesChangedRequest : IRequest { }
+    [OmniSharpEndpoint(OmniSharpEndpoints.FilesChanged, typeof(IEnumerable<FilesChangedRequest>), typeof(FilesChangedResponse))]
+    public class FilesChangedRequest : Request
+    {
+        [JsonConverter(typeof(FileChangeTypeConverter))]
```
david-driscoll: Couldn't this just be [`StringEnumConverter`](https://github.com/JamesNK/Newtonsoft.Json/blob/f012ba857f36fe75b1294a210b9104130a4db4d5/Src/Newtonsoft.Json/Converters/StringEnumConverter.cs)?

src/OmniSharp.Abstractions/FileWatching/IFileSystemWatcher.cs
```diff
@@ -1,12 +1,20 @@
...
-        void Watch(string path, Action<string> callback);
+        void Watch(string path, Action<string, FileChangeType?> callback);
 
-        void TriggerChange(string path);
+        /// <summary>
+        /// Called when a file is created, changed, or deleted.
+        /// </summary>
+        /// <param name="path">The path to the file</param>
+        /// <param name="verb">The type of change. Hosts are not required to pass a change type</param>
+        void TriggerChange(string path, FileChangeType? verb);
```
DustinCampbell: IMO, the terms `type` and `verb` should match. So, if the type is `FileChangeType`, the parameter here should be `type`. Or, the type should be `FileChangeVerb`.

src/OmniSharp.MSBuild/MSBuildProjectSystem.cs
```diff
@@ -400,6 +406,31 @@ private void UpdateSourceFiles(Project project, IList<string> sourceFiles)
             }
         }
 
+        private void WatchDirectoryContainingFile(string sourceFile) => _fileSystemWatcher.WatchDirectory(Path.GetDirectoryName(sourceFile), OnDirectoryFileChanged);
+
+        private void OnDirectoryFileChanged(string path, FileChangeType? changeType)
+        {
+            // Hosts may not have passed through a file change type
+            if (changeType == null && !File.Exists(path) || changeType.GetValueOrDefault() == FileChangeType.Delete)
```
david-driscoll: Minor nit it's more obvious that the value is Nullable<> if we use `!changeType.HasValue`.

src/OmniSharp.DotNet/DotNetProjectSystem.cs
```diff
@@ -166,13 +166,13 @@ private void UpdateProject(string projectDirectory)
             _projectStates.Update(projectDirectory, contexts, AddProject, RemoveProject);
 
             var projectFilePath = contexts.First().ProjectFile.ProjectFilePath;
-            _fileSystemWatcher.Watch(projectFilePath, file =>
+            _fileSystemWatcher.Watch(projectFilePath, (file, verb) =>
```
DustinCampbell: See my comment above about `type` vs. `verb`. If you go with `type`, be sure to update this as well.

src/OmniSharp.DotNet/DotNetProjectSystem.cs
```diff
@@ -166,13 +166,13 @@ private void UpdateProject(string projectDirectory)
...
             {
                 _logger.LogInformation($"Watcher: {file} updated.");
                 Update(allowRestore: true);
             });
 
-            _fileSystemWatcher.Watch(Path.ChangeExtension(projectFilePath, "lock.json"), file =>
+            _fileSystemWatcher.Watch(Path.ChangeExtension(projectFilePath, "lock.json"), (file, verb) =>
```
DustinCampbell: ditto

src/OmniSharp.Host/FileWatching/ManualFileSystemWatcher.cs
```diff
@@ -1,24 +1,37 @@
-﻿using System;
+﻿using OmniSharp.Models.FilesChanged;
+using System;
 using System.Collections.Generic;
+using System.IO;
```
DustinCampbell: Sort usings so `System` namespaces are first.

src/OmniSharp.Host/Services/WorkspaceHelper.cs
```diff
@@ -1,81 +0,0 @@
...
-using System.Composition.Hosting;
-using Microsoft.Extensions.Configuration;
-using Microsoft.Extensions.Logging;
-using OmniSharp.Options;
-using OmniSharp.Roslyn;
-using OmniSharp.Roslyn.Options;
-
-namespace OmniSharp.Services
-{
-    public class WorkspaceHelper
```
DustinCampbell: Good catch! This is old code.

src/OmniSharp.MSBuild/MSBuildProjectSystem.cs
```diff
@@ -232,14 +236,14 @@ private void WatchProject(ProjectFileInfo project)
         {
             // TODO: This needs some improvement. Currently, it tracks both deletions and changes
             // as "updates". We should properly remove projects that are deleted.
-            _fileSystemWatcher.Watch(project.FilePath, file =>
+            _fileSystemWatcher.Watch(project.FilePath, (file, verb) =>
```
DustinCampbell: Same comment about `type` vs. `verb`.

src/OmniSharp.MSBuild/MSBuildProjectSystem.cs
```diff
@@ -232,14 +236,14 @@ private void WatchProject(ProjectFileInfo project)
...
             {
                 OnProjectChanged(project.FilePath, allowAutoRestore: true);
             });
 
             if (!string.IsNullOrEmpty(project.ProjectAssetsFile))
             {
-                _fileSystemWatcher.Watch(project.ProjectAssetsFile, file =>
+                _fileSystemWatcher.Watch(project.ProjectAssetsFile, (file, verb) =>
```
DustinCampbell: Same thing here.

src/OmniSharp.MSBuild/MSBuildProjectSystem.cs
DustinCampbell: I don't think you need the `GetValueOrDefault()` here.

src/OmniSharp.MSBuild/MSBuildProjectSystem.cs
```diff
@@ -400,6 +406,31 @@ private void UpdateSourceFiles(Project project, IList<string> sourceFiles)
...
+            // Hosts may not have passed through a file change type
+            if (changeType == null && !File.Exists(path) || changeType.GetValueOrDefault() == FileChangeType.Delete)
+            {
+                foreach (var documentId in _workspace.CurrentSolution.GetDocumentIdsWithFilePath(path))
+                {
+                    _workspace.RemoveDocument(documentId);
+                }
+            }
+
+            if (changeType == null && File.Exists(path) || changeType.GetValueOrDefault() == FileChangeType.Create)
```
DustinCampbell: Or here

src/OmniSharp.Abstractions/Models/v1/FilesChanged/ChangeTypeConverter.cs
```diff
@@ -0,0 +1,25 @@
...
+                    return FileChangeType.Change;
+                case "Create":
+                    return FileChangeType.Create;
+                case "Delete":
+                    return FileChangeType.Delete;
+                default:
+                    return null;
+            }
+        }
+        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer) => throw new NotImplementedException();
```
david-driscoll: If we keep this we should override `CanWrite` and return false, would hate to have this throw unexpectedly.

src/OmniSharp.MSBuild/MSBuildProjectSystem.cs
```diff
@@ -400,6 +406,31 @@ private void UpdateSourceFiles(Project project, IList<string> sourceFiles)
...
+            {
+                // Only add .cs files to the workspace
+                if (Path.GetExtension(path) == ".cs")
```
DustinCampbell: Should be a case-insensitive comparison.

tests/OmniSharp.Roslyn.CSharp.Tests/FilesChangedFacts.cs
```diff
@@ -0,0 +1,44 @@
...
+using OmniSharp.Models.FilesChanged;
+using OmniSharp.Roslyn.CSharp.Services.Files;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using TestUtility;
+using Xunit;
+using Xunit.Abstractions;
```
DustinCampbell: Usings should be sorted with `System` on top.

tests/OmniSharp.Roslyn.CSharp.Tests/FilesChangedFacts.cs
```diff
@@ -0,0 +1,44 @@
...
+                string filePath = null;
+                FileChangeType? ct = null;
+                watcher.WatchDirectory("", (path, changeType) => { filePath = path; ct = changeType; });
+
+
+                var handler = GetRequestHandler(host);
+                handler.Handle(new[] { new FilesChangedRequest() { FileName = "FileName.cs", FileChangeType = FileChangeType.Create } });
+
+                Assert.Equal("FileName.cs", filePath);
+                Assert.Equal(FileChangeType.Create, ct);
```
DustinCampbell: Is it worth checking to see if the file is now included in the workspace? I believe you can get to it with `host.Workspace`.

src/OmniSharp.Abstractions/Models/v1/FilesChanged/FilesChangedRequest.cs
DustinCampbell: I think that's right.

src/OmniSharp.Abstractions/FileWatching/IFileSystemWatcher.cs
filipw: it somehow feels that an enum with default value would be more intuitive than nullable enum? at least that's what [MSDN](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/enum) would tell us 😀

rchande: Fixed names and added a default value.

src/OmniSharp.Abstractions/Models/v1/FilesChanged/ChangeTypeConverter.cs
rchande: Removed this and switched to the automatic serialization offered by json.net, which is case insensitive.

DustinCampbell: Great!

rchande:
Incorporated feedback. Want to sign off @DustinCampbell @david-driscoll ?

@DustinCampbell regarding the test, it looks like all of the existing test helpers create projects that use the "dotnet" project system. I think it might be worth going through and updating the tests to use the "msbuild" project system, or to run against both project systems.

src/OmniSharp.MSBuild/MSBuildProjectSystem.cs
```diff
@@ -20,6 +15,13 @@
 using OmniSharp.MSBuild.SolutionParsing;
 using OmniSharp.Options;
 using OmniSharp.Services;
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Composition;
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
```
DustinCampbell: Sort usings with `System` namespaces on top.

src/OmniSharp.Host/FileWatching/ManualFileSystemWatcher.cs
```diff
@@ -1,24 +1,37 @@
...
+using OmniSharp.Models.FilesChanged;
 
 namespace OmniSharp.FileWatching
 {
     public class ManualFileSystemWatcher : IFileSystemWatcher
     {
-        private readonly Dictionary<string, Action<string>> _callbacks = new Dictionary<string, Action<string>>();
+        private readonly Dictionary<string, Action<string, FileChangeType>> _callbacks = new Dictionary<string, Action<string, FileChangeType>>();
+        private readonly Dictionary<string, Action<string, FileChangeType>> _directoryCallBacks = new Dictionary<string, Action<string, FileChangeType>>();
```
DustinCampbell: One concern I have here is that only a single callback can be called for a particular path. And, the last callback registered will win. I don't think this will be a problem so far, but it may be in the future -- especially since we're watching directories. I can easily imagine some other project system (like script or cake) wanting to listen to the same directories that the MSBuild project system does. Should we combine callbacks using `Delegate.Combine(...)` and `Delegate.Remove(...)`?

src/OmniSharp.Roslyn.CSharp/Services/Files/OnFilesChangedService.cs
```diff
@@ -1,16 +1,15 @@
...
-using System.Composition;
-using System.Threading.Tasks;
 using Microsoft.CodeAnalysis;
 using OmniSharp.FileWatching;
 using OmniSharp.Mef;
-using OmniSharp.Models;
 using OmniSharp.Models.FilesChanged;
+using System.Collections.Generic;
+using System.Composition;
+using System.Threading.Tasks;
```
DustinCampbell: Sort usings with `System` namespaces on top please.

tests/OmniSharp.Roslyn.CSharp.Tests/FilesChangedFacts.cs
```diff
@@ -0,0 +1,39 @@
+﻿using OmniSharp.FileWatching;
+using OmniSharp.Models.FilesChanged;
+using OmniSharp.Roslyn.CSharp.Services.Files;
+using TestUtility;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace OmniSharp.Roslyn.CSharp.Tests
+{
+
```
DustinCampbell: Blank line

DustinCampbell:
> Regarding the test, it looks like all of the existing test helpers create projects that use the "dotnet" project system. I think it might be worth going through and updating the tests to use the "msbuild" project system, or to run against both project systems.

The appropriate way to do this is *not* to create projects. Instead, you can use one of the test projects like we do in the OmniSharp.MSBuild tests like so: https://github.com/OmniSharp/omnisharp-roslyn/blob/master/tests/OmniSharp.MSBuild.Tests/WorkspaceInformationTests.cs#L18-L32.

src/OmniSharp.Abstractions/Models/v1/FilesChanged/FilesChangedRequest.cs
```diff
@@ -1,8 +1,11 @@
-using System.Collections.Generic;
 using OmniSharp.Mef;
+using System.Collections.Generic;
```
DustinCampbell: Sort `System` namespaces first please. You've got the option set incorrectly in VS (this is also the preference in Roslyn).

src/OmniSharp.MSBuild/MSBuildProjectSystem.cs
rchande: I swear my option is resetting underneath me.

DustinCampbell: bummer 😞

DustinCampbell: Maybe you've discovered a VS bug via dogfooding?

src/OmniSharp.Host/FileWatching/ManualFileSystemWatcher.cs
rchande: That's a good point--will add this

rchande:
@DustinCampbell Incorporated your latest feedback

src/OmniSharp.Host/FileWatching/ManualFileSystemWatcher.cs
```diff
@@ -1,24 +1,44 @@
...
+        public void Watch(string path, Action<string, FileChangeType> callback)
         {
             _callbacks[path] = callback;
         }
+
+        public void WatchDirectory(string path, Action<string, FileChangeType> callback)
+        {
+            if (_directoryCallBacks.TryGetValue(path, out var existingCallback))
+            {
+                _directoryCallBacks[path] = (Action<string, FileChangeType>)Delegate.Combine(callback, existingCallback);
```
DustinCampbell: I'm OK leaving this as is, but it's also possible to just write:

```C#
_directoryCallBacks[path] = callback + existingCallback;
```

Believe it or not, `+` handles combination of delegates in C#. :smile:

**The More You Know**

rchande: So much nicer. Fixing.

DustinCampbell:
@david-driscoll : Does this look better to you?

rchande:
Found some issues while testing--let's hold off on merging this for a bit.

src/OmniSharp.Abstractions/Models/v1/FilesChanged/FilesChangedRequest.cs
```diff
@@ -3,6 +3,9 @@
 
 namespace OmniSharp.Models.FilesChanged
 {
-    [OmniSharpEndpoint(OmniSharpEndpoints.FilesChanged, typeof(IEnumerable<Request>), typeof(FilesChangedResponse))]
-    public class FilesChangedRequest : IRequest { }
+    [OmniSharpEndpoint(OmniSharpEndpoints.FilesChanged, typeof(IEnumerable<FilesChangedRequest>), typeof(FilesChangedResponse))]
+    public class FilesChangedRequest : Request
+    {
+        public FileChangeType ChangeType { get; set; }
```
david-driscoll: This will serialize as an int (I think, I didn't see any where we configure `Newtonsoft.Json`)

rchande:
This is ready to merge. Can you hit the button @DustinCampbell ?

PR #988, opened 2017-10-25
Re-architect how MSBuild is loaded by OmniSharp

DustinCampbell:
With MSBuild 15, locating MSBuild on the user's box can be a significant challenge. OmniSharp provides a version of MSBuild as a fallback, but that can often only work with .NET Core/Standard SDK-style projects; projects targeting .NET Framework can run into significant issues.

This PR introduces an `IMSBuildLocator` service that is used to discover instances of MSBuild 15 installed on the user's machine. It can discover MSBuild 15 using a few methods:

1. (Windows only) Using the COM-based Visual Studio setup API.
2. (Windows only) Checking environment variables that are set by a Visual Studio Developer Prompt
3. (OSX/Linux only) Looking for a global Mono installation that contains MSBuild 15.
4. If MSBuild 15 can't be discovered using the three methods above, it falls back to the standalone MSBuild included in OmniSharp.

Once MSBuild is located, OmniSharp will install an `AppDomain.AssemblyResolve` event to ensure that it loads the `Microsoft.Build.*` assemblies from that location. This ensures that the MSBuild used by OmniSharp will be compatible with whatever MSBuild toolset is used to process the user's project. In addition, this can speed up processing depending on whether the `Microsoft.Build.*` assemblies are NGen'd or AOT'd in the location they're loaded from.

Note that, because we're using `AssemblyResolve` to load `Microsoft.Build.*` assemblies, the event *must* be hooked before OmniSharp is composed with MEF. Otherwise, MEF will fail to compose.

In addition, as part of this change, I've reorganized the `.msbuild` folder to layout more closely to how it's laid out in Visual Studio.

A few things I want to do before merging this:

- [x] Delete the `MSBuildEnvironment` class. This is still there because there's still logic in this class that hasn't been replicated in an MSBuild discoverer yet.
- [x] More testing on OSX/Linux with and without Mono installed.
- [x] Testing on OSX/Linux when an older Mono is installed.
- [x] Add more logging in MSBuild discovery to help identify failures (especially with Mono).

Final note: There's an issue with Mono and AssemblyResolve where it can get into a bad state an attempt is made to load an assembly before the AssemblyResolve event is hooked. On desktop CLR, this works as expected. The core issue here is that xUnit uses `Assembly.Load` on every assembly in the directory to try and locate xUnit types and that happens before our AssemblyResolve is installed. Currently, this is working OK on OSX/Linux, but it's going to be a problem at some point. I've filed the following bug on Mono to get this fixed: https://bugzilla.xamarin.com/show_bug.cgi?id=60100.


.travis.yml
```diff
@@ -24,6 +24,12 @@ addons:
       - zlib1g
       - curl
 
+before_install:
+  - |
+    if [ "$TRAVIS_OS_NAME" == "osx" ]; then
+      rvm install ruby-2.3.3
+    fi
+
```
DustinCampbell: This is old code. Will remove this.

src/OmniSharp.Host/CompositionHostBuilder.cs
```diff
@@ -49,15 +50,19 @@ public CompositionHost Build()
...
-
             var fileSystemWatcher = new ManualFileSystemWatcher();
             var metadataHelper = new MetadataHelper(assemblyLoader);
 
+            // We must register an MSBuild instance before composing MEF to ensure that
+            // our AssemblyResolve event is hooked up first.
+            var msbuildLocator = _serviceProvider.GetRequiredService<IMSBuildLocator>();
+            var instances = msbuildLocator.GetInstances();
+            var instance = instances.FirstOrDefault();
+            if (instance != null)
```
rchande: Are we supposed to do anything if we couldn't locate any instances?

src/OmniSharp.Host/MSBuild/Discovery/MSBuildLocator.cs
```diff
@@ -0,0 +1,150 @@
...
+                    : null;
+
+                if (result != null)
+                {
+                    _logger.LogDebug($"Resolved '{assemblyName.Name}' to '{assemblyPath}'");
+                    return result;
+                }
+            }
+
+            return null;
```
rchande: For my own edification, what happens if Resolve returns null?

src/OmniSharp.Host/MSBuild/Discovery/Providers/StandAloneInstanceProvider.cs
```diff
@@ -0,0 +1,79 @@
...
+                // If the 'msbuild' folder does not exists beneath "OmniSharp.exe", this is likely a development scenario,
+                // such as debugging or running unit tests. In that case, we try to locate the ".msbuild" folder at the
+                // solution level.
+                msbuildDirectory = FindLocalMSBuildFromSolution();
+            }
+
+            return msbuildDirectory;
+        }
+
+        private static string FindLocalMSBuildFromSolution()
```
rchande: What about adding a command line argument that we can pass while debugging that specifies the standalone msbuild location? That way we don't have to talk about "OmniSharp.sln" in the product code.

src/OmniSharp.Host/MSBuild/Discovery/Providers/VisualStudioInstanceProvider.cs
```diff
@@ -0,0 +1,89 @@
...
+
+                var builder = ImmutableArray.CreateBuilder<MSBuildInstance>();
+
+                var instanceEnum = configuration.EnumAllInstances();
+
+                int fetched;
+                var instances = new ISetupInstance[1];
+                do
+                {
+                    instanceEnum.Next(1, instances, out fetched);
```
rchande: Does this only enumerate Dev15 versions?

src/OmniSharp.Host/MSBuild/Discovery/Providers/DevConsoleInstanceProvider.cs
```diff
@@ -0,0 +1,53 @@
...
+
+            if (version == null)
+            {
+                versionString = Environment.GetEnvironmentVariable("VisualStudioVersion");
+                Version.TryParse(versionString, out version);
+            }
+
+            return ImmutableArray.Create(
+                new MSBuildInstance(
+                    "DEVCONSOLE",
```
rchande: nameof(DiscoveryType.DevConsole)?

src/OmniSharp.Host/CompositionHostBuilder.cs
DustinCampbell: We should probably report an error to the log. Will add that.

src/OmniSharp.Host/MSBuild/Discovery/MSBuildLocator.cs
DustinCampbell: The runtime will continue to the next AssemblyResolve handler it finds. From https://msdn.microsoft.com/en-us/library/system.appdomain.assemblyresolve%28v=vs.110%29.aspx?f=255&MSPPError=-2147217396:

> If more than one event handler is registered for this event, the event handlers are called in order until an event handler returns a value that isn't null. Subsequent event handlers are ignored.

src/OmniSharp.Host/MSBuild/Discovery/Providers/DevConsoleInstanceProvider.cs
DustinCampbell: Good idea

src/OmniSharp.Host/MSBuild/Discovery/Providers/StandAloneInstanceProvider.cs
DustinCampbell: Maybe, but where would it get populated from? Note that the product code already referenced "OmniSharp.sln". This PR just moved that code to here.

src/OmniSharp.Host/MSBuild/Discovery/Providers/VisualStudioInstanceProvider.cs
DustinCampbell: That's all it can do. This is the API for the Willow installer.

src/OmniSharp.Host/MSBuild/Discovery/Providers/StandAloneInstanceProvider.cs
rchande: We would add the argument to the arguments passed by the debugger or by tests. Not a huge deal either way.

DustinCampbell: I suppose we could add some MSBuild logic to add an argument passed to the debugger. I'm not sure how it'd work for tests, but I'm guessing we could figure something out. I'm just not sure it's all that valuable.

DustinCampbell:
@david-driscoll, @filipw, @rchande : I've done a fair amount of manual testing with this change on OSX and am feeling pretty confident about it. The core MSBuild project scenarios are still working, some scenarios work better, and some new scenarios have starting working. Here are the scenarios I tested:

* No mono installed + run OmniSharp standalone:
    * Omnisharp-roslyn does not work (expected due to .NETFramework, 4.6 being missing)
    * .NET Core MVC project works with .NET Core SDK installed
    * Unity does not work (expected due to .NETFramework 2.0 being missing)
* Mono 4.8 installed + run OmniSharp standalone:
    * Omnisharp-roslyn works
    * .NET Core MVC project works
    * Unity project works
* Mono 5.2 installed + run OmniSharp standalone:
    * Omnisharp-roslyn works
    * .NET Core MVC project works
    * Unity project works
    * PCL project does not work (unfortunately, this is expected at the moment - cf. https://github.com/OmniSharp/omnisharp-vscode/issues/1779)
    * Xamarin forms project with iOS and Android heads works
* Mono 5.2 installed + run OmniSharp on installed Mono:
    * Omnisharp-roslyn works
    * .NET Core MVC project works
    * Unity project works
    * PCL project works (https://github.com/OmniSharp/omnisharp-vscode/issues/1779)
    * Xamarin forms project with iOS and Android heads works
* Mono 5.4 installed + run OmniSharp standalone:
    * Omnisharp-roslyn works
    * .NET Core MVC project works
    * Unity project works
    * PCL project does not work (unfortunately, this is expected at the moment - cf. https://github.com/OmniSharp/omnisharp-vscode/issues/1779)
    * Xamarin forms project with iOS and Android heads works
* Mono 5.4 installed + run OmniSharp on installed Mono:
    * Omnisharp-roslyn works
    * .NET Core MVC project works
    * Unity project works
    * PCL project works (https://github.com/OmniSharp/omnisharp-vscode/issues/1779)
    * Xamarin forms project with iOS and Android heads works

src/OmniSharp.Host/MSBuild/Discovery/MSBuildLocator.cs
```diff
@@ -0,0 +1,158 @@
...
+        {
+            var assemblyName = new AssemblyName(e.Name);
+
+            _logger.LogDebug($"Attempting to resolve '{assemblyName}'");
+
+            if (s_msbuildAssemblies.Contains(assemblyName.Name))
+            {
+                var assemblyPath = Path.Combine(_registeredInstance.MSBuildPath, assemblyName.Name + ".dll");
+                var result = File.Exists(assemblyPath)
+                    ? Assembly.LoadFrom(assemblyPath)
```
filipw: should this use https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp.Host/Services/AssemblyLoader.cs#L62 ?

DustinCampbell:
@filipw: I haven't specifically tested that scenario, but it should already work. The MSBuild 15.0 build tools use the same installer as VS 2017. So, the MSBuild discovery that looks for Visual Studio using the setup API should also work for the Build Tools as well.

src/OmniSharp.Host/MSBuild/Discovery/Providers/VisualStudioInstanceProvider.cs
david-driscoll: Which gets picked if someone has dev15 and a dev15 preview installed?

src/OmniSharp.Host/MSBuild/Discovery/Providers/MonoInstanceProvider.cs
```diff
@@ -20,6 +21,32 @@ public override ImmutableArray<MSBuildInstance> GetInstances()
                 return NoInstances;
             }
 
+            // Don't resolve to MSBuild assemblies under the installed Mono unless OmniSharp
+            // is actually running under the installed Mono.
+            var monoRuntimePath = PlatformHelper.GetMonoRuntimePath();
```
rchande: Is this obvious? Maybe explain a bit further what goes wrong if this is not the case?

rchande:
Thanks for testing this so thoroughly and documenting what you tested @DustinCampbell. cc @piotrpMSFT https://github.com/OmniSharp/omnisharp-roslyn/pull/988#issuecomment-339947876 explains the size of the test matrix.

DustinCampbell:
FWIW, I don't normally test all of those scenarios, but they were necessary for this particular change. Too often, changes like these have broken unity support.

src/OmniSharp.Host/MSBuild/Discovery/Providers/VisualStudioInstanceProvider.cs
DustinCampbell: It's a bit arbitrary. We don't have a way for the user to select one, so it's whichever is first in the list. However, we check for the Visual Studio tied to a developer command prompt first. So, if the user wants to launch OmniSharp with a specific Visual Studio they can launch it from the command prompt that was installed for that VS. Alternatively, they can set the environment variables that the developer command prompt uses.

src/OmniSharp.Host/MSBuild/Discovery/MSBuildLocator.cs
DustinCampbell: Yes, you're right. I didn't use it initially because I was still trying to figure out when/how the `IMSBuildLocator` would be initialized. New commit adds this.

src/OmniSharp.Host/MSBuild/Discovery/Providers/MonoInstanceProvider.cs
DustinCampbell: None of this is obvious. :smile: But you're right, I'll add a little bit more detail about how it can go wrong.

DustinCampbell:
OK. I think it's good now. I'm going to go ahead and merge. Let me know if you think of anything else.

filipw:
excellent!

enricosada:
awesome. i am going to steal it. thx @DustinCampbell !

casz:
@DustinCampbell Did you try mono mdk? the package provided by Xamarin people through brew? 👍

DustinCampbell:
I installed the Mono MDK with the installer

casz:
👍 Awesome

DustinCampbell:
FWIW, I'll check `brew install cask mono-mdk` on Monday.

casz:
SAME 👍

corngood:
This seems to have broken mono/windows, see #1001

DustinCampbell:
@corngood: Why are you running OmniSharp on Mono/Windows? We've never tested that scenario. The assumption has generally been that desktop CLR on Windows is the way to go for OmniSharp.

DustinCampbell:
@casz: I just checked `brew cask install mono-mdk` and it works fine.

corngood:
@DustinCampbell I'm working around a bug w/ net46.  We can discuss on #1001

casz:
@DustinCampbell I tested it with our dotnet 4.5 project and it did not work 😢

DustinCampbell:
@casz: Bummer! Could you file an issue with your OmniSharp log?

casz:
@DustinCampbell did I have to do something special or was downloading c# extension enough for VS Code?
Using 1.12.1 c# extension in vscode

casz:
Oh found the beta install guide... my bad

casz:
@DustinCampbell it works almost https://github.com/OmniSharp/omnisharp-vscode/issues/1821

Issue #968, opened 2017-09-25
[Epic] Language Server Protocol


david-driscoll:
For history see: #576
For initial PR see: #969

Lsp packages have been pushed to nuget [here](https://www.nuget.org/packages/OmniSharp.Extensions.JsonRpc/) and [here](https://www.nuget.org/packages/OmniSharp.Extensions.LanguageServerProtocol/).

This starts consuming them, the branch `lsp` may be long lived, or we may merge into master a few times, we haven't discussed this yet.  

Note: not all endpoints currently map 1:1 to LSP, more information on lsp can be found [here](https://github.com/OmniSharp/csharp-language-server-protocol) and the C# implementation [here](https://github.com/OmniSharp/csharp-language-server-protocol)

For testing I was using the test extension from the csharp-lsp repository, and simplying making sure that `OmniSharp.exe` was invoked with `-lsp` or `--languageserver` and it will negotiate with language server mode automatically.  It would be great to get others feedback using this against language server 2.0 vs 3.0 (that vscode implements).

Parts done:
* [x] Wired in language server protocol to the default host
* Endpoints configured so far :
  * [x] `/gotodefinition`
  * [ ] `/findsymbols`
  * [x] `/updatebuffer`
  * [x] `/changebuffer`
  * [x] `/codecheck`
  * [ ] `/filesChanged`
  * [ ] `/formatAfterKeystroke`
  * [ ] `/formatRange`
  * [ ] `/codeformat`
  * [ ] `/highlight`
  * [ ] `/autocomplete`
  * [ ] `/findimplementations`
  * [ ] `/findusages`
  * [ ] `/gotofile`
  * [ ] `/gotoregion`
  * [ ] `/navigateup`
  * [ ] `/navigatedown`
  * [x] `/typelookup`
  * [ ] `/getcodeactions`
  * [ ] `/runcodeaction`
  * [x] `/rename`
  * [x] `/signatureHelp`
  * [ ] `/currentfilemembersastree`
  * [ ] `/currentfilemembersasflat`
  * [ ] `/gettestcontext`
  * [ ] `/metadata`
  * [ ] `/packagesource`
  * [ ] `/packagesearch`
  * [ ] `/packageversion`
  * [ ] `/projects`
  * [ ] `/project`
  * [ ] `/fixusings`
  * [ ] `/checkalivestatus`
  * [ ] `/checkreadystatus`
  * [ ] `/stopserver`
  * [x] `/open`
[...]

PR #997, opened 2017-10-27
Adds RenameHandler

RLittlesII:
This pull request adds RenameHandler to #968 

@david-driscoll Let me know what should return in the event the underlying service returns an error.  Should the handler throw an exception or just return null?


david-driscoll:
Letting it throw should be fine the lsp runtime (should...🤔) properly send the error response back to the client.

PR #1003, opened 2017-11-02
Reload and update MSBuild projects in response to dotnet restore

DustinCampbell:
Fixes https://github.com/OmniSharp/omnisharp-vscode/issues/1583, https://github.com/OmniSharp/omnisharp-vscode/issues/1661, https://github.com/OmniSharp/omnisharp-vscode/issues/1785

This is a fairly substantial change in order to address a problem where MSBuild projects would not be updated in response to a dotnet restore. Essentially, the MSBuild project system must watch four different files per project in order to respond to a dotnet restore:

* project.asset.json
* <project-file>.nuget.cache
* <project-file>.nuget.g.props
* <project-file>.nuget.g.targets

To ensure that we don't reload and update a project multiple times in response to multiple file
changes, this PR introduces a simple queue and processing loop using a TPL DataFlow BufferBlock. The implication of this is that the MSBuild project system is now considered "initialized" only once the initial project files are queued for an update. So, OmniSharp is ready to respond to requests more quickly, but the projects won't necessarily be loaded yet.

In addition, I've made a few other changes while fixing this issue:

1. It turns out that metadata references were not properly removed from the workspace due to the fact that Roslyn's `MetadataReference` does not implement value equality. To address this, I create our own `IEqualityComparer<MetadataReference>` implementation.
2. I've simplified file watching/notification by unifying file and directory watching, and fixed two issues:
    a. Paths are now treated case-insensitively as they are in the rest of OmniSharp.
    b. Access to the internal callback dictionary is now serialized.
3. Most code in ProjectFileInfo and MSBuildProjectSystem have moved into two new classes: ProjectLoader and ProjectManager. This will facilitate some MSBuild performance improvements I'll be making later.


src/OmniSharp.MSBuild/PackageDependencyChecker.cs
```diff
@@ -5,17 +5,55 @@
...
-    internal class PackageDependencyResolver
+    internal class PackageDependencyChecker
     {
         private readonly ILogger _logger;
+        private readonly IEventEmitter _eventEmitter;
+        private readonly DotNetCliService _dotNetCli;
+        private readonly MSBuildOptions _options;
 
-        public PackageDependencyResolver(ILoggerFactory loggerFactory)
+        public PackageDependencyChecker(ILoggerFactory loggerFactory, IEventEmitter eventEmitter, DotNetCliService dotNetCli, MSBuildOptions options)
```
filipw: when we resolve inline package references (`#r "nuget: Foo, 1.0.0"`) in scripts, we create behind the scenes a temp `csproj` and restore using .NET CLI to resolve them. However, currently, we don't handle changes after project system init anymore. Looks like we'll be able to reuse this logic there 👍

src/OmniSharp.MSBuild/ProjectManager.cs
```diff
@@ -0,0 +1,469 @@
...
+                }
+            }
+
+            foreach (var existingProjectReference in existingProjectReferences)
+            {
+                _workspace.RemoveProjectReference(project.Id, existingProjectReference);
+            }
+        }
+
+        private class MetadataReferenceComparer : IEqualityComparer<MetadataReference>
```
filipw: I think this needs to be extracted into a reusable place - it will be useful i.e. here https://github.com/OmniSharp/omnisharp-roslyn/blob/d3930049529fe3342de8d206a18c6823dd071d96/src/OmniSharp.Script/ScriptProjectSystem.cs#L97

DustinCampbell: Yes, good idea. I'll make it a utility in OmniSharp.Roslyn.

DustinCampbell:
Note: Because this PR changes the MSBuild project system at a fundamental level, I'm going to hold off on merging this until we can get a C# for VS Code release out early next week.

PR #1015, opened 2017-11-09
Use --zero-based-indices argument in Stdio and Http hosts

daviwil:
This change fixes an issue where the --zero-based-indices command
line argument was not being passed along to Configuration.ZeroBasedIndices
so that location/range serialization uses zero-based indices. This
fix wires up the argument to configure the serializer correctly.

### DIFF
--- a/src/OmniSharp.Abstractions/AssemblyInfo.cs
+++ b/src/OmniSharp.Abstractions/AssemblyInfo.cs
@@ -1,5 +1,6 @@
 ﻿using System.Runtime.CompilerServices;
 
+[assembly: InternalsVisibleTo("OmniSharp")]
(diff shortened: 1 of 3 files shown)


daviwil:
Man, TravisCI is super slow today, looks like the build jobs timed out while the tests were running.

filipw:
public static field 🙄 🤐 😄

DustinCampbell:
@daviwil : yeah, the OSX jobs get super-slow during US business hours.

DustinCampbell:
I'm going to go ahead and force merge this. It doesn't *actually* have conflicts with master

daviwil:
Thanks a lot Dustin!

PR #1016, opened 2017-11-09
Fix 1.26.1 bugs

### LINKED ISSUE #1011: OmniSharp should not use Mono MSBuild if it is missing
On Debian, there is a separate `msbuild` package for installing MSBuild into Mono. Until that's installed, MSBuild is not actually present and OmniSharp can't use it, even if the directory structure has been created.

Related to https://github.com/OmniSharp/omnisharp-vscode/issues/1847

### LINKED ISSUE #1014: OmniSharp should not Visual Studio 2017 MSBuild if it is VS 2017 RTM
Related to https://github.com/OmniSharp/omnisharp-vscode/issues/1848

DustinCampbell:
Fixes #1011 
Fixes #1014 

This PR addresses a couple of MSBuild discovery issues:

1. Don't use Mono MSBuild if `Microsoft.Build.dll` is missing. This can happen on Linux layouts of Mono when the user hasn't installed the `msbuild` package yet.
2. Don't use Visual Studio 2017 MSBuild if it is VS 2017 RTM. In this case, the user is encouraged to update VS 2017.

### DIFF
--- a/src/OmniSharp.Abstractions/Utilities/Platform.cs
+++ b/src/OmniSharp.Abstractions/Utilities/Platform.cs
@@ -49,7 +49,20 @@ namespace OmniSharp.Utilities
             var os = OperatingSystem.Unknown;
             var architecture = Architecture.Unknown;
 
-            // Simple check to see if this is Windows
+            // Simple check to see if this is Windows. Note: this check is derived from the fact that the
+            // System.PlatformID enum has six values (https://msdn.microsoft.com/en-us/library/3a8hyw88.aspx)
+            //
+            //   * Win32 = 0
+            //   * Win32Windows = 1
+            //   * Win32NT = 2
+            //   * WinCE = 3
+            //   * Unix = 4
+            //   * Xbox = 5
+            //   * MacOSX = 6
+            //
+            // Essentially, we check to see if this is one of the "windows" values or Xbox. The other values
+            // can be a little unreliable, so we'll shell out to 'uname' for Linux and macOS.
+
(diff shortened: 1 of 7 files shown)


src/OmniSharp.Abstractions/Utilities/Platform.cs
```diff
@@ -0,0 +1,172 @@
...
+        }
+
+        private static Platform GetCurrentPlatform()
+        {
+            var os = OperatingSystem.Unknown;
+            var architecture = Architecture.Unknown;
+
+            // Simple check to see if this is Windows
+            var platformId = (int)Environment.OSVersion.Platform;
+            if (platformId <= 3 || platformId == 5)
```
rchande: Can you add a comment explaining where this knowledge came from?

DustinCampbell: I'll try to remember where it came from. This is derived from `scripts/platform.cake`

DustinCampbell: Oh right. It was [this](https://msdn.microsoft.com/en-us/library/3a8hyw88(v=vs.110).aspx):

| Member       | Underlying value |
|--------------|------------------|
| Win32S       | 0                |
| Win32Windows | 1                |
| Win32NT      | 2                |
| WinCE        | 3                |
| Unix         | 4                |
| Xbox         | 5                |
| MacOSX       | 6                |

Will add a comment

DustinCampbell:
cc @akshita31 and @TheRealPiotrP

PR #1017, opened 2017-11-09
Load MSBuild assemblies up front

DustinCampbell:
This works around what appears to be an issue with Mono when `AppDomain.AssemblyResolve` events are now triggered on ThreadPool threads due to my recent changes to MSBuild project loading.


DustinCampbell:
cc @akshita31 as well

rchande:
What went wrong with mono?

DustinCampbell:
Mono didn't like it. A type initializer failed when it was called on a different thread than the one where the AssemblyResolve happened.

PR #1019, opened 2017-11-10
VS Code Issue 1057 - Problem with Newlines in displaying XML Documentation

akshita31:
Issue : https://github.com/OmniSharp/omnisharp-vscode/issues/1057

The issue was of absence of New Lines in Parsed XML Documentation. The problem was that VS Code is displaying a  newline character( like "\n" or "\r\n" ) as a space and not a newline. It is rather accepting two newline characters as a newline ("\n\n" or "\r\n\r\n" ). So the TypeLookUp handler has been modified to pass the appropriate lineEnding parameter. Accordingly a test has also been added.

### DIFF
--- a/src/OmniSharp.Roslyn.CSharp/Services/Types/TypeLookup.cs
+++ b/src/OmniSharp.Roslyn.CSharp/Services/Types/TypeLookup.cs
@@ -1,4 +1,5 @@
-﻿using System.Composition;
+﻿using System;
+using System.Composition;
 using System.Threading.Tasks;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.FindSymbols;
@@ -45,7 +46,9 @@ namespace OmniSharp.Roslyn.CSharp.Services.Types
 
                     if (request.IncludeDocumentation)
                     {
-                        response.Documentation = DocumentationConverter.ConvertDocumentation(symbol.GetDocumentationCommentXml(), _formattingOptions.NewLine);
+                        string newLine = Environment.NewLine + Environment.NewLine;
+                        //VS Code renders a single new line for two newline sequences in the response, hence two new lines are passed in the lineEnding parameter 
+                        response.Documentation = DocumentationConverter.ConvertDocumentation(symbol.GetDocumentationCommentXml(),newLine);
                     }
                 }
             }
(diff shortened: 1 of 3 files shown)


src/OmniSharp.Roslyn.CSharp/Services/Types/TypeLookup.cs
```diff
@@ -45,7 +46,9 @@ public TypeLookupService(OmniSharpWorkspace workspace, FormattingOptions formatt
 
                     if (request.IncludeDocumentation)
                     {
-                        response.Documentation = DocumentationConverter.ConvertDocumentation(symbol.GetDocumentationCommentXml(), _formattingOptions.NewLine);
+                        string newLine = Environment.NewLine + Environment.NewLine;
+                        //VS Code renders a single new line for two newline sequences in the response, hence two new lines are passed in the lineEnding parameter
```
DustinCampbell: If this is a VS Code-specific tweak, why does it belong in omnisharp-roslyn? Shouldn't the fix be in omnisharp-vscode? After all, other editors use omnisharp-roslyn (e.g. Emacs, Vim, Atom, Sublime, etc.). Wouldn't this be a breaking change for them?

src/OmniSharp.Roslyn.CSharp/Services/Types/TypeLookup.cs
```diff
@@ -45,7 +46,9 @@ public TypeLookupService(OmniSharpWorkspace workspace, FormattingOptions formatt
 
                     if (request.IncludeDocumentation)
                     {
-                        response.Documentation = DocumentationConverter.ConvertDocumentation(symbol.GetDocumentationCommentXml(), _formattingOptions.NewLine);
+                        string newLine = Environment.NewLine + Environment.NewLine;
+                        //VS Code renders a single new line for two newline sequences in the response, hence two new lines are passed in the lineEnding parameter 
+                        response.Documentation = DocumentationConverter.ConvertDocumentation(symbol.GetDocumentationCommentXml(),newLine);
```
DustinCampbell: nit: add space before 'newLine'

DustinCampbell:
Also, please squash unnecessary commits like these:

* VS Code Iss 1057 fix
* VS Code Issue 1057 fix
* Issue 1057 fix

src/OmniSharp.Roslyn.CSharp/Services/Types/TypeLookup.cs
TheRealPiotrP: Yep, if this is a VS Code-specific bug then we should move the fix to omnisharp-vscode. We can capture the output coming from O# and map it to what VS Code expects.

PR #1020, opened 2017-11-10
Build: Wait for run scripts to idle rather than terminate them after a timeout

DustinCampbell:
This change ensures that we get to see the whole run script before terminating it during a build. Recently, a bug was masked because the run scripts were terminated before the bug would have manifested.


DustinCampbell:
> the real question is - did you use Omnisharp to make this Cake change?

I did. :smile:

filipw:
on a potentially related note - we are in the process of [discussing](https://github.com/filipw/dotnet-script/issues/156#issuecomment-341990138) how `#load` of a script file directly from Nuget should work against Roslyn resolvers. Hopefully one day it will allow us to share reusable pieces like the run helpers here across various scripts easier.

PR #1021, opened 2017-11-10
Add back code that sets MSBuildSDKsPath environment variable for each project

DustinCampbell:
Unfortunately, it seems that there are some [reliability issues](https://github.com/Microsoft/msbuild/issues/2532) with the MSBuild SDK Resolvers that we haven't gotten to the bottom of yet. So, I'm adding back code to set the `MSBuildSDKsPath` environment variable per project based on running the `dotnet` in that project's directory.

This should address issues like https://github.com/OmniSharp/omnisharp-vscode/issues/1849 and https://github.com/OmniSharp/omnisharp-vscode/issues/1846.

Note that this is not a long term solution. It will work for now, but there are several future concerns that we need to be mindful of:

1. Setting an environment variable will be problematic if we ever want to run design-time builds in parallel. Currently, that's blocked by several issues, but maybe someday.
2. If the `dotnet --info` output changes such that `Base Path` no longer appears in the text, this will break and need to be updated.
3. This requires a process to be launched for every project. This isn't out biggest performance issue at the moment and it's what we were doing before, but this is something to keep in mind.

cc @nguerrera, @livarcocc

### DIFF
--- a/src/OmniSharp.MSBuild/MSBuildProjectSystem.cs
+++ b/src/OmniSharp.MSBuild/MSBuildProjectSystem.cs
@@ -33,6 +33,7 @@ namespace OmniSharp.MSBuild
         private readonly OmniSharpWorkspace _workspace;
         private readonly MSBuildInstance _msbuildInstance;
         private readonly DotNetCliService _dotNetCli;
+        private readonly SdksPathResolver _sdksPathResolver;
         private readonly MetadataFileReferenceCache _metadataFileReferenceCache;
         private readonly IEventEmitter _eventEmitter;
         private readonly IFileSystemWatcher _fileSystemWatcher;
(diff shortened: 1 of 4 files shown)


DustinCampbell:
Thanks!

haifengmeiITC:
@DustinCampbell can you also apply this change to the omnisharp-vscode project?

DustinCampbell:
Yes. It will take time though. These builds go through some process and get uploaded to a couple of different CDNs first.

nguerrera:
Have you ever been able to repro the reliability issue?

DustinCampbell:
I haven't been able to repro it at all, but there are enough reports that I want to add this back in the meantime.

nguerrera:
I suspect Microsoft/msbuild#2532 is different by the way. See my comments there.

DustinCampbell:
Yeah, I saw that. This issue has me a bit mystified though. We don't do anything special with the SdkResolvers -- just copy them in place. Then, we just use standard MSBuild APIs to evaluate and build a project. No funny business. 😀

Users are reporting problems on multiple OSs.

PR #1023, opened 2017-11-10
Support rename/openDocument operations from CodeActions

rchande:
If a CodeAction returns a RenameDocumentOperation, use reflection to extract the file rename information and apply the requested changes to the workspace.
Enable host support by returning new versions of ModifiedFileRequests that indicate the files have been renamed, or in the case of OpenDocumentOperation, opened.


rchande:
Tag @DustinCampbell @david-driscoll @TheRealPiotrP

rchande:
VSCode side: https://github.com/OmniSharp/omnisharp-vscode/pull/1854

rchande:
cc @akshita31

src/OmniSharp.Abstractions/Models/v1/FileOperation/ModifiedFileResponse.cs
```diff
@@ -2,16 +2,13 @@
 
 namespace OmniSharp.Models
 {
-    public class ModifiedFileResponse
+    public class ModifiedFileResponse : FileOperationResponse
```
DustinCampbell: Good idea

src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/RunCodeActionService.cs
```diff
@@ -71,6 +86,29 @@ public class RunCodeActionService : BaseCodeActionService<RunCodeActionRequest,
...
 
+        private static string GetNewFilePath(string newFileName, string currentFilePath)
+        {
+            var directory = Path.GetDirectoryName(currentFilePath);
+            return Path.Combine(directory, newFileName);
+        }
+
+        bool IsRenameDocumentOperation(CodeActionOperation o, out DocumentId oldDocumentId, out DocumentId newDocumentId, out string name)
+        {
+            var type = Type.GetType("Microsoft.CodeAnalysis.CodeActions.RenameDocumentOperation, Microsoft.CodeAnalysis.Workspaces, Version=2.4.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35");
```
DustinCampbell: I don't think you need to include the version number, culture or publickeytoken. This assembly should already be loaded and I'd hate to see this fail (and be another place to update) each time we update Roslyn.

DustinCampbell: Actually, you might be able to just grab it from `Configuration.RoslynWorkspaces` ([src](https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp.Abstractions/Configuration.cs#L12)).

See, https://github.com/OmniSharp/omnisharp-roslyn/blob/e18913e887144119c41d60f1842e49f8e9bfcf72/src/OmniSharp.Roslyn/MetadataHelper.cs for how we typically handle reflection scenarios in OmniSharp. We also have some reflection extension methods which are intended to fail quickly so we can easily tell when Roslyn internal surface area breaks something ((https://github.com/OmniSharp/omnisharp-roslyn/blob/e18913e887144119c41d60f1842e49f8e9bfcf72/src/OmniSharp.Abstractions/Utilities/ReflectionExtensions.cs) ).

src/OmniSharp.Roslyn/OmniSharpWorkspace.cs
```diff
@@ -142,5 +145,59 @@ public override bool CanApplyChange(ApplyChangesKind feature)
...
+
+        private void DeleteDocumentFile(DocumentId id, string filePath)
+        {
+            try
+            {
+                File.Delete(filePath);
+            }
+            catch (IOException) { }
+            catch (NotSupportedException) { }
+            catch (UnauthorizedAccessException) { }
```
DustinCampbell: Do we want to log any of these for debugging purposes?

src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/RunCodeActionService.cs
david-driscoll: my 👀🔥🔥🔥

src/OmniSharp.Roslyn/OmniSharpWorkspace.cs
```diff
@@ -142,5 +145,59 @@ public override bool CanApplyChange(ApplyChangesKind feature)
...
+                }
+
+                using (var writer = new StreamWriter(fullPath, append: false, encoding: encoding))
+                {
+                    newText.Write(writer);
+                }
+            }
+            catch (IOException)
+            {
+            }
```
DustinCampbell: log for debugging purposes?

src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/RunCodeActionService.cs
```diff
@@ -71,6 +86,29 @@ public class RunCodeActionService : BaseCodeActionService<RunCodeActionRequest,
...
+            if (o.GetType() == type)
+            {
+                oldDocumentId = (DocumentId)type.GetField("_oldDocumentId", System.Reflection.BindingFlags.NonPublic | System.Reflection.BindingFlags.Instance).GetValue(o);
+                newDocumentId = (DocumentId)type.GetField("_newDocumentId", System.Reflection.BindingFlags.NonPublic | System.Reflection.BindingFlags.Instance).GetValue(o);
+                name = (string)type.GetField("_newFileName", System.Reflection.BindingFlags.NonPublic | System.Reflection.BindingFlags.Instance).GetValue(o);
+                return true;
```
DustinCampbell: We'll be in a weird state if the reflection for any of the fields above fails, which could easily happen in a Roslyn update.

DustinCampbell:
@david-driscoll , @rchande : Per David's feedback, I agree with the concern that we'll now be sending operations to editors that won't know how to handle them. In addition to David's options, what if we add a new property to `RunCodeActionRequest` to allow editor's to opt-in to the new behavior? If they don't set the property, they only get file modifications. That's another approach we've taken in the past and should be relatively cheap.

david-driscoll:
yeah an opt-in property like `WantAllOperations` would solve it. (feel free to name it as you see fit)

rchande:
@david-driscoll @DustinCampbell That's a great point. I'll add some property hosts can set if they want to be advised of changes other than modifications. Thanks!

src/OmniSharp.Roslyn/OmniSharpWorkspace.cs
rchande: Good point, I'll log these.

src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/RunCodeActionService.cs
filipw: yeah +1 for using the existing lazy extensions

src/OmniSharp.Abstractions/Models/v1/FileOperation/FileOperationResponse.cs
```diff
@@ -0,0 +1,19 @@
...
+    {
+        public FileOperationResponse(string fileName, FileModificationType type)
+        {
+            FileName = fileName;
+            ModificationType = type;
+        }
+
+        public string FileName { get; }
+
+        [JsonConverter(typeof(StringEnumConverter))]
```
filipw: at the moment I believe none of the enums we have on the responses are stringified. I think it would be better to stay consistent

src/OmniSharp.Abstractions/Models/v1/FileOperation/FileModificationType.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace OmniSharp.Models
+{
+    public enum FileModificationType
```
filipw: there is already a public `FileChangeType` enum for the watcher, isn't it a bit confusing to add `FileModificationType` too?

DustinCampbell: This is pretty different though. They're not really interchangeable.

rchande:
@DustinCampbell @david-driscoll Incorporated your suggestions. Want to take another look?

src/OmniSharp.Roslyn/OmniSharpWorkspace.cs
```diff
@@ -142,5 +149,74 @@ public override bool CanApplyChange(ApplyChangesKind feature)
...
+            {
+                LogDeletionException(e, fullPath);
+            }
+            catch (UnauthorizedAccessException e)
+            {
+                LogDeletionException(e, fullPath);
+            }
+        }
+
+        private void LogDeletionException(Exception e, string filePath)
```
DustinCampbell: Missed opportunity for local function. :smile:

DustinCampbell:
Note that you have a build failure.

DustinCampbell:
@rchande: Note that your test failed on Ubuntu: https://travis-ci.org/OmniSharp/omnisharp-roslyn/jobs/301540682

DustinCampbell:
@rchande : I'm pretty sure the error is that you're trying to add a document to the workspace called `/somepath/buffer.cs`. However, `/somepath` is a top-level directory.

rchande:
@DustinCampbell Yep, my fix was to have a filepath that didn't start with /

DustinCampbell:
FWIW, it already failed with the same error. AFAICT, it's happening when the workspace applies changes:

```
OmniSharp.Roslyn.CSharp.Tests.CodeActionsV2Facts.Can_send_rename_and_fileOpen_responses_when_codeAction_renames_file [FAIL]
      System.UnauthorizedAccessException : Access to the path "/somepath" is denied.
      Stack Trace:
          at System.IO.Directory.CreateDirectoriesInternal (System.String path) [0x0005e] in <c9f8153c41de4f8cbafd0e32f9bf6b28>:0 
          at System.IO.Directory.CreateDirectory (System.String path) [0x0008f] in <c9f8153c41de4f8cbafd0e32f9bf6b28>:0 
          at OmniSharp.OmniSharpWorkspace.SaveDocumentText (Microsoft.CodeAnalysis.DocumentId id, System.String fullPath, Microsoft.CodeAnalysis.Text.SourceText newText, System.Text.Encoding encoding) [0x00017] in <4e1f5f2b866e4b60b7f726dcc5fc2898>:0 
          at OmniSharp.OmniSharpWorkspace.ApplyDocumentAdded (Microsoft.CodeAnalysis.DocumentInfo info, Microsoft.CodeAnalysis.Text.SourceText text) [0x00048] in <4e1f5f2b866e4b60b7f726dcc5fc2898>:0 
          at Microsoft.CodeAnalysis.Workspace.ApplyProjectChanges (Microsoft.CodeAnalysis.ProjectChanges projectChanges) [0x0027f] in <751d22bb910d48a5907708a3f5126eb1>:0 
          at Microsoft.CodeAnalysis.Workspace.TryApplyChanges (Microsoft.CodeAnalysis.Solution newSolution, Microsoft.CodeAnalysis.Shared.Utilities.IProgressTracker progressTracker) [0x000e9] in <751d22bb910d48a5907708a3f5126eb1>:0 
          at Microsoft.CodeAnalysis.Workspace.TryApplyChanges (Microsoft.CodeAnalysis.Solution newSolution) [0x00007] in <751d22bb910d48a5907708a3f5126eb1>:0 
          at OmniSharp.Roslyn.CSharp.Services.Refactoring.V2.RunCodeActionService+<Handle>d__8.MoveNext () [0x00571] in <68d476fb278945f6a5561ccee2049324>:0 
```

DustinCampbell:
IOW, it's happening when `RunRefactoringsAsync`, but you changed `FindRefactoringsAsync`

DustinCampbell:
FWIW, since you've got a situation where need to actually update files on disk, you might consider creating or reusing a test project. [These](https://github.com/OmniSharp/omnisharp-roslyn/blob/master/tests/OmniSharp.MSBuild.Tests/WorkspaceInformationTests.cs) tests use that infrastructure if you need to model it on something.

rchande:
@DustinCampbell Thanks--started using a test project and that seems to make tests pass. Note that the mac build failed on "installing .NET SDK" :(

DustinCampbell:
Looking good @rchande! You probably want to add your new test project to the list of `TestAssets` in `build.json`: https://github.com/OmniSharp/omnisharp-roslyn/blob/master/build.json#L39

PR #1030, opened 2017-11-23
Intellisense not showing methods from the Base class in Signature Help

akshita31:
Issue : https://github.com/OmniSharp/omnisharp-vscode/issues/1440

During the search for the signature help, only the immediate Type that contains the given method  was being searched for possible method overloads hence the overloads present in the Base Classes were not appearing. Added the code to search for members using the GetMemberGroup. Added a test for the same. 

Please review @DustinCampbell @TheRealPiotrP @rchande

### DIFF
--- a/src/OmniSharp.Roslyn.CSharp/OmniSharp.Roslyn.CSharp.csproj
+++ b/src/OmniSharp.Roslyn.CSharp/OmniSharp.Roslyn.CSharp.csproj
@@ -5,6 +5,14 @@
     <PlatformTarget>AnyCPU</PlatformTarget>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
+    <LangVersion>7.1</LangVersion>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <LangVersion>7.1</LangVersion>
+  </PropertyGroup>
+
   <ItemGroup>
     <ProjectReference Include="..\OmniSharp.Abstractions\OmniSharp.Abstractions.csproj" />
     <ProjectReference Include="..\OmniSharp.Roslyn\OmniSharp.Roslyn.csproj" />
--- a/src/OmniSharp.Roslyn.CSharp/OmniSharp.Roslyn.CSharp.csproj
+++ b/src/OmniSharp.Roslyn.CSharp/OmniSharp.Roslyn.CSharp.csproj
@@ -5,14 +5,6 @@
     <PlatformTarget>AnyCPU</PlatformTarget>
   </PropertyGroup>
 
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
-    <LangVersion>7.1</LangVersion>
-  </PropertyGroup>
-
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
-    <LangVersion>7.1</LangVersion>
-  </PropertyGroup>
-
   <ItemGroup>
     <ProjectReference Include="..\OmniSharp.Abstractions\OmniSharp.Abstractions.csproj" />
     <ProjectReference Include="..\OmniSharp.Roslyn\OmniSharp.Roslyn.csproj" />
--- a/src/OmniSharp.Roslyn.CSharp/Services/Signatures/CheckStaticInvocation.cs
+++ b/src/OmniSharp.Roslyn.CSharp/Services/Signatures/CheckStaticInvocation.cs
@@ -1,5 +1,5 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
-
+// Adapted from https://github.com/dotnet/roslyn/blob/master/src/Workspaces/CSharp/Portable/Extensions/SyntaxNodeExtensions.cs
 using System;
 using System.Collections.Generic;
 using System.Linq;
--- a/src/OmniSharp.Roslyn.CSharp/Services/Signatures/CheckStaticInvocation.cs
+++ b/src/OmniSharp.Roslyn.CSharp/Services/Signatures/CheckStaticInvocation.cs
@@ -1,0 +1,175 @@
+﻿using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+namespace OmniSharp.Roslyn.CSharp.Services.Signatures
+{
+    static internal class CheckForStatic
+    {
+        public static bool IsInStaticContext(this SyntaxNode node)
+        {
+            // this/base calls are always static.
+            if (node.FirstAncestorOrSelf<ConstructorInitializerSyntax>() != null)
+            {
+                return true;
+            }
+
+            var memberDeclaration = node.FirstAncestorOrSelf<MemberDeclarationSyntax>();
+            if (memberDeclaration == null)
+            {
+                return false;
+            }
+
+            switch (memberDeclaration.Kind())
+            {
+                case SyntaxKind.MethodDeclaration:
+                case SyntaxKind.ConstructorDeclaration:
+                case SyntaxKind.EventDeclaration:
+                case SyntaxKind.IndexerDeclaration:
+                    return GetModifiers(memberDeclaration).Any(SyntaxKind.StaticKeyword);
+
+                case SyntaxKind.PropertyDeclaration:
+                    return GetModifiers(memberDeclaration).Any(SyntaxKind.StaticKeyword) ||
+                        node.IsFoundUnder((PropertyDeclarationSyntax p) => p.Initializer);
+
+                case SyntaxKind.FieldDeclaration:
+                case SyntaxKind.EventFieldDeclaration:
+                    // Inside a field one can only access static members of a type (unless it's top-level).
+                    return !memberDeclaration.Parent.IsKind(SyntaxKind.CompilationUnit);
+
+                case SyntaxKind.DestructorDeclaration:
+                    return false;
+            }
+
+            // Global statements are not a static context.
+            if (node.FirstAncestorOrSelf<GlobalStatementSyntax>() != null)
+            {
+                return false;
+            }
+
+            // any other location is considered static
+            return true;
+        }
+        public static SyntaxTokenList GetModifiers(SyntaxNode member)
+        {
+            if (member != null)
+            {
+                switch (member.Kind())
+                {
+                    case SyntaxKind.EnumDeclaration:
+                        return ((EnumDeclarationSyntax)member).Modifiers;
+                    case SyntaxKind.ClassDeclaration:
+                    case SyntaxKind.InterfaceDeclaration:
+                    case SyntaxKind.StructDeclaration:
+                        return ((TypeDeclarationSyntax)member).Modifiers;
+                    case SyntaxKind.DelegateDeclaration:
+                        return ((DelegateDeclarationSyntax)member).Modifiers;
+                    case SyntaxKind.FieldDeclaration:
+                        return ((FieldDeclarationSyntax)member).Modifiers;
+                    case SyntaxKind.EventFieldDeclaration:
+                        return ((EventFieldDeclarationSyntax)member).Modifiers;
+                    case SyntaxKind.ConstructorDeclaration:
+                        return ((ConstructorDeclarationSyntax)member).Modifiers;
+                    case SyntaxKind.DestructorDeclaration:
+                        return ((DestructorDeclarationSyntax)member).Modifiers;
+                    case SyntaxKind.PropertyDeclaration:
+                        return ((PropertyDeclarationSyntax)member).Modifiers;
+                    case SyntaxKind.EventDeclaration:
+                        return ((EventDeclarationSyntax)member).Modifiers;
+                    case SyntaxKind.IndexerDeclaration:
+                        return ((IndexerDeclarationSyntax)member).Modifiers;
+                    case SyntaxKind.OperatorDeclaration:
+                        return ((OperatorDeclarationSyntax)member).Modifiers;
+                    case SyntaxKind.ConversionOperatorDeclaration:
+                        return ((ConversionOperatorDeclarationSyntax)member).Modifiers;
+                    case SyntaxKind.MethodDeclaration:
+                        return ((MethodDeclarationSyntax)member).Modifiers;
+                    case SyntaxKind.GetAccessorDeclaration:
+                    case SyntaxKind.SetAccessorDeclaration:
+                    case SyntaxKind.AddAccessorDeclaration:
+                    case SyntaxKind.RemoveAccessorDeclaration:
+                        return ((AccessorDeclarationSyntax)member).Modifiers;
+                }
+            }
+
+            return default;
+        }
+        public static bool IsFoundUnder<TParent>(this SyntaxNode node, Func<TParent, SyntaxNode> childGetter)
+           where TParent : SyntaxNode
+        {
+            var ancestor = node.GetAncestor<TParent>();
+            if (ancestor == null)
+            {
+                return false;
+            }
+
+            var child = childGetter(ancestor);
+
+            // See if node passes through child on the way up to ancestor.
+            return node.GetAncestorsOrThis<SyntaxNode>().Contains(child);
+        }
+        public static TNode GetAncestor<TNode>(this SyntaxNode node)
+           where TNode : SyntaxNode
+        {
+            var current = node.Parent;
+            while (current != null)
+            {
+                if (current is TNode tNode)
+                {
+                    return tNode;
+                }
+
+                current = current.GetParent();
+            }
+
+            return null;
+        }
+        private static SyntaxNode GetParent(this SyntaxNode node)
+        {
+            return node is IStructuredTriviaSyntax trivia ? trivia.ParentTrivia.Token.Parent : node.Parent;
+        }
+
+        public static TNode FirstAncestorOrSelfUntil<TNode>(this SyntaxNode node, Func<SyntaxNode, bool> predicate)
+            where TNode : SyntaxNode
+        {
+            for (var current = node; current != null; current = current.GetParent())
+            {
+                if (current is TNode tnode)
+                {
+                    return tnode;
+                }
+
+                if (predicate(current))
+                {
+                    break;
+                }
+            }
+
+            return default;
+        }
+
+        public static TNode GetAncestorOrThis<TNode>(this SyntaxNode node)
+                where TNode : SyntaxNode
+        {
+            return node?.GetAncestorsOrThis<TNode>().FirstOrDefault();
+        }
+
+        public static IEnumerable<TNode> GetAncestorsOrThis<TNode>(this SyntaxNode node)
+            where TNode : SyntaxNode
+        {
+            var current = node;
+            while (current != null)
+            {
+                if (current is TNode tNode)
+                {
+                    yield return tNode;
+                }
+
+                current = current.GetParent();
+            }
+        }
+    }
+}
+
--- a/src/OmniSharp.Roslyn.CSharp/Services/Signatures/CheckStaticInvocation.cs
+++ b/src/OmniSharp.Roslyn.CSharp/Services/Signatures/CheckStaticInvocation.cs
@@ -9,7 +9,7 @@ using Microsoft.CodeAnalysis.CSharp.Syntax;
 
 namespace OmniSharp.Roslyn.CSharp.Services.Signatures
 {
-    //TO DO: Reomove this class once a public API for Signature Help from Roslyn is available
+    //TO DO: Remove this class once a public API for Signature Help from Roslyn is available
     internal static class CheckForStaticExtension
     {
         public static bool IsInStaticContext(this SyntaxNode node)
--- a/src/OmniSharp.Roslyn.CSharp/Services/Signatures/CheckStaticInvocation.cs
+++ b/src/OmniSharp.Roslyn.CSharp/Services/Signatures/CheckStaticInvocation.cs
@@ -1,4 +1,6 @@
-﻿using System;
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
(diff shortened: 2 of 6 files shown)


src/OmniSharp.Roslyn.CSharp/Services/Signatures/SignatureHelpService.cs
```diff
@@ -133,7 +133,15 @@ private IEnumerable<IMethodSymbol> GetMethodOverloads(SemanticModel semanticMode
                 return new IMethodSymbol[] { };
             }
 
-            return symbol.ContainingType.GetMembers(symbol.Name).OfType<IMethodSymbol>();
+            var MethodOverloads = symbol.ContainingType.GetMembers(symbol.Name).OfType<IMethodSymbol>();
```
filipw: local variables should start with lowercase

src/OmniSharp.Roslyn.CSharp/Services/Signatures/SignatureHelpService.cs
```diff
@@ -133,7 +133,15 @@ private IEnumerable<IMethodSymbol> GetMethodOverloads(SemanticModel semanticMode
...
+            var BaseType = symbol.ContainingType.BaseType;
+            while(BaseType!=null && BaseType.ContainingNamespace.Name!= "System")
```
filipw: spacing is off here. 
Also, perhaps it would be better to check for `baseTypeSymbol.SpecialType != SpecialType.System_Object` and terminate there instead rather than on "System" namespace

akshita31: The above mentioned change is failing for the case of attribute constructor as the System.Attribute class doesn't have the SpecialType as System_Object, hence for that case we will also get an Attribute.Attribute method which is not required.This is the reason the System Namespace has been used.

src/OmniSharp.Roslyn.CSharp/Services/Signatures/SignatureHelpService.cs
```diff
@@ -133,7 +133,15 @@ private IEnumerable<IMethodSymbol> GetMethodOverloads(SemanticModel semanticMode
                 return new IMethodSymbol[] { };
             }
 
-            return symbol.ContainingType.GetMembers(symbol.Name).OfType<IMethodSymbol>();
+            var methodOverloads = symbol.ContainingType.GetMembers(symbol.Name).OfType<IMethodSymbol>();
+            var baseTypeSymbol = symbol.ContainingType.BaseType;
+            while(baseTypeSymbol != null && baseTypeSymbol.ContainingNamespace.Name != "System")
+            {
+                methodOverloads = methodOverloads.Concat(baseTypeSymbol.GetMembers(symbol.Name).OfType<IMethodSymbol>());
```
rchande: I think this will add inaccessible members with the same name.

tests/OmniSharp.Roslyn.CSharp.Tests/SignatureHelpFacts.cs
```diff
@@ -474,13 +474,40 @@ public Program(Program p, int n)
...
 }";
-
             var actual = await GetSignatureHelp(source);
             Assert.Equal(3, actual.Signatures.Count());
             Assert.Equal(1, actual.ActiveParameter);
             Assert.Contains("ctor2", actual.Signatures.ElementAt(actual.ActiveSignature).Documentation);
         }
 
+        [Fact]
+        public async Task SignatureHelpForOverloadedMethodsInheritance()
```
rchande: Can you add a case for methods with the same name that shouldn't be accessible? Eg:
```C#
    class Program
    {
        static void Main(string[] args)
        {
        }

        private void Foo() { }
    }

    class G : Program
    {
        protected void Foo() { }
    }
```

src/OmniSharp.Roslyn.CSharp/Services/Signatures/SignatureHelpService.cs
akshita31: Yes it is returning all the methods regardless of their accessibility.

tests/OmniSharp.Roslyn.CSharp.Tests/SignatureHelpFacts.cs
```diff
@@ -474,13 +474,69 @@ public Program(Program p, int n)
...
+        c1.MyMethod($$);
+    }
+
+ }";
+            var actual = await GetSignatureHelp(source);
+            Assert.Equal(4, actual.Signatures.Count());
+        }
+
+        [Fact]
+        public async Task SignatureHelpForOverloadedInaccesibleMethods()
```
rchande: Can you also add a test to verify behavior if there's an extension method with the same name?

tests/OmniSharp.Roslyn.CSharp.Tests/SignatureHelpFacts.cs
```diff
@@ -474,13 +474,127 @@ public Program(Program p, int n)
...
+        c1.MyMethod($$);
+    }
+
+ }";
+            var actual = await GetSignatureHelp(source);
+            Assert.Equal(4, actual.Signatures.Count());
+        }
+
+        [Fact]
+        public async Task SignatureHelpForInheritedInaccesibleMethods()
```
DustinCampbell: I'd like to see a few more tests for variants of this scenario:

1. An inherited protected method appears in signature help within the a sub type.

    ```C#
    class A
    {
        protected void M1() { }
    }

    class B : A
    {
        void M2()
        {
            $$ // Both M1() and M2()
        }
    }
    ```

2. An inherited protected method appears in signature help after `this.`.

    ```C#
    class A
    {
        protected void M1() { }
    }

    class B : A
    {
        void M2()
        {
            this.$$ // Both M1() and M2()
        }
    }
    ```

3. An inherited protected method appears in signature help after `base.`.

    ```C#
    class A
    {
        protected void M1() { }
    }

    class B : A
    {
        void M2()
        {
            base.$$ // Only M1()
        }
    }
    ```

4. Only static methods appear in a static context.

    ```C#
    class A
    {
        protected static void M1() { }
        public void M2() { }
    }

    class B : A
    {
        static void M3()
        {
            A.$$ // Only M1()
        }
        public void M4() { }
    }
    ```

    ```C#
    class A
    {
        protected static void M1() { }
        public void M2() { }
    }

    class B : A
    {
        static void M3()
        {
            B.$$ // Both M1() and M3()
        }
        public void M4() { }
    }
    ```

5. Static and instance methods appear in an instance context.

    ```C#
    class A
    {
        protected static void M1() { }
        public void M2() { }
    }

    class B : A
    {
        static void M3()
        {
            $$ // M1(), M2(), M3() and M4() appear
        }
        public void M4() { }
    }
    ```

src/OmniSharp.Roslyn.CSharp/OmniSharp.Roslyn.CSharp.csproj
```diff
@@ -5,6 +5,14 @@
     <PlatformTarget>AnyCPU</PlatformTarget>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
+    <LangVersion>7.1</LangVersion>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <LangVersion>7.1</LangVersion>
+  </PropertyGroup>
```
DustinCampbell: Instead of adding this to a particular project, could you remove this properties here and just add `<LangVersion>7.1</LangVersion>` to https://github.com/OmniSharp/omnisharp-roslyn/blob/master/build/Settings.props so that all projects are using C# 7.1?

src/OmniSharp.Roslyn.CSharp/Services/Signatures/CheckStaticInvocation.cs
```diff
@@ -0,0 +1,175 @@
+﻿using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using System;
+using System.Collections.Generic;
+using System.Linq;
```
DustinCampbell: Sort usings. `System` namespaces go first. (Note that this is a setting in Visual Studio)

src/OmniSharp.Roslyn.CSharp/Services/Signatures/CheckStaticInvocation.cs
```diff
@@ -0,0 +1,175 @@
...
+namespace OmniSharp.Roslyn.CSharp.Services.Signatures
```
DustinCampbell: Missing a line between the using directives and the namespace.

src/OmniSharp.Roslyn.CSharp/Services/Signatures/CheckStaticInvocation.cs
```diff
@@ -0,0 +1,175 @@
...
+{
+    static internal class CheckForStatic
```
DustinCampbell: Sort modifiers. "internal" before "static"

src/OmniSharp.Roslyn.CSharp/Services/Signatures/CheckStaticInvocation.cs
```diff
@@ -0,0 +1,175 @@
...
+            // Global statements are not a static context.
+            if (node.FirstAncestorOrSelf<GlobalStatementSyntax>() != null)
+            {
+                return false;
+            }
+
+            // any other location is considered static
+            return true;
+        }
+        public static SyntaxTokenList GetModifiers(SyntaxNode member)
```
DustinCampbell: Missing a blank line between methods.

src/OmniSharp.Roslyn.CSharp/Services/Signatures/CheckStaticInvocation.cs
```diff
@@ -0,0 +1,175 @@
...
+                {
+                    yield return tNode;
+                }
+
+                current = current.GetParent();
+            }
+        }
+    }
+}
+
```
DustinCampbell: Is this a bunch of code copied from Roslyn? If so, this should be commented per the MIT license in the Roslyn code base. It might also be worth adding a TODO comment to remove this once we've added a proper public API for Signature Help to Roslyn.

src/OmniSharp.Roslyn.CSharp/Services/Signatures/CheckStaticInvocation.cs
```diff
@@ -0,0 +1,175 @@
...
+                {
+                    return tNode;
+                }
+
+                current = current.GetParent();
+            }
+
+            return null;
+        }
+        private static SyntaxNode GetParent(this SyntaxNode node)
```
DustinCampbell: Missing a blank line before this method.

src/OmniSharp.Roslyn.CSharp/Services/Signatures/CheckStaticInvocation.cs
```diff
@@ -0,0 +1,175 @@
...
+            {
+                return false;
+            }
+
+            var child = childGetter(ancestor);
+
+            // See if node passes through child on the way up to ancestor.
+            return node.GetAncestorsOrThis<SyntaxNode>().Contains(child);
+        }
+        public static TNode GetAncestor<TNode>(this SyntaxNode node)
```
DustinCampbell: Missing a blank line before this method.

src/OmniSharp.Roslyn.CSharp/Services/Signatures/CheckStaticInvocation.cs
```diff
@@ -0,0 +1,175 @@
...
+                    case SyntaxKind.SetAccessorDeclaration:
+                    case SyntaxKind.AddAccessorDeclaration:
+                    case SyntaxKind.RemoveAccessorDeclaration:
+                        return ((AccessorDeclarationSyntax)member).Modifiers;
+                }
+            }
+
+            return default;
+        }
+        public static bool IsFoundUnder<TParent>(this SyntaxNode node, Func<TParent, SyntaxNode> childGetter)
```
DustinCampbell: Missing a blank line before this method.

src/OmniSharp.Roslyn.CSharp/Services/Signatures/CheckStaticInvocation.cs
DustinCampbell: By convention, classes containing extension methods are typically named with an `Extensions` suffix?

src/OmniSharp.Roslyn.CSharp/Services/Signatures/SignatureHelpService.cs
```diff
@@ -59,7 +60,25 @@ public SignatureHelpService(OmniSharpWorkspace workspace)
...
                 var types = invocation.ArgumentTypes;
-                foreach (var methodOverload in GetMethodOverloads(invocation.SemanticModel, invocation.Receiver))
+                ISymbol throughSymbol = null;
+                ISymbol throughType = null;
+                var methodGroup = invocation.SemanticModel.GetMemberGroup(invocation.Receiver).OfType<IMethodSymbol>();
+                if (invocation.Receiver is MemberAccessExpressionSyntax)
+                {
+                    var throughExpression = ((MemberAccessExpressionSyntax)invocation.Receiver).Expression;
+                    throughSymbol = invocation.SemanticModel.GetSpeculativeSymbolInfo(invocation.Position, throughExpression, SpeculativeBindingOption.BindAsExpression).Symbol;
+                    throughType = invocation.SemanticModel.GetSpeculativeTypeInfo(invocation.Position, throughExpression, SpeculativeBindingOption.BindAsTypeOrNamespace).Type;
```
DustinCampbell: Why do we need to retrieve speculative symbol and type info here?

src/OmniSharp.Roslyn.CSharp/Services/Signatures/SignatureHelpService.cs
```diff
@@ -59,7 +60,25 @@ public SignatureHelpService(OmniSharpWorkspace workspace)
...
+                    var includeInstance = throughSymbol != null && !(throughSymbol is ITypeSymbol);
+                    var includeStatic = (throughSymbol is INamedTypeSymbol) || throughType != null;
```
DustinCampbell: Why are the two clauses in the statements above reversed? Was that intentional? Why are we checking for null? Doesn't the `is` check already do that for you?

src/OmniSharp.Roslyn.CSharp/Services/Signatures/SignatureHelpService.cs
```diff
@@ -59,7 +60,25 @@ public SignatureHelpService(OmniSharpWorkspace workspace)
...
+                    methodGroup = methodGroup.Where(m => (m.IsStatic && includeStatic) || (!m.IsStatic && includeInstance));
+                }
+
+                else if (invocation.Receiver is SimpleNameSyntax && invocation.IsStatic)
```
DustinCampbell: Unnecessary blank line above this.

tests/OmniSharp.Roslyn.CSharp.Tests/SignatureHelpFacts.cs
```diff
@@ -474,38 +474,294 @@ public Program(Program p, int n)
...
+    public void foo()
     {
+        Class1 c1 = new Class1();
+        c1.MyMethod($$);
     }
-    public bool B(this Program p, int n)
+
+ }";
+            var actual = await GetSignatureHelp(source);
+            Assert.Single(actual.Signatures);
```
DustinCampbell: Should we verify which signature we got?

tests/OmniSharp.Roslyn.CSharp.Tests/SignatureHelpFacts.cs
```diff
@@ -474,38 +474,294 @@ public Program(Program p, int n)
...
+    void M1(int a,int b,int c)
+    {
+        M1($$)
+    }
+    static void M1(int a,int b,int c,int d) { }
+}";
+            var actual = await GetSignatureHelp(source);
+            Assert.Equal(4, actual.Signatures.Count());
+        }
+        [Fact]
```
DustinCampbell: Missing a blank line before this method

src/OmniSharp.Roslyn.CSharp/Services/Signatures/InvocationContext.cs
```diff
@@ -13,23 +13,26 @@ internal class InvocationContext
         public SyntaxNode Receiver { get; }
         public IEnumerable<TypeInfo> ArgumentTypes { get; }
         public IEnumerable<SyntaxToken> Separators { get; }
+        public bool IsStatic { get; }
```
rchande: IsInStaticContext

src/OmniSharp.Roslyn.CSharp/Services/Signatures/SignatureHelpService.cs
rchande: Our heuristic is to show instance members if the thing left of the dot can bind as a non-type and to show static members if the thing left of the dot can bind as a type (and both, in the case of Color Color).

src/OmniSharp.Roslyn.CSharp/Services/Signatures/SignatureHelpService.cs
```diff
@@ -94,19 +113,19 @@ public SignatureHelpService(OmniSharpWorkspace workspace)
                 if (node is InvocationExpressionSyntax invocation && invocation.ArgumentList.Span.Contains(position))
                 {
                     var semanticModel = await document.GetSemanticModelAsync();
-                    return new InvocationContext(semanticModel, position, invocation.Expression, invocation.ArgumentList);
+                    return new InvocationContext(semanticModel, position, invocation.Expression, invocation.ArgumentList, invocation.IsInStaticContext());
```
rchande: Can InvocationContext take care of calling IsInStaticContext()? You've making the same call in all the initializations.

src/OmniSharp.Roslyn.CSharp/Services/Signatures/SignatureHelpService.cs
```diff
@@ -94,19 +113,19 @@ public SignatureHelpService(OmniSharpWorkspace workspace)
...
                 }
 
                 if (node is ObjectCreationExpressionSyntax objectCreation && objectCreation.ArgumentList.Span.Contains(position))
                 {
                     var semanticModel = await document.GetSemanticModelAsync();
-                    return new InvocationContext(semanticModel, position, objectCreation, objectCreation.ArgumentList);
+                    return new InvocationContext(semanticModel, position, objectCreation, objectCreation.ArgumentList,objectCreation.IsInStaticContext());
```
rchande: Missing comma between objectCreation.ArgumentList and objectCreation

src/OmniSharp.Roslyn.CSharp/Services/Signatures/SignatureHelpService.cs
akshita31: In that case we will have to pass the invocation object to InvocationContext and since there are three different types of such objects we will have to add another constructor in InvocationContext. Which is better of the two ?

src/OmniSharp.Roslyn.CSharp/Services/Signatures/SignatureHelpService.cs
akshita31: In the first case if we remove the null check,then if throughSymbol is null,the second statement will evaluate to true (and we want it to be false on a null).
In the second case one statement is checking throughSymbol and the other throughType, so we need to have both of them.

DustinCampbell: I see it now -- thanks!

src/OmniSharp.Roslyn.CSharp/Services/Signatures/SignatureHelpService.cs
DustinCampbell: Is this just copied and pasted from Roslyn?

tests/OmniSharp.Roslyn.CSharp.Tests/SignatureHelpFacts.cs
akshita31: Yes we should.Added that.

src/OmniSharp.Roslyn.CSharp/Services/Signatures/CheckStaticInvocation.cs
```diff
@@ -1,4 +1,6 @@
-﻿using System;
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
```
DustinCampbell: You might add a comment for where this was copied from.

src/OmniSharp.Roslyn.CSharp/Services/Signatures/SignatureHelpService.cs
akshita31: Roslyn uses a different API for doing the same thing.

rchande: Roslyn does awkward type lookup to achieve the same effect: http://source.roslyn.io/#Microsoft.CodeAnalysis.CSharp.Features/SignatureHelp/InvocationExpressionSignatureHelpProvider_MethodGroup.cs,44. @DustinCampbell does what they are doing seem materially different to you?

src/OmniSharp.Roslyn.CSharp/Services/Signatures/SignatureHelpService.cs
rchande: Ok, I'll leave it up to you.

src/OmniSharp.Roslyn.CSharp/Services/Signatures/CheckStaticInvocation.cs
```diff
@@ -0,0 +1,183 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace OmniSharp.Roslyn.CSharp.Services.Signatures
+{
+    //TO DO: Reomove this class once a public API for Signature Help from Roslyn is available
```
rchande: Remove

src/OmniSharp.Roslyn.CSharp/Services/Signatures/SignatureHelpService.cs
DustinCampbell: Yup. It's similar. This looks fine.

PR #1037, opened 2017-11-29
Ensure that we don't load multiple compilation assemby versions

seesharper:
This PR ensures that we don't load multiple versions of the same compilation assembly. This PR only affects loading metadata for scripts targeting .Net Core.

This PR also resolves the issue discussed here. 
https://github.com/filipw/dotnet-script/issues/194

### DIFF
--- a/src/OmniSharp.Script/ScriptProjectSystem.cs
+++ b/src/OmniSharp.Script/ScriptProjectSystem.cs
@@ -112,7 +112,7 @@ namespace OmniSharp.Script
             }
             else
             {
-                HashSet<string> loadedFiles = new HashSet<string>();
+                HashSet<string> loadedFiles = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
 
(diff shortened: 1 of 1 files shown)


src/OmniSharp.Script/ScriptProjectSystem.cs
```diff
@@ -112,10 +112,15 @@ public void Initalize(IConfiguration configuration)
             }
             else
             {
+                HashSet<string> loadedFiles = new HashSet<string>();
```
DustinCampbell: Should this be case-insensitive? (e.g. using `StringComparer.OrdinalIgnoreCase`)

seesharper: That certainly would not hurt 👍  Fixing

filipw:
thanks a lot, we may need to make it more robust (i.e. group by TFM and select reliably) but this will address the immediate issue

PR #1038, opened 2017-11-30
TypeLookUp Structured Documentation Feature to be used by VS Code

akshita31:
Issue : https://github.com/OmniSharp/omnisharp-vscode/issues/1057

To display the XML Documentation properly, we need omnisharp-roslyn to send a documentation object to omnisharp-vscode with different parts of the documentation in separate fields so that newlines can then be added to the documentation appropriately on the vscode side. So a V2 endpoint has been created that will be invoked by omnisharp-vscode and appropriate test cases have been added.

Edit : Instead of a V2 endpoint, a property has been added to V1 with the structuredDocumentation
Please Review @DustinCampbell @TheRealPiotrP @rchande


src/OmniSharp.Roslyn.CSharp/Services/DocumentationConverter.cs
```diff
@@ -126,5 +128,130 @@ private static string GetCref(string cref)
...
+                                    ParamInstance.Append(": ");
+                                    CurrentElementText = ParamInstance;
+                                    Param.Add(ParamInstance);
+                                    break;
+                                case "typeparamref":
+                                    CurrentElementText.Append(xml["name"]);
+                                    CurrentElementText.Append(" ");
+                                    break;
+                                case "typeparam":
+                                    StringBuilder TypeParamInstance = new StringBuilder();
```
rchande: lowecase.

src/OmniSharp.Roslyn.CSharp/Services/DocumentationConverter.cs
```diff
@@ -126,5 +128,130 @@ private static string GetCref(string cref)
...
+                                    TypeParamInstance.Append(TrimMultiLineString(xml["name"], lineEnding));
```
rchande: You can initialize a StringBuilder with some text.

src/OmniSharp.Roslyn.CSharp/Services/DocumentationConverter.cs
```diff
@@ -126,5 +128,130 @@ private static string GetCref(string cref)
...
+                                CurrentElementText.Append(TrimMultiLineString(xml.Value, lineEnding));
+                            }
+                        }
+                    } while (xml.Read());
+                }
+                catch (Exception)
+                {
+                    return null;
+                }
+                docComment.RemarksText = RemarksText.ToString();
```
rchande: Insert blank line after closing brace.

src/OmniSharp.Roslyn.CSharp/Services/DocumentationConverter.cs
```diff
@@ -126,5 +128,130 @@ private static string GetCref(string cref)
...
+            if (string.IsNullOrEmpty(xmlDocumentation))
+                return null;
+            var docComment = new DocumentationComment();
+            var reader = new StringReader("<docroot>" + xmlDocumentation + "</docroot>");
+            StringBuilder RemarksText = new StringBuilder();
+            StringBuilder ExampleText = new StringBuilder();
+            StringBuilder ReturnsText = new StringBuilder();
+            StringBuilder SummaryText = new StringBuilder();
+            StringBuilder ValueText = new StringBuilder();
+            List<StringBuilder> Param = new List<StringBuilder>();
```
rchande: ParamElements?

src/OmniSharp.Roslyn.CSharp/Services/Types/V2/TypeLookUp.cs
```diff
@@ -0,0 +1,56 @@
...
+using OmniSharp.Roslyn.CSharp.Services.Documentation;
+
+namespace OmniSharp.Roslyn.CSharp.Services.Types.V2
+{
+    [OmniSharpHandler(OmniSharpEndpoints.V2.TypeLookup, LanguageNames.CSharp)]
+    public class TypeLookupService : IRequestHandler<TypeLookupRequest, TypeLookupResponse>
+    {
+        private readonly FormattingOptions _formattingOptions;
+        private readonly OmniSharpWorkspace _workspace;
+        private static readonly SymbolDisplayFormat DefaultFormat = SymbolDisplayFormat.FullyQualifiedFormat.
```
rchande: Where did this come from?

src/OmniSharp.Roslyn.CSharp/Services/Types/V2/TypeLookUp.cs
```diff
@@ -0,0 +1,56 @@
...
+            var response = new TypeLookupResponse();
+            if (document != null)
+            {
+                var semanticModel = await document.GetSemanticModelAsync();
+                var sourceText = await document.GetTextAsync();
+                var position = sourceText.Lines.GetPosition(new LinePosition(request.Line, request.Column));
+                var symbol = await SymbolFinder.FindSymbolAtPositionAsync(semanticModel, position, _workspace);
+                if (symbol != null)
+                {
+                    response.Type = symbol.Kind == SymbolKind.NamedType ?
```
rchande: How much of this is duplicated from the v1 endpoint? Wondering if there's a refactoring we can do here.

tests/OmniSharp.Roslyn.CSharp.Tests/TypeLookUpV2Facts.cs
```diff
@@ -0,0 +1,326 @@
...
+    /// <param name=""tagName"">Name of the tag </param>
+    public static bool C$$ompare(int gameObject, string tagName)
+    {
+    }
+}";
+            var response = await GetTypeLookUpResponse(content);
+            var expected =
+            @"gameObject: The game object.";
+            Assert.Equal(2,response.DocComment.Param.Length);
+            Assert.Equal(expected, response.DocComment.Param[0]);
```
rchande: You can combine this test and the one below it and check both parameters.

tests/OmniSharp.Roslyn.CSharp.Tests/TypeLookUpV2Facts.cs
```diff
@@ -0,0 +1,326 @@
...
+    public static T[] m$$kArray<T>(int n)
+    {
+        return new T[n];
+    }
+}";
+            var response = await GetTypeLookUpResponse(content);
+            var expected =
+            @"T: The element type of the array";
+            Assert.Single(response.DocComment.TypeParam);
+            Assert.Equal(expected, response.DocComment.TypeParam[0]);
```
rchande: Can you verify it works for more than 1 type parameter?

tests/OmniSharp.Roslyn.CSharp.Tests/TypeLookUpV2Facts.cs
```diff
@@ -0,0 +1,326 @@
...
+
+    /// <summary>The Name property represents the employee's name.</summary>
+    /// <value>The Name property gets/sets the value of the string field, _name.</value>
+    public string Na$$me
+    {
+    }
+}
+";
+            var response = await GetTypeLookUpResponse(content);
+            var expected =
```
rchande: Why isn't the summary text in the response?

tests/OmniSharp.Roslyn.CSharp.Tests/TypeLookUpV2Facts.cs
```diff
@@ -0,0 +1,326 @@
...
+            }
+}";
+        var response = await GetTypeLookUpResponse(content);
+            var expected =
+@"Summary: DoWork is a method in the TestClass class.
+See also: TestClass.Main ";
+            Assert.Equal(expected.Replace("\r",""), response.DocComment.SummaryText);
+        }
+
+    }
```
rchande: I like that you have tests for each type of section or tag, but there aren't many tests that validate that multiple sections all show up correctly together? Could you add some tests that verify what happens if you have a <summary>, <params> and <returns> all in one doc comment (for example).

Can you also add tests that validate the behavior if <params> elements are interleaved with other elements? eg <param>... <typeparam> ... <param> ...

src/OmniSharp.Roslyn.CSharp/Services/Types/V2/TypeLookUp.cs
filipw: @rchande here is the background https://github.com/OmniSharp/omnisharp-roslyn/pull/665

filipw:
The entire new handler is really a copy of the old one, except for the one line that computes documentation.
Wouldn't it be simpler to just extend the old endpoint with something like `WantStructuredDocumentation` on the request object and just use that?

DustinCampbell:
Is this new endpoint actually necessary? Can't these properties (or a property returning a richer object) by returned by the v1 endpoint?

rchande:
I think @TheRealPiotrP wanted us to start being more diligent about versioning endpoints.

DustinCampbell:
@rchande : That seems reasonable, but it's a change in approach that the OmniSharp community has not taken in the past. I don't think copying and pasting entire endpoints to add a property is worthwhile. :smile: @david-driscoll, how do you feel about this?

rchande:
@DustinCampbell, there's definitely a refactoring that needs to happen. I think @akshita31 is working on it.

DustinCampbell:
FWIW, I'm totally comfortable adding new V2 end points if they make sense. However, general policy decisions around versioning of the protocol need to happen with the OmniSharp community.

filipw:
While proper versioning is always welcome, at the same time don't think it's sustainable to have a new version with every new property being added - it only would make sense for breaking changes. 

Since this (structured documentation) doesn't have to replace the old documentation property and therefore can be seen as an additive and non-breaking change, IMHO it should just be added to the existing endpoint and the complexity of the PR drastically reduced.

rchande:
Fair enough, I don't have super strong feelings about this one way or the other.

david-driscoll:
Yeah, if we can get by using a new property instead of fracturing endpoints it would be preferable.  I'm fine making breaking changes when required, because I want to avoid causing problems for consumers that may be integrating slower, or can't/won't make use of the new endpoint.

src/OmniSharp.Roslyn.CSharp/Services/DocumentationConverter.cs
```diff
@@ -126,5 +128,131 @@ private static string GetCref(string cref)
...
+                            }
+                        }
+                    } while (xml.Read());
+                }
+                catch (Exception)
+                {
+                    return null;
+                }
+
+                docComment.RemarksText = remarksText.ToString();
```
rchande: How do you feel about adding a constructor to docComment that takes all of these, rather than having to remember to set all of the properties? (Maybe makes it easier to not forget a section if we refactor).

PR #1042, opened 2017-12-05
Added Opening Brace as trigger character for formatAfterKeyStroke

akshita31:
Issue : https://github.com/OmniSharp/omnisharp-vscode/issues/1591

Added { as trigger character for formatting changes request.

### DIFF
--- a/src/OmniSharp.Roslyn.CSharp/Workers/Formatting/FormattingWorker.cs
+++ b/src/OmniSharp.Roslyn.CSharp/Workers/Formatting/FormattingWorker.cs
@@ -25,7 +25,7 @@ namespace OmniSharp.Roslyn.CSharp.Workers.Formatting
                     return await GetFormattingChanges(document, targetLine.Start, targetLine.End);
                 }
             }
-            else if (character == '}' || character == ';')
+            else if (character == '}' || character == ';' || character == '{')
             {
                 // format after ; and }
                 var root = await document.GetSyntaxRootAsync();


akshita31:
omnisharp-vscode side : https://github.com/OmniSharp/omnisharp-vscode/pull/1906

DustinCampbell:
I'm a little worried about this change. What happens in the following case?

1. Set `"editor.autoClosingBrackets": false` in the VS Code settings.
2. Start with this code;
    ```C#
    class C
    {
        void M1()
        $$

        void M2()
        {
        }
    }
    ```
3. Type `{` at the caret position (`$$`)

Does the declaration of M2() get indented? If so, this might not be the right approach.

akshita31:
ezgif com-video-to-gif

It doesn't indent M2()

akshita31:
Just noticed a problem with this change. After the format the caret symbol should appear just before the closing brace so that on hitting enter we have a proper indentation

ezgif com-video-to-gif1

DustinCampbell:
Maybe debug and see if it's got the correct span for the format?

akshita31:
Tried to edit the span on hitting enter (for this particular case) to include three lines - the previous line(opening brace) , the current line and the nextLine (closing brace), but doing that deletes the spaces in the current line.
incorrect_cursor

I think this is related to auto-indent in vscode. Hence I have filed a bug [here](https://github.com/Microsoft/vscode/issues/40235)

lastthyme (approved):

lastthyme (approved):

gabrieldechichi:
Any updates on this?

PR #1044, opened 2017-12-05
Don't lock Cake host object assembly.

mholo65:
Don't lock Cake host object assembly (Cake.Core) when loading host object type. We'd like to avoid locking file in workspace.


src/OmniSharp.Host/Services/AssemblyLoader.cs
```diff
@@ -77,5 +81,31 @@ public Assembly LoadFrom(string assemblyPath)
...
+                try
+                {
+                    var bytes = File.ReadAllBytes(assemblyPath);
+                    assembly = Assembly.Load(bytes);
+                }
+                catch (Exception ex)
+                {
+                    _logger.LogError(ex, $"Failed to load assembly from path: {assemblyPath}");
+                }
+                if (assembly != null)
```
DustinCampbell: tiny nit: I'd put a line break before to visually separate the if statement from the catch handler.

src/OmniSharp.Host/Services/AssemblyLoader.cs
```diff
@@ -77,5 +81,31 @@ public Assembly LoadFrom(string assemblyPath)
             _logger.LogTrace($"Assembly loaded from path: {assemblyPath}");
             return assembly;
         }
+
+        public Assembly LoadFrom(string assemblyPath, bool lockAssembly)
```
DustinCampbell: Should the two `LoadFrom` overloads be merged with `lockAssembly` becoming an optional parameter? Also, from an API perspective, I'd sort of prefer to see `dontLockAssemblyOnDisk` since the goal in calling this method is to avoid locking.

src/OmniSharp.Host/Services/AssemblyLoader.cs
```diff
@@ -1,14 +1,18 @@
...
 using System.Reflection;
 using Microsoft.Extensions.Logging;
+using OmniSharp.Utilities;
 
 namespace OmniSharp.Services
 {
     internal class AssemblyLoader : IAssemblyLoader
     {
+        private static readonly ConcurrentDictionary<string, Assembly> AssemblyCache = new ConcurrentDictionary<string, Assembly>(
+            PlatformHelper.IsWindows ? StringComparer.OrdinalIgnoreCase : StringComparer.Ordinal);
```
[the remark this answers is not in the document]
mholo65: Ok, I'll update it to a `StringComparer.OrdinalIgnoreCase` then.

src/OmniSharp.Host/Services/AssemblyLoader.cs
mholo65: I could merge if you like. And yeah, the naming `dontLockAssemblyOnDisk` sounds better, with a default to `false`.

src/OmniSharp.Host/Services/AssemblyLoader.cs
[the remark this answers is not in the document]
mholo65: Additionally, wan't the cache to cover all methods and *not* only when loading from `byte[]`?

PR #1045, opened 2017-12-05
Check if file exist before trying to create MetadataReference.

mholo65:
Just to avoid throwing any uneccesary exceptions.

### DIFF
--- a/src/OmniSharp.Cake/CakeProjectSystem.cs
+++ b/src/OmniSharp.Cake/CakeProjectSystem.cs
@@ -148,7 +148,7 @@ namespace OmniSharp.Cake
                 var document = solution.GetDocument(documentId);
                 var project = document.Project;
 
-                var metadataReferences = e.References.Select(reference => MetadataReference.CreateFromFile(reference, documentation: GetDocumentationProvider(reference)));
+                var metadataReferences = GetMetadataReferences(e.References);
                 var fileReferencesToRemove = project.MetadataReferences;
 
                 foreach (var reference in metadataReferences)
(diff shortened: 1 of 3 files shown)


src/OmniSharp.Cake/CakeProjectSystem.cs
```diff
@@ -148,7 +148,7 @@ private void ScriptReferencesChanged(object sender, ReferencesChangedEventArgs e
                 var document = solution.GetDocument(documentId);
                 var project = document.Project;
 
-                var metadataReferences = e.References.Select(reference => MetadataReference.CreateFromFile(reference, documentation: GetDocumentationProvider(reference)));
+                var metadataReferences = GetMetadataReferences(e.References);
```
filipw: you may want to use https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp.Roslyn/Utilities/MetadataReferenceEqualityComparer.cs to avoid duplicates

DustinCampbell:
Just asking out of curiosity -- should you really just be using the [MetadataFileCacheService](https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp.Roslyn/Services/MetadataFileReferenceCache.cs)?

mholo65:
@DustinCampbell yes, I should be using that! 😄 Will rework this PR to use that instead.

src/OmniSharp.Cake/CakeProjectSystem.cs
mholo65: Actually, `e.References` should be `ISet<string>` and not `IReadOnlyCollection<string>`, so duplicates *should* already be avoided.

DustinCampbell: @filipw is referring to the issue of adding the same assembly reference to a project twice. This can happen if there are multiple copies of the same assembly with different file paths.

david-driscoll:
AppVeyor seems to have died getting nuget pacakges.  kicking.

PR #1047, opened 2017-12-07
Feature/improve bakery resolver

mholo65:
Support resolving from both OmniSharp options and PATH.


david-driscoll:
The mac build looked like a legitimate error, however cryptic.  Restarting to see if that is the case.

PR #1055, opened 2017-12-11
Roslyn 2.6.0 + C# 7.2

filipw:
- upgraded to Roslyn 2.6.0
 - added C# 7.2 support
 - added a C# 7.2 test for scripting (scripting by design should default to latest language version)
 - fixed some tests that contained invalid C#

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 All changes to the project will be documented in this file.
 
+## [not yet released]
+* Updated to Roslyn 2.6.1 packages - C# 7.2 support, PR: [#1055](https://github.com/OmniSharp/omnisharp-roslyn/pull/1055)
+
 ## [1.28.0] - 2017-12-14
 
 * Fixed issue with loading XML documentation for `#r` assembly references in CSX scripts ([#1026](https://github.com/OmniSharp/omnisharp-roslyn/issues/1026), PR: [#1027](https://github.com/OmniSharp/omnisharp-roslyn/pull/1027))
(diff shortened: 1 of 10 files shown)


filipw:
@DustinCampbell  thanks!  there is however a weird issue on Travis which seems legit. any idea what could cause this weird exception on Mono? 

```
[39;49m[31m    OmniSharp.Roslyn.CSharp.Tests.RenameFacts.Rename_DoesTheRightThingWhenDocumentIsNotFound [FAIL]
[39;49m[37m      System.AggregateException : One or more errors occurred.
[39;49m[37m      ---- System.ArgumentException : checksum must be a SHA-1 hash
```

The only related thing I found is this https://github.com/dotnet/roslyn/pull/22794, could it be a regression on Mono of some sort?

DustinCampbell:
I think you're right. The exception is coming from here: https://github.com/dotnet/roslyn/blob/de8b37a43d13aeea3a7a8b6bdf33e8363ada4da2/src/Workspaces/Core/Portable/Workspace/Solution/Checksum.cs#L30.

I've pinged that PR to see it can be ported to Roslyn master. I don't think this is a regression in Mono. It looks more like some subtle change in Roslyn's solution checksum caused it. I'm starting a mail thread internally to raise the priority.

filipw:
Adding a blocked tag - pending https://github.com/dotnet/roslyn/issues/23722

DustinCampbell:
@filipw : The 2.6.1 Roslyn packages are now available on NuGet.

filipw:
perfect! I updated this PR and now the checks are all passing 🎉

filipw:
@david-driscoll @DustinCampbell  I'm punching this in then 😇

PR #1056, opened 2017-12-11
 Watch .csx files in order to add/remove script projects to workspace.

mholo65:
Builds upon PR #1053. (Merge that first, and then rebase this...)

Add created .csx files to workspace and remove deleted .csx files from workspace.

ping @filipw

### DIFF
--- a/src/OmniSharp.Script/ScriptProjectSystem.cs
+++ b/src/OmniSharp.Script/ScriptProjectSystem.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Composition;
 using System.IO;
@@ -12,13 +13,14 @@ using Microsoft.CodeAnalysis.Scripting;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyModel;
 using Microsoft.Extensions.Logging;
+using OmniSharp.FileWatching;
 using OmniSharp.Models.WorkspaceInformation;
 using OmniSharp.Services;
 using OmniSharp.Roslyn.Utilities;
 using LogLevel = Dotnet.Script.DependencyModel.Logging.LogLevel;
 
 namespace OmniSharp.Script
-{   
+{
     [Export(typeof(IProjectSystem)), Shared]
     public class ScriptProjectSystem : IProjectSystem
     {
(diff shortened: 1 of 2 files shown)


src/OmniSharp.Script/ScriptProjectSystem.cs
```diff
@@ -128,37 +135,69 @@ public void Initalize(IConfiguration configuration)
...
+            if (changeType == FileChangeType.Unspecified && !File.Exists(filePath) ||
+                changeType == FileChangeType.Delete)
+            {
+                RemoveFromWorkspace(filePath);
+            }
+
+            if (changeType == FileChangeType.Unspecified && File.Exists(filePath) ||
+                changeType == FileChangeType.Create)
+            {
+                AddToWorkspace(filePath);
```
filipw: Here we'll want to run `TryGetCompilationDependencies` again, because that would resolve `#r nuget:` and `#load nuget:` if these are used inside of the file. However, there is no API at the moment to do it other than for a current folder so this needs to go on a TODO list for later. I plan to adjust some APIs to facilitate that soon.

mholo65: Ah, I see. Dependencies are not added automatically by the NuGet Resolvers (`NuGetSourceReferenceResolver` and `NuGetMetadataReferenceResolver`)?

filipw: That is the long term goal, but at the moment they can't due to a current limitation in Roslyn which prevents a reference from resolving into more than a single `MetadataReference`.

mholo65: Ok, in that case, will it add any value to run `TryGetCompilationDependencies` again here? Since a recently added file will most of the time be blank... Unless, it’s a rename, move or copy from within VSCode?

filipw: I think for now let's just leave it as is - and when we upgrade to "self contained" resolvers the problem will fix itself. If that doesn't come soon enough, we'll do something smarter here - what you have in this PR is a big step forward anyway 👍

filipw: Not to mention for the primary scenario we want to support - CSI - it will already work as expected.

mholo65:
@DustinCampbell changelog updated.

filipw:
@mholo65 there is a changelog conflict that needs resolving now 😄

filipw:
thanks!

mholo65:
@filipw resolved conflict using Github GUI.

Issue #1046, opened 2017-12-07
Space after <c> tag is removed


SIkebe:
In xml documents, a space after `<c>` tag is removed.
`<c>DoWork</c> is a method in the <c>TestClass</c> class.` returns `DoWorkis a method in the TestClassclass.`.

![xml-doc-c-tag](https://user-images.githubusercontent.com/17608272/33692338-2ce55102-db30-11e7-8836-048eb176be41.png)

Here's the omnisharp logs.

```
        ************ Request ************
{"Type":"request","Seq":714,"Command":"/typelookup","Arguments":{"FileName":"<file path>","Line":83,"Column":35,"IncludeDocumentation":true}}
[dbug]: OmniSharp.Stdio.Host
        ************  Response ************ 
{"Request_seq":714,"Command":"/typelookup","Running":true,"Success":true,"Message":null,"Body":{"Type":"void TestClass.DoWork(int Int1)","Documentation":"DoWorkis a method in the TestClassclass.\n"},"Seq":2541,"Type":"response"}
```

rchande:
@akshita31 Can you take a look?

akshita31:
This issue has been fixed in the 1.14 milestone. We just released a beta of 1.14 - https://github.com/OmniSharp/omnisharp-vscode/releases/tag/v1.14.0-beta4. Follow [these instructions](https://github.com/OmniSharp/omnisharp-vscode/releases/tag/v1.14.0-beta4) to install a beta release of C# for VS Code.

PR #1064, opened 2017-12-14
Retain single leading space when appending string in TypeLookUp

akshita31:
Fixes: #1046 

Before concatenating the string, a TrimStart was being called that removed all the whitespaces. Added a call to a function that retains a single space if the string begins with a whitespace. Added a test for the same and modified the existing tests to match the behaviour.

Please review : @DustinCampbell @TheRealPiotrP @rchande

### DIFF
--- a/src/OmniSharp.Abstractions/Models/v1/TypeLookup/DocumentationComment.cs
+++ b/src/OmniSharp.Abstractions/Models/v1/TypeLookup/DocumentationComment.cs
@@ -140,7 +140,25 @@ namespace OmniSharp.Models.TypeLookup
                     return null;
                 }
             }
-            return new DocumentationComment(summaryText.ToString(), typeParamElements.Select(s => s.ToString()).ToArray(), paramElements.Select(s => s.ToString()).ToArray(), returnsText.ToString(), remarksText.ToString(), exampleText.ToString(), valueText.ToString(), exception.Select(s => s.ToString()).ToArray());
+            /* return new DocumentationComment(
+                 TrimMultipleSpaces(summaryText),
+                 typeParamElements.Select(s => TrimMultipleSpaces(s)).ToArray(),
+                 paramElements.Select(s => TrimMultipleSpaces(s)).ToArray(),
+                 TrimMultipleSpaces(returnsText),
+                 TrimMultipleSpaces(remarksText),
+                 TrimMultipleSpaces(exampleText),
+                 TrimMultipleSpaces(valueText),
+                 exception.Select(s => TrimMultipleSpaces(s)).ToArray());*/
+            return new DocumentationComment(
+               summaryText.ToString(),
+               typeParamElements.Select(s => s.ToString()).ToArray(),
+               paramElements.Select(s => s.ToString()).ToArray(),
+               returnsText.ToString(),
+               remarksText.ToString(),
+               exampleText.ToString(),
+               valueText.ToString(),
+               exception.Select(s => s.ToString()).ToArray());
+
         }
 
         private static string TrimMultiLineString(string input, string lineEnding)
@@ -176,6 +194,11 @@ namespace OmniSharp.Models.TypeLookup
             return " " + input.Substring(offset);
         }
 
-        //private static string TrimMultipleSpaces()
+        // Replace multiple spaces occuring together with a single space
+        private static string TrimMultipleSpaces(string input)
+        {
+            //var convertedString = input.ToString();
+            return System.Text.RegularExpressions.Regex.Replace(input, @"[ ]+", " ");
+        }
     }
 }
--- a/src/OmniSharp.Abstractions/Models/v1/TypeLookup/DocumentationComment.cs
+++ b/src/OmniSharp.Abstractions/Models/v1/TypeLookup/DocumentationComment.cs
@@ -147,7 +147,7 @@ namespace OmniSharp.Models.TypeLookup
         private static string TrimMultiLineString(string input, string lineEnding)
         {
             var lines = input.Split(new string[] { "\n", "\r\n" }, StringSplitOptions.RemoveEmptyEntries);
-            return string.Join(lineEnding, lines.Select(l => RetainSingleLeadingSpace(l)));
+            return string.Join(lineEnding, lines.Select(l => TrimStartRetainingSingleLeadingSpace(l)));
         }
 
         private static string GetCref(string cref)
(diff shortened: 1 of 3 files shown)


src/OmniSharp.Abstractions/Models/v1/TypeLookup/DocumentationComment.cs
```diff
@@ -166,6 +166,16 @@ private static string GetCref(string cref)
             }
             return cref + " ";
         }
+
+        private static string RetainSingleLeadingSpace(string input)
```
jonsequitur: Would `TrimStartRetainingSingleLeadingSpace` be clearer in terms of the intent?

src/OmniSharp.Abstractions/Models/v1/TypeLookup/DocumentationComment.cs
```diff
@@ -166,6 +166,16 @@ private static string GetCref(string cref)
...
+        {
+            if (string.IsNullOrWhiteSpace(input))
+                return string.Empty;
+            if (!Char.IsWhiteSpace(input[0]))
+                return input;
+            int offset = input.TakeWhile(c => char.IsWhiteSpace(c)).Count();
```
rchande: I would just write this as a simple for loop instead of linq.

rchande: (It's faster).

src/OmniSharp.Abstractions/Models/v1/TypeLookup/DocumentationComment.cs
```diff
@@ -166,6 +166,16 @@ private static string GetCref(string cref)
...
+            return " " + input.Substring(offset);
```
jonsequitur: The use of LINQ caught my eye here. It looks like this will be a pretty hot path so we might do a little perf tuning.

Would deleting the LINQ expression and doing something like this satisfy the goal?

```csharp
return $" {input.TrimStart()}";
```

src/OmniSharp.Abstractions/Models/v1/TypeLookup/DocumentationComment.cs
```diff
@@ -147,7 +147,7 @@ public static DocumentationComment From(string xmlDocumentation, string lineEndi
         private static string TrimMultiLineString(string input, string lineEnding)
         {
             var lines = input.Split(new string[] { "\n", "\r\n" }, StringSplitOptions.RemoveEmptyEntries);
-            return string.Join(lineEnding, lines.Select(l => l.TrimStart()));
+            return string.Join(lineEnding, lines.Select(l => RetainSingleLeadingSpace(l)));
```
rchande: What happens if you put `  ` (space space) in your list of strings to split on? I think that will give you strings that start with 0 or 1 spaces...

rchande: (Just spitballing, you don't *have* to do this)

src/OmniSharp.Abstractions/Models/v1/TypeLookup/DocumentationComment.cs
rchande: FWIW, we allocate one string fewer by not calling TrimStart.

rchande: Oops, @jonsequitur is right--I don't know why I thought string.substring didn't allocate a string. I like his solution.

tests/OmniSharp.Roslyn.CSharp.Tests/TypeLookupFacts.cs
```diff
@@ -499,7 +499,7 @@ public class TestClass
 }";
             var response = await GetTypeLookUpResponse(content);
             var expected =
-            @"This sample shows how to call the TestClass.GetZero method.
```
TheRealPiotrP: why did we eliminate these spaces?

tests/OmniSharp.Roslyn.CSharp.Tests/TypeLookupFacts.cs
```diff
@@ -499,7 +499,7 @@ public class TestClass
...
+@"This sample shows how to call the TestClass.GetZero  method.
```
TheRealPiotrP: Why are we adding extra whitespace here?

tests/OmniSharp.Roslyn.CSharp.Tests/TypeLookupFacts.cs
```diff
@@ -452,7 +452,7 @@ public class TestClass
 }";
             var response = await GetTypeLookUpResponse(content);
             var expected =
-            @"DoWork is a method in the TestClass class. System.Console.WriteLine(System.String) for information about output statements.";
+            @"DoWork is a method in the TestClass class. System.Console.WriteLine(System.String)  for information about output statements.";
```
TheRealPiotrP: why are we adding extra whitespace here?

tests/OmniSharp.Roslyn.CSharp.Tests/TypeLookupFacts.cs
akshita31: That doesn't matter because the spaces were not a part of the string anyways. Just left-aligned it

src/OmniSharp.Abstractions/Models/v1/TypeLookup/DocumentationComment.cs
akshita31: I dont think that will work. Suppose the user separates two works with 3 spaces "hello     hello1" , then these parts will also be split and joined with a `lineEnding` . That will be unnecessary splitting and incorrect joining.

tests/OmniSharp.Roslyn.CSharp.Tests/TypeLookupFacts.cs
akshita31: Initially `return cref + " ";` in GetCref was appending a space after `TestClass.GetZero` and the space before `method` was being trimmed so overall there was a single space. Now we are retaining the space before `method` as well, hence the output consists of 2 spaces. But this shouldn't be an issue as vscode collapses the multiple spaces in the text (as Hover API uses a MarkDown string)

tests/OmniSharp.Roslyn.CSharp.Tests/TypeLookupFacts.cs
akshita31: Same as above

src/OmniSharp.Abstractions/Models/v1/TypeLookup/DocumentationComment.cs
```diff
@@ -166,6 +166,15 @@ private static string GetCref(string cref)
             }
             return cref + " ";
         }
+
+        private static string TrimStartRetainingSingleLeadingSpace(string input)
+        {
+            if (string.IsNullOrWhiteSpace(input))
+                return string.Empty;
+            if (!Char.IsWhiteSpace(input[0]))
```
rchande: Style: `char` instead of `Char`.

Issue #1036, opened 2017-11-29
The given path's format is not supported


jmbeach:
When trying to run Omnisharp:

`Omnisharp.exe -p 2000 -s "D:\Path\to\my\soluntion.sln"`

I get the error:

```
{"Event":"log","Body":{"LogLevel":"INFORMATION","Name":"OmniSharp.MSBuild.ProjectSystem","Message":"Detecting projects in 'D:\\Path\\to\\my\\solution.sln'."},"Seq":10,"Type":"event"}
{"Event":"log","Body":{"LogLevel":"ERROR","Name":"OmniSharp.Stdio.Host","Message":"The project system 'OmniSharp.MSBuild.ProjectSystem' threw exception during initialization.\r\nSystem.NotSupportedException: The given path's format is not supported.\r\n   at System.Security.Permissions.FileIOPermission.EmulateFileIOPermissionChecks(String fullPath)\r\n   at System.Security.Permissions.FileIOPermission.QuickDemand(FileIOPermissionAccess access, String fullPath, Boolean checkForDuplicates, Boolean needFullPath)\r\n   at OmniSharp.MSBuild.ProjectSystem.GetProjectPathsFromSolution(String solutionFilePath)\r\n   at OmniSharp.MSBuild.ProjectSystem.Initalize(IConfiguration configuration)\r\n   at OmniSharp.WorkspaceInitializer.Initialize(IServiceProvider serviceProvider, CompositionHost compositionHost, IConfiguration configuration, ILogger logger)"},"Seq":11,"Type":"event"}
```

Not sure what's wrong here.

jmbeach:
I did realize I was using the stdin executable instead of http like I meant to, but same issue with http.

```
fail: OmniSharp.Http.Startup[0]
      The project system 'OmniSharp.MSBuild.ProjectSystem' threw exception during initialization.
System.NotSupportedException: The given path's format is not supported.
   at System.Security.Permissions.FileIOPermission.EmulateFileIOPermissionChecks(String fullPath)
   at System.Security.Permissions.FileIOPermission.QuickDemand(FileIOPermissionAccess access, String fullPath, Boolean checkForDuplicates, Boolean needFullPath)
   at OmniSharp.MSBuild.ProjectSystem.GetProjectPathsFromSolution(String solutionFilePath)
   at OmniSharp.MSBuild.ProjectSystem.Initalize(IConfiguration configuration)
[...]

PR #1066, opened 2017-12-15
don't parse ASP.NET websites

jmbeach:
Fixes issue "The given path's format is not supported" #1036 when you have an ASP.NET website in your solution.

### DIFF
--- a/src/OmniSharp.MSBuild/SolutionParsing/ProjectBlock.cs
+++ b/src/OmniSharp.MSBuild/SolutionParsing/ProjectBlock.cs
@@ -43,6 +43,11 @@ namespace OmniSharp.MSBuild.SolutionParsing
             Sections = sections;
         }
 
+        public bool IsAspWebsite()
+        {
+            return RelativePath.StartsWith("http://");
+        }
+
         public static ProjectBlock Parse(string headerLine, Scanner scanner)
         {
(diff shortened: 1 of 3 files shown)


jmbeach:
Travis ci timed out...

src/OmniSharp.MSBuild/SolutionParsing/ProjectBlock.cs
```diff
@@ -43,6 +43,11 @@ private ProjectBlock(string projectTypeGuid, string projectName, string relative
             Sections = sections;
         }
 
+        public bool IsAspWebsite()
+        {
+            return RelativePath.StartsWith("http://");
```
filipw: it should probably be case insensitive. can this be `https://` too? 
btw, do you have a sample project like this? I don't think it's possible to create these types of projects in VS for quite a while now

jmbeach: My project was made in Visual Studio 2013 for my work.

Steps to make one:

- Right click solution item in solution explorer
- Add
    - New website

The option is still there in Visual Studio 2017 Community Edition. Don't know if it works the same.

david-driscoll:
Yay mac builds, going to see if this one passes, if it doesn't I'm sure things change wasn't the culprit and I'll merge anyway.

PR #1069, opened 2017-12-19
allow plugins to be specified by a dll file path

jonsequitur:
This allows plugins to be specified using the path to a dll. This adds to the existing behavior of allowing plugins to be specified using an assembly name.

### DIFF
--- a/src/OmniSharp.LanguageServerProtocol/LanguageServerHost.cs
+++ b/src/OmniSharp.LanguageServerProtocol/LanguageServerHost.cs
@@ -3,7 +3,6 @@ using System.Composition.Hosting;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
-using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Configuration;
@@ -91,10 +90,12 @@ namespace OmniSharp.LanguageServerProtocol
 
             var eventEmitter = new LanguageServerEventEmitter(_server);
             var plugins = _application.CreatePluginAssemblies();
+
+            var assemblyLoader = _serviceProvider.GetRequiredService<IAssemblyLoader>();
             var compositionHostBuilder = new CompositionHostBuilder(_serviceProvider, _environment, eventEmitter)
                 .WithOmniSharpAssemblies()
                 .WithAssemblies(typeof(LanguageServerHost).Assembly)
-                .WithAssemblies(plugins.AssemblyNames.Select(Assembly.Load).ToArray());
+                .WithAssemblies(assemblyLoader.LoadByAssemblyNameOrPath(plugins.AssemblyNames).ToArray());
 
             _compositionHost = compositionHostBuilder.Build();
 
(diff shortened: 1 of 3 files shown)


jonsequitur:
@david-driscoll I could alternately PR this to master. Rebasing from there to the lsp branch will result in a small merge conflict. Let me know if you have a preference.

src/OmniSharp.LanguageServerProtocol/LanguageServerHost.cs
```diff
@@ -94,7 +94,7 @@ private void CreateCompositionHost(InitializeParams initializeParams)
             var compositionHostBuilder = new CompositionHostBuilder(_serviceProvider, _environment, eventEmitter)
                 .WithOmniSharpAssemblies()
                 .WithAssemblies(typeof(LanguageServerHost).Assembly)
-                .WithAssemblies(plugins.AssemblyNames.Select(Assembly.Load).ToArray());
+                .WithAssemblies(plugins.LoadAssemblies().ToArray());
```
filipw: i think you could do `var assemblyLoader = _serviceProvider.GetRequiredService<IAssemblyLoader>();` here and use the OmniSharp assembly loader instead, this way the assembly loading logic would be always the same

jonsequitur: Thanks, I didn't notice that. 

`IAssemblyLoader` has different methods for `Load` (which in the one implementation calls `Assembly.Load`) and `LoadFrom` (which calls `Assembly.LoadFrom`) so the method for deciding which to call based on file existence still needs to be added somewhere. It could be added to `IAssemblyLoaderExtensions` but would need a clearer name (e.g. `LoadByAssemblyNameOrPath` would be clear but verbose.)

The question is, since this is rationalizing command-line arguments for the `-pl` option, maybe resolving this ambiguity belongs closer to the parsing operation anyway?

jonsequitur: @filipw Thoughts on the latest changes?

filipw: sorry for the delay - what I actually meant is that this logic would ideally utilize the already existing methods of the loader (or its extension methods) - as they seem to cover everything needed.

 For example there is already a method to load an assembly by string name (here https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp.Abstractions/Services/IAssemblyLoader.cs#L23), by multiple names (here https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp.Abstractions/Services/IAssemblyLoader.cs#L34) or to load it from a given path (here https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp.Host/Services/AssemblyLoader.cs#L64). So you don't need to add any new ones, you just need to make the decision if it's a file or not a file and call the relevant existing method.

The advantage of this is also that the existing methods handle exceptions so they would prevent the logic from crashing already at startup and they log so we could easily see in the server logs which assemblies have been successfully loaded as plugins.

PR #1072, opened 2017-12-21
AssemblyLoader.Load should not throw if loading fails.

mholo65:
@filipw mentioned this in Slack.

### DIFF
--- a/src/OmniSharp.Host/Services/AssemblyLoader.cs
+++ b/src/OmniSharp.Host/Services/AssemblyLoader.cs
@@ -20,18 +20,17 @@ namespace OmniSharp.Services
 
         public Assembly Load(AssemblyName name)
         {
-            Assembly result;
+            Assembly result = null;
             try
             {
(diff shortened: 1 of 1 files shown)


src/OmniSharp.Host/Services/AssemblyLoader.cs
```diff
@@ -20,18 +20,17 @@ public AssemblyLoader(ILoggerFactory loggerFactory)
 
         public Assembly Load(AssemblyName name)
         {
-            Assembly result;
+            Assembly result = null;
             try
             {
                 result = Assembly.Load(name);
+                _logger.LogTrace($"Assembly loaded: {name}");
```
rchande: Looks like the only consumer of this method already checks for null; sounds good.

rchande:
(opening and closing to retrigger travis)

DustinCampbell:
Looks like the macOS machines are slower and slower. I kicked Travis directly.

PR #1075, opened 2017-12-22
Added attribute span property for fileMemberElement

akshita31:
Issue : https://github.com/OmniSharp/omnisharp-vscode/issues/429

Added properties to contain the attribute span in a file member element, which will be used by the editor to display the "# of references" before the attributes.

### DIFF
--- a/src/OmniSharp.Abstractions/Models/v1/MembersTree/FileMemberElement.cs
+++ b/src/OmniSharp.Abstractions/Models/v1/MembersTree/FileMemberElement.cs
@@ -9,6 +9,10 @@ namespace OmniSharp.Models.MembersTree
 
         public QuickFix Location { get; set; }
 
+        public int AttributeSpanStart { get; set; }
+
+        public int AttributeSpanEnd { get; set; }
+
         public string Kind { get; set; }
         
         public ICollection<SyntaxFeature> Features { get; } = new List<SyntaxFeature>();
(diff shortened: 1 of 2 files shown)


akshita31:
Omnisharp-vscode side : https://github.com/OmniSharp/omnisharp-vscode/pull/1938

src/OmniSharp.Abstractions/Models/v1/MembersTree/FileMemberElement.cs
```diff
@@ -9,6 +9,10 @@ public class FileMemberElement : IComparable<FileMemberElement>
 
         public QuickFix Location { get; set; }
 
+        public int AttributeSpanStart { get; set; }
```
rchande: Can we make these some kind of range or span type?

akshita31: We could pass both AttributeSpanStart and AttributeSpanEnd in a Span type class. But in that case we will also have to define this class in protocol.ts in omnisharp-vscode. Should that be done ?

rchande: Protocol.ts defines a range type: https://github.com/OmniSharp/omnisharp-vscode/blob/master/src/omnisharp/protocol.ts#L450

akshita31: @rchande Yes it does, but the range has the start and end as `Point` type. And I couldn't spot a function in omnisharp-roslyn that converts from a `span` to `range`

Issue #758, opened 2017-02-08
Properly order Code Actions


DustinCampbell:
Today, code actions are ordered with some [very suspicious logic](https://github.com/OmniSharp/omnisharp-roslyn/blob/v1.9-beta25/src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/BaseCodeActionService.cs#L75-L76).

Roslyn defines an [ExtensionOrderAttribute](https://github.com/dotnet/roslyn/blob/version-2.0.0-rc3/src/Workspaces/Core/Portable/ExtensionOrderAttribute.cs) that is applied to code fixes to allow them to state their order dependencies. OmniSharp should use this to create a sensible order.

david-driscoll:
... hey the `.Reverse()` usually orders them in alphabetical order... 👍

DustinCampbell:
:smile:

filipw:
would be great if [this](https://github.com/dotnet/roslyn/blob/version-2.0.0/src/Workspaces/Core/Portable/Shared/Utilities/ExtensionOrderer.cs) was public

DustinCampbell:
Yeah, I know. It's a bit silly.

akshita31:
This issue has been fixed in the 1.14 milestone. We just released a beta of 1.14 - https://github.com/OmniSharp/omnisharp-vscode/releases/tag/v1.14.0-beta4. Follow [these instructions](https://github.com/OmniSharp/omnisharp-vscode/wiki/Installing-Beta-Releases) to install a beta release of C# for VS Code.

PR #1078, opened 2018-01-02
Order code actions

akshita31:
Fixes : #758 

To order code actions we can order the providers based on the ExtensionOrder attribute in roslyn.There are two types of providers - CodeFixProvider and CodeRefactoringProvider. For each of these categories we perform the following on a list of providers : 

1. For each provider create a ProviderNode that includes the Before and After strings and the ProviderName
2. Create a dictionary that includes a mapping from the provider name to the providernode
3. Create a graph based on the Before and After properties of the node using the dictionary created above
4. Perform topological sorting on the nodes. 

Please review @DustinCampbell @TheRealPiotrP @rchande @filipw


src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/CodeActionsOrder.Graph.cs
```diff
@@ -0,0 +1,75 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace OmniSharp.Roslyn.CSharp.Services.Refactoring.V2
+{
+    internal class Graph<T>
```
rchande: Is this adapted from Roslyn?

src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/BaseCodeActionService.cs
```diff
@@ -128,26 +125,54 @@ private TextSpan GetTextSpan(ICodeActionRequest request, SourceText sourceText)
...
+            foreach (var provider in this.Providers)
+            {
+                foreach (var codeFixProvider in provider.CodeFixProviders)
+                {
+                    nodesList.Add(ProviderNode<CodeFixProvider>.From(codeFixProvider));
                 }
             }
+
+            var graph = Graph<CodeFixProvider>.GetGraph(nodesList);
+            return graph.TopologicalSort();
```
rchande: Can we cache this? (Eg, does the total set of CodeFixProviders change)

src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/CodeActionsOrder.Graph.cs
akshita31: Yes, this is taken from [Extension Orderer](https://github.com/dotnet/roslyn/blob/614299ff83da9959fa07131c6d0ffbc58873b6ae/src/Workspaces/Core/Portable/Shared/Utilities/ExtensionOrderer.cs) and modified a little for the current case.

rchande: Ok, we might want to add a comment stating as such.

filipw: @DustinCampbell can we get that API opened up publicly? 😀

src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/BaseCodeActionService.cs
```diff
@@ -23,6 +23,8 @@ public abstract class BaseCodeActionService<TRequest, TResponse> : IRequestHandl
     {
         protected readonly OmniSharpWorkspace Workspace;
         protected readonly IEnumerable<ICodeActionProvider> Providers;
+        protected List<CodeFixProvider> OrderedCodeFixProviders;
```
rchande: You can make this a `Lazy<IEnumerable<CodeFixProvider>` and initialize it like
```
_lazyFixProviders = new Lazy(() => GetOrderedProviders)```

That will let you avoid some of the crazy exception handling below.

File: src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/BaseCodeActionService.cs [hunk 4]
```diff
@@ -51,6 +55,41 @@ protected BaseCodeActionService(OmniSharpWorkspace workspace, CodeActionHelper h
             }
         }
 
+        private void SetOrderedProviders()
+        {
+            try
+            {
```
rchande: Why is this in a try/catch? What are you expecting to fail?

akshita31: the GetSortedCodeFixProviders() will return an exception if there are cycles in the graph. This try catch will handle the exception and return the unordered list if that exception occurs

filipw:
green!

PR #1084, opened 2018-01-11
Detect legacy ASP.NET Websites via project GUID

filipw:
This is a follow up to #1066.

I just noticed the logic there wasn't sufficient as these projects don't necessarily start with `http://` - for example it could be `localhost` or any other string really (up to the user).
 
However, the legacy ASP.NET Websites have a common well-known project GUID so we can use that to exclude it from OmniSharp (since they don't have csproj files, we can't support them at the moment).

I threw in a test just to be on the safe side.


filipw:
thanks - it's green now, so I will merge this to unclog the PR pipeline a bit

PR #1085, opened 2018-01-12
Using StructuredDocumentation for Signature Help and appropriate documentation for Parameter symbols

akshita31:
Changes:

1. Added `StructuredDocumentation` field to SignatureHelpResponse to display signature help documentation properly.
2. Added code for returning documentation for parameters, typeparameters and alias types, by adding a class `DocumentationHelper` and the corresponding methods.
3. Modified DocumentationComment to return empty object instead of null if the xmlDocumentation is empty.
4. Modified the `GetStructuredDocumentation` - to be used by both TypeLookUp and SignatureHelp to check for the type of symbol and return the `DocumentationComment` object accordingly.
5. Modified SignatureHelp to use the `StructuredDocumentation` for the signature and get the correct documentation for the parameter.
6. Modified TypeLookUp according to the changes.
7. Added test cases for the SignatureHelp and TypeLookUp.

Please review : @DustinCampbell @rchande @TheRealPiotrP @david-driscoll @filipw


akshita31:
With this change the Hover in VSCode will show documentation for parameters : 
hover_param_doc_showing

akshita31:
Omnisharp-vscode side : https://github.com/OmniSharp/omnisharp-vscode/pull/1958

src/OmniSharp.Roslyn.CSharp/Services/Signatures/SignatureHelpService.cs
```diff
@@ -168,19 +169,19 @@ private static SignatureHelpItem BuildSignature(IMethodSymbol symbol)
             signature.Documentation = symbol.GetDocumentationCommentXml();
```
akshita31: @rchande @DustinCampbell Do we want to return the Parsed xml for other editors as well. Currently GetDocumentationCommentXml() gives the xml with the tags.

src/OmniSharp.Abstractions/Models/v1/TypeLookup/DocumentationComment.cs
```diff
@@ -30,8 +30,22 @@ private DocumentationComment(string summaryText, DocumentationItem[] typeParamEl
             Exception = exception;
         }
 
+        private DocumentationComment()
```
rchande: You could make a `this()` constructor call and call your existing constructor with the default values (to avoid having to write the assignments over and over again).
Alternatively, you could remove this constructor and make all the parameters optional in the one above this.

src/OmniSharp.Abstractions/Models/v1/TypeLookup/DocumentationComment.cs
```diff
@@ -9,16 +9,16 @@ namespace OmniSharp.Models.TypeLookup
...
-        public DocumentationItem[] TypeParamElements { get; }
-        public DocumentationItem[] ParamElements { get; }
-        public string ReturnsText { get; }
-        public string RemarksText { get; }
-        public string ExampleText { get; }
-        public string ValueText { get; }
-        public DocumentationItem[] Exception { get; }
-
-        private DocumentationComment(string summaryText, DocumentationItem[] typeParamElements, DocumentationItem[] paramElements, string returnsText, string remarksText, string exampleText, string valueText, DocumentationItem[ ] exception)
+        public string SummaryText { get; private set; }
```
rchande: Why setters? You should be able to set get-only in the constructor.

src/OmniSharp.Abstractions/Models/v1/TypeLookup/DocumentationComment.cs
```diff
@@ -30,8 +30,22 @@ private DocumentationComment(string summaryText, DocumentationItem[] typeParamEl
...
+            RemarksText = string.Empty;
+            ExampleText = string.Empty;
+            ValueText = string.Empty;
+            Exception = Array.Empty<DocumentationItem>();
+        }
+
         public static DocumentationComment From(string xmlDocumentation, string lineEnding)
         {
+            if (string.IsNullOrEmpty(xmlDocumentation))
+                return Empty;
```
rchande: Newline.

src/OmniSharp.Abstractions/Models/v1/TypeLookup/DocumentationComment.cs
```diff
@@ -175,6 +189,32 @@ private static string TrimStartRetainingSingleLeadingSpace(string input)
                 return input;
             return $" {input.TrimStart()}";
         }
+
+        //Gets the parameter documentation from this object
+        public string GetParameterText(string name)
+        {
+            var requiredParam = Array.Find(ParamElements, parameter => parameter.Name == name);
```
rchande: You could do an expression body here:
`GetParameterText() => Array.Find(...)?.Documentation ?? string.Empty.

src/OmniSharp.Abstractions/Models/v1/TypeLookup/DocumentationComment.cs
```diff
@@ -175,6 +189,32 @@ private static string TrimStartRetainingSingleLeadingSpace(string input)
...
+        {
+            var requiredParam = Array.Find(TypeParamElements, typeParam => typeParam.Name == name);
+            if (requiredParam != null)
+                return requiredParam.Documentation;
+            return string.Empty;
+        }
+
+        public static DocumentationComment WithSummaryText(string summaryText)
+        {
+            var docComment = new DocumentationComment();
```
rchande: Yeah, let's just make the arguments to the first constructor optional.

src/OmniSharp.Roslyn.CSharp/Services/DocumentationConverter.cs
```diff
@@ -131,10 +132,40 @@ private static string GetCref(string cref)
...
-            if (string.IsNullOrEmpty(xmlDocumentation))
-                return null;
             return DocumentationComment.From(xmlDocumentation, lineEnding);
         }
+
+        public static DocumentationComment GetStructuredDocumentation(ISymbol symbol, string lineEnding = "\n")
+        {
+            switch (symbol)
+            {
+                case IParameterSymbol parameter: return DocumentationComment.WithSummaryText(DocumentationHelper.GetParameterDocumentation(parameter, lineEnding));
```
rchande: Typically the case code goes on the line after the case label.

rchande: Wouldn't this be nicer if it was `New DocumentationComment(summaryText = DocumentationHelper.GetParameterer(...`?

src/OmniSharp.Roslyn.CSharp/Services/DocumentationConverter.cs
```diff
@@ -131,10 +132,40 @@ private static string GetCref(string cref)
...
+                case ITypeParameterSymbol typeParam: return DocumentationComment.WithSummaryText(DocumentationHelper.GetTypeParameterDocumentation(typeParam, lineEnding));
+                case IAliasSymbol alias: return DocumentationComment.WithSummaryText(DocumentationHelper.GetAliasDocumentation(alias, lineEnding));
+                default: return GetStructuredDocumentation(symbol.GetDocumentationCommentXml(), lineEnding);
+            }
+        }
+    }
+
+    public static class DocumentationHelper
+    {
+        public static string GetParameterDocumentation(IParameterSymbol parameter, string lineEnding = "\n")
```
rchande: What's the advantage of putting these in another class versus inside DocumentationConverter? Do you imagine other code using these methods?

rchande: NVM, used below. Can you move this class to its own file?

tests/OmniSharp.Roslyn.CSharp.Tests/SignatureHelpFacts.cs
```diff
@@ -758,6 +758,63 @@ bool LocalFunction(int i)
...
+        return true;
+    }
+}";
+            var actual = await GetSignatureHelp(source);
+            Assert.Single(actual.Signatures);
+
+            var signature = actual.Signatures.ElementAt(0);
+            Assert.Equal(2, signature.Parameters.Count());
+            Assert.Equal("gameObject", signature.Parameters.ElementAt(0).Name);
+            Assert.Equal("The game object.", signature.Parameters.ElementAt(0).Documentation);
```
rchande: What happens if there are additional tags inside the parameter documentation?

src/OmniSharp.Roslyn.CSharp/Services/Signatures/SignatureHelpService.cs
```diff
@@ -168,19 +169,19 @@ private static SignatureHelpItem BuildSignature(IMethodSymbol symbol)
...
             signature.Name = symbol.MethodKind == MethodKind.Constructor ? symbol.ContainingType.Name : symbol.Name;
             signature.Label = symbol.ToDisplayString(SymbolDisplayFormat.MinimallyQualifiedFormat);
+            signature.StructuredDocumentation = DocumentationConverter.GetStructuredDocumentation(symbol);
```
rchande: It looks like this code is calling symbol.GetDocumentationCommentXml once for the method itself and once for every parameter in the method. Is there a way we can just parse the XML once?

tests/OmniSharp.Roslyn.CSharp.Tests/SignatureHelpFacts.cs
akshita31: The StructuredDocumentation removes all the tags and appends the text to the appropriate section hence any additional tags inside the parameter documentation will be processed accordingly.
Added a test to check for the same.

src/OmniSharp.Abstractions/Models/v1/TypeLookup/DocumentationComment.cs
```diff
@@ -175,6 +186,15 @@ private static string TrimStartRetainingSingleLeadingSpace(string input)
                 return input;
             return $" {input.TrimStart()}";
         }
+
+        //Gets the parameter documentation from this object
```
rchande: Make this a doc comment or remove?

src/OmniSharp.Abstractions/Models/v1/TypeLookup/DocumentationComment.cs
```diff
@@ -18,20 +18,31 @@ public class DocumentationComment
         public string ValueText { get; }
         public DocumentationItem[] Exception { get; }
 
-        private DocumentationComment(string summaryText, DocumentationItem[] typeParamElements, DocumentationItem[] paramElements, string returnsText, string remarksText, string exampleText, string valueText, DocumentationItem[ ] exception)
+        public DocumentationComment
+            (string summaryText = "",
```
rchande: Nit: paren should be on the same line as DocumentationComment

src/OmniSharp.Abstractions/Models/v1/TypeLookup/DocumentationComment.cs
```diff
@@ -18,20 +18,31 @@ public class DocumentationComment
...
+            string returnsText = "",
+            string remarksText = "",
+            string exampleText = "",
+            string valueText = "",
+            DocumentationItem[] exception = null)
         {
             SummaryText = summaryText;
-            TypeParamElements = typeParamElements;
-            ParamElements = paramElements;
+            TypeParamElements = typeParamElements ?? new DocumentationItem[0];
```
rchande: Array.Empty

src/OmniSharp.Roslyn.CSharp/Services/DocumentationHelper.cs
```diff
@@ -0,0 +1,26 @@
...
+using OmniSharp.Roslyn.CSharp.Services.Documentation;
+
+namespace OmniSharp.Roslyn.CSharp.Services
+{
+    public static class DocumentationHelper
+    {
+        public static string GetParameterDocumentation(IParameterSymbol parameter, string lineEnding = "\n")
+        {
+            var contaningSymbolDef = parameter.ContainingSymbol.OriginalDefinition;
+            return DocumentationConverter.GetStructuredDocumentation(contaningSymbolDef.GetDocumentationCommentXml(), lineEnding).GetParameterText(parameter.Name);
```
rchande: I would wrap a line of this length.

src/OmniSharp.Roslyn.CSharp/Services/Signatures/SignatureHelpService.cs
```diff
@@ -168,19 +169,19 @@ private static SignatureHelpItem BuildSignature(IMethodSymbol symbol)
...
 
             signature.Parameters = symbol.Parameters.Select(parameter =>
             {
                 return new SignatureHelpParameter()
                 {
                     Name = parameter.Name,
                     Label = parameter.ToDisplayString(SymbolDisplayFormat.MinimallyQualifiedFormat),
-                    Documentation = parameter.GetDocumentationCommentXml()
+                    Documentation = signature.StructuredDocumentation.GetParameterText(parameter.Name)
```
rchande: Ah, now there's only one caller for each of those methods in DocumentationHelper. Can they move inside StructuredDocumentation?

PR #1095, opened 2018-01-24
Attempt to fix rename file same as type issue.

savpek:
Fixes: https://github.com/OmniSharp/omnisharp-vscode/issues/1384

However i have question: 

Issue was that this code was never invoked because `WantsAllCodeActionOperations` was not set. Is this correct fix or is real underlaying issue that vscode-omnisharp extension doesn't set that true on request? I have no idea what WantsAllCodeActionOperations is for.


rchande:
@savpek This needs a change on the VS Code side to work (https://github.com/OmniSharp/omnisharp-vscode/pull/1854/). I'm planning to merge that today.

DustinCampbell:
@rchande : Is this particular change unnecessary with your omnisharp-vscode change?

rchande:
Yes-- @savpek I don't think this change is necessary any more.

PR #1103, opened 2018-02-02
Feature/script supersede lower versions

filipw:
Conflicting assembly versions in scripting are typically resolved via [identity comparer](https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp.Script/ScriptHelper.cs#L63), which we already use. 

However, scripting scenarios in Roslyn additionally set [a special flag in the ReferenceManager](https://github.com/dotnet/roslyn/blob/version-2.6.0-beta3/src/Compilers/Core/Portable/ReferenceManager/CommonReferenceManager.State.cs#L473-L484) to enable additional tie-breaking behavior in case of conflicting references. This flag [is propagated from](https://github.com/dotnet/roslyn/blob/version-2.6.0-beta3/src/Compilers/CSharp/Portable/Compilation/CSharpCompilation.cs#L199) the compilation options, and is always implicitly injected for script compilations.

To mirror 1:1 the runtime behavior of scripts, we should ensure the same flag is set at tooling level (at the moment we have it set to to `false` which can lead to edge case false-positive intellisense errors). Unfortunately the flag is internal so it's set with reflection (along with a few other things we do with reflection in script tooling ☹️).

### DIFF
--- a/src/OmniSharp.Script/ScriptHelper.cs
+++ b/src/OmniSharp.Script/ScriptHelper.cs
@@ -16,6 +16,7 @@ namespace OmniSharp.Script
     {
         private const string BinderFlagsType = "Microsoft.CodeAnalysis.CSharp.BinderFlags";
         private const string TopLevelBinderFlagsProperty = "TopLevelBinderFlags";
+        private const string ReferencesSupersedeLowerVersionsProperty = "ReferencesSupersedeLowerVersions_internal_protected_set";
         private const string IgnoreCorLibraryDuplicatedTypesField = "IgnoreCorLibraryDuplicatedTypes";
         private const string RuntimeMetadataReferenceResolverType = "Microsoft.CodeAnalysis.Scripting.Hosting.RuntimeMetadataReferenceResolver";
         private const string ResolverField = "_resolver";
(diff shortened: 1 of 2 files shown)


src/OmniSharp.Script/ScriptHelper.cs
```diff
@@ -73,6 +74,14 @@ private CSharpCompilationOptions CreateCompilationOptions()
                 topLevelBinderFlagsProperty?.SetValue(compilationOptions, ignoreCorLibraryDuplicatedTypesValue);
             }
 
+            // in scripts, the option to supersede lower versions is ALWAYS enabled
+            // see: https://github.com/dotnet/roslyn/blob/version-2.6.0-beta3/src/Compilers/CSharp/Portable/Compilation/CSharpCompilation.cs#L199
+            var referencesSupersedeLowerVersionsProperty = typeof(CompilationOptions).GetProperty(ReferencesSupersedeLowerVersionsProperty, BindingFlags.Instance | BindingFlags.NonPublic);
+            if (referencesSupersedeLowerVersionsProperty != null)
+            {
+                referencesSupersedeLowerVersionsProperty?.SetValue(compilationOptions, true);
```
DustinCampbell: Is the `?.` needed after checking for null above?

david-driscoll: Technically the if could be removed couldn't it?

filipw: you are both right 😀

filipw:
I will merge this now 😀

screen shot 2018-02-13 at 14 40 40

PR #1106, opened 2018-02-05
A couple of tweaks to platform-sniffing code

DustinCampbell:
1. Trim the osName and osArch values returned by `uname`
2. Ensure that the `/etc/os-release` file exists before reading it

### DIFF
--- a/src/OmniSharp.Abstractions/Utilities/Platform.cs
+++ b/src/OmniSharp.Abstractions/Utilities/Platform.cs
@@ -88,8 +88,8 @@ namespace OmniSharp.Utilities
                 }
 
                 var values = output.Split(' ');
-                var osName = values[0];
-                var osArch = values[1];
+                var osName = values[0].Trim();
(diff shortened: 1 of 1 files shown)


src/OmniSharp.Abstractions/Utilities/Platform.cs
```diff
@@ -88,8 +88,8 @@ private static Platform GetCurrentPlatform()
                 }
 
                 var values = output.Split(' ');
-                var osName = values[0];
-                var osArch = values[1];
+                var osName = values[0].Trim();
+                var osArch = values[1].Trim();
```
david-driscoll: Will the array length ever possibly be `1`?

DustinCampbell: Yes, if `uname` fails. Although, that's already a potential problem here and I haven't seen anything fail yet due to that.

PR #1107, opened 2018-02-07
Disable cake diagnostics endpoint for requests that don't specify a file

rchande:
Cake's diagnostics engine calls the generic CodeCheckService, which generates all diagnostics for the solution if called without a specific file path. This results in duplicate errors in C# projects when VS Code requests project level diagnostics. The solution is to only compute diagnostics for specified files.

Fixes https://github.com/OmniSharp/omnisharp-vscode/issues/1830

Diagnostics window in VS Code now looks like:
image

cc @mholo65

### DIFF
--- a/src/OmniSharp.Cake/Services/RequestHandlers/CakeRequestHandler.cs
+++ b/src/OmniSharp.Cake/Services/RequestHandlers/CakeRequestHandler.cs
@@ -54,15 +54,14 @@ namespace OmniSharp.Cake.Services.RequestHandlers
 
             request = await TranslateRequestAsync(request);
 
-            var response = await HandleCore(request, service);
+            var response = IsValid(request)
+                ? await service.Handle(request)
+                : default(TResponse)
 
             return await TranslateResponse(response, request);
         }
 
-        public virtual  Task<TResponse> HandleCore(TRequest request, IRequestHandler<TRequest, TResponse> service)
-        {
-            return service.Handle(request);
-        }
+        public virtual bool IsValid(TRequest request) => true;
 
         protected virtual async Task<TRequest> TranslateRequestAsync(TRequest req)
         {
--- a/src/OmniSharp.Cake/Services/RequestHandlers/CakeRequestHandler.cs
+++ b/src/OmniSharp.Cake/Services/RequestHandlers/CakeRequestHandler.cs
@@ -54,9 +54,14 @@ namespace OmniSharp.Cake.Services.RequestHandlers
 
             request = await TranslateRequestAsync(request);
 
-            var response = await service.Handle(request);
+            var response = await HandleCore(request, service);
 
             return await TranslateResponse(response, request);
+        }
+
+        public virtual  Task<TResponse> HandleCore(TRequest request, IRequestHandler<TRequest, TResponse> service)
+        {
+            return service.Handle(request);
         }
 
         protected virtual async Task<TRequest> TranslateRequestAsync(TRequest req)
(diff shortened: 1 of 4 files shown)


src/OmniSharp.Cake/Services/RequestHandlers/CakeRequestHandler.cs
```diff
@@ -54,11 +54,16 @@ protected CakeRequestHandler(OmniSharpWorkspace workspace)
 
             request = await TranslateRequestAsync(request);
 
-            var response = await service.Handle(request);
+            var response = await HandleCore(request, service);
 
             return await TranslateResponse(response, request);
         }
 
+        public virtual  Task<TResponse> HandleCore(TRequest request, IRequestHandler<TRequest, TResponse> service)
```
mholo65: Is it necessary to introduce this? `Task<TResponse> Handle(TRequest request)` is already `virtual`.

src/OmniSharp.Cake/Services/RequestHandlers/Diagnostics/CodeCheckHandler.cs
```diff
@@ -14,5 +15,15 @@ public class CodeCheckHandler : CakeRequestHandler<CodeCheckRequest, QuickFixRes
             : base(workspace)
         {
         }
+
+        public override Task<QuickFixResponse> HandleCore(CodeCheckRequest request, IRequestHandler<CodeCheckRequest, QuickFixResponse> service)
```
mholo65: Override `Task<TResponse> Handle(TRequest request)` instead.

src/OmniSharp.Cake/Services/RequestHandlers/Diagnostics/CodeCheckHandler.cs
```diff
@@ -14,5 +15,15 @@ public class CodeCheckHandler : CakeRequestHandler<CodeCheckRequest, QuickFixRes
...
+        {
+            if (string.IsNullOrEmpty(request.FileName))
+            {
+                return Task.FromResult(new QuickFixResponse());
+            }
+
+            return service.Handle(request);
```
mholo65: Override Task<TResponse> Handle(TRequest request) and call `base.Handle(request)` instead?

rchande:
@mholo65 It seems like Handle() does a bunch of pre- and post-processing on the request and result type that every overrider shouldn't have to be concerned with. I therefore added HandleCore where endpoints only have to be concerned with the actual call to the underlying service. Thoughts?

mholo65:
@rchande that’s a valid point. Instead of adding a new `HandleCore` method, what do you think of adding a `bool IsValid(TRequest request)` with the intention to validate the request?

rchande:
@mholo65 Good idea, I'll do that.

tests/OmniSharp.Cake.Tests/CodeCheckFacts.cs
```diff
@@ -0,0 +1,71 @@
...
+
+        [Fact]
+        public async Task ShouldNotCallCodeCheckServiceIfRequestDoesNotSpecifyFileName()
+        {
+            const string input = @"zzz$$";
+
+            var diagnostics = await FindDiagnostics(input, includeFileName: false);
+            Assert.Null(diagnostics);
+        }
+        private async Task<QuickFixResponse> FindDiagnostics(string contents, bool includeFileName)
```
DustinCampbell: blank line above

src/OmniSharp.Cake/Services/RequestHandlers/CakeRequestHandler.cs
```diff
@@ -54,11 +54,15 @@ protected CakeRequestHandler(OmniSharpWorkspace workspace)
 
             request = await TranslateRequestAsync(request);
 
-            var response = await service.Handle(request);
+            var response = IsValid(request)
```
mholo65: Maybe check validity before `TranslateRequest`?

PR #1112, opened 2018-02-15
Support RSP-based configuration for scripting project system

### LINKED ISSUE #1024: Support .rsp files for Scripting Project System bootstrapping
Spun off from https://github.com/OmniSharp/omnisharp-vscode/issues/1856

It would be useful for people not using CSI or dotnet-script - especially those using custom CSX hosts - to be able to predefine using statements and assembly references via `.rsp` files. Those are standard files already used by the compiler itself and look like this: https://github.com/dotnet/roslyn/blob/version-2.4.0/src/Interactive/csi/csi.rsp

In our case, we'd pick up the references used there and use them as globally imported namespaces / references when constructing CSX project system. Configuration proposal - inside `omnisharp.json`, under `script` node, as `rspPath` property. For example:

Since `omnisharp.json` can be global, you could even - in one sweep - configure your entire DEV box if you wanted.

filipw:
Fixes #1024 

This PR introduces the ability to configure a path to RSP file which is picked up by the scripting project system.
The RSP is not supported in the compiler-specific traditional way (where it can configure almost all of the compilation options) but rather is just a vessel to provide custom assembly references (`/r:...`) and global namespaces (`/u:...`). This is consistent with CSI behavior, which also uses RSP files only for these two types of input (when it comes to compilation options at least).

The obvious user benefit of all of this is, if I built my own custom script runner using Roslyn Scripting Engine, I could use OmniSharp for intellisense and language services against it (or at least to certain degree). That's due to the fact that I'd be able to predefine my implicitly available namespaces and assembly references using an RSP file now.

The way it all works:

 - I can specify a path to an RSP file in my `omnisharp.json` file - at global or local level, relative or absolute

```
{
    "Script": {
        "EnableScriptNuGetReferences": true,
        "RspFilePath": "C:/omnisharp-ext/foo.rsp"
    }
}
```

 - if there are any namespaces in the RSP file, they **replace** the default set we normally have (the CSI-set) 
 - if there are any assembly references in the RSP file, they **are merged** with default set we have (the one we have is really only aiming to bring in `object` reference - this is default requirement of C# scripting anyway, hence the merge not replace)
 - this feature relies on an API that @DustinCampbell opened up recently https://github.com/dotnet/roslyn/pull/23435. However, that hasn't shipped yet, so the PR still is based on reflection. Not sure whether we want to wait with it until next Roslyn release is out?

As a bonus, I added strongly typed scripting configuration (it was dictionary based).
I will try to invent some creative way to test this, at the moment the PR comes without any tests unfortunately.

### DIFF
--- a/src/OmniSharp.Script/ScriptHelper.cs
+++ b/src/OmniSharp.Script/ScriptHelper.cs
@@ -48,8 +48,9 @@ namespace OmniSharp.Script
         private readonly ScriptOptions _scriptOptions;
         private readonly IOmniSharpEnvironment _env;
         private readonly ILogger _logger;
+        private readonly bool _isDesktopClr;
 
-        public ScriptHelper(ScriptOptions scriptOptions, IOmniSharpEnvironment env, ILoggerFactory loggerFactory)
+        public ScriptHelper(ScriptOptions scriptOptions, IOmniSharpEnvironment env, ILoggerFactory loggerFactory, bool isDesktopClr)
         {
             _scriptOptions = scriptOptions ?? throw new ArgumentNullException(nameof(scriptOptions));
             _env = env ?? throw new ArgumentNullException(nameof(env));
@@ -57,6 +58,7 @@ namespace OmniSharp.Script
             _logger = loggerFactory.CreateLogger<ScriptProjectSystem>();
             _compilationOptions = new Lazy<CSharpCompilationOptions>(CreateCompilationOptions);
             _commandLineArgs = new Lazy<CSharpCommandLineArguments>(CreateCommandLineArguments);
+            _isDesktopClr = isDesktopClr;
             InjectXMLDocumentationProviderIntoRuntimeMetadataReferenceResolver();
         }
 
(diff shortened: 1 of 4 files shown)


david-driscoll:
There is a conflict that needs merging. 😄

src/OmniSharp.Script/ScriptHelper.cs
```diff
@@ -43,26 +43,65 @@ public class ScriptHelper
...
+            _logger = loggerFactory.CreateLogger<ScriptProjectSystem>();
             _compilationOptions = new Lazy<CSharpCompilationOptions>(CreateCompilationOptions);
-            InjectXMLDocumentationProviderIntoRuntimeMetadataReferenceResolver();
+            _commandLineArgs = new Lazy<CSharpCommandLineArguments>(CreateCommandLineArguments);
+            _isDesktopClr = isDesktopClr;
+        }
+
+        private CSharpCommandLineArguments CreateCommandLineArguments()
+        {
+            var scriptRunnerParserProperty = typeof(CSharpCommandLineParser).GetProperty("ScriptRunner", BindingFlags.Static | BindingFlags.NonPublic);
```
DustinCampbell: FWIW, I don't think you need this anymore. `CSharpCommandLineParser.Script` appears to be public in Roslyn 2.7.0.

PR #1117, opened 2018-02-21
Write out versionInfo file

rchande:
Create versionInfo file so it can be used by vscode downloader to determine what version to use.


rchande:
Tag @david-driscoll -- I think our uploading code in travis.yml already uploads everything from the deployment folder. Any other places I need to update?

david-driscoll:
Yeah, I think that's the case.

PR #1128, opened 2018-03-06
Share a test host instance between tests that are part of the same collection in OmniSharp.Roslyn.CSharp tests

### LINKED ISSUE #1127: Improve test duration
At the moment [this code to create a test host](https://github.com/OmniSharp/omnisharp-roslyn/blob/master/tests/TestUtility/OmniSharpTestHost.cs#L94-L143) runs for every test.

On my machine it adds an extra ~500ms for every test. Since most of the test only differ by test files, it would be good to reuse the same instance and just reset the workspace files if its possible (it will not be possible everywhere).

filipw:
Fixes #1127 

Within the context of one test collection (so effectively, in xUnit, a test class), we now reuse a single instance of `OmnisharpTestWorkspace`. 
It is initialized once per collection (instead of once per test now) and then the documents are removed before each test (so that each test populates it with the documents it needs).

Sharing a test host is not mandatory, so for several tests where it doesn't make sense (i.e. it wants to modify file watchers and wants them to be cleaned up), the old pattern can still be used.

On my machine, this improved the test suite run time by 3 minutes 15 seconds.


tests/TestUtility/SharedOmniSharpHostFixture.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace TestUtility
+{
+    public class SharedOmniSharpHostFixture : IDisposable
+    {
+        public SharedOmniSharpHostFixture()
+        {
+        }
+
+        public void Dispose()
```
rchande: Doesn't look like we're calling this anywhere-does this happen automatically though some xunit magic?

filipw: correct, it's an [xunit feature](https://xunit.github.io/docs/shared-context.html)
`Dispose()` in a test class will run at the end of each test, while `Dispose()` in a fixture class will run at the end of the collection.

PR #1133, opened 2018-03-13
Consume latest MSBuild in OmniSharp

rchande:
This should get us some of the performance fixes that went into VS 15.6. cc @jonsequitur


rchande:
@DustinCampbell Great, that's exactly the information I was looking for.

DustinCampbell:
I just dug into this a bit and it's going to be more work than just updating version numbers and binaries. MSBuild has taken a harder dependency on NuGet during all of the performance work that happened over the past six months. We're going to need to package up some of the NuGet build tasks along with our MSBuild to make it work with the latest and greatest.

DustinCampbell:
I'm closing this PR. I'll have a new PR out shortly that updates OmniSharp's MSBuild to the latest.

PR #1134, opened 2018-03-15
Fix loading of Xamarin.Android solution

grendello:
Xamarin.Android solution (`Xamarin.Android.sln` in
https://github.com/xamarin/xamarin-android/) fails to load with the current
version of omnisharp-roslyn because, somehow, one of the project reference file
paths is null or empty. I don't know which project that is (there are around
1500 of them in the XA repo) and, unfortunately, I had no time to investigate
that further. However, the change in this diff makes it work and I think it's a
good safeguard to have this check in place.


DustinCampbell:
This is fine. It'd be nice to know how this can happen though. :smile:

DustinCampbell:
FWIW, it turns out that this could've happen in the projects created for .csx files. Maybe that was the issue? The problem for .csx files was recently fixed by https://github.com/OmniSharp/omnisharp-roslyn/commit/95fc7f47c99165d0b80f473b64dd8ab07fd1c75c.

PR #1142, opened 2018-04-04
Script projects with no project filename on disk

akshita31:
For a script system, a filePath doesnot exist, hence when the following statement is executed an null reference exception occurs when the following code is executed - https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp.Roslyn/ProjectEventForwarder.cs#L70

Fix : Pass the path to the csxfile as the filePath when creating the project.

Please review @rchande @filipw @DustinCampbell @TheRealPiotrP


DustinCampbell:
Frustratingly, it seems that MyGet is reporting a 503 for some feeds, which is causing builds to fail.

PR #1143, opened 2018-04-05
Fix generate file

rchande:
* Use the modified Omnisharp solution, because the apply change operation solution does not have filepaths for new files
* Use the documentId requested by the apply change operation


DustinCampbell:
The new test is failing.

test-assets/test-projects/ProjectWithMissingType/HelloWorld.csproj
```diff
@@ -0,0 +1,21 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
DustinCampbell: You'll probably want to add this to build.json so it gets restored and built before the test runs: https://github.com/OmniSharp/omnisharp-roslyn/blob/master/build.json#L40.

rchande: Will do. Thanks!

tests/OmniSharp.Roslyn.CSharp.Tests/CodeActionsV2Facts.cs
```diff
@@ -153,11 +152,40 @@ private static void NewMethod()
...
+                    Identifier = "Generate class 'Z' in new file",
+                    WantsTextChanges = true,
+                    WantsAllCodeActionOperations = true
+                };
+
+                var response = await requestHandler.Handle(request);
+                var changes = response.Changes.ToArray();
+                Assert.Equal(2, changes.Length);
+                Assert.NotNull(changes[0].FileName);
+                Assert.NotNull(changes[1].FileName);
```
DustinCampbell: Should you also assert that the file was generated on disk?

DustinCampbell:
FYI: I've removed cli-deps in my MSBuild PR, which is what is breaking the build here. I have one more fix that I'm testing to ensure Unity projects keep working as expected. Once I'm done, I'll push that change and merge my PR after CI completes. So, if I go first, you can update your PR with my changes and CI will work for you.

rchande:
@DustinCampbell Sounds good!

DustinCampbell:
This is more fallout from the MyGet issues that happened yesterday. 😞

rchande:
Fixes https://github.com/OmniSharp/omnisharp-vscode/issues/2112

PR #1145, opened 2018-04-05
Update OmniSharp to latest MSBuild

DustinCampbell:
This change updates OmniSharp with the latest MSBuild binaries. Recent performance changes for MSBuild now require NuGet.Build.Tasks, which requires a set of NuGet binaries be included as well. Because the versions of these NuGet binaries might be different (actually, they *are* different) than the binaries that OmniSharp uses, there's a bit of duplication.


DustinCampbell:
I've tested a number of scenarios with the new MSBuild, including Xamarin Forms (with iOS and Android), Unity and .NET Core 2.1 (KestrelHttpServer). All are working well.

PR #1148, opened 2018-04-09
Further build performance improvements

DustinCampbell:
These aren't huge improvements, but they help a bit.

* Ensure that we set DOTNET_SKIP_FIRST_TIME_EXPERIENCE=true for ALL CI builds to avoid the first run population of the NuGet cache. Previously, we only did this for Travis. Now we do it for AppVeyor as well.
* Don't install openssl on OSX builds. At one time, this was needed for the legacy .NET Core SDK, but we don't even run those tests on OSX builds anymore.
* Disable caching in CI builds.

    Now that the build performance has been improved a great deal, it's clear that the build caching itself is taking a big chunk of time. On Linux, it takes ~90 seconds to set up the build cache and about ~130 to pack and upload the cache. From what I can tell, it does this for every single build, adding around 3.5 minutes on Linux. Disabling the caching wins back ~2-3 minutes on Linux builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,6 +40,7 @@ addons:
 install:
 - |
   if [ "$TRAVIS_OS_NAME" == "linux" ]; then
+    # On Linux, we build libuv from source
     sudo apt-get install make automake libtool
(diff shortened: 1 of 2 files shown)


DustinCampbell:
Before:

image

After:

image

DustinCampbell:
Before:

image

After:

image

PR #1156, opened 2018-04-11
Fix subtle issue that breaks handling of discovered referenced projects

DustinCampbell:
Recently, we added some additional logging that dumps the ProjectFileInfo using Json.net when a project is added for debugging purposes. Unfortunately, this throws when the ProjectFileInfo was created for a referenced project that is outside of the current solution or OmniSharp target directory. The exception thrown is an InvalidOperationException during Json.net serialization because several of the properties on the ProjectFileInfo are "unintialized" `ImmutableArray<T>` values. This happens specifically for project references outside of the solution/targetdir because OmniSharp just creates a barebones ProjectFileInfo and queues an update to fill it with real values later.

I'm fixing the issue be ensuring that ProjectFileInfo `ImmutableArray<T>` properties are always initialized to `ImmutableArray<T>.Empty`. I've also added a test to cover the particular scenario of referenced projects outside of the solution/targetdir. We didn't have a test already, which is why this wasn't caught sooner.


tests/OmniSharp.MSBuild.Tests/WorkspaceInformationTests.cs
```diff
@@ -122,6 +122,25 @@ public WorkspaceInformationTests(ITestOutputHelper output)
             }
         }
 
+        [Fact]
+        public async Task TestProjectWithReferencedProjectOutsideOfOmniSharp()
+        {
+            using (var testProject = await TestAssets.Instance.GetTestProjectAsync("TwoProjectsWithSolution"))
```
rchande: What log level are we running OmniSharp at in these tests? If it's not "Debug", the logging isn't getting tested.

DustinCampbell: When testing, we use a logger that captures *all* logging, regardless of level. When there's a failure, you get a dump of all logged messages.

DustinCampbell: Also, FWIW, the logging *does* get tested in the sense that the call to the logger certainly happens. Whether or not we output it or not is a different concern.

rchande: Ah, ok. Cool.

PR #1157, opened 2018-04-11
Set MSBuild property to allow the XAML markup compiler task to run

DustinCampbell:
Fixes https://github.com/OmniSharp/omnisharp-vscode/issues/2173

We need to set the AlwaysCompileMarkupFilesInSeparateDomain property to false. Otherwise, the XAML markup compiler task will try to create an AppDomain to run the markup compiler in. In normal MSBuild scenarios, this is fine. However, in OmniSharp, our AppDomain.AssemblyResolve handler that's used to locate MSBuild will not be installed in the marker compiler task's AppDomain, causing it to fail.


filipw:
thanks for the explanation, I learnt something today 😇
I actually saw this error before but had no clue what was happening.

DustinCampbell:
I tracked this down when working on MSBuildWorkspace

PR #1159, opened 2018-04-11
Fix transient documents not disappearing on project update

SirIntruder:
PR aims to fix these common issues with Unity projects:
https://github.com/OmniSharp/omnisharp-vscode/issues/2161
https://github.com/OmniSharp/omnisharp-vscode/issues/2040
https://github.com/OmniSharp/omnisharp-vscode/issues/1970

Context: project model can contain documents which are not part of the project on disk, but were added as a temporary guess ("transient documents"). When project file gets modified on disk, there is a diff run between previous and new document list, to check which docs need to be added, and which removed. But if a newly included document was already added as a transient guess, diff check will not detect this addition as a real change, and project continues on using transient document without informing anyone.

This PR would have MSBuild ProjectManager add a new documentInfo in such cases instead of old transient one, so BufferManager can remove transient documents it created on resulting DocumentAdded event (as it looks like it was idea). All other wrong guesses are removed this way, and faux errors in Unity projects are gone with them, without having to reload omnisharp/vscode.

There is an existing test for this which was disabled, I reverted c28276a8740bc5d5d9a88f082fbda13932b8b820 commit to re-enable it. That commit both skipped test and added host.Workspace.TryApplyChanges(newSolution); - this line caused problems for reasons I do not understand fully. Test now passes on my machine, but I would like if somebody could sanity check this.


src/OmniSharp.Roslyn/OmniSharpWorkspace.cs
```diff
@@ -194,7 +194,7 @@ private void LogDeletionException(Exception e, string filePath)
         protected override void ApplyDocumentAdded(DocumentInfo info, SourceText text)
         {
             var project = this.CurrentSolution.GetProject(info.Id.ProjectId);
-            var fullPath = info.FilePath;
+            var fullPath = Path.GetFullPath(info.FilePath);
```
SirIntruder: When I was trying to make host.Workspace.TryApplyChanges(newSolution) work, it was sometimes throwing exceptions here since info.FilePath was directly in project folder (no directory), I kept this since it looks like a reasonable change. Also, is this method used only in tests or am I missing something?

david-driscoll: This is called on the roslyn side, it's possible that this may some weird problems.  It might make more sense for the test to be changed, so that it has a proper path (and even make sure the file exists)

tests/OmniSharp.Tests/UpdateBufferFilterFacts.cs
```diff
@@ -89,7 +89,6 @@ public UpdateBufferFilterFacts(ITestOutputHelper output)
                     filePath: "transient.cs");
 
                 var newSolution = host.Workspace.CurrentSolution.AddDocument(document);
-                host.Workspace.TryApplyChanges(newSolution);
```
david-driscoll: With this being removed, I'm not sure if the test is actually testing the behavior it's supposed to.

If you're new to Roslyn, then one thing to understand is that everything is immutable by default.  This allows for some really powerful features, but also is confusing if you're not familiar with it.  With the `TryApplyChanges` being removed then it's not actually adding the file the workspaces solution.

SirIntruder: Thanks for your time to explain this! 

IIRC, once I got that path issue out of the way, it looked to me problem was caused because WorkspaceChange events being raised async, which meant test is checking count of docs before BufferManager can get the memo it needs to get rid of transient docs, explaining where the 3 docs are coming from. I tried with a magic Thread.Sleep, and then it had only doc - which may be what should be a correct result after all! I just didn't want to jump to conclusions since test was working like this for such a long time.

If what I assume is correct, test should wait for event to be raised after this line (preferably without magic thread sleep 😆)

SirIntruder: I am not sure why there should be 2 docs with "transient.cs" file path after initial buffer update - unless those are the same file added to 2 different projects which exist in the test workspace. If real docuiment gets added, it should kill both copies, by my reading of BufferManager.WorkspaceChanged code.

tests/OmniSharp.Tests/UpdateBufferFilterFacts.cs
```diff
@@ -89,9 +89,18 @@ public UpdateBufferFilterFacts(ITestOutputHelper output)
...
 
                 var newSolution = host.Workspace.CurrentSolution.AddDocument(document);
+                host.Workspace.TryApplyChanges(newSolution);
 
+                // 2 transient + 1 real document.
                 docIds = host.Workspace.CurrentSolution.GetDocumentIdsWithFilePath("transient.cs");
-                Assert.Equal(2, docIds.Length);
+                Assert.Equal(3, docIds.Length);
+
+                // wait for event to be raised.
```
SirIntruder: @david-driscoll I pushed this version of test to illustrate what I think is going on (workspaceChange is async, test was checking too early). If we agree this is expected end result and find a more... elegant way of awaiting raising event, I will update test. I think previous test was not really testing much,

DustinCampbell:
@david-driscoll : Thanks for explaining things. I understand the whole transient system a bit better now.

tests/OmniSharp.Tests/UpdateBufferFilterFacts.cs
DustinCampbell: I'd rather see a loop that awaits a `Task.Delay(...)` or a `Task.Yield(...)` and verifies that the event triggered rather than a single `Thread.Sleep(...)`.

tests/OmniSharp.Tests/UpdateBufferFilterFacts.cs
DustinCampbell: Without the `TryApplyChanges`, the updated solution will never get applied to the workspace.

DustinCampbell: FWIW, I think this whole test needs rewritten. The fact that it's trying to write files to disk is a bit suspicious. I think it should be rewritten to use a proper test project like most of our other tests. Right now, we're trying to write the contents of "transient.cs" to disk when the project it belongs to doesn't even exist on disk.

DustinCampbell:
@SirIntruder: One thing I'm not clear on is how this fixes the issues that are linked here. If I understand those issues, a key part of the problem is that the .csproj does not get updated with new `<Compile Include=""/>` items. Am I missing something?

SirIntruder:
@DustinCampbell currently, when unity updates its . csproj files, omnisharp doesn't handle update correctly  because transient files are not cleared - requiring vscode reload. With this fix, faux errors will disappear as soon as unity updates project files.

In practice, this means just focusing unity window will get rid of the errors. This is short of "it just works" solution, which would require vscode to correctly guess how unity wants to update csprojs - and from previous discussions I accepted that would be a bit to specific for omnisharp.

tests/OmniSharp.Tests/UpdateBufferFilterFacts.cs
david-driscoll: I think for now switching it from `Thread.Sleep` to `await Task.Delay(100)` is a better choice.  

You could also add a loop like with a cancellation token...

```csharp
var cts = new CancellationTokenSource();
cts.CancelAfter(TimeSpan.FromSeconds(10));
while (!cts.IsCancellationRequested) {
  await Task.Yield();
  docIds = host.Workspace.CurrentSolution.GetDocumentIdsWithFilePath("transient.cs");
  if (docIds.Count() <= 1) break;
}
Assert.Single(docIds);
```

It's a little (okay its a lot) janky, but it would ensure that test is a little more stable if it takes more than 100ms to update the workspace.

src/OmniSharp.Roslyn/OmniSharpWorkspace.cs
david-driscoll: Still not sure about this, if you could try taking it out once the test has been fixed, and let us know what the errors your seeing are, it's probably something that needs to be fixed.

SirIntruder: This is an exception without this change:

```
OmniSharp.Tests.UpdateBufferFilterFacts.UpdateBuffer_TransientDocumentsDisappearWhenProjectAddsThem [FAIL]
      System.ArgumentException : Path cannot be the empty string or all whitespace.
      Stack Trace:
           at System.IO.Directory.CreateDirectory(String path)
           at OmniSharp.OmniSharpWorkspace.SaveDocumentText(DocumentId id, String fullPath, SourceText newText, Encoding encoding)
```

this is the code it causes it:
```
var dir = Path.GetDirectoryName(fullPath);
if (!Directory.Exists(dir))
{
    Directory.CreateDirectory(dir);
}
```
without Path.GetFullPath, this would also work:
```
var dir = Path.GetDirectoryName(fullPath);
if (!string.IsNullOrEmpty(dir) && !Directory.Exists(dir))
{
    Directory.CreateDirectory(dir);
}
```

Changing this specific test to use full path would fix the issue, but then all the other tests use filenames like "foo.cs" "bar.cs" etc. If DocumentInfo.FileName is allowed to be relative path in general, some check should be in order. Documentation is vague: us/dotnet/api/microsoft.codeanalysis.documentinfo?view=roslyn-dotnet.

DustinCampbell:
Thinking about this further. I think it'd be good to add a test that exercises this particular scenario. If I understand it correctly, the actually fix is in the MSBuild project loader, which doesn't even run in the test that we've been talking about.

SirIntruder:
That's true. Ideally we would have .csproj setup like in those issues (I attached zip with test project in one of those issues), simulate editing csproj. and check if a diagnostic check still reports errors. If everyone is busy I may try this in next couple of days.

Another thing, unity-like error can happen every time transient guess adds the new file to a group of projects, with one of those having no reference to other projects in group. Reducing target group of projects (code in BufferManager.cs) to those with most in-group dependencies would probably be most user-friendly.

DustinCampbell:
I suspect we can simulate the scenario with one of the .NET Core test projects that we have.

PR #1160, opened 2018-04-12
Ensure that we don't subscribe the same file-watcher callback multiple times

DustinCampbell:
While refactoring some code, I noticed that we were queuing the same callback with the `IFileSystemWatcher` many, many times in order to watch source file directories. That means that these callbacks would be called multiple times for any notification. This change expands the `ManualFileSystemWatcher` to only add the same callback for a particular extension, file or directory once. Note that the callback is only checked for reference equality, so it's necessary to stash your callback in a field if you don't want it added multiple times.

### DIFF
--- a/src/OmniSharp.Host/FileWatching/ManualFileSystemWatcher.Callbacks.cs
+++ b/src/OmniSharp.Host/FileWatching/ManualFileSystemWatcher.Callbacks.cs
@@ -1,0 +1,29 @@
+﻿using System.Collections.Generic;
+
+namespace OmniSharp.FileWatching
+{
+    internal partial class ManualFileSystemWatcher
+    {
+        private class Callbacks
+        {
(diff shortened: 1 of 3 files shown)


src/OmniSharp.Host/FileWatching/ManualFileSystemWatcher.Callbacks.cs
```diff
@@ -0,0 +1,29 @@
...
+            {
+                if (_callbackSet.Add(callback))
+                {
+                    _callbacks.Add(callback);
+                }
+            }
+
+            public void Invoke(string filePath, FileChangeType changeType)
+            {
+                foreach (var callback in _callbacks)
```
rchande: Is there a reason we need to guarantee the order the callbacks are called?

DustinCampbell: Not really. It was easy enough to do though, so I did. :smile:

PR #1161, opened 2018-04-12
Added support for excluding search paths via globbing patterns

filipw:
This PR is intended to fix https://github.com/OmniSharp/omnisharp-roslyn/issues/896

It introduces a new `omnisharp.json` setting to control which paths OmniSharp should skip (paths can be explicit or use globbing patterns).
Globbing is provided via `Microsoft.Extensions.FileSystemGlobbing` which we conveniently already referenced 😄

By default the following set up is used:

```
{
  "fileOptions": {
    "excludeSearchPatterns": [
      "**/node_modules/**/*",
      "**/bin/**/*",
      "**/obj/**/*"
    ]
  }
}
```

So no files from `bin`, `obj` or `node_modules` are discovered, at any level.

The default setup can be overridden using a global or local `omnisharp.json` file. 
For example:
```
{
  "fileOptions": {
    "excludeSearchPatterns": [
      "build/**/*",
      "**/TestFixtures/*.csproj"
    ]
  }
}
```

This should not find anything from a `build` folder relative to currently open OmniSharp and any `csproj` file directly inside `TestFixtures`.
The new service is used in place of the old `SearchOption.AllDirectories` scanning.


src/OmniSharp.Abstractions/Options/FileOptions.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace OmniSharp.Options
+{
+    public class FileOptions
+    {
+        public string[] ExcludeSearchPatterns { get; set; } = new[] { "**/node_modules/**/*", "**/bin/**/*", "**/obj/**/*" };
```
david-driscoll: Should we also add search inclusions?  (in the future perhaps?

Kinda feels like it would be sort of similar to how we used to be able to do things with project.json.

david-driscoll:
Going to merge tomorrow afternoon if there are no other objections

filipw:
One thing that might improve the usability of this feature is perhaps we should split the exclusion patterns into 2 separate properties that you can override and customize.

The way inheritance rules in our hierarchical configuration work, is that you can override everything, but you completely negate the setup from the previous source.
What I mean is that at the moment if you want to to exclude i.e. `/foo` locally, you'd do this:

```
{
  "fileOptions": {
    "excludeSearchPatterns": [
      "foo/**/*"
    ]
  }
}
```

The problem is, this overrides the default setup and `bin`/`obj`/`node_modules` are no longer excluded anymore. 
So to make it work in a way that the defaults are still observed, you'd have to technically carry over the defaults, like this:

```
{
  "fileOptions": {
    "excludeSearchPatterns": [
      "**/node_modules/**/*",
      "**/bin/**/*",
      "**/obj/**/*"
      "foo/**/*"
    ]
  }
}
```

This is not the greatest experience. 
However, we could have 2 properties to control all this. Something like: `userExcludeSearchPatterns` and `systemSearchPatterns`. The default setup would be:

```
{
  "fileOptions": {
    "systemExcludeSearchPatterns": [
      "**/node_modules/**/*",
      "**/bin/**/*",
      "**/obj/**/*"
    ],
    "userExcludeSearchPatterns": []
  }
}
```

Then, you'd only override `userExcludeSearchPatterns`, and the `systemExcludeSearchPatterns` carry over automatically.

```
{
  "fileOptions": {
    "userExcludeSearchPatterns": [
      "foo/**/*"
    ]
  }
}
```

If this makes sense, I will push this change before we merge.

src/OmniSharp.Abstractions/Options/FileOptions.cs
DustinCampbell: What about `"**/.git/**/*"`?

filipw: that's a good one 👍

david-driscoll:
@filipw I like the idea of having a system level setting, and a user setting.  🚲 🏠 should the user property by `userExcludeSearchPatterns` or `excludeSearchPatterns`.

I tend to favor `excludeSearchPatterns` because most users won't want to even know that `systemExcludeSearchPatterns` exists, however by setting it up in such a way it makes the system exclusions less discoverable.  I'm good to :shipit: either way.

filipw:
OK thanks. So based on the convo, I added `systemExcludeSearchPatterns` and moved the "default" excludes there, and made the `excludeSearchPatterns` empty. 

This way you can inject your own excludes into `excludeSearchPatterns` and still benefit from the `systemExcludeSearchPatterns` as at the end of the day, they get merged. 
If anyone, for whatever reason, is eager to load projects from i.e. `node_modules`, `systemExcludeSearchPatterns` is also overridable.

DustinCampbell:
@filipw: should we update the WIki page with this information? https://github.com/OmniSharp/omnisharp-roslyn/wiki/Configuration-Options

filipw:
yes - you are right! I'll do that 📚

PR #1166, opened 2018-04-18
Report better version for VS preview consoles

rainersigwald:
For prereleases, the Visual Studio command prompt sets the environment
variable `VSCMD_VER` to a value like `15.7.0-pre.3.0`, which is not
parseable by System.Version.

This could be handled by using a version parser that can handle this
SemVer 2 version as well as Windows 4-part versions, but that would be
another dependency, so I just special-cased the problematic dash.

### DIFF
--- a/src/OmniSharp.Host/MSBuild/Discovery/Providers/DevConsoleInstanceProvider.cs
+++ b/src/OmniSharp.Host/MSBuild/Discovery/Providers/DevConsoleInstanceProvider.cs
@@ -38,6 +38,17 @@ namespace OmniSharp.MSBuild.Discovery.Providers
 
             if (version == null)
             {
+                var dashIndex = versionString.IndexOf('-');
+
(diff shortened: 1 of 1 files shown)


rainersigwald:
Port of Microsoft/MSBuildLocator#32.

I noticed this originally when I was mid-update on my preview VS, which caused the Setup API to not report that instance, and OmniSharp to use its own packaged MSBuild because it didn't want the "15.0" devconsole (pointed to preview) nor the (actual) 15.0 instance.

That won't be a normal case, but the bug would prevent OmniSharp from respecting the devconsole instance for previews, which could be a problem if you had a preview + other installations and ran from inside the preview devconsole, depending on the ordering of the installations.

@DustinCampbell I see you already approved the Locator version of this :)

DustinCampbell:
Excellent! Thanks @rainersigwald!

PR #1169, opened 2018-04-22
when fixing usings, return namespaces associated with ambiguous result

tadesegha:
the potential using statements are added to the Text of the QuickFix response, allowing the user manually choose which using statement to employ.

### DIFF
--- a/tests/OmniSharp.Roslyn.CSharp.Tests/FixUsingsFacts.cs
+++ b/tests/OmniSharp.Roslyn.CSharp.Tests/FixUsingsFacts.cs
@@ -250,7 +250,7 @@ namespace OmniSharp
                     Line = point.Line,
                     Column = point.Offset,
                     FileName = TestFileName,
-                    Text = "`classX` is ambiguous"
+                    Text = "`classX` is ambiguous. Namespaces: using nsA; using nsB;",
                 }
             };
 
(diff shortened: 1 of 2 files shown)


tadesegha:
Hi Ravi,

I did add a test addressing the scenario in question.

I use omnisharp-vim with omnisharp-roslyn as a dependency. When I run the fix usings  command, sometimes I get a response indicating there is an ambiguous reference. This PR would update the message that comes back to not only indicate an ambiguous reference, but also indicate the ambiguous namespaces. That way, I as the user can decide which reference I want to use.

If you take a look at the commits you should see a test addressing this scenario.

Thanks,

Tolu

________________________________

tests/OmniSharp.Roslyn.CSharp.Tests/FixUsingsFacts.cs
```diff
@@ -250,7 +250,7 @@ public method1()
                     Line = point.Line,
                     Column = point.Offset,
                     FileName = TestFileName,
-                    Text = "`classX` is ambiguous"
+                    Text = "`classX` is ambiguous. Namespaces: using nsA; using nsB;",
```
rchande: @tadesegha This isn't your fault, but I don't think this test does what we think it does;

* I don't see any code in this test that asserts on the expected content
* When pasting this code into VS Code, I get the error "The type or namepspace ClassX could not be found" and quick fixes to add usings for either of the namespaces. 

Can you fix up the test to actually assert something? Thanks!

rchande:
@tadesegha I'm sorry, you did add a test. I missed it since it was at the bottom of the diff. I do think there are some issues with the test you modified. Mind taking a look? I'm happy to help out if you have any questions.

tadesegha:
@rchande the last line of the test 'await AssertUnresolvedReferencesAsync(content.Code, expectedUnresolved);' has assertions. If you update the test without updating the code you should get a failing test.

PR #1170, opened 2018-04-23
Update readme for vscode "latest" option

akshita31:
Fixes: https://github.com/OmniSharp/omnisharp-vscode/issues/2204
Update the readme to enable the usage of the vscode -" Use latest OmniSharp option" : https://github.com/OmniSharp/omnisharp-vscode/blob/master/CHANGELOG.md#specify-omnisharp-version

Please review : @rchande @TheRealPiotrP @DustinCampbell

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,16 +59,14 @@ For more details, see [Build](https://github.com/OmniSharp/omnisharp-roslyn/blob
 
 Add the following setting to your [User Settings or Workspace Settings](https://code.visualstudio.com/Docs/customization/userandworkspace).
 
-To consume the latest build of omnisharp from "master": 
-
 ``` JSON
 {
-  "omnisharp.path": "latest"
+  "omnisharp.path": "<Path to the omnisharp executable>"
 }
 ```
-The above option can also set to the path to an OmniSharp executable on disk or a specific version number like `1.29.2-beta.60`
-
-In order to use the latest build from the master branch, add
+The above option can also be set to:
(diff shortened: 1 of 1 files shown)


akshita31:
Yes we should wait for 1.15.

README.md
```diff
@@ -59,11 +59,14 @@ For more details, see [Build](https://github.com/OmniSharp/omnisharp-roslyn/blob
 
 Add the following setting to your [User Settings or Workspace Settings](https://code.visualstudio.com/Docs/customization/userandworkspace).
 
+To consume the latest build of omnisharp from "master": 
+
 ``` JSON
 {
-  "omnisharp.path": "<Path to the omnisharp executable>"
+  "omnisharp.path": "latest"
```
DustinCampbell: The only concern I have here is that it doesn't help with debugging or development of OmniSharp, which I think is the goal of these instructions. In other words, these instructions are to help folks who have built a version of OmniSharp on their box so they can try it out and debug it. Could you expand this?

DustinCampbell: I meant, could you expand the description to include information about `latest` rather than replacing the existing instructions with `latest`? Since the instructions here are meant for developers who are actively working on OmniSharp, it feels like `latest` might not be what they want. Instead, they'll want to debug the version of OmniSharp that they just built on their dev machine. Does that make sense?

akshita31: Yes it does. I'll make the changes. Thanks!

PR #1181, opened 2018-05-08
Handle CSProj files that are not C# projects that 

dmgonch:
The proposed changes allow OmniSharp not to stop processing projects when it encounters a file with CSProj extension that is not a C# project. At the moment I'm working with a repo where out of 736 CSProj files 229 are of this sort. They typically are TypeScript projects, projects that copy files into final release folder and project files used only as data inputs for test cases. With my changes, OmniSharp invoked from VSCode was able to successfully process all valid C# projects while ignoring the rest.

### DIFF
--- a/src/OmniSharp.MSBuild/ProjectManager.cs
+++ b/src/OmniSharp.MSBuild/ProjectManager.cs
@@ -43,7 +43,7 @@ namespace OmniSharp.MSBuild
         private readonly MetadataFileReferenceCache _metadataFileReferenceCache;
         private readonly PackageDependencyChecker _packageDependencyChecker;
         private readonly ProjectFileInfoCollection _projectFiles;
-        private readonly ConcurrentDictionary<string, int> _failedToLoadProjectFiles;
+        private readonly HashSet<string> _failedToLoadProjectFiles;
         private readonly ProjectLoader _projectLoader;
         private readonly OmniSharpWorkspace _workspace;
 
@@ -63,7 +63,7 @@ namespace OmniSharp.MSBuild
             _metadataFileReferenceCache = metadataFileReferenceCache;
             _packageDependencyChecker = packageDependencyChecker;
             _projectFiles = new ProjectFileInfoCollection();
-            _failedToLoadProjectFiles = new ConcurrentDictionary<string, int>(StringComparer.OrdinalIgnoreCase);
+            _failedToLoadProjectFiles = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
             _projectLoader = projectLoader;
             _workspace = workspace;
 
@@ -151,13 +151,13 @@ namespace OmniSharp.MSBuild
                     projectList.Add(currentProject);
 
                     // update or add project
-                    _failedToLoadProjectFiles.TryRemove(currentProject.FilePath, out int notUsed);
+                    _failedToLoadProjectFiles.Remove(currentProject.FilePath);
                     if (_projectFiles.TryGetValue(currentProject.FilePath, out var projectFileInfo))
                     {
                         projectFileInfo = ReloadProject(projectFileInfo);
                         if (projectFileInfo == null)
                         {
-                            _failedToLoadProjectFiles.TryAdd(currentProject.FilePath, 0 /*not used*/);
+                            _failedToLoadProjectFiles.Add(currentProject.FilePath);
                             continue;
                         }
                         _projectFiles[currentProject.FilePath] = projectFileInfo;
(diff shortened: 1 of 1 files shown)


src/OmniSharp.MSBuild/ProjectManager.cs
```diff
@@ -148,14 +151,25 @@ private void ProcessQueue(CancellationToken cancellationToken)
                     projectList.Add(currentProject);
 
                     // update or add project
+                    _failedToLoadProjectFiles.TryRemove(currentProject.FilePath, out int notUsed);
```
DustinCampbell: `out int nonUsed` can just be `out _` in C# 7.

src/OmniSharp.MSBuild/ProjectManager.cs
```diff
@@ -42,6 +43,7 @@ public ProjectToUpdate(string filePath, bool allowAutoRestore)
         private readonly MetadataFileReferenceCache _metadataFileReferenceCache;
         private readonly PackageDependencyChecker _packageDependencyChecker;
         private readonly ProjectFileInfoCollection _projectFiles;
+        private readonly ConcurrentDictionary<string, int> _failedToLoadProjectFiles;
```
DustinCampbell: Is a `ConcurrentDictionary` really needed? MSBuild project loader really doesn't do much in the way of concurrency since it can't use MSBuild concurrently. All of the methods where `_failedToLoadProjectFiles` is updated also update `_projectFiles`, which is just a wrapper around a `List` and a `Dictionary`. In other words, the majority of this code is already synchronous. :smile:

src/OmniSharp.MSBuild/ProjectManager.cs
```diff
@@ -193,31 +207,36 @@ private ProjectFileInfo LoadOrReloadProject(string projectFilePath, Func<(Projec
...
 
             try
             {
+                ImmutableArray<MSBuildDiagnostic> diagnostics;
                 (projectFileInfo, diagnostics) = loadFunc();
 
-                if (projectFileInfo == null)
+                if (projectFileInfo != null)
+                {
+                    _logger.LogWarning($"Successfully loaded project file '{projectFilePath}'.");
```
DustinCampbell: Why would this be a warning?

src/OmniSharp.MSBuild/ProjectManager.cs
```diff
@@ -193,31 +207,36 @@ private ProjectFileInfo LoadOrReloadProject(string projectFilePath, Func<(Projec
...
                 {
                     _logger.LogWarning($"Failed to load project file '{projectFilePath}'.");
                 }
+
+                _eventEmitter.MSBuildProjectDiagnostics(projectFilePath, diagnostics);
             }
             catch (Exception ex)
             {
-                _logger.LogWarning($"Failed to load project file '{projectFilePath}'.", ex);
+                _logger.LogWarning($"Failed to load project file '{projectFilePath}' : {ex.ToString()}.");
```
DustinCampbell: Why switch over to `ex.ToString()`? Passing `ex` should still log the exception.

DustinCampbell: Also, it feels like this should probably be `LogError`.

src/OmniSharp.MSBuild/ProjectManager.cs
dmgonch: Updated

src/OmniSharp.MSBuild/ProjectManager.cs
dmgonch: I was hoping that this could the first step towards making OmniSharp to be able loading multiple projects in parallel. But for the sake of consistency I will go ahead and change this to Dictionary.

src/OmniSharp.MSBuild/ProjectManager.cs
dmgonch: Good point - fixed.

src/OmniSharp.MSBuild/ProjectManager.cs
dmgonch: In my testing it didn't print the exception. But I will go ahead and revert this line to the original this is not related to the issue being fixed by these changes.

src/OmniSharp.MSBuild/ProjectManager.cs
DustinCampbell: FWIW, loading projects in parallel is a big effort. The issue here is that MSBuild handles parallel builds by launching multiple processes to serve as additional nodes. However, doing that cross platform is challenging.

src/OmniSharp.MSBuild/ProjectManager.cs
DustinCampbell: Hmmm... we should check and see if the exception's not printing. I've definitely seen exceptions print in the past, but I wonder if it's not printing because this is a warning. I'll take a look.

Issue #1138, opened 2018-03-19
Fails to load a project that uses an additional SDK reference


DustinCampbell:
_From @timjroberts on March 19, 2018 20:21_

The extension fails to load .csproj files where additional SDKs have been specified as per the MSBUILD 15.6 "Sdk" Design concept.

## Environment data
`dotnet --info` output:
```
.NET Command Line Tools (2.1.300-preview1-008174)

Product Information:
 Version:            2.1.300-preview1-008174
 Commit SHA-1 hash:  b8df89a54f

Runtime Environment:
 OS Name:     Mac OS X
 OS Version:  10.13
 OS Platform: Darwin
 RID:         osx.10.12-x64
 Base Path:   /usr/local/share/dotnet/sdk/2.1.300-preview1-008174/

Microsoft .NET Core Shared Framework Host

  Version  : 2.1.0-preview1-26216-03
  Build    : f2c3216183d20416568a4bbf5bb7d153e826f153
```
VS Code version: **1.21.1**
C# Extension version: **1.14.0**

## Steps to reproduce
Create a .csproj file that references an additional SDK. In my particular example, I have:
```xml
<Project Sdk="Microsoft.NET.Sdk;TypeScript.Sdk">
    ....
</Project>
```
In this example, `TypeScript.Sdk` has been manually installed into the SDK base path (identified in the dotnet --info output above), however, I've also tried using `TypeScript.Sdk/1.0.0` which should prompt the NuGet resolver in MSBUILD to download and make the SDK available.

## Expected  behavior
The project should load successfully and code-lens/autocompletion support work as normal.

## Actual behavior
I receive the following output in the Output window:
```
[info]: OmniSharp.MSBuild.ProjectManager
        Loading project: /Users/timroberts/test3/Core/Core.csproj
Can't find custom attr constructor image: /usr/local/share/dotnet/sdk/2.1.300-preview1-008174/Sdks/TypeScript.Sdk/lib/netstandard2.0/TypeScript.Sdk.Tasks.dll mtoken: 0x0a00000a due to: Could not load file or assembly 'netstandard, Version=2.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51' or one of its dependencies. assembly:netstandard, Version=2.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51 type:<unknown type> member:<none>
```
At that point I receive no code-lens or autocompletion support for any project in the solution.

[...]

PR #1192, opened 2018-05-18
Remove MSBuildSDKsPath hack that blocks MSBuild SDK resolvers

### LINKED ISSUE #1190: Cannot load projects that use an MSBuild SDK package
MSBuild 15.6 added a new SDK resolver that can use NuGet packages as SDKs.  When opening a project that uses a NuGet-based SDK package, OmniSharp cannot load the project.

**Repro**
git clone https://github.com/dotnet/arcade
Open VS Code

**Expected**
OmniSharp uses the same SDK resolution that VS 15.6 and the .NET Core CLI 2.1.105 use.

**Actual**
Project fails to load with:
```
[warn]: OmniSharp.MSBuild.ProjectManager
        Failed to load project file '/Users/namc/dev/dotnet/arcade/src/Microsoft.DotNet.Build.Tasks.Feed/Microsoft.DotNet.Build.Tasks.Feed.csproj'.
/Users/namc/dev/dotnet/arcade/src/Microsoft.DotNet.Build.Tasks.Feed/Microsoft.DotNet.Build.Tasks.Feed.csproj(1,1)
Microsoft.Build.Exceptions.InvalidProjectFileException: The SDK 'RoslynTools.RepoToolset' specified could not be found.  /Users/namc/dev/dotnet/arcade/src/Microsoft.DotNet.Build.Tasks.Feed/Microsoft.DotNet.Build.Tasks.Feed.csproj
  at Microsoft.Build.Evaluation.ProjectCollection.LoadProject (System.String fileName, System.String toolsVersion) [0x00000] in <61115f75067146fab35b10183e6ee379>:0 
  at OmniSharp.MSBuild.ProjectLoader.EvaluateProjectFileCore (System.String filePath) [0x0003e] in <64f2a38974d54ecaaa93e3697628b738>:0 
  at OmniSharp.MSBuild.ProjectLoader.BuildProject (System.String filePath) [0x0000d] in <64f2a38974d54ecaaa93e3697628b738>:0 
  at OmniSharp.MSBuild.ProjectFile.ProjectFileInfo.Load (System.String filePath, OmniSharp.MSBuild.ProjectLoader loader) [0x00014] in <64f2a38974d54ecaaa93e3697628b738>:0 
  at OmniSharp.MSBuild.ProjectManager+<>c__DisplayClass23_0.<LoadProject>b__0 () [0x00000] in <64f2a38974d54ecaaa93e3697628b738>:0 
  at (wrapper delegate-invoke) System.Func`1[System.ValueTuple`2[OmniSharp.MSBuild.ProjectFile.ProjectFileInfo,System.Collections.Immutable.ImmutableArray`1[OmniSharp.MSBuild.Logging.MSBuildDiagnostic]]].invoke_TResult()
  at OmniSharp.MSBuild.ProjectManager.LoadOrReloadProject (System.String projectFilePath, System.Func`1[TResult] loadFunc) [0x0001b] in <64f2a38974d54ecaaa93e3697628b738>:0 
```
**Details**
 
See https://github.com/microsoft/msbuildsdks and https://docs.microsoft.com/en-us/visualstudio/msbuild/how-to-use-project-sdk#how-project-sdks-are-resolved for more details

VS Code C# extension: 1.15.2
VS Code: 1.23.1

DustinCampbell:
Fixes #1190

This change ended up being a bit larger than I'd intended. If you're reviewing, I recommend walking through each commit at a time.

At it's core, this change removes an old hack that set the `MSBuildSDKsPath` environment variable to a path that was computed by launching `dotnet` in the project folder. This hack is an override that blocks MSBuild's SDK resolvers, which is the system that *should* be used to locate the SDK(s) for a project. However, when we tried removing this hack in the past, we were bit by the fact that the base MSBuild SDK resolve doesn't work with .NET Core SDKs <= 1.0.3. So, we added the hack back. Since there weren't really any custom MSBuild SDK resolvers out there, the override really didn't have much of an impact.

Now that MSBuild has shipped the NuGet-based SDK resolver that can download SDKs from NuGet, our override hack is a significant problem. So, I've removed the hack and the base SDK resolver is used instead. If anybody wants to the old behavior (because they're using an old .NET Core SDK), they can set the following option in an `omnisharp.json` file:

```JSON
{
    "MSBuild": {
        "UseLegacySdkResolver": true
    }
}
```

Unfortunately, removing the hack forced several other changes, primarily to get tests working without the hack. I had to rationalize the `IDotNetCliService` and elevated `IOmniSharpEnvironment` and `IEventEmitter` to required services. Finally, because the base SDK resolver is a netstandard2.0 library, I needed to package `netstandard.dll` with our standalone Mono install. As part of this, I went ahead and updated the Mono and MSBuild bits for Mono as well.

Note that there is further work to package the NuGet MSBuild SDK resolver with our standalone MSBuild bits. However, that's a change for another day. For now, custom MSBuild SDKs should work if the user has a newer VS or Mono installed.

cc @nguerrera.

### DIFF
--- a/build.json
+++ b/build.json
@@ -8,12 +8,12 @@
   "LegacyDotNetVersion": "1.0.0-preview2-1-003177",
   "RequiredMonoVersion": "5.8.0.0",
   "DownloadURL": "https://omnisharpdownload.blob.core.windows.net/ext",
-  "MonoRuntimeMacOS": "mono.osx-5.10.1.20.zip",
-  "MonoRuntimeLinux32": "mono.linux-x86-5.10.1.20.zip",
-  "MonoRuntimeLinux64": "mono.linux-x86_64-5.10.1.20.zip",
-  "MonoFramework": "framework-5.10.1.20.zip",
-  "MonoMSBuildRuntime": "Microsoft.Build.Runtime.Mono-mono-5.10.1.20.zip",
-  "MonoMSBuildLib": "Microsoft.Build.Lib.Mono-mono-5.10.1.20.zip",
+  "MonoRuntimeMacOS": "mono.osx-5.12.0.226.zip",
+  "MonoRuntimeLinux32": "mono.linux-x86-5.12.0.226.zip",
+  "MonoRuntimeLinux64": "mono.linux-x86_64-5.12.0.226.zip",
+  "MonoFramework": "framework-5.12.0.226.zip",
+  "MonoMSBuildRuntime": "Microsoft.Build.Runtime.Mono-5.12.0.226.zip",
+  "MonoMSBuildLib": "Microsoft.Build.Lib.Mono-5.12.0.226.zip",
   "HostProjects": [
     "OmniSharp.Stdio.Driver",
     "OmniSharp.Http.Driver"
--- a/src/OmniSharp.Abstractions/Services/IDotNetCliService.cs
+++ b/src/OmniSharp.Abstractions/Services/IDotNetCliService.cs
@@ -1,0 +1,59 @@
+﻿using NuGet.Versioning;
+using System;
+using System.Diagnostics;
+using System.Threading.Tasks;
+
+namespace OmniSharp.Services
+{
+    public interface IDotNetCliService
+    {
+        /// <summary>
+        /// The path used to launch the .NET CLI. By default, this is "dotnet".
+        /// </summary>
+        string DotNetPath { get; }
+
+        /// <summary>
+        /// Launches "dotnet --info" in the given wotking directory and returns a
+        /// <see cref="DotNetInfo"/> representing the returned information text.
+        /// </summary>
+        DotNetInfo GetInfo(string workingDirectory = null);
+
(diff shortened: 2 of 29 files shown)


DustinCampbell:
This should also address #1138, but I'm not able to set up that repro up because I don't where to find the `TypeScript.Sdk` referenced in the repro.

DustinCampbell:
One other change here that I forgot to mention is the update to the MSBuildLocator AssemblyResolve handler. Now it will try to load assemblies *other* than the core MSBuild assemblies. However, it will only load them if they match by name, version and public key token. This is necessary to load dependencies for NuGet MSBuild SDK resolver, which live in the MSBuild tools path on Mono.

DustinCampbell:
> Also, the embedded Mono bits got bumped, but the way I understand it, the global Mono minimum version stays the same with this PR, correct?

Correct. And, the version of Mono needed to build OmniSharp has not changed.

PR #1194, opened 2018-05-18
Disable DotNetProjectSystem by default

DustinCampbell:
This change threads through a new `IProjectSystem.EnabledByDefault` property and sets it to false for the legacy project.json project system.


filipw:
We also have a first level of configuration via an optional `config.json` file [loaded here](https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp.Host/ConfigurationBuilder.cs#L33). 

We used to ship it in the past, but then at some point it got lost (I really don't know what happened to it, I think it genuinely got lost) and it's not there anymore (though it was only setting some defaults for dotnet project system AFAIR and the defaults are hardcoded anyway).
 
So technically the same effect could be achieved by re-shipping the `config.json` at OmniSharp root again, with the following content:

```
{
  "dotnet": {
     "enabled": false
  }
}
```

Regardless, I'm voting with my hand and feet to have the dotnet project system disabled by default 😀

filipw:
that one is a good reason to celebrate 😃

PR #1195, opened 2018-05-18
Fix bug when multi-TFM project reference is outside OmniSharp's target path

DustinCampbell:
Fixes https://github.com/OmniSharp/omnisharp-vscode/issues/2295

When OmniSharp loads a project reference outside the OmniSharp path, it does not attempt to do a full design-time build of the project. Instead, it just uses the MSBuild project evaluation. However, an important code path was being skipped where a target framework is selected in the case that the project has multiple target frameworks. Later, this causes an exception to be thrown because several properties did not evaluate properly. This change ensures that a target framework is selected in this case.

### DIFF
--- a/src/OmniSharp.MSBuild/ProjectLoader.cs
+++ b/src/OmniSharp.MSBuild/ProjectLoader.cs
@@ -110,7 +110,11 @@ namespace OmniSharp.MSBuild
 
             toolsVersion = GetLegalToolsetVersion(toolsVersion, projectCollection.Toolsets);
 
-            return projectCollection.LoadProject(filePath, toolsVersion);
+            var project = projectCollection.LoadProject(filePath, toolsVersion);
+
+            SetTargetFrameworkIfNeeded(project);
+
+            return project;
         }
 
         private static void SetTargetFrameworkIfNeeded(MSB.Evaluation.Project evaluatedProject)
@@ -127,10 +131,7 @@ namespace OmniSharp.MSBuild
                 // do better and potentially allow OmniSharp hosts to select a target framework.
                 targetFramework = targetFrameworks[0];
                 evaluatedProject.SetProperty(PropertyNames.TargetFramework, targetFramework);
-            }
-            else if (!string.IsNullOrWhiteSpace(targetFramework) && targetFrameworks.Length == 0)
-            {
-                targetFrameworks = ImmutableArray.Create(targetFramework);
+                evaluatedProject.ReevaluateIfNecessary();
             }
         }
 
--- a/build.json
+++ b/build.json
@@ -37,7 +37,8 @@
     "TwoProjectsWithSolution",
     "ProjectWithGeneratedFile",
     "CSharpAndFSharp",
-    "ProjectWithMismatchedFileName"
+    "ProjectWithMismatchedFileName",
+    "ProjectWithMultiTFMLib"
   ],
   "LegacyTestAssets": [
     "LegacyNUnitTestProject",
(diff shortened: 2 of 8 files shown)


DustinCampbell:
@david-driscoll / @filipw : @rchande is currently on vacation. Do either of you have confidence to sign off on this change?

src/OmniSharp.MSBuild/ProjectLoader.cs
```diff
@@ -110,7 +110,11 @@ private MSB.Evaluation.Project EvaluateProjectFileCore(string filePath)
 
             toolsVersion = GetLegalToolsetVersion(toolsVersion, projectCollection.Toolsets);
 
-            return projectCollection.LoadProject(filePath, toolsVersion);
+            var project = projectCollection.LoadProject(filePath, toolsVersion);
+
+            SetTargetFrameworkIfNeeded(project);
```
DustinCampbell: Every project goes through this code path, where we first evaluate the project with MSBuild. At this point, we need to set the `<TargetFramework/>` property if there are multiple `<TargetFrameworks/>` to ensure that the project evaluates with a TFM. Otherwise, several MSBuild properties (like `TargetFrameworkMoniker`) are missing.

src/OmniSharp.MSBuild/ProjectLoader.cs
```diff
@@ -127,10 +131,7 @@ private static void SetTargetFrameworkIfNeeded(MSB.Evaluation.Project evaluatedP
                 // do better and potentially allow OmniSharp hosts to select a target framework.
                 targetFramework = targetFrameworks[0];
                 evaluatedProject.SetProperty(PropertyNames.TargetFramework, targetFramework);
-            }
-            else if (!string.IsNullOrWhiteSpace(targetFramework) && targetFrameworks.Length == 0)
```
DustinCampbell: This check is not needed here. We used to pass `targetFrameworks` as an out parameter but that was removed long ago. So, this is essentially unnecessary code.

src/OmniSharp.MSBuild/ProjectLoader.cs
```diff
@@ -127,10 +131,7 @@ private static void SetTargetFrameworkIfNeeded(MSB.Evaluation.Project evaluatedP
...
-            {
-                targetFrameworks = ImmutableArray.Create(targetFramework);
+                evaluatedProject.ReevaluateIfNecessary();
```
DustinCampbell: Here, we tell the project to re-evaluate itself if it's dirty, which will be the case because we set the `<TargetFramework>` property.

DustinCampbell:
I added some comments to describe what the code changes do. The rest is just test code.

src/OmniSharp.MSBuild/ProjectLoader.cs
```diff
@@ -127,10 +131,7 @@ private static void SetTargetFrameworkIfNeeded(MSB.Evaluation.Project evaluatedP
                 // do better and potentially allow OmniSharp hosts to select a target framework.
                 targetFramework = targetFrameworks[0];
```
filipw: how could we allow this to be overrideable in the future? 
allow the user to set `preferred TFM` via `MSBuildOptions` in the config?

david-driscoll: We need create a singleton service or something that we can toggle the tfm for a given project, and then also enhance services to also understand that...

DustinCampbell: Yes, we need the same thing for setting the active solution/project configuration as well.

PR #1197, opened 2018-05-22
added a note about script RSP support to a CHANGELOG.md

filipw:
It was missing from the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,8 +39,8 @@ All changes to the project will be documented in this file.
     }
     ```   
     OmniSharp will automatically include the predefined DLLs and namespaces in the language services for all the scripts in the given folder (in case of a local `omnisharp.json`) or on the machine (in case of a global `omnisharp.json`). Note that the reference to `mscorlib`/`System.Runtime` is always there anyway and doesn't need to be specified again in the `.rsp` file. ([#1024](https://github.com/OmniSharp/omnisharp-roslyn/issues/1024), PR: [#1112](https://github.com/OmniSharp/omnisharp-roslyn/issues/1112))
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -18,6 +18,29 @@ All changes to the project will be documented in this file.
...
+    ```
+    {
+        "Script": {
+            "RspFilePath": "path/to/my.rsp"
+        }
+    }
+    ```   
+    OmniSharp will automatically include the predefined DLLs and namespaces in the language services for all the scripts in the given folder (in case of a local `omnisharp.json`) or on the machine (in case of a global `omnisharp.json`). Note that the reference to `mscorlib`/`System.Runtime` is always there anyway and doesn't need to be specified again in the `.rsp` file. ([#1024](https://github.com/OmniSharp/omnisharp-roslyn/issues/1024), PR: [#1112](https://github.com/OmniSharp/omnisharp-roslyn/issues/1112))
+      * Note that the reference to `mscorlib`/`System.Runtime` is always there anyway and doesn't need to be specified again in the `.rsp` file
+      * only imports and references are supported as part of the `.rsp` file (scripting doesn't support other compiler settings passed using the `.rsp` file). In the future, depending on whether the [feature is available in Roslyn](https://github.com/dotnet/roslyn/issues/23421), OmniSharp may also support defining a scripting globals type via `.rsp` file.
```
DustinCampbell: I think this bullet needs to be indented by four spaces instead of six to be rendered as a bullet. GitHub's markdown preview is showing this all jumbled up.

filipw: you are right, fixed it now. 
I am using [MacDown](https://macdown.uranusjr.com) for md editing and Markdown output likes to differ from parser to parser 😀

PR #1201, opened 2018-05-23
Fix diagnostics from loaded cake files shown in the current file

mholo65:
Fixes issue reported [here](https://github.com/cake-build/bakery/issues/90)

### DIFF
--- a/src/OmniSharp.Cake/Services/RequestHandlers/Diagnostics/CodeCheckHandler.cs
+++ b/src/OmniSharp.Cake/Services/RequestHandlers/Diagnostics/CodeCheckHandler.cs
@@ -1,8 +1,12 @@
-﻿using System.Composition;
+﻿using System;
+using System.Composition;
+using System.IO;
+using System.Linq;
 using System.Threading.Tasks;
 using OmniSharp.Mef;
 using OmniSharp.Models;
 using OmniSharp.Models.CodeCheck;
+using OmniSharp.Utilities;
 
 namespace OmniSharp.Cake.Services.RequestHandlers.Diagnostics
 {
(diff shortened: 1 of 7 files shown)


mholo65:
ping @jnm2

src/OmniSharp.Cake/Services/RequestHandlers/Diagnostics/CodeCheckHandler.cs
```diff
@@ -18,5 +22,33 @@ public class CodeCheckHandler : CakeRequestHandler<CodeCheckRequest, QuickFixRes
...
+                return Task.FromResult(response);
+            }
+
+            var quickFixes = response.QuickFixes.Where(x => PathsAreEqual(x.FileName, request.FileName));
+            response.QuickFixes = quickFixes;
+            return Task.FromResult(response);
+
+            bool PathsAreEqual(string x, string y)
+            {
+                if (x == null && y == null)
```
jnm2: Could be `ReferenceEquals(x, y)` to be even more efficient in case the same string instance happens to be reused.

src/OmniSharp.Cake/Services/RequestHandlers/Diagnostics/CodeCheckHandler.cs
```diff
@@ -18,5 +22,33 @@ public class CodeCheckHandler : CakeRequestHandler<CodeCheckRequest, QuickFixRes
...
+                {
+                    return true;
+                }
+                if (x == null || y == null)
```
jnm2: Another tip I had from a contributor at corefx was that `x == null | y == null` performs better than the branch introduced by `||`. Not that this is performance-critical though.

PR #1209, opened 2018-05-31
Add a /blockstructure endpoint that uses Roslyn's outlining implemention

rchande:
We'll use this to implement language-specific folding in the C# extension


src/OmniSharp.Roslyn.CSharp/Services/Structure/BlockStructureService.cs
```diff
@@ -0,0 +1,76 @@
...
+using OmniSharp.Models;
+using OmniSharp.Services;
+using OmniSharp.Utilities;
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Composition;
+using System.Reflection;
+using System.Threading;
+using System.Threading.Tasks;
```
DustinCampbell: Sort with System namespaces on top.

src/OmniSharp.Roslyn.CSharp/Services/Structure/BlockStructureService.cs
```diff
@@ -0,0 +1,76 @@
...
+            _featureAssembly = _loader.LazyLoad(Configuration.RoslynFeatures);
+
+            _blockStructureService = _featureAssembly.LazyGetType("Microsoft.CodeAnalysis.Structure.BlockStructureService");
+            _blockStructure = _featureAssembly.LazyGetType("Microsoft.CodeAnalysis.Structure.BlockStructure");
+            _blockSpan = _featureAssembly.LazyGetType("Microsoft.CodeAnalysis.Structure.BlockSpan");
+
+            _getBlockStructure = _blockStructureService.LazyGetMethod("GetBlockStructure");
+            _getSpans = _blockStructure.Value.GetProperty("Spans");
+            _getIsCollpasible = _blockSpan.Value.GetProperty("IsCollapsible");
+            _getTextSpan = _blockSpan.Value.GetProperty("TextSpan");
```
DustinCampbell: Could you file an issue on https://github.com/dotnet/roslyn/issues to add a public API here?

src/OmniSharp.Roslyn.CSharp/Services/Structure/BlockStructureService.cs
```diff
@@ -0,0 +1,76 @@
...
+        public async Task<IEnumerable<QuickFix>> Handle(Request request)
+        {
+            var document = _workspace.GetDocument(request.FileName);
+            var text = await document.GetTextAsync();
+            var lines = text.Lines;
+
+            var service = _blockStructureService.LazyGetMethod("GetService").InvokeStatic(new[] { document });
+
+            var structure = _getBlockStructure.Invoke<object>(service, new object[] { document, CancellationToken.None });
+            IEnumerable spans = _getSpans.GetMethod.Invoke<IEnumerable>(structure, Array.Empty<object>());
```
DustinCampbell: var?

src/OmniSharp.Roslyn.CSharp/Services/Structure/BlockStructureService.cs
```diff
@@ -0,0 +1,76 @@
...
+
+            List<QuickFix> outliningSpans = new List<QuickFix>();
```
DustinCampbell: var>

src/OmniSharp.Roslyn.CSharp/Services/Structure/BlockStructureService.cs
```diff
@@ -0,0 +1,76 @@
...
+            foreach (var span in spans)
+            {
+                if (_getIsCollpasible.GetMethod.Invoke<bool>(span, Array.Empty<object>()))
+                {
+                    var textSpan = _getTextSpan.GetMethod.Invoke<TextSpan>(span, Array.Empty<object>());
+                    outliningSpans.Add(new QuickFix()
+                    {
+                        Line = lines.GetLineFromPosition(textSpan.Start).LineNumber,
+                        EndLine = lines.GetLineFromPosition(textSpan.End).LineNumber
+                    });
```
DustinCampbell: Is there are reason not to add `Column` and `EndColumn` too?

src/OmniSharp.Roslyn.CSharp/Services/Structure/BlockStructureService.cs
```diff
@@ -0,0 +1,76 @@
...
+using System.Collections.Generic;
+using System.Composition;
+using System.Reflection;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace OmniSharp.Roslyn.CSharp.Services.Structure
+{
+    [OmniSharpHandler(OmniSharpEndpoints.BlockStructure, LanguageNames.CSharp)]
+    public class BlockStructureService : IRequestHandler<Request, IEnumerable<QuickFix>>
```
DustinCampbell: Is `QuickFix` enough? I have two thoughts:

1. VS Code's `FoldingRange` also has a `FoldingRangeKind` that we might want to expose, so we might want a custom model.
2. `QuickFix` is a "kitchen sink" model with lots of baggage that even unnecessarily allocates a `List<string>`. I've never been a fan of overusing it in the protocol. If most of the properties on `QuickFix` aren't interesting, it's probably cleaner to just define a new `BlockStructure` model.

DustinCampbell:
This aligns well with the `/codestructure` end point that I'm working on that'll serve as a replacement for `/currentfilemembersastree` and `/currentfilemembersasflat`

src/OmniSharp.Abstractions/OmniSharpEndpoints.cs
```diff
@@ -25,6 +25,7 @@ public static class OmniSharpEndpoints
         public const string Rename = "/rename";
         public const string SignatureHelp = "/signatureHelp";
         public const string MembersTree = "/currentfilemembersastree";
+        public const string BlockStructure = "/blockstructure";
```
DustinCampbell: Should this be a `V2` endpoint like other new endpoints that have been added?

tests/OmniSharp.Roslyn.CSharp.Tests/BlockStructureFacts.cs
```diff
@@ -0,0 +1,58 @@
...
+        {
+        }
+    }
+}");
+            var testFile = new TestFile("bar.cs", @"class Bar {}");
+
+            var lineSpans = (await GetResponseAsync(testFile1)).
+                    Select(l => (l.Line, l.EndLine)).ToArray();
+
+            var expected = new [] { (0, 8), (2, 7), (4, 6) };
```
DustinCampbell: There is a markup feature for tests in OmniSharp that is similar to Roslyn's that you can use instead of raw numbers. You can see it used in places like this: https://github.com/OmniSharp/omnisharp-roslyn/blob/master/tests/OmniSharp.Roslyn.CSharp.Tests/FormattingFacts.cs#L141

src/OmniSharp.Roslyn.CSharp/Services/Structure/BlockStructureService.cs
rchande: https://github.com/dotnet/roslyn/issues/27332

rchande:
@DustinCampbell I incorporated your feedback. Want to sign off?

src/OmniSharp.Abstractions/Models/v2/BlockSpan.cs
```diff
@@ -0,0 +1,30 @@
...
+            TextSpan = textSpan;
+            HintSpan = hintSpan;
+            BannerText = bannerText;
+            Type = type;
+        }
+
+        /// <summary>
+        /// The span of text to collapse.
+        /// </summary>
+        public Range TextSpan { get; }
```
DustinCampbell: We might consider using "Range" terminology here. I don't think "Span" appears anywhere in OmniSharp's protocol.

src/OmniSharp.Abstractions/Models/v2/BlockSpan.cs
```diff
@@ -0,0 +1,30 @@
...
+
+        /// <summary>
+        /// The span of text to display in the hint on mouse hover.
+        /// </summary>
+        public Range HintSpan { get; }
```
DustinCampbell: Ditto here.

src/OmniSharp.Abstractions/Models/v2/BlockSpan.cs
```diff
@@ -0,0 +1,30 @@
+﻿namespace OmniSharp.Models.V2
+{
+    public class BlockSpan
```
DustinCampbell: Should this be more specific to the purpose its indeed? E.g. `CodeFoldingBlock` or something like that?

src/OmniSharp.Abstractions/Models/v2/BlockSpan.cs
```diff
@@ -0,0 +1,30 @@
...
+        /// The span of text to display in the hint on mouse hover.
+        /// </summary>
+        public Range HintSpan { get; }
+
+        /// <summary>
+        /// The text to display inside the collapsed region.
+        /// </summary>
+        public string BannerText { get; }
+
+        public string Type { get; }
```
DustinCampbell: What is this? What values are legal?

src/OmniSharp.Abstractions/Models/v2/Point.cs
```diff
@@ -8,5 +8,21 @@ public class Point
...
+
+        public override bool Equals(object obj)
+        {
+            var point = obj as Point;
+            return point != null &&
+                   Line == point.Line &&
+                   Column == point.Column;
+        }
+
+        public override int GetHashCode()
```
DustinCampbell: This will conflict with my PR. Sorry about that!

src/OmniSharp.Abstractions/Models/v2/Range.cs
```diff
@@ -1,8 +1,26 @@
...
+
+        public override bool Equals(object obj)
+        {
+            var range = obj as Range;
+            return range != null &&
+                   EqualityComparer<Point>.Default.Equals(Start, range.Start) &&
+                   EqualityComparer<Point>.Default.Equals(End, range.End);
+        }
+
+        public override int GetHashCode()
```
DustinCampbell: The `Equals` and `GetHashCode` will conflict with my PR. Sorry!

src/OmniSharp.Roslyn.CSharp/Services/Structure/BlockStructureService.cs
```diff
@@ -0,0 +1,92 @@
...
+            var document = _workspace.GetDocument(request.FileName);
+            var text = await document.GetTextAsync();
+            var lines = text.Lines;
+
+            var service = _blockStructureService.LazyGetMethod("GetService").InvokeStatic(new[] { document });
+
+            var structure = _getBlockStructure.Invoke<object>(service, new object[] { document, CancellationToken.None });
+            var spans = _getSpans.Invoke<IEnumerable>(structure, Array.Empty<object>());
+
+
```
DustinCampbell: nit: extra blank line.

src/OmniSharp.Roslyn.CSharp/Services/Structure/BlockStructureService.cs
```diff
@@ -0,0 +1,92 @@
...
+                    var line = lines.GetLineFromPosition(textSpan.Start);
+                    var column = textSpan.Start - line.Start;
+                    var endLine = lines.GetLineFromPosition(textSpan.End);
+                    var endColumn = textSpan.End - endLine.Start;
+
+                    outliningSpans.Add(new BlockSpan(
+                        textSpan.ToRange(lines),
+                        hintSpan: _getHintSpan.Invoke<TextSpan>(span, Array.Empty<object>()).ToRange(lines),
+                        bannerText: _getBannerText.Invoke<string>(span, Array.Empty<object>()),
+                        type: _getType.Invoke<string>(span, Array.Empty<object>())));
```
DustinCampbell: What is type?

src/OmniSharp.Roslyn.CSharp/Services/Structure/BlockStructureService.cs
```diff
@@ -0,0 +1,92 @@
...
+            {
+                if (_getIsCollpasible.Invoke<bool>(span, Array.Empty<object>()))
+                {
+                    var textSpan = _getTextSpan.Invoke<TextSpan>(span, Array.Empty<object>());
+                    var line = lines.GetLineFromPosition(textSpan.Start);
+                    var column = textSpan.Start - line.Start;
+                    var endLine = lines.GetLineFromPosition(textSpan.End);
+                    var endColumn = textSpan.End - endLine.Start;
+
+                    outliningSpans.Add(new BlockSpan(
```
DustinCampbell: If I recall correctly, Roslyn's block structure API handles spans for code folding *and* vertical lines. Do we want to include all of the inner spans that are intended for vertical lines in the editor? Or, should this be filtered to just the block spans that are used for outlining?

src/OmniSharp.Roslyn.CSharp/Services/Structure/BlockStructureService.cs
```diff
@@ -0,0 +1,92 @@
...
+                        hintSpan: _getHintSpan.Invoke<TextSpan>(span, Array.Empty<object>()).ToRange(lines),
+                        bannerText: _getBannerText.Invoke<string>(span, Array.Empty<object>()),
+                        type: _getType.Invoke<string>(span, Array.Empty<object>())));
+                }
+            }
+
+            return new BlockStructureResponse() { Spans = outliningSpans };
+        }
+
+
```
DustinCampbell: nit: unnecesssary blank lines.

tests/OmniSharp.Roslyn.CSharp.Tests/BlockStructureFacts.cs
```diff
@@ -0,0 +1,58 @@
...
+}|]");
+            var text = testFile.Content.Text;
+
+            var lineSpans = (await GetResponseAsync(testFile)).Spans
+                .Select(b => b.TextSpan)
+                .ToArray();
+
+
+            var expected = testFile.Content.GetSpans()
+                .Select(t => t.ToRange(text.Lines)).ToArray();
```
DustinCampbell: Note that there's already a `TestContent.GetRangeFromSpan(...)` helper that you can use. Ultimately that calls an existing extension method in TestUtility on `SourceText`: https://github.com/OmniSharp/omnisharp-roslyn/blob/master/tests/TestUtility/SourceTextExtensions.cs.

src/OmniSharp.Roslyn/Extensions/TextSpanExtensions.cs
```diff
@@ -0,0 +1,22 @@
+﻿using Microsoft.CodeAnalysis.Text;
+using OmniSharp.Models.V2;
+
+namespace OmniSharp.Roslyn.Extensions
+{
+    public static class TextSpanExtensions
+    {
+        public static Range ToRange(this TextSpan textSpan, TextLineCollection lines)
```
DustinCampbell: I don't think you need to add this extension method. See my comment below.

tests/OmniSharp.Roslyn.CSharp.Tests/BlockStructureFacts.cs
```diff
@@ -0,0 +1,58 @@
...
+        protected override string EndpointName => OmniSharpEndpoints.V2.BlockStructure;
+
+        [Fact]
+        public async Task UsesRoslynBlockStructureService()
+        {
+            var testFile = new TestFile("foo.cs", @"class Foo[|
+{
+    void M()[|
+    {
+        if (false)[|
```
DustinCampbell: Note that this isn't collapsible in Visual Studio. I think we might be conflating the vertical indicators in VS with code folding.

tests/OmniSharp.Roslyn.CSharp.Tests/BlockStructureFacts.cs
```diff
@@ -0,0 +1,58 @@
...
+
+            var expected = testFile.Content.GetSpans()
+                .Select(t => t.ToRange(text.Lines)).ToArray();
+
+            Assert.Equal(expected, lineSpans);
+        }
+
+        private async Task<BlockStructureResponse> GetResponseAsync(TestFile testFile)
+        {
+            SharedOmniSharpTestHost.AddFilesToWorkspace(new[] { testFile });
```
DustinCampbell: nit: you don't need to construct an array here. It's a params array parameter.

tests/OmniSharp.Roslyn.CSharp.Tests/BlockStructureFacts.cs
```diff
@@ -0,0 +1,58 @@
...
+            var request = new BlockStructureRequest
+            {
+                FileName = testFile.FileName,
+            };
+
+            var requestHandler = GetRequestHandler(SharedOmniSharpTestHost);
+            return await requestHandler.Handle(request);
```
DustinCampbell: There's no need to make this method async and await here. Just return the Task.

tests/OmniSharp.Roslyn.CSharp.Tests/BlockStructureFacts.cs
rchande: Well, actually 😄.   : 
image

tests/OmniSharp.Roslyn.CSharp.Tests/BlockStructureFacts.cs
rchande: That extension returns `TestUtiltity.TextRange`, but my product code returns `OmniSharp.Models.V2.Range`. These types have essentially the same shape and there are even duplicated test/product `Point` classes. I don't really want to write code that knows how to compare test and product Range/Points. Thoughts on the best way to resolve this?

src/OmniSharp.Abstractions/Models/v2/BlockSpan.cs
rchande: This is property on the internal Roslyn types that's also stringly-typed. I wasn't sure if we would need to use it at some point so I left it in, but I can also remove it.

DustinCampbell: For example, we should provide the possible values here so that OmniSharp consumers understand what it's used for.

src/OmniSharp.Abstractions/Models/v2/BlockStructureRequest.cs
```diff
@@ -0,0 +1,9 @@
+﻿using OmniSharp.Mef;
+
+namespace OmniSharp.Models.v2
+{
+    [OmniSharpEndpoint(OmniSharpEndpoints.V2.BlockStructure, typeof(BlockStructureRequest), typeof(BlockStructureResponse))]
+    public class BlockStructureRequest : Request
```
DustinCampbell: This should probably just inherit from `SimpleFileRequest`. It doesn't need to support all of the properties from `Request`.

src/OmniSharp.Abstractions/Models/v2/CodeFoldingBlock.cs
```diff
@@ -0,0 +1,29 @@
...
+        /// <summary>
+        /// If the block is one of the types specified in <see cref="CodeFoldingBlockKind"/>, that type.
+        /// Otherwise, null.
+        /// </summary>
+        public string Type { get; }
+    }
+
+    public class CodeFoldingBlockKind
+    {
+        public static readonly string Comment = nameof(Comment).ToLowerInvariant();
```
rchande: I chose to include these values because the VS Code folding API has special handling for these types of blocks

rchande:
@DustinCampbell I think I got everything, care to sign off now?

PR #1193, opened 2018-05-18
Added tests for CSX project system

filipw:
This PR - as suggested by @rchande on Slack 😀- adds several tests for the CSX project system (a simple CSI script, 2 CSI scripts together, a simple .NET Core script, a .NET Core script with a Nuget reference). 

The tests mimic some of the project tests done by the MsBuild project system - the workspace gets initialized, and then we use `WorkspaceInformationRequest` to retrieve the project system info about what OmniSharp has found there.

This gives us a long, long overdue (I know 😓) "first level" of an automated check - in case some change in the future breaks the project system. 
Irrespective of this PR, I'd still like to add an integration test in the VS Code extension repo.

PR #1210, opened 2018-06-01
Initialize CSX file watcher also when no CSX files are available at startup

### LINKED ISSUE #1199: Working with *.csx and intellisense is messing up
I'm working on an asp.net core solution. In that solution the user can add a c# script via monaco editor. For better usage I'm running Omnisharp as a service and send the /autocomplete to omnisharp.
Omnisharp is started this way: Omnisharp.exe -s Path
The scripts are stored on the file system as a xxx.csx file. When starting omnisharp when a csx file is present it works well. Added own assemblies as #r "XY.dll". 
If omnisharp starts without the *.csx file the intellisense is not working and from /codecheck I'm also getting warnings that #r can only be used in scripts.

What would be a smart way to handle this?

filipw:
Fixes #1199 
At the moment we only initialized CSX file watcher (reacting to new or removed files) when there were any CSX files discovered at startup, which was mentioned and discussed in #1199 

This PR fixes that, and we also listen when no CSX files are available at startup, so as soon as you add your *first* one, language services light up.
This whole change would have been a single line https://github.com/OmniSharp/omnisharp-roslyn/compare/master...filipw:bugfix/1199?expand=1#diff-ed5a3c116a19423c18b1e630f80455b9R70 but I took a liberty to rationalize some of the code around scripting.

For example renamed `ScriptHelper` to `ScriptProjectProvider`, and extracted a `ScriptContext` and `ScriptContextProvider` from the spaghetti project system. While this made the diff rather big (sorry), it shouldn't affect anything.

Finally, ignored the ordering on one of the flaky tests (https://github.com/OmniSharp/omnisharp-roslyn/pull/1193#pullrequestreview-125128489) and added a test for this new watching feature.


CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 # Changelog
 All changes to the project will be documented in this file.
 
+## [1.32.0] - not yet released
+* Fixed a bug where lanugage services for newly created CSX files were not provided if no CSX files existed at the moment OmniSharp was started ([#1199](https://github.com/OmniSharp/omnisharp-roslyn/issues/1199), PR: [#1210](https://github.com/OmniSharp/omnisharp-roslyn/pull/1210))
```
DustinCampbell: typo: "lanugage" -> "language"

tests/TestUtility/DisposableFile.cs
```diff
@@ -0,0 +1,69 @@
+﻿using System;
+using System.IO;
+
+namespace TestUtility
+{
+    public class DisposableFile : IDisposable
```
DustinCampbell: Is there anything here that can be shared with the code in `TestProject`?

filipw: good point, I think I will remove this completely, and instead allow a new file to be added to an existing `TestProject`. If the project was shadowcopied, the file will disappear with it, and it if it wasn't we will delete it during the disposal of `TestProject`. This will keep all I/O clean up operations in `TestProject` which is our unit of work in these tests

PR #1211, opened 2018-06-01
New /codestructure endpoint

DustinCampbell:
This change adds a new V2 `/codestructure` OmniSharp endpoint that is intended to replace the `/currentfilemembersastree`. The `/currentfilemembersastree` endpoint is used for a number of things by C# for VS Code, including code lens, test method detection, etc. However, it has a lot of baggage associated with it and lots of bugs. For example:

* Many code elements were not included, such as namespaces, destructors, operators, event declarations (as opposed to field-like events), indexers, etc.
* The "range" provide by `/currentfilemembersastree` is actually the span of a member's identifier name rather than the span of the whole member. This results in various VS Code extensions that use the `documentSymbolProvider` API with C# to get odd results where members aren't contained within one another (https://github.com/patrys/vscode-code-outline/issues/47).
* There's lots of missing information that might be useful, such as accessibility, instance vs. static, display name, etc.

Honestly, it feels like `/currentfilemembersastree` was thrown together quickly to make a handful of features work and then forgotten.

 `/codestructure` should be a much cleaner design.

* All C# types and members are included.
* The models are clean (no usage of `QuickFix` anywhere :smile:).
* Ranges are a first class concept. Currently, code elements can support two ranges: `"attributes"` and `"full"`. However, adding others (e.g. the "name" range) should be easy if needed and won't impact existing client code.
* Lot's more information is included, such as a display name, accessibility, etc.
* It no longer unnecessarily walks the whole tree with a `CSharpSyntaxWalker` when only the top-level declarations are needed.
* The extensibility point that allows other features to provide properties for a given code element is now rational.


src/OmniSharp.Abstractions/Models/v2/CodeStructure/CodeElement.cs
```diff
@@ -0,0 +1,31 @@
...
+
+namespace OmniSharp.Models.V2.CodeStructure
+{
+    public partial class CodeElement
+    {
+        public string Kind { get; }
+        public string Name { get; }
+        public string DisplayName { get; }
+        public IReadOnlyList<CodeElement> Children { get; }
+        public IReadOnlyDictionary<string, Range> Ranges { get; }
```
rchande: I just noticed this while working on BlockStructure--seems like the Point type has a "ZeroBasedIndexJsonConverter" on its fields which results in the serialized values from OmniSharp being one-based instead of zero-based. Very annoying if you're consuming in C# extension...

DustinCampbell: Yes, but that's the right thing. OmniSharp is one-based by default so we should keep it consistent. There's an option that can passed to OmniSharp to make everything zero-based. We should probably just invest in C# for VS Code to convert it over to using OmniSharp in a zero-based fashion.

rchande: Ah, good to know. Thanks!

src/OmniSharp.Abstractions/Models/v2/CodeStructure/CodeElementKinds.cs
```diff
@@ -0,0 +1,27 @@
...
+    public static class CodeElementKinds
+    {
+        // types
+        public static readonly string Class = nameof(Class).ToLowerInvariant();
+        public static readonly string Delegate = nameof(Delegate).ToLowerInvariant();
+        public static readonly string Enum = nameof(Enum).ToLowerInvariant();
+        public static readonly string Interface = nameof(Interface).ToLowerInvariant();
+        public static readonly string Struct = nameof(Struct).ToLowerInvariant();
+
+        // members
```
filipw: should this include `IndexerDeclaration` too?

DustinCampbell: Yup. I just noticed that. Fixed.

src/OmniSharp.Roslyn.CSharp/Services/Structure/CodeStructureService.cs
```diff
@@ -0,0 +1,411 @@
...
+        {
+            var accessibility = GetAccessibility(symbol);
+            if (accessibility != null)
+            {
+                builder.AddProperty(CodeElementPropertyNames.Accessibility, accessibility);
+            }
+
+            builder.AddProperty(CodeElementPropertyNames.Static, symbol.IsStatic);
+
+            foreach (var propertyProvider in _propertyProviders)
```
filipw: I like the extensibility, much cleaner than the existing strange bolt-on 🙂

src/OmniSharp.Abstractions/Models/v2/CodeStructure/CodeElement.cs
DustinCampbell: Yeah, here's the flag: https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp.Host/CommandLineApplication.cs#L34

We can pass that in C# for VS Code when we launch OmniSharp, but we'll need to update a ton of places in VS Code where we convert from one-based.

filipw:
out of curiosity, is this going to include local functions?

src/OmniSharp.Roslyn.CSharp/Services/Structure/CodeStructureService.cs
DustinCampbell: I was not a fan of the bolt on. :smile:

DustinCampbell:
> out of curiosity, is this going to include local functions?

I wasn't intending to do that yet. It raises questions about whether locals and parameter should appear as well.

src/OmniSharp.Abstractions/Models/v2/CodeStructure/CodeElement.cs
david-driscoll: Indeed the original editors vim/emacs were all one based, hence that's how OmniSharp has always behaved.  I added the converter to make it easier to write against Roslyn but lets OmniSharp respect the editors design.

DustinCampbell: Yup! Well said @david-driscoll

PR #1212, opened 2018-06-04
GH1205: Add support for code actions in Cake

### LINKED ISSUE #1205: [Cake] Code actions not working
Code actions are returned properly, but when executing them, nothing happens.

image

We need to implement similar handlers as can be found [here](https://github.com/OmniSharp/omnisharp-roslyn/tree/92afdca886b344e27690d83c472636e0d08648fe/src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2) for `.cake` files.

mholo65:
- Fixes #1205

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ All changes to the project will be documented in this file.
         }
     }
     ```
+* Added support for code actions in `.cake` files. ([#1205](https://github.com/OmniSharp/omnisharp-roslyn/issues/1205), PR: [#1212](https://github.com/OmniSharp/omnisharp-roslyn/pull/1212))
 
 ## [1.31.1] - 2018-05-28
 * Fixed bug where diagnostics from loaded `.cake` files was shown in the current file. (PR: [#1201](https://github.com/OmniSharp/omnisharp-roslyn/pull/1201))
--- a/src/OmniSharp.Cake/Extensions/ResponseExtensions.cs
+++ b/src/OmniSharp.Cake/Extensions/ResponseExtensions.cs
@@ -7,6 +7,7 @@ using OmniSharp.Models;
 using OmniSharp.Models.Navigate;
 using OmniSharp.Models.MembersTree;
 using OmniSharp.Models.Rename;
+using OmniSharp.Models.V2;
 
 namespace OmniSharp.Cake.Extensions
 {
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 # Changelog
 All changes to the project will be documented in this file.
 
+## [1.32.0] - _Not Yet Released_
+* Added support for code actions in `.cake` files. ([#1205](https://github.com/OmniSharp/omnisharp-roslyn/issues/1205))
```
DustinCampbell: I won the merge race. This will probably have a conflict.

src/OmniSharp.Cake/Extensions/ResponseExtensions.cs
```diff
@@ -85,6 +86,44 @@ public static Task<QuickFixResponse> TranslateAsync(this QuickFixResponse respon
...
+                    await PopulateModificationsAsync(modifiedFile, workspace, changes);
+                }
+
+                fileOperations.Add(fileOperation);
+            }
+
+            foreach (var change in changes)
+            {
+
+                if (!(fileOperations.FirstOrDefault(x => x.FileName == change.Key &&
```
DustinCampbell: You could just invert this to avoid having a `continue` and an ugly `!` before a pattern match. E.g.

```C#
if (fileOperations.FirstOrDefault(x => x.FileName == change.Key &&x.ModificationType == FileModificationType.Modified) is ModifiedFileResponse modifiedFile)
{
    modifiedFile.Changes = change.Value;
}
```

src/OmniSharp.Cake/Extensions/ResponseExtensions.cs
```diff
@@ -85,6 +86,44 @@ public static Task<QuickFixResponse> TranslateAsync(this QuickFixResponse respon
...
+                                                         x.ModificationType == FileModificationType.Modified)
+                    is ModifiedFileResponse modifiedFile))
+                {
+                    continue;
+                }
+
+                modifiedFile.Changes = change.Value;
```
DustinCampbell: This means that `modifiedFile.Changes` will only ever have the last matching value. Is that intentional or should there by some sort of merging?

mholo65: No, `change` is of type `KeyValuePair<string, List<LinePositionSpanTextChange>>` and the dictionary it's contained in is populated above, using the `PopulateModificationsAsync`.

This is the same thing we do with rename responses also. it's ugly, but due to the fact that loaded scripts are "merged" into the base script we might end up in a situation where a `ModifiledFileResponse` contains modifications belonging to a different file. This is what we are trying to fix in this method, e.g. loop through all modifications, map line numbers and assign them to the correct `ModifiedFileResponse`.

Someday, Cake should not "copy & paste" loaded script contents in the base file and instead use `#load` as provided by Roslyn scripting.

src/OmniSharp.Cake/Extensions/ResponseExtensions.cs
mholo65: Yes, will fix this!

CHANGELOG.md
mholo65: Good that I caught my eyes on the `/codestructure` PR. We need to create a Cake endpoint for that too.. Created issue here #1214

src/OmniSharp.Cake/Extensions/ResponseExtensions.cs
DustinCampbell: got it. Thanks!

mholo65:
@DustinCampbell addressed feedback and rebased!

src/OmniSharp.Cake/Extensions/ResponseExtensions.cs
mholo65: @filipw I see somebody has already written a blog post that will help me getting started :wink: https://www.strathweb.com/2016/06/implementing-custom-load-behavior-in-roslyn-scripting/

PR #1215, opened 2018-06-05
Clean up V2 models and helpers

DustinCampbell:
This change just restructures the V2 models (now that we're adding more V2 end points) and cleans up some helpers and extension methods.


src/OmniSharp.Roslyn/Extensions/SymbolExtensions.cs
```diff
@@ -1,11 +1,185 @@
 using System;
+using System.Collections.Generic;
+using System.ComponentModel;
 using Microsoft.CodeAnalysis;
+using OmniSharp.Models.V2;
+using OmniSharp.Utilities;
 
 namespace OmniSharp.Extensions
 {
     public static class SymbolExtensions
```
filipw: shouldn't this be called `ISymbolExtensions` instead? technically those are `ISymbol` extensions. I believe Roslyn also has it's own `ISymbolExtensions`

src/OmniSharp.Roslyn/Utilities/SymbolDisplayFormats.cs
```diff
@@ -0,0 +1,39 @@
+﻿using Microsoft.CodeAnalysis;
+
+namespace OmniSharp.Roslyn.Utilities
+{
+    public static class SymbolDisplayFormats
```
filipw: good idea

src/OmniSharp.Roslyn/Extensions/SymbolExtensions.cs
DustinCampbell: This was actually a source of minor disagreement on the Roslyn team years ago. :smile: After all, interfaces are supposed to start with an `I` and extension methods can be called as normal static methods (e.g. `ISymbolExtensions.GetKindString(...)`.

In Roslyn, we're inconsistent:

* SymbolExtensions: https://github.com/dotnet/roslyn/blob/a73beac38d1563b2133108dab664a08d14981fa2/src/Test/Utilities/Portable/Extensions/SymbolExtensions.cs
* ISymbolExtensions: https://github.com/dotnet/roslyn/blob/cb21f0e9d5687b128a6c3188b348516c2b7b572d/src/Compilers/Core/Portable/Symbols/ISymbolExtensions.cs

I could go either way on this, but I chose with `SymbolExtensions` because this includes extension methods for several `ISymbol` derivatives as well (like `INamedTypeSymbol`). In other words, these are extension methods for "symbols", not a particular type.

PR #1216, opened 2018-06-07
Fix zero/one-based range and point models

DustinCampbell:
I apparently lost my mind the other day and added code to increment zero-based indeces to one-based indeces in the `OmniSharp.Models.V2.Point` type. Thankfully, @mholo65 found this while battling some other code and pointed it out. I've fixed this, added a few more helper extension methods for converting lines, offsets, spans, ranges and points, and added several tests for the `ZeroBasedIndexConverter` using specific OmniSharp protocol types.


filipw:
thanks - looks good! I'll merge it so that we have the correct behavior on master soonest

DustinCampbell:
Thanks!

PR #1217, opened 2018-06-07
GH1214: Add cake handler for /codestructure endpoint.

### LINKED ISSUE #1214: [Cake] Support V2 /codestructure endpoint in Cake
Now that #1211 is merged, we need to support this endpoint in Cake project system as well.

mholo65:
-Fixes #1214

### DIFF
--- a/src/OmniSharp.Cake/Extensions/ResponseExtensions.cs
+++ b/src/OmniSharp.Cake/Extensions/ResponseExtensions.cs
@@ -6,7 +6,9 @@ using OmniSharp.Models;
 using OmniSharp.Models.Navigate;
 using OmniSharp.Models.MembersTree;
 using OmniSharp.Models.Rename;
+using OmniSharp.Models.V2;
 using OmniSharp.Models.V2.CodeActions;
+using OmniSharp.Models.V2.CodeStructure;
 
 namespace OmniSharp.Cake.Extensions
 {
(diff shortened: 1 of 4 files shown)


src/OmniSharp.Cake/Extensions/ResponseExtensions.cs
```diff
@@ -121,6 +124,85 @@ public static Task<QuickFixResponse> TranslateAsync(this QuickFixResponse respon
...
+            foreach (var range in element.Ranges ?? Enumerable.Empty<KeyValuePair<string, Range>>())
+            {
+                builder.AddRange(range.Key, await range.Value.TranslateAsync(workspace, request));
+            }
+
+            foreach (var childElement in element.Children ?? Enumerable.Empty<CodeElement>())
+            {
+                var translatedElement = await childElement.TranslateAsync(workspace, request);
+
+                // This is plain stupid, but someone might put a #load directive inside a method or class
```
filipw: is this supported in Cake?
In scripting this would generate `CS8098 "Cannot use #load after first token in file"`
https://github.com/dotnet/roslyn/blob/ca2f6ad9ab68518344c2d3e25c0731c83d136701/src/Compilers/CSharp/Portable/Errors/ErrorCode.cs#L1329

mholo65: Yes it is supported since Cake only includes the content of `#load`:ed files. It doesn't rely on the Roslyn `#load` directive currently. I'm about to look into changing that behavior however 😄

PR #1230, opened 2018-06-18
Pin Nuget package versions using the new SDK feature

filipw:
This PR replaces the package versions that were pinned via MsBuild variables (`<MicrosoftExtensionsDependencyModelVersion>2.0.0</MicrosoftExtensionsDependencyModelVersion>` and so on) with the new MsBuild (shipped with SDK 2.1.300) pinning via `Directory.Build.targets` and the `<PackageReference Update="Microsoft.Extensions.DependencyModel" Version="2.0.0" />` style syntax.

It's a net gain on having less code and a bit easier on the eye.

I also took the liberty to remove the `OmniSharp.Nuget` project from the source control (we excluded it from the solution and don't package it or do anything with it, yet it was still unnecessarily participating in the build process via i.e. restore). It doesn't make sense to keep code around 'just in case'. If it's needed in the future we can always reach back into the history 😀

I also updated the xUnit test project to 2.3.1 since it was running on 2.2.0. As already widely documented in various PRs here, there is the need to pass the `targetFrameworkVersion` to vstest since xUnit 2.3.0 so needed to update the tests accordingly.


DustinCampbell:
I hadn't seen this feature in action yet. So the trick is to define `<PackageReference/>`s with `Update`?

filipw:
Yeah, at project level you just do `<PackageReference Include="Foo" />`, without the version. 
And then you globally set the versions using `<PackageReference/>` with `Update`.

I learned this from David [here](https://github.com/davkean/maket). He also mentioned that this feature is likely going to be incorporated into VS for 1st class support from the GUI.

filipw:
There was an additional bug in more complex scripting setups, with multiple `#load nuget:...` directives, that was described here https://github.com/filipw/dotnet-script/issues/294.

This scenario is now fixed in OmniSharp via this PR as I have updated the `Dotnet.Script.*` libraries to the latest versions.

Issue #1225, opened 2018-06-17
"The type 'class_name' exists in both..." error shows up in VSCode for types declared in project with delay signing enabled


dmgonch:
Repro steps:
1. Unzip ConsoleAppSigned.zip to say c:\ConsoleAppSigned (on Windows)
2. Open the solution in Visual Studio 2017 and build it (or do the build from Developer Command Prompt for VS2017)
3. Open c:\ConsoleAppSigned in VSCode with OmniSharp extension installed.
4. Observe the following error in Program.cs : "The type 'Class1' exists in both 'ClassLibrary1, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null' and 'ClassLibrary1, Version=1.0.0.0, Culture=neutral, PublicKeyToken=a5d85a5baa39a6a6'"

Expected: no error since there is actually only a single type ClassLibrary1 in the solution, from developer code of view at least.

[ConsoleAppSigned.zip](https://github.com/OmniSharp/omnisharp-roslyn/files/2108614/ConsoleAppSigned.zip)

![image](https://user-images.githubusercontent.com/30739723/41503842-aee74be4-7192-11e8-8f43-eb9f85f44e32.png)

PR #1234, opened 2018-06-25
Fixes bug https://github.com/OmniSharp/omnisharp-roslyn/issues/1225

dmgonch:
One potential controversial issue with the changes is that I tricked git to include bin/obj folders in the new test-assets\test-projects\DotNetCoreAppSigned. If desired I can zip the test files and unzip the project during the test.


dmgonch:
To clarify the need of the extra files in test assets: the assembly is delay signed and I couldn’t find a good repro without these extra files in obj/bin that Visual Studio generated.

src/OmniSharp.MSBuild/ProjectManager.cs
```diff
@@ -483,22 +483,31 @@ private void UpdateReferences(Project project, ImmutableArray<string> referenceP
...
                 {
                     _logger.LogWarning($"Unable to resolve assembly '{referencePath}'");
+                    continue;
                 }
-                else
+
+                // There is no need to explicitly add assembly to workspace when the assembly is produced by a project reference.
+                // Doing so actually can cause /codecheck request to return errors like below for types in the referenced project if it is for example signed:
+                // The type 'TestClass' exists in both 'SignedLib, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null' and 'ClassLibrary1, Version=1.0.0.0, Culture=neutral, PublicKeyToken=a5d85a5baa39a6a6'
+                if (TryFindLoadedProjectReferenceWithTargetPath(referencePath, projectReferencePaths, project.Name, out ProjectFileInfo projectReferenceWithTarget))
```
DustinCampbell: Should we just build up a `HashSet<string>` of project target paths ahead of time to avoid searching the list of project references for each metadata reference?

test-assets/test-projects/DotNetStdSigned/CalleeLib/obj/project.assets.json
```diff
@@ -0,0 +1,235 @@
...
+  "packageFolders": {
+    "C:\\Users\\dgonch\\.nuget\\packages\\": {},
+    "C:\\Program Files\\dotnet\\sdk\\NuGetFallbackFolder": {}
+  },
+  "project": {
+    "version": "1.0.0",
+    "restore": {
+      "projectUniqueName": "C:\\temp\\DotNetStdSigned\\CalleeLib\\CalleeLib.csproj",
+      "projectName": "CalleeLib",
+      "projectPath": "C:\\temp\\DotNetStdSigned\\CalleeLib\\CalleeLib.csproj",
```
DustinCampbell: Note that `project.assets.json` files contain absolute file paths to your machine. So, they're not actually valid on the test machine. Why are these needed?

dmgonch:
Now that you pointed me to build.json I see that there is no point in checking in the test outputs. Removed. Thanks!.

test-assets/test-projects/DotNetStdSigned/CalleeLib/obj/project.assets.json
dmgonch: Indeed after switching to use build.json as per your other comment I don't need test outputs in the sources. Removed.

src/OmniSharp.MSBuild/ProjectManager.cs
dmgonch: Given that typically the list of project references is relatively short IMO it is acceptable here to sacrifice a little bit of perf for avoiding managing an additional piece of workspace state, particularly for the cases when the list of project is updated.
Please advise if you still would prefer me to look into adding this optimization.

DustinCampbell: No need to make the change. Just asking.

PR #1237, opened 2018-06-30
Upgrade dependencies

david-driscoll:
NOTE: These changes are required for the latest LSP library.


README.md
```diff
@@ -8,7 +8,7 @@ omnisharp-roslyn
 
 OmniSharp-Roslyn is a .NET development platform based on [Roslyn](https://github.com/dotnet/roslyn) workspaces. It provides project dependencies and language syntax to various IDE and plugins.
 
-OmniSharp-Roslyn is built with the [.NET Core SDK](https://dot.net/) on Windows and [Mono](http://www.mono-project.com/) on OSX/Linux. It targets the __net46__ target framework. OmniSharp requires __mono__ (>=5.2.0) if it is run on a platform other than Windows.
+OmniSharp-Roslyn is built with the [.NET Core SDK](https://dot.net/) on Windows and [Mono](http://www.mono-project.com/) on OSX/Linux. It targets the __net461__ target framework. OmniSharp requires __mono__ (>=5.2.0) if it is run on a platform other than Windows.
```
mholo65: Are some of the new dependencies `netstandard` only (>1.3), or what was the reason for targeting `net461`?

david-driscoll: This is to allow for netstandard2.0 loading and support.  In addition https://github.com/OmniSharp/csharp-language-server-protocol is now netstandard2.0 (which is what caused me to start this separately)

build/Packages.props
```diff
@@ -2,57 +2,57 @@
...
+    <PackageReference Update="Cake.Scripting.Transport" Version="0.3.0" />
+
     <PackageReference Update="Dotnet.Script.DependencyModel" Version="0.6.1" />
     <PackageReference Update="Dotnet.Script.DependencyModel.NuGet" Version="0.6.2" />
 
-    <PackageReference Update="Microsoft.AspNetCore.Diagnostics" Version="1.1.0" />
-    <PackageReference Update="Microsoft.AspNetCore.Hosting" Version="1.1.0" />
-    <PackageReference Update="Microsoft.AspNetCore.Http.Features" Version="1.1.0" />
-    <PackageReference Update="Microsoft.AspNetCore.Server.Kestrel" Version="1.1.0" />
+    <PackageReference Update="McMaster.Extensions.CommandLineUtils" Version="2.2.4" />
```
filipw: nice. got caught by surprise with this one the other day when upgrading another project - could not find the new version, only to discover it is not supported by MS anymore 😀

src/OmniSharp.Abstractions/OmniSharp.Abstractions.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net46</TargetFramework>
+    <TargetFramework>net461</TargetFramework>
     <PlatformTarget>AnyCPU</PlatformTarget>
```
filipw: we could have all the libs as netstandard2.0 and just keep the entry projects net461

mholo65: :+1: on this!

DustinCampbell: be careful with this. There are known issues with netstandard2.0. I'd recommend sticking with net461 if there isn't a reason to switch.

src/OmniSharp.Abstractions/Logging/LoggerExtensions.cs
```diff
@@ -1,22 +0,0 @@
...
-        {
-            logger.LogError(0, ex, message, args);
-        }
-
-        public static void LogTrace(this ILogger logger, Exception ex, string message, params object[] args)
-        {
-            logger.LogTrace(0, ex, message, args);
-        }
-    }
-}
```
DustinCampbell: Cool! Are these now in the forked CommandLine package?

filipw: they are finally part of `Microsoft.Extensions.Logging`, we can now use them since we moved from 1.1.0 to 2.1.1 https://github.com/aspnet/Logging/pull/392

src/OmniSharp.Abstractions/OmniSharp.Abstractions.csproj
filipw: technically there is no real reason - other than maybe when we publish Nuget packages there might be easier to consume as `netstandard2.0`. That said I think the only reason anyone would consume OmniSharp packages would be to build things for OmniSharp so it wouldn't matter anyway

david-driscoll:
@filipw thanks I started on this the other day then git distracted by a :shipit: 😭

filipw:
🎉👍

PR #1239, opened 2018-07-03
Debug only when test manager is connected

akshita31:
Fixes: https://github.com/OmniSharp/omnisharp-vscode/issues/2400

If there is a Build failure the test manager fails to connect and the required variables for the debug process are not initialized. However since there is no check that the TestManager is connected or not, we go ahead with the debug request and the system fails with the exception `_writer is null`. However, since we entered the `StartSession` method, the required variables are set and it appears to omnisharp-roslyn that a debug session has started whereas there is no such session. Hence any subsequent debug request is rejected with the error - `Debug session started`.

Hence the fix is to verify whether the testManager is connected before we call the StartSession ,else fail with the relevant exception.

### DIFF
--- a/src/OmniSharp.DotNetTest/Services/DebugTestService.cs
+++ b/src/OmniSharp.DotNetTest/Services/DebugTestService.cs
@@ -1,4 +1,5 @@
-﻿using System.Composition;
+﻿using System;
+using System.Composition;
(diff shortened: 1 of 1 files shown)


rchande:
This is probably interesting enough to do a changelog entry :)

Issue #207, opened 2015-05-04
Files without a project


jrieken:
In VSCode we encounter solo `cs`-files which don't belong to a project. While we OmniSharp cannot provide semantic features (IntelliSense, type check, etc) we could enable syntactical features which require a parse tree only (Formatting, outline, etc)

nosami:
Heh... nice to hear you talk about VSCode on here :) 

We should be able to support this scenario already, by specifying the folder to the cs file. We do this in other editors already.

filipw:
@nosami This doesn't work right now AFAIK because the project systems (asp.net 5 and msbuild) try to find projects in the folder (sln and project.json respectively) and if they don't find any projects, they just exit without populating workspace with anything.

scriptcs project system is the only one that discovers loose files and adds them to workspace but it's `csx` only

nosami:
Ah, ok. I was getting confused with the old server which does work this way.

jrieken:
@filipw - and you go with mscorlib which comes with desktop clr/mono right? So no happy path for core clr users? wondering if we could funnel in some default lib for this case...

filipw:
yeah right now it just uses desktop/mono mscorlib. 
in fact, the whole csx thing is #ifdeffed against ASPNET50 too

I have started a separate repo so that all the same information could be retrieved from CSX files when omnisharp itself runs on coreclr https://github.com/filipw/scriptcs-dnx this is very rough outline now, but this approach should be replacing the embedded scriptcs dlls in the future

btw, one more thing, at the moment roslyn scripting doesn't support core clr, in fact it doesn't even work on mono, so scriptcs uses roslyn binaries on windows and mono.csharp on mono

but going back to the original topic, I think orphan `*.cs` files could be treated similarly to `*.csx` files and just added to workspace in a loose way; that would like you say require determining the runtime, plus some organizational decisions regarding how to code it in - would that be a new specialized project system that runs at the end or would that be part of msbuildprojectsystem for example, to basically gather all the loose files it didn't need as part of msbuild projects

jrieken:
[...]

PR #1252, opened 2018-07-30
Add support for Miscellaneous Files

akshita31:
Fixes: #207


filipw:
very promising 😃

src/OmniSharp.Abstractions/Mef/ExportIProjectSystemAttribute.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+using System.Composition;
+using Microsoft.CodeAnalysis;
+using OmniSharp.Services;
+
+namespace OmniSharp.Mef
+{
+    [MetadataAttribute]
+    [AttributeUsage(AttributeTargets.Class, AllowMultiple = false)]
+    public class ExportIProjectSystemAttribute: ExportAttribute
```
DustinCampbell: Naming: remove the "I" and go with "ExportProjectSystemAttribute".

src/OmniSharp.MiscellanousFiles/MiscellanousFiles.cs
```diff
@@ -0,0 +1,115 @@
...
+                   language: Language);
+
+                _workspace.AddProject(projectInfo);
+                _workspace.AddMiscellanousFileDocument(projectInfo.Id, filePath);
+                _projects[filePath] = projectInfo;
+                _logger.LogInformation($"Successfully added file '{filePath}' to workspace");
+            }
+        }
+
+        private void OnMiscellanousFileChanged(string filePath, FileChangeType changeType)
```
DustinCampbell: How does this interact with the BufferManager logic where new files get placed in an existing project?

src/OmniSharp.MiscellanousFiles/MiscellanousFiles.cs
```diff
@@ -0,0 +1,115 @@
...
+                   id: ProjectId.CreateNewId(),
+                   version: VersionStamp.Create(),
+                   name: Path.GetFileName(filePath),
+                   metadataReferences: new MetadataReference[] { MetadataReference.CreateFromFile((typeof(object).Assembly).Location) },
+                   assemblyName: assemblyName,
+                   language: Language);
+
+                _workspace.AddProject(projectInfo);
+                _workspace.AddMiscellanousFileDocument(projectInfo.Id, filePath);
+                _projects[filePath] = projectInfo;
```
DustinCampbell: It strikes me as overkill to add a project for every single miscellaneous file. In Visual Studio, I believe there's a single miscellaneous project that all misc files are added to. @jasonmalinowski would know best.

src/OmniSharp.Abstractions/OmniSharp.Abstractions.csproj
```diff
@@ -7,6 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.CodeAnalysis" Version="2.8.2" />
```
DustinCampbell: I think this breaks layering. IIRC, `OmniSharp.Abstractions` was not supposed to have a dependency on Roslyn. @david-driscoll, is that right?

src/OmniSharp.MiscellanousFiles/OmniSharp.MiscellanousFiles.csproj
```diff
@@ -0,0 +1,13 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <TargetFramework>net461</TargetFramework>
+        <PlatformTarget>AnyCPU</PlatformTarget>
+    </PropertyGroup>
+
+    <ItemGroup>
+      <ProjectReference Include="..\OmniSharp.Abstractions\OmniSharp.Abstractions.csproj" />
+      <ProjectReference Include="..\OmniSharp.MSBuild\OmniSharp.MSBuild.csproj" />
```
DustinCampbell: This shouldn't need to have a dependency on OmniSharp.MSBuild. Instead, I'd put a list of well-known project system names at a lower layer rather than referencing OmniSharp.MSBuild directly just to get the name of the project system.

src/OmniSharp.Abstractions/OmniSharp.Abstractions.csproj
filipw: that was definitely the idea, that's also why `OmniSharp.Roslyn` is there, as an extra layer over `OmniSharp.Roslyn.CSharp`

src/OmniSharp.MiscellanousFiles/MiscellanousFiles.cs
jasonmalinowski: In Visual Studio we do create a separate project per individual file. This was largely to replicate the previous C# language service that did effectively the same in my opinion. I've long thought about moving them into one project though, so opportunistically things like go to def will move between files if you happen to have multiple files from the same project open.

src/OmniSharp.MiscellanousFiles/OmniSharp.MiscellanousFiles.csproj
akshita31: We need the project system type here, because we need to know when the project system finished initialization, before we can say that a particular document is a miscellanous document. So I have added this method - https://github.com/OmniSharp/omnisharp-roslyn/pull/1252/files#diff-63fe9836b6472be5b5ab510254cf5bb5R214 to the MSBuild project system and hence we need to know the `ProjectSystem` type for which we need the MSBuild reference.

akshita31: There are two possible approaches that we can take here:

1. Add `HasCompletedUpdateRequest` to the interface `IProjectSystem` interface
2. Make the GetWorkspaceModelAsync method public so that we can call it from the MiscellanousFiles system.

@DustinCampbell What do you think would be the best approach?

DustinCampbell: I wouldn't go with option 2. Instead, I think we should work to break the coupling of the project systems, so the first option is the better approach.

src/OmniSharp.MSBuild/ProjectSystem.cs
```diff
@@ -209,5 +210,10 @@ async Task<object> IProjectSystem.GetProjectModelAsync(string filePath)
 
             return new MSBuildProjectInfo(projectFileInfo);
         }
+
+        public async Task HasCompletedUpdateRequest()
```
DustinCampbell: I think this needs a better name. "HasCompletedUpdateRequest" implies that this would return a boolean and doesn't really indicate that it's asynchronous.

How about `WaitForInitializationToCompletionAsync`?

src/OmniSharp.MiscellanousFiles/MiscellanousFiles.cs
```diff
@@ -0,0 +1,119 @@
...
+            foreach (var filePath in allFiles)
+                AddIfMiscellanousFile(filePath);
+
+            _fileSystemWatcher.Watch(miscFileExtension, OnMiscellanousFileChanged);
+        }
+
+        private async void AddIfMiscellanousFile(string filePath)
+        {
+            //wait for the project system to get initialised
+            await _projectSystem.HasCompletedUpdateRequest();
```
DustinCampbell: Shouldn't this be in the `Initialize` method rather than every miscellaneous file that gets added?

src/OmniSharp.MiscellanousFiles/OmniSharp.MiscellanousFiles.csproj
DustinCampbell: Another option might be to introduce another interface that can be used to check a project system for pending updates, implement it on the MSBuild project system, and then cast to the new interface in the miscellaneous files project system.

src/OmniSharp.MSBuild/ProjectSystem.cs
DustinCampbell: Or, if you decide to introduce a different interface for this behavior, maybe `HasPendingUpdatesAsync` that returns a `Task<bool>`?

src/OmniSharp.MiscellanousFiles/MiscellanousFiles.cs
```diff
@@ -0,0 +1,119 @@
...
+using OmniSharp.FileSystem;
+using OmniSharp.FileWatching;
+using OmniSharp.Mef;
+using OmniSharp.Models.WorkspaceInformation;
+using OmniSharp.MSBuild;
+using OmniSharp.Services;
+
+namespace OmniSharp.MiscellanousFiles
+{
+    [ExtensionOrder(After = ProjectSystemNames.MSBuildProjectSystem)]
```
DustinCampbell: How does the interact with the legacy `DotNetProjectSystem` which handles old project.json projects?

src/OmniSharp.MiscellanousFiles/OmniSharp.MiscellanousFiles.csproj
akshita31: @DustinCampbell Done. Added an IUpdates interface. Though I think we can have a better name for that. Any suggestion ?

src/OmniSharp.MiscellanousFiles/MiscellanousFiles.cs
akshita31: After any file is added we need to wait for the MSBuild project system to signal that it has completed the update for that particular file. So, I think we need it here.

src/OmniSharp.MiscellanousFiles/MiscellanousFiles.cs
akshita31: The MSBuildProjectSystem interacts with the BufferManager and if a file is added to a project that it is already watching, then the BufferManager is called to add the new document. Only once MSBuild has completed this update process, does the Miscellanous FIle system add a new file, because of this await here: https://github.com/OmniSharp/omnisharp-roslyn/pull/1252/files/dd73d39631d118010ea073878b54755555e5f4ea#diff-b0a00a01200d50b36022179ac672964dR71. Thoughts ?

src/OmniSharp.MiscellanousFiles/MiscellanousFiles.cs
akshita31: As of now we only await the MSBuild project system has completed the update, no interaction with the Dotnet Project System.

akshita31: Should we add the similar `IUpdates` interface and await the update for the DotnetProjectSystem ?

DustinCampbell:
This will also address https://github.com/OmniSharp/omnisharp-vscode/issues/47

akshita31:
@DustinCampbell We've been observing [test failures](https://travis-ci.org/OmniSharp/omnisharp-roslyn/jobs/411453548) on FilesChangedFacts after adding the Misc File support, because those tests were invoking a files changed request for a random file, with a relative path: https://github.com/OmniSharp/omnisharp-roslyn/blob/master/tests/OmniSharp.Roslyn.CSharp.Tests/FilesChangedFacts.cs#L52. Initailly these tests would pass as this file didnot trigger any project system to add the file to the workspace. However now with the Miscellanous File system, we care about this file and try to add it to the workspace , hence getting an error from the [OmnisharpTextLoader](https://github.com/OmniSharp/omnisharp-roslyn/blob/e18913e887144119c41d60f1842e49f8e9bfcf72/src/OmniSharp.Roslyn/OmniSharpTextLoader.cs#L23).

To workaround this we have added the [code](https://github.com/OmniSharp/omnisharp-roslyn/pull/1252/commits/c243202b31f2f39fe2d4e061b483234d6acbe5a7) in miscellanous file system that coverts the path to absolute and also checks if the file exists on disk before adding the document to the workspace. Does that look like a reasonable approach ?

SirIntruder:
This looks very nice 😄 

I have a questions. If a file which was initialy an oprhan becomes "adopted" later on by some project (say .csproj files get modified by the user to include that file), will OmnisharpWorkspace.RemoveMiscellanousFileDocument be invoked? Without calling this method, looks like ex-orphan will be stuck with an orphan status, and "IsCapableOfSemanticDiagnostics" will return false on that file, even though it should be treated as a normal part of its project now. I see deleting a misc file will invoke RemoveMiscellanousFileDocument, but I see no code that would handle the "adoption" scenario.

akshita31:
@SirIntruder Thank you so much for pointing that out. That scenario didnt cross my mind at all. You are right. In the adoption scenario, we will still classify the document as miscellaneous and not show semantic diagnostics for it. But, I think I would like to take that up as a separate work item once this change is merged. Do you think that is a very common use case scenario?

src/OmniSharp.Abstractions/Services/IUpdates.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System.Threading.Tasks;
+
+namespace OmniSharp.Services
+{
+    public interface IUpdates
```
rchande: Why not derive an IWaitableProjectSystem from IProjectSystem and have MSBuild and Dotnet implement that?

src/OmniSharp.MiscellanousFiles/MiscellanousFiles.cs
```diff
@@ -0,0 +1,132 @@
...
+    [ExtensionOrder(After = ProjectSystemNames.DotNetProjectSystem)]
+    [ExportProjectSystem(ProjectSystemNames.MiscellanousFilesProjectSystem), Shared]
+    public class MiscellanousFilesProjectSystem : IProjectSystem
+    {
+        private readonly string miscFileExtension = ".cs";
```
rchande: const

src/OmniSharp.MiscellanousFiles/MiscellanousFiles.cs
```diff
@@ -0,0 +1,132 @@
...
+        public string Key { get; } = "MiscellanousFiles";
```
rchande: You can an expression body property for these if you want

src/OmniSharp.MiscellanousFiles/MiscellanousFiles.cs
```diff
@@ -0,0 +1,132 @@
...
+        {
+            _workspace = workspace;
+            _fileSystemWatcher = fileSystemWatcher ?? throw new ArgumentNullException(nameof(fileSystemWatcher));
+            _fileSystemHelper = fileSystemHelper;
+            _logger = loggerFactory.CreateLogger<MiscellanousFilesProjectSystem>();
+            _projectSystems = projectSystems.
+                Where
+                (ps => ps.Metadata.Name == ProjectSystemNames.MSBuildProjectSystem || 
+                ps.Metadata.Name == ProjectSystemNames.DotNetProjectSystem).
+                Select(ps => ps.Value).
```
rchande: . on next line

src/OmniSharp.MiscellanousFiles/MiscellanousFiles.cs
```diff
@@ -0,0 +1,132 @@
...
+                Select(ps => ps as IUpdates)
```
rchande: There's actually a .Cast method in linq

src/OmniSharp.MiscellanousFiles/MiscellanousFiles.cs
```diff
@@ -0,0 +1,132 @@
...
+        public MiscellanousFilesProjectSystem(OmniSharpWorkspace workspace, IFileSystemWatcher fileSystemWatcher, FileSystemHelper fileSystemHelper,
+            ILoggerFactory loggerFactory, [ImportMany] IEnumerable<Lazy<IProjectSystem, ProjectSystemMetadata>> projectSystems)
+        {
+            _workspace = workspace;
+            _fileSystemWatcher = fileSystemWatcher ?? throw new ArgumentNullException(nameof(fileSystemWatcher));
+            _fileSystemHelper = fileSystemHelper;
+            _logger = loggerFactory.CreateLogger<MiscellanousFilesProjectSystem>();
+            _projectSystems = projectSystems.
+                Where
+                (ps => ps.Metadata.Name == ProjectSystemNames.MSBuildProjectSystem ||
```
rchande: Wrapping is weird

src/OmniSharp.MiscellanousFiles/MiscellanousFiles.cs
```diff
@@ -0,0 +1,132 @@
...
+            foreach (var filePath in allFiles)
+                AddIfMiscellanousFile(filePath);
+
+            _fileSystemWatcher.Watch(miscFileExtension, OnMiscellanousFileChanged);
+        }
+
+        private async void AddIfMiscellanousFile(string filePath)
+        {
+            //wait for the project systems to finish processing the updates
+            foreach(var projectSystem in _projectSystems)
```
rchande: Do we need to wait for the other project systems each time we process a file, or can we do it once outside this method?

src/OmniSharp.MiscellanousFiles/MiscellanousFiles.cs
```diff
@@ -0,0 +1,132 @@
...
+        void IProjectSystem.Initalize(IConfiguration configuration)
+        {
+            var allFiles = _fileSystemHelper.GetFiles("**/*.cs");
+            foreach (var filePath in allFiles)
+                AddIfMiscellanousFile(filePath);
+
+            _fileSystemWatcher.Watch(miscFileExtension, OnMiscellanousFileChanged);
+        }
+
+        private async void AddIfMiscellanousFile(string filePath)
```
rchande: spelling

src/OmniSharp.MiscellanousFiles/MiscellanousFiles.cs
```diff
@@ -0,0 +1,132 @@
...
+        {
+            //wait for the project systems to finish processing the updates
+            foreach(var projectSystem in _projectSystems)
+            {
+                await projectSystem.WaitForUpdatesAsync();
+            }
+
+            var absoluteFilePath = new FileInfo(filePath).FullName;
+            if (!File.Exists(absoluteFilePath))
+                return;
```
rchande: newline

src/OmniSharp.MiscellanousFiles/MiscellanousFiles.cs
```diff
@@ -0,0 +1,132 @@
...
+                    var projectInfo = ProjectInfo.Create(
+                   id: ProjectId.CreateNewId(),
+                   version: VersionStamp.Create(),
+                   name: "MiscellanousFiles",
+                   metadataReferences: new MetadataReference[] { MetadataReference.CreateFromFile((typeof(object).Assembly).Location) },
+                   assemblyName: assemblyName,
+                   language: Language);
+
+                    _workspace.AddProject(projectInfo);
+                    _projectId = projectInfo.Id;
```
rchande: Extract all of this project creation stuff into another method

src/OmniSharp.Roslyn/OmniSharpWorkspace.cs
```diff
@@ -83,6 +86,19 @@ public void AddDocument(DocumentInfo documentInfo)
             OnDocumentAdded(documentInfo);
         }
 
+        public DocumentId AddMiscellanousFileDocument(ProjectId projectId, string filePath)
```
rchande: Given that this workspace already understands how to handle misc files, is there a reason clients need to be able to control what project they wind up in?

tests/OmniSharp.MiscellanousFiles.Tests/AssemblyInfo.cs
```diff
@@ -0,0 +1 @@
+﻿[assembly: Xunit.CollectionBehavior(DisableTestParallelization = true)]
```
rchande: Why?

rchande:
@SirIntruder Can you remind me how I can make omnisharp respond to a csproj change without restarting it? It's hard to remember exactly what operations are supported.

src/OmniSharp.MiscellanousFiles/MiscellanousFiles.cs
akshita31: If we add a new file, that the project systems care about, then we will have to wait for them to get updated right, before we treat the file as a miscellaneous file

src/OmniSharp.Abstractions/Services/IUpdates.cs
akshita31: Done

src/OmniSharp.Roslyn/OmniSharpWorkspace.cs
akshita31: In that case, we will have to add a "MiscFileProject" to the workspace and the first time we see a misc file (or even in the constructor of the MiscFileSystem) we will have to pass it to the workspace. From then on, every misc file will be added to the same project. Does that seem a better approach ?

SirIntruder:
@rchande o# responds fine as long as there is one csproj, but with two or more csprojs, BufferManager screws things up by making one orphan copy for each project, and does not let go of them until o# is restarted.

In theory, BufferManager should recover - it listens to "DocumentAdded" event and removes orphans when they get adopted. In practice, however, when csproj gets modified, "DocumentAdded" event never gets raised, because other projects regard orphans as normal parts of themselves. The new system would not have this problem, if it decides on similar recovery mechanism.

rchande:
@SirIntruder Cool, makes sense. @akshita31, basically you need a way to have your file watching code determine if the file got added to a different project, and if it did, not add it to misc files.

tests/OmniSharp.MiscellanousFiles.Tests/AssemblyInfo.cs
DustinCampbell: It's in other tests projects as well. The problem is that we share the same MEF composition and OmniSharpWorkspace across a lot of tests. If tests run in parallel, unexpected things can happen.

src/OmniSharp.MiscellaneousFile/MiscellaneousFileProjectSystem.cs
```diff
@@ -0,0 +1,115 @@
...
+
+namespace OmniSharp.MiscellaneousFile
+{
+    [ExtensionOrder(After = ProjectSystemNames.MSBuildProjectSystem)]
+    [ExtensionOrder(After = ProjectSystemNames.DotNetProjectSystem)]
+    [ExportProjectSystem(ProjectSystemNames.MiscellaneousFilesProjectSystem), Shared]
+    public class MiscellaneousFilesProjectSystem : IProjectSystem
+    {
+        private const string miscFileExtension = ".cs";
+        public string Key => "MiscellaneousFiles";
```
rchande: ProjectSystemNames.MiscFiles?

src/OmniSharp.MiscellaneousFile/MiscellaneousFileProjectSystem.cs
```diff
@@ -0,0 +1,115 @@
...
+        private readonly FileSystemHelper _fileSystemHelper;
+        private readonly List<IWaitableProjectSystem> _projectSystems;
+        private readonly ILogger _logger;
+
+        [ImportingConstructor]
+        public MiscellaneousFilesProjectSystem(OmniSharpWorkspace workspace, IFileSystemWatcher fileSystemWatcher, FileSystemHelper fileSystemHelper,
+            ILoggerFactory loggerFactory, [ImportMany] IEnumerable<Lazy<IProjectSystem, ProjectSystemMetadata>> projectSystems)
+        {
+            _workspace = workspace;
+            _fileSystemWatcher = fileSystemWatcher ?? throw new ArgumentNullException(nameof(fileSystemWatcher));
```
rchande: Why do we only null check this one?

src/OmniSharp.MiscellaneousFile/MiscellaneousFileProjectSystem.cs
```diff
@@ -0,0 +1,115 @@
...
+            _fileSystemHelper = fileSystemHelper;
+            _logger = loggerFactory.CreateLogger<MiscellaneousFilesProjectSystem>();
+            _projectSystems = projectSystems
+                .Where(ps => ps.Metadata.Name == ProjectSystemNames.MSBuildProjectSystem ||
+                ps.Metadata.Name == ProjectSystemNames.DotNetProjectSystem)
```
rchande: Indent

src/OmniSharp.MiscellaneousFile/MiscellaneousFileProjectSystem.cs
```diff
@@ -0,0 +1,115 @@
...
+            foreach (var projectSystem in _projectSystems)
+            {
+                await projectSystem.WaitForUpdatesAsync();
+            }
+
+            var absoluteFilePath = new FileInfo(filePath).FullName;
+            if (!File.Exists(absoluteFilePath))
+                return;
+
+            if (_workspace.GetDocument(filePath) == null)
```
rchande: This check is no longer needed, right? Workspace will do the right thing for us

src/OmniSharp.Roslyn/OmniSharpWorkspace.cs
```diff
@@ -80,9 +85,51 @@ public void RemoveMetadataReference(ProjectId projectId, MetadataReference metad
 
         public void AddDocument(DocumentInfo documentInfo)
         {
+            var documentId = GetDocumentId(documentInfo.FilePath);
+            if (documentId != null && _miscellanousFiles.Contains(documentId))
+            {
+                //if the file has already been added as a misc file,
+                //because of a possible race condition between the updates of the project systems,
+                //remove the misc file and add it the document as required
```
rchande: space after //
delete `it`

src/OmniSharp.Roslyn/OmniSharpWorkspace.cs
```diff
@@ -21,12 +21,17 @@ public class OmniSharpWorkspace : Workspace
...
+
+        private ProjectInfo MiscFilesProjectInfo;
+
         [ImportingConstructor]
         public OmniSharpWorkspace(HostServicesAggregator aggregator, ILoggerFactory loggerFactory)
             : base(aggregator.CreateHostServices(), "Custom")
         {
             BufferManager = new BufferManager(this);
             _logger = loggerFactory.CreateLogger<OmniSharpWorkspace>();
+            _miscellanousFiles = new HashSet<DocumentId>();
```
rchande: Spelling

src/OmniSharp.Roslyn/OmniSharpWorkspace.cs
```diff
@@ -21,12 +21,17 @@ public class OmniSharpWorkspace : Workspace
 
         private readonly ILogger<OmniSharpWorkspace> _logger;
 
+        private HashSet<DocumentId> _miscellanousFiles;
+
+        private ProjectInfo MiscFilesProjectInfo;
```
rchande: lowercase m.

src/OmniSharp.Roslyn/OmniSharpWorkspace.cs
```diff
@@ -224,5 +271,11 @@ private void SaveDocumentText(DocumentId id, string fullPath, SourceText newText
                 _logger.LogError(e, $"Error saving document {fullPath}");
             }
         }
+
+        public bool IsCapableOfSemanticDiagnostics(Document document)
+        {
+            var documentId = GetDocumentId(document.FilePath);
+            return !_miscellanousFiles.Contains(documentId);
```
rchande: Could we get rid of the _miscellaneousFiles hashset but checking to see if the projectId is the one we keep a reference to?

tests/OmniSharp.MiscellaneousFile.Tests/EndpointFacts.cs
```diff
@@ -0,0 +1,309 @@
...
+            : base(output)
+        {
+        }
+
+        [Fact]
+        public async Task Returns_only_syntactic_diagnotics()
+        {
+            using (var testProject = await TestAssets.Instance.GetTestProjectAsync("EmptyProject"))
+            {
+
```
rchande: Extra line

src/OmniSharp.Roslyn/OmniSharpWorkspace.cs
```diff
@@ -80,9 +85,51 @@ public void RemoveMetadataReference(ProjectId projectId, MetadataReference metad
...
+                _miscellanousFiles.Remove(documentId);
+                RemoveDocument(documentId);
+            }
+           
             OnDocumentAdded(documentInfo);
         }
 
+        public DocumentId AddMiscellaneousFileDocument(string filePath, string language)
+        {
+            if (MiscFilesProjectInfo == null)
```
rchande: Yeah, move the check if the doc is already a regular document into this method :)

src/OmniSharp.MiscellaneousFile/MiscellaneousFileProjectSystem.cs
```diff
@@ -0,0 +1,116 @@
...
+            }
+
+            var absoluteFilePath = new FileInfo(filePath).FullName;
+            if (!File.Exists(absoluteFilePath))
+                return;
+
+            var documentId = _workspace.TryAddMiscellaneousFileDocument(absoluteFilePath, Language);
+            if (documentId!= null)
+            {
+                _documents[absoluteFilePath] = documentId;
```
rchande: Do we still need to track these documents in the project system? Seems like the workspace now knows how to do the right thing?

src/OmniSharp.Roslyn/OmniSharpWorkspace.cs
```diff
@@ -80,9 +85,54 @@ public void RemoveMetadataReference(ProjectId projectId, MetadataReference metad
...
+            _miscellaneousFiles.Add(documentId);
+            return documentId;
+        }
+
+        private ProjectInfo CreateMiscFilesProject(string language)
+        {
+            string assemblyName = Guid.NewGuid().ToString("N");
+            //If not project exists for the Misc files, create one
+            return ProjectInfo.Create(
+           id: ProjectId.CreateNewId(),
```
rchande: Indenting

src/OmniSharp.Roslyn/OmniSharpWorkspace.cs
```diff
@@ -80,9 +85,54 @@ public void RemoveMetadataReference(ProjectId projectId, MetadataReference metad
...
+
+            var documentId = AddDocument(miscFilesProjectInfo.Id, filePath);
+            _miscellaneousFiles.Add(documentId);
+            return documentId;
+        }
+
+        private ProjectInfo CreateMiscFilesProject(string language)
+        {
+            string assemblyName = Guid.NewGuid().ToString("N");
+            //If not project exists for the Misc files, create one
```
rchande: Slightly strange place for this comment

src/OmniSharp.Roslyn/OmniSharpWorkspace.cs
```diff
@@ -224,5 +274,11 @@ private void SaveDocumentText(DocumentId id, string fullPath, SourceText newText
                 _logger.LogError(e, $"Error saving document {fullPath}");
             }
         }
+
+        public bool IsCapableOfSemanticDiagnostics(Document document)
+        {
+            var documentId = GetDocumentId(document.FilePath);
+            return !_miscellaneousFiles.Contains(documentId);
```
rchande: documentId.ProjectId == miscFilesProjectInfo.Id?

src/OmniSharp.Roslyn/OmniSharpWorkspace.cs
```diff
@@ -80,9 +83,53 @@ public void RemoveMetadataReference(ProjectId projectId, MetadataReference metad
 
         public void AddDocument(DocumentInfo documentInfo)
         {
+            var documentId = GetDocumentId(documentInfo.FilePath);
```
SirIntruder: Can 86-93 be a call to TryRemoveMiscellaneousDocument(documentInfo.filePath)?  Looks same functionality wise

akshita31: @SirIntruder Done. Thanks!

src/OmniSharp.Roslyn/Utilities/ExtensionOrderer.Graph.cs
```diff
@@ -0,0 +1,88 @@
+﻿// Adapted from ExtensionOrderer in Roslyn
```
filipw: @DustinCampbell https://github.com/OmniSharp/omnisharp-roslyn/issues/758#issuecomment-285273319 😃

OmniSharp.sln
```diff
@@ -68,6 +69,10 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "OmniSharp.Stdio.Driver", "s
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "OmniSharp.Script.Tests", "tests\OmniSharp.Script.Tests\OmniSharp.Script.Tests.csproj", "{9E4BA68C-7F4B-429A-A0C7-8CE7D41D610F}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "OmniSharp.MiscellaneousFile", "src\OmniSharp.MiscellaneousFile\OmniSharp.MiscellaneousFile.csproj", "{23A91D15-B178-45A9-8F7A-B0CE1F7A8DAA}"
```
filipw: why is the project + folder called `OmniSharp.MiscellaneousFile` (singular) but everywhere we refer to  `MiscellaneousFiles`? I believe for consistency, the project and folder should be renamed to `OmniSharp.MiscellaneousFiles` too

src/OmniSharp.Host/WorkspaceInitializer.cs
```diff
@@ -24,9 +27,11 @@ public class WorkspaceInitializer
 
             var projectEventForwarder = compositionHost.GetExport<ProjectEventForwarder>();
             projectEventForwarder.Initialize();
+            var projectSystems = compositionHost.GetExports<Lazy<IProjectSystem, ProjectSystemMetadata>>();
```
filipw: maybe we should hide that logic in some generic extension method? this way it could be reused in the future against other potentially ordered look ups like i.e. https://github.com/OmniSharp/omnisharp-roslyn/blob/5129ba5a7b5b401935cf234b32a45611bb1e3b13/src/OmniSharp.Host/WorkspaceInitializer.cs#L74

src/OmniSharp.Roslyn/OmniSharpWorkspace.cs
```diff
@@ -80,9 +83,49 @@ public void RemoveMetadataReference(ProjectId projectId, MetadataReference metad
...
+        }
+
+        private ProjectInfo CreateMiscFilesProject(string language)
+        {
+            string assemblyName = Guid.NewGuid().ToString("N");
+            var projectInfo = ProjectInfo.Create(
+                   id: ProjectId.CreateNewId(),
+                   version: VersionStamp.Create(),
+                   name: "MiscellaneousFiles",
+                   metadataReferences: new MetadataReference[] { MetadataReference.CreateFromFile((typeof(object).Assembly).Location) },
```
filipw: scripting project system does something similar, where it adds some default CLR references. This is the code https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp.Script/ScriptContextProvider.cs#L120-L128 that ensures the mandatory "common" types are loaded and also work on Mono (where you have System.Runtime etc). Perhaps you want to use that here?

filipw:
looks good! I just left some small comments 😀

src/OmniSharp.Roslyn/Utilities/DefaultMetadataReferencesHelper.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace OmniSharp.Roslyn.Utilities
+{
+    public static class DefaultMetadataReferencesHelper
```
rchande: `DefaultMetadataReferenceHelper`

src/OmniSharp.Roslyn/Utilities/DefaultMetadataReferencesHelper.cs
```diff
@@ -0,0 +1,42 @@
...
+    {
+        public static IEnumerable<string> GetDefaultMetadataReferencesLocations()
```
rchande: `GetDefaultMetadataReferenceLocations`

SirIntruder:
I built this, and omnisharp.exe now takes 5.5GB of ram to process my work project, (~200 files marked as misc) 😲
after
before it looked like this:
before

SirIntruder:
I did a quick prototype to replace ConcurrentDictionary<string, ProjectInfo> with one ProjectInfo and a lock, and the problem is gone. Any reason why one-project-per-file is needed? Just having one project file seems both simpler, more performant and a better service to the user (cross references between misc files work)

src/OmniSharp.MiscellanousFiles/MiscellanousFiles.cs
DustinCampbell: I'm not sure I understood your answer to my question. And your link sends me to a giant diff rather than a specific location. The BufferManager has code which checks to see any files that it is notified about (through a `/changeBuffer` or `/updateBuffer` request) are already included in the OmniSharpWorkspace. If a file is not found, that code tries to find a logical project to include it within. I suspect that there may be cases with the MiscellaneousFileProjectSystem will have already added the file to the the "misc files" project and the BufferManager will never place new files in the correct project. This will be a regression.

DustinCampbell:
@rchande: FYI I would have appreciated another opportunity to take a look at this before it was merged. @akshita31 had asked me, @filipw and @david-driscoll if it was OK last Friday. Then, it got merged after the weekend on Monday?

I'm concerned that this will cause a significant regression: https://github.com/OmniSharp/omnisharp-roslyn/pull/1252/commits/81321cd3e99831cbeaf6db702fbe593bfb0b2ee0#r210019117

rchande:
@DustinCampbell My apologies. Next time, we'll be more careful about waiting for everyone to get a chance to review. @akshita31 and I are happy to address any post-facto feedback you have on this one. It's hard to tell exactly from the link you added but I think the concern is that the MiscFiles project system is eagerly scanning for every .cs file under the cone? I suppose the way to make this lazier would be to allow editors to drive the behavior. I guess this could work at least a couple of ways:
* Requests (copmletion, etc) sent for files not in the Workspace result in O# automatically adding that file as a misc file
* We add a new endpoint "FileOpened" or similar that editors can send requests to. If a file is opened that's not in the workspace already, it gets added as a misc file.

Thoughts?

DustinCampbell:
>  It's hard to tell exactly from the link you added but I think the concern is that the MiscFiles project system is eagerly scanning for every .cs file under the cone? I suppose the way to make this lazier would be to allow editors to drive the behavior.

My concern is that the MiscFilesWorkspace eagerly adds files based on file changed events. Consider the following scenario:

* The user opens an existing C# project in OmniSharp. The C# project has file globs to determine what files are included (for example, an ASP.NET Core project).
* The user creates a new C# file in VS Code in their project folder, expecting it to be included in the project.
* They are surprised to find that their editor experience in that file is terrible because it didn't get included in their project -- it got included in the MiscFilesWorkspace.
* The user has to restart OmniSharp to fix the problem.

To be clear, the MiscFiles workspace should be a fallback. As implemented, I believe it will "steal" files that belong to projects.

Prior to this change, newly-created files wouldn't be included in any project until the user edited them. When OmniSharp's BufferManager got notified of the edit, it would locate a project in which to include the file if it wasn't in a project already. Since the MiscFilesWorkspace will have already included it in a project, I suspect that code won't work.

Does that make sense? I _think_ there's a significant experience regression that'll be caused by this.

akshita31:
@DustinCampbell @rchande 
We tried to address this problem by adding a [wait](https://github.com/OmniSharp/omnisharp-roslyn/blob/426ba087907c16d464c95b3841a11172d3123e7e/src/OmniSharp.MiscellaneousFiles/MiscellaneousFilesProjectSystem.cs#L75) for the update of the project systems in the MiscFilesSystem. 
But I now understand that if the update for the MiscFiles is invoked first, then the system will add the file as a MiscFile and the BufferManager will not be able to do the right thing with it.

An alternative that comes to my mind is that instead of creating a whole project system for the MiscFiles we could add the document as a "misc document" when the buffer manager is unable to add it as a transient document: https://github.com/OmniSharp/omnisharp-roslyn/blob/7e684d4a7c8d0ca29313fb19b267dfc121c3cd73/src/OmniSharp.Roslyn/BufferManager.cs#L131. However if we do it that way, how do we handle the deletion of these files?

To the best of my understanding, currently all the project systems handle the deletions for the files that they "care" about. But if we add the handling of addition of files to the buffer manager, who will handle the deletion? Is it possible for the BufferManager to handle it?

SirIntruder:
Omnisharp now crashes every time you open deleted file in git side bar ._.

> Unhandled Exception: System.NotSupportedException: The given path's format is not supported.
>    at System.Security.Permissions.FileIOPermission.EmulateFileIOPermissionChecks(String fullPath)
>    at System.Security.Permissions.FileIOPermission.QuickDemand(FileIOPermissionAccess access, String fullPath, Boolean checkForDuplicates, Boolean needFullPath)
>    at System.IO.FileInfo.Init(String fileName, Boolean checkHost)
>    at OmniSharp.MiscellaneousFile.MiscellaneousFilesProjectSystem.<TryAddMiscellaneousFile>d__19.MoveNext()
> --- End of stack trace from previous location where exception was thrown ---
>    at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
>    at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
>    at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
>    at System.Threading.QueueUserWorkItemCallback.System.Threading.IThreadPoolWorkItem.ExecuteWorkItem()
>    at System.Threading.ThreadPoolWorkQueue.Dispatch()
>

akshita31:
@SirIntruder The crash was because the extension was sending incorrect file paths for deleted git files. I have created a PR in omnisharp-vscode to address that: https://github.com/OmniSharp/omnisharp-vscode/pull/2465.

akshita31:
@DustinCampbell Can you take a look ?

DustinCampbell:
What you suggested in that comment makes sense to me. I think a project system is probably overkill and introduces some ordering concerns.

gulbanana:
Is there a setting to disable miscellaneous files support? I have some solutions where I want OmniSharp only to load certain projects (because the others are Windows-specific).

akshita31:
@gulbanana Miscellaneous files support doesn't load any projects. It is only meant to provide basic intellisense support for files without a project or files that have not been loaded as part of the currently loaded project in omnisharp. Also, the misc file is added to the workspace only when the user edits the file, hence if there are files in your workspace that you are not editing, misc files support won't do anything about it.

beeuhtricks:
Hi all,

I'm working on a bug in (OmniSharp/omnisharp-emacs#464) that is almost identical to OmniSharp/omnisharp-vscode#785. I've hunted through the elisp files and I think this bug is upstream. I found this PR while looking through diffs in the 1.32.x series because it seemed relevant. Is it possible that new buffers are getting initially parsed as misc files, causing namespace duplication errors when the current solution finishes loading? These bugs do not seem to be a problem with files that are actually outside of a project structure.

rchande:
@beeuhtricks We recently took a fix (https://github.com/OmniSharp/omnisharp-roslyn/pull/1357) for this issue. Can try out one of our latest builds and see if that fix works for you? Thanks!

beeuhtricks:
@rchande I tried it with 1.32.9-beta.20/omnisharp-osx.tar.gz and it seems to have fixed the problem. Thank you so much!

beeuhtricks:
Hi there,
We’re getting reports that this issue persists, albeit occasionally, with v1.32.10. I would be happy to work with you on a solution if necessary.


________________________________

PR #1259, opened 2018-08-02
revert from McMaster.Extensions.CommandLineUtils to Microsoft.Extensions.CommandLineUtils

filipw:
This should fix #1256 (at least on my local box it works on the embedded Mono).
We do not take advantage of any features of `McMaster.Extensions.CommandLineUtils` over the old `Microsoft.Extensions.CommandLineUtils` so it might be a bit excessive to have to update the embedded Mono to include more stuff just for the sake of a library upgrade.

PR #1260, opened 2018-08-06
Add packaging scripts for embedded Mono and MSBuild bits

DustinCampbell:
This change adds two scripts, `copy-mono.sh` and `copy-msbuild.sh`, that can be used to create embedded Mono and MSBuild packages for the operating system that they're run on (macOS, linux-x64 and linux-x86).

Note that I've gone ahead and included the binary that was causing trouble with the `McMaster.Extensions.CommandLineUtils` package.


DustinCampbell:
Thanks all! This *should* fix the problem that https://github.com/OmniSharp/omnisharp-roslyn/pull/1259 was addressing.

PR #1267, opened 2018-08-14
Fix MiscFile system creating new project for each file

SirIntruder:
Looks like https://github.com/OmniSharp/omnisharp-roslyn/pull/1252 had an intent to create one project for misc files per language, but bug caused it to create new project for each file, causing a skyrocket of memory consumption in larger projects. This should fix the issue


rchande:
Good catch! Thanks @SirIntruder!

PR #1268, opened 2018-08-14
Add environment variable to try and stabilize OSX builds

DustinCampbell:
Apparently, there is a bug in .NET Core SDK 2.1.301 that can cause HTTP flakiness on OSX. The .NET team suggested that we try setting `DOTNET_SYSTEM_NET_HTTP_USESOCKETSHTTPHANDLER=0` and see if that addresses the problem.


rchande:
Thanks!

PR #1272, opened 2018-08-17
 construct scripting workspace information directly from the ProjectInfo

filipw:
This is a follow up to #1112 which we shipped but it didn't have a real test 😃

In this PR:
 - I fixed a "bug" where RSP based references/namespaces were not respected by the Workspace information endpoint in scripting (construct that info from `ProjectInfo` directly, which is most reliable way)
 - better late than never - added an integration test for the RSP based configuration of scripting project system
 - fixed some suspicious test code

### DIFF
--- a/src/OmniSharp.Script/ScriptContextModel.cs
+++ b/src/OmniSharp.Script/ScriptContextModel.cs
@@ -1,22 +1,32 @@
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
 
 namespace OmniSharp.Script
 {
     public class ScriptContextModel
     {
-        public ScriptContextModel(string csxPath, ProjectInfo project, HashSet<string> implicitAssemblyReferences)
+        public ScriptContextModel(string csxPath, ProjectInfo project)
         {
             Path = csxPath;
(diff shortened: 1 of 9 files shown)


src/OmniSharp.Script/ScriptContextModel.cs
```diff
@@ -1,22 +1,32 @@
...
+        public ScriptContextModel(string csxPath, ProjectInfo project)
         {
             Path = csxPath;
-            ImplicitAssemblyReferences = implicitAssemblyReferences;
-            CommonUsings = ScriptProjectProvider.DefaultNamespaces;
+            AssemblyReferences = project.MetadataReferences.Select(a =>
+            {
+                if (a is PortableExecutableReference portableExecutableReference)
+                {
+                    return portableExecutableReference.FilePath ?? portableExecutableReference.Display;
```
david-driscoll: Is this ever null?

david-driscoll:
Updated to ensure that everything is passing with the latest stuff

src/OmniSharp.Script/ScriptContextModel.cs
filipw: not at the moment, but once we support plugins to meddle with this project system it might 😃

filipw:
I added a changelog update

Issue #1281, opened 2018-08-31
Duplicate "find symbols" results


rchande:
This looks like a regression in the 1.32 release...

![image](https://user-images.githubusercontent.com/3751401/44924633-c74b3480-ad00-11e8-9dd0-4f1e86516123.png)

mholo65:
Same problem we had with `/codecheck` a while ago? What happened to `/findsymbols`? Does it check whole solution instead of file nowadays?

rchande:
@mholo65 I don't believe there was a behavior change to findSymbols. I'm guessing this got regressed when we added the cake handler.

rchande:
@mholo65 I would guess so--I could be wrong about it being new in 1.32. The find symbols implementation hasn't changed since April 2017.

PR #1282, opened 2018-08-31
Don't duplicate find symbols results from C# files

rchande:
Fixes https://github.com/OmniSharp/omnisharp-roslyn/issues/1281

### DIFF
--- a/src/OmniSharp.Cake/Services/RequestHandlers/Navigation/FindSymbolsHandler.cs
+++ b/src/OmniSharp.Cake/Services/RequestHandlers/Navigation/FindSymbolsHandler.cs
@@ -26,8 +26,7 @@ namespace OmniSharp.Cake.Services.RequestHandlers.Navigation
                 ? candidate.IsValidCompletionFor(request.Filter)
                 : true;
 
-            return  Workspace.CurrentSolution.FindSymbols(isMatch,
-                p => p.Name.EndsWith(LanguageNames.Cake, StringComparison.OrdinalIgnoreCase));
+            return Workspace.CurrentSolution.FindSymbols(isMatch, LanguageNames.Cake);
         }
 
         protected override Task<QuickFixResponse> TranslateResponse(QuickFixResponse response, FindSymbolsRequest request)
--- a/src/OmniSharp.Cake/Services/RequestHandlers/Navigation/FindSymbolsHandler.cs
+++ b/src/OmniSharp.Cake/Services/RequestHandlers/Navigation/FindSymbolsHandler.cs
@@ -6,6 +6,7 @@ using OmniSharp.Extensions;
 using OmniSharp.Mef;
 using OmniSharp.Models;
 using OmniSharp.Models.FindSymbols;
+using static OmniSharp.Cake.Constants;
 
 namespace OmniSharp.Cake.Services.RequestHandlers.Navigation
 {
--- a/src/OmniSharp.Cake/Services/RequestHandlers/Navigation/FindSymbolsHandler.cs
+++ b/src/OmniSharp.Cake/Services/RequestHandlers/Navigation/FindSymbolsHandler.cs
@@ -26,7 +26,7 @@ namespace OmniSharp.Cake.Services.RequestHandlers.Navigation
                 : true;
 
             return  Workspace.CurrentSolution.FindSymbols(isMatch,
-                p => p.Name.EndsWith(".cake", StringComparison.OrdinalIgnoreCase));
+                p => p.Name.EndsWith(LanguageNames.Cake, StringComparison.OrdinalIgnoreCase));
         }
 
         protected override Task<QuickFixResponse> TranslateResponse(QuickFixResponse response, FindSymbolsRequest request)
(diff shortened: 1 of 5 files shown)


src/OmniSharp.Cake/Services/RequestHandlers/Navigation/FindSymbolsHandler.cs
```diff
@@ -16,6 +18,17 @@ public FindSymbolsHandler(OmniSharpWorkspace workspace)
...
 
+        public override Task<QuickFixResponse> Handle(FindSymbolsRequest request)
+        {
+            Func<string, bool> isMatch =
+                candidate => request != null
+                ? candidate.IsValidCompletionFor(request.Filter)
+                : true;
+
+            return  Workspace.CurrentSolution.FindSymbols(isMatch,
+                p => p.Name.EndsWith(".cake", StringComparison.OrdinalIgnoreCase));
```
mholo65: Could we reuse the extension as defined in the project system?

src/OmniSharp.Roslyn.CSharp/Services/Navigation/FindSymbolsService.cs
```diff
@@ -30,57 +27,8 @@ public async Task<QuickFixResponse> Handle(FindSymbolsRequest request = null)
                 ? candidate.IsValidCompletionFor(request.Filter)
                 : true;
 
-            return await FindSymbols(isMatch);
+            return await _workspace.CurrentSolution.FindSymbols(isMatch,
+                p => !p.Name.EndsWith(".cake", StringComparison.OrdinalIgnoreCase));
```
mholo65: Could we try avoiding to use `.cake` extension in `OmniSharp.Roslyn.CSharp`, instead allow only `.cs` and `.csx`..? Would be difficult to get from project system, since `OmniSharp.Roslyn.CSharp` doesn't have (and shouldn't) have a dependency on `OmniSharp.Scripting`.

src/OmniSharp.Cake/Services/RequestHandlers/Navigation/FindSymbolsHandler.cs
DustinCampbell: Maybe the project system and this handler should point to the same constant?

src/OmniSharp.Roslyn.CSharp/Services/Navigation/FindSymbolsService.cs
DustinCampbell: Yeah, this feels weird. It would certainly fix the issue, but it feels a bit dirty and easily broken.

rchande:
@mholo65 Done; thanks. It does kind of look like this has been present for as long as cake...

rchande:
@mholo65 @DustinCampbell Another approach could be to simply remove the "cake" flavored handler. Ctrl-t in VS Code doesn't specify a language, so the C# extension doesn't pass one to OmniSharp, thus running both handlers. 

However, the symbol provider (https://code.visualstudio.com/docs/extensionAPI/vscode-api#DocumentSymbolProvider) gets access to the current document, so we could specify a language in the request based on the current document. Of course, that means you could only search for Cake symbols from a Cake file and vice versa. Thoughts?

src/OmniSharp.Cake/Services/RequestHandlers/Navigation/FindSymbolsHandler.cs
```diff
@@ -16,6 +19,17 @@ public FindSymbolsHandler(OmniSharpWorkspace workspace)
...
         }
 
+        public override Task<QuickFixResponse> Handle(FindSymbolsRequest request)
+        {
+            Func<string, bool> isMatch =
+                candidate => request != null
+                ? candidate.IsValidCompletionFor(request.Filter)
+                : true;
+
+            return  Workspace.CurrentSolution.FindSymbols(isMatch,
```
DustinCampbell: Should the `FindSymbols` extension method just take the file extension? It seems error prone to always remember to specify `StringComparison.OrdinalIgnoreCase`.

rchande: Good idea.

src/OmniSharp.Roslyn.CSharp/Services/Navigation/FindSymbolsService.cs
david-driscoll: Yeah I think being explicit about the file types supported makes sense, instead of excluding things.  If for example we ever add vb support, then this will start to leak again.

DustinCampbell:
> @mholo65 @DustinCampbell Another approach could be to simply remove the "cake" flavored handler. Ctrl-t in VS Code doesn't specify a language, so the C# extension doesn't pass one to OmniSharp, thus running both handlers.

> However, the symbol provider (https://code.visualstudio.com/docs/extensionAPI/vscode-api#DocumentSymbolProvider) gets access to the current document, so we could specify a language in the request based on the current document. Of course, that means you could only search for Cake symbols from a Cake file and vice versa. Thoughts?

I think there should be a _single_ FindSymbolsHandler that lives in the `OmniSharp.Roslyn.CSharp` project that returns results from the entire workspace for all project systems. The cake handlers really just translate locations don't they? This could be done by calling into a method exposed by the project system from the findsymbols endpoint.

rchande:
In the interest of being able to release today, I'm going to merge this PR. I filed https://github.com/OmniSharp/omnisharp-roslyn/issues/1287 to track cleaning this up further.

Issue #1243, opened 2018-07-10
"Go to Symbol in Workspace…" is slow with moderately sized solution containing only C# projects


dmgonch:
REPRO: I opened a solution in VS Code containing 68 C# projects (according to the response for "/projects" request in OmniSharp output log). Once all projects were loaded (took about a minute), I pressed Ctrl-T and had to wait for at least 15 seconds for the list to be populated. This happened every time Ctrl-T is pressed - not only the first time. Additionally, typing a single character freezes the list again for few seconds which makes the feature only marginally useful in this case.

EXPECTED: IMO OmniSharp could:
 - Cache results of recently issued "/findsymbols" requests.
 - Apply a strict timeout when processing "/findsymbols" requests and return cached or partial or simply empty result when the timeout is reached.

SirIntruder:
IIRC Problem is in vscode rendering thousands of symbols in the ui after server returns them, not on the server side.

dmgonch:
Interesting. I was only judging by OmniSharp output log because I see a significant delay after "/findsymbols" request printed and before the response appears. In any case I would argue that "/findsymbols" processing should be limited in both time and the amount of data it is allowed to return. A user IMO is unlikely to just scroll through thousands of entries and s/he is more likely to continue typing the filter to narrow the result to only few entries.

dmgonch:
BTW, this issue I opened sounds like a dupe of https://github.com/OmniSharp/omnisharp-vscode/issues/1808

SirIntruder:
Yup that's it. I remember I just modified "workspaceSymbolProvider.js" manually so I never pass on insanely big symbol lists to vscode, and it solved all the problems. Wanted to make a PR but it just flew over my head.

dmgonch:
This is definitely could be a good quick fix, but long term IMO ideally each supported file system (MSBuild, Cake, etc) limits as well how many items it attempts to discover and send back with /findsymbols response.

PR #1284, opened 2018-09-03
Introduce configuration for "Go to Symbol in Workspace"

dmgonch:
Allow configuring number of chars user must to type in for "Go to Symbol in Workspace" command to return any results (default is 0 to preserve existing behavior). The reasoning here is that when lots of projects are loaded by OmniSharp it makes little sense to return all symbols discovered. This change should largely address these issues as well: https://github.com/OmniSharp/omnisharp-roslyn/issues/1243, https://github.com/OmniSharp/omnisharp-vscode/issues/1808.
Additionally allow configuring max number of items returned by "Go to Symbol in Workspace" command. Similarly to the above in a big repo returning pages and pages of items in most cases will be wasteful. But default is 0 which is indicates no limit to preserve existing behavior.


src/OmniSharp.Roslyn.CSharp/Services/Navigation/FindSymbolsService.cs
```diff
@@ -51,6 +59,11 @@ private async Task<QuickFixResponse> FindSymbols(Func<string, bool> predicate)
                 {
                     symbolLocations.Add(ConvertSymbol(symbol, location));
                 }
+
+                if (_options.MaxItemsToReturn > 0 && symbolLocations.Count >= _options.MaxItemsToReturn)
```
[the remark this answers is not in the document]
dmgonch: At some point though the resulting list of items is so long that VSCode doesn't respond. I personally prefer faster response rather then waiting for tons of results to be processed just to be discarded as I typed another character. Your guys call though.

src/OmniSharp.Roslyn.CSharp/Services/Navigation/FindSymbolsService.cs
```diff
@@ -9,22 +9,30 @@
...
 {
     [OmniSharpHandler(OmniSharpEndpoints.FindSymbols, LanguageNames.CSharp)]
     public class FindSymbolsService : IRequestHandler<FindSymbolsRequest, QuickFixResponse>
     {
         private OmniSharpWorkspace _workspace;
+        private FindSymbolsOptions _options;
 
         [ImportingConstructor]
-        public FindSymbolsService(OmniSharpWorkspace workspace)
+        public FindSymbolsService(OmniSharpWorkspace workspace, OmniSharpOptions omniSharpOptions)
```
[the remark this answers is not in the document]
dmgonch: Sounds good - I will try that model.

src/OmniSharp.Roslyn.CSharp/Services/Navigation/FindSymbolsService.cs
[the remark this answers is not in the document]
rchande: I guess, what value do you think we'll recommend to users?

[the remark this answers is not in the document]
dmgonch: I would suggest minFindSymbolsFilterLength == 1 and maxFindSymbolsItems == 1000 by default.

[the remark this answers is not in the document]
dmgonch: Went with minFindSymbolsFilterLength == 0 and maxFindSymbolsItems == 1000 in the last update to preserve the current user experience as much as possible why tackling scale issue with the default settings. Let me know what you think of this option.

tests/OmniSharp.Roslyn.CSharp.Tests/FindSymbolsFacts.cs
```diff
@@ -7,13 +7,15 @@
...
+using System.Collections.Generic;
+using OmniSharp.Options;
 
 namespace OmniSharp.Roslyn.CSharp.Tests
 {
     public class FindSymbolsFacts : AbstractSingleRequestHandlerTestFixture<FindSymbolsService>
     {
-        public FindSymbolsFacts(ITestOutputHelper output, SharedOmniSharpHostFixture sharedOmniSharpHostFixture)
-            : base(output, sharedOmniSharpHostFixture)
+        public FindSymbolsFacts(ITestOutputHelper output)
```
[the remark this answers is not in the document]
dmgonch: done in the last update

src/OmniSharp.Abstractions/Models/v1/FindSymbols/FindSymbolsRequest.cs
```diff
@@ -7,5 +7,7 @@ public class FindSymbolsRequest : IRequest
     {
         public string Language { get; set; }
         public string Filter { get; set; }
+        public int MinFilterLength { get; set; }
```
rchande: @DustinCampbell @david-driscoll What's our versioning strategy here? Should this become a "V2" endpoint if we add properties?

src/OmniSharp.Cake/Services/RequestHandlers/Navigation/FindSymbolsHandler.cs
```diff
@@ -21,12 +21,18 @@ public FindSymbolsHandler(OmniSharpWorkspace workspace)
 
         public override Task<QuickFixResponse> Handle(FindSymbolsRequest request)
         {
+            if (request != null && request.Filter != null && request.Filter.Length < request.MinFilterLength)
```
rchande: This changes how the API works, so we probably need to implement this as a V2 API

src/OmniSharp.Cake/Services/RequestHandlers/Navigation/FindSymbolsHandler.cs
```diff
@@ -21,12 +21,18 @@ public FindSymbolsHandler(OmniSharpWorkspace workspace)
...
+                return Task.FromResult(new QuickFixResponse { QuickFixes = Array.Empty<QuickFix>() });
+            }
+
             Func<string, bool> isMatch =
                 candidate => request != null
                 ? candidate.IsValidCompletionFor(request.Filter)
                 : true;
 
-            return Workspace.CurrentSolution.FindSymbols(isMatch, LanguageNames.Cake);
+            int maxItemsToReturn = request == null ? 0 : request.MaxItemsToReturn;
```
rchande: Request is already not null here

src/OmniSharp.Abstractions/Models/v1/FindSymbols/FindSymbolsRequest.cs
dmgonch: IMO, adding these properties safe because their are forward and backward compatible and independent from other ones.

src/OmniSharp.Cake/Services/RequestHandlers/Navigation/FindSymbolsHandler.cs
dmgonch: Hmm, actually when invoked from UTs it WAS null and logic of "isMatch" checks for null as well. Are you implying that IRequestHandler.Handle can never receive a null?

rchande: Don't you check for not-null on line 24?

dmgonch: The previous check is more broad though: (request != null && request.Filter != null && request.Filter.Length < request.MinFilterLength. It will still allow to request to be null when maxItemsToReturn  is calculated. I did this to prevent changing existing behavior of the handler when MinFilterLength is not provided or doesn't need to be applied. Would you suggest to always return empty result when request == null? Should this be done then across the board for all requests?

rchande: Right--reading comprehension :)

src/OmniSharp.Abstractions/Models/v1/FindSymbols/FindSymbolsRequest.cs
DustinCampbell: IMO. this doesn't need to be a new "V2" endpoint because it's just an additional property. However, I wonder if it should be `int?` to signal that it's optional?

dmgonch: Your guys call. I see both approaches are used in the current APIs, e.g. https://github.com/OmniSharp/omnisharp-roslyn/blob/6836fadb9c35a88d4695276d14302336a460b841/src/OmniSharp.Abstractions/Models/v1/GotoDefinition/GotoDefinitionRequest.cs#L8 and https://github.com/OmniSharp/omnisharp-roslyn/blob/6836fadb9c35a88d4695276d14302336a460b841/src/OmniSharp.Abstractions/Models/v1/CodeAction/CodeActionRequest.cs#L9

dmgonch: ... but I personally try to avoid allowing nulls in general and in APIs particularly since they are usually confusing and error prone.

dmgonch: @DustinCampbell : I will go with int? then. Thanks.

dmgonch: @DustinCampbell, @rchande  : done

tests/OmniSharp.Cake.Tests/FindSymbolsFacts.cs
```diff
@@ -23,25 +24,41 @@ public FindSymbolsFacts(ITestOutputHelper testOutput) : base(testOutput)
...
         {
-            var symbols = await FindSymbols("CakeProject");
+            var symbols = await FindSymbols("CakeProject", null, null);
             Assert.NotEmpty(symbols.QuickFixes);
         }
 
+        [Fact]
+        public async Task ShouldNotFindSymbolsInCakeProjectsDueToEmptyFilter()
+        {
+            var symbols = await FindSymbols("CakeProject", 1, 0);
```
rchande: Nit: can you use named args for the integer consts/nulls?

dmgonch: updated

filipw:
@rchande @akshita31 I somehow missed this PR before.. what is the reason for `MinFilterLength` to be introduced *on a request*? If that comes from the request, then the sender of the request should already know better whether a given value should result in request being sent or not.

this specific check doesn't make much sense, we are validating sender's input, based on sender's condition.. https://github.com/OmniSharp/omnisharp-roslyn/pull/1284/files#diff-030b76f11f6d3bc42e70a3cde2a74be9R24

I think the min length should be implemented on the client side in the VS Code extension, and here we could at best have a global setting that would come from `omnisharp.json` not from the request

akshita31:
@filipw Agreed. Apologies that this got missed out somehow.

dmgonch:
@filipw Thanks for the feedback! 
@akshita31 I will look into making the change unless this is something you are planning to work on right away.

filipw:
I sent a PR here https://github.com/OmniSharp/omnisharp-vscode/pull/2625

PR #1290, opened 2018-09-09
Fix missing signatures for some cases

SirIntruder:
A fix for https://github.com/OmniSharp/omnisharp-vscode/issues/2495. Provides missing signature help for compiler literals and typeof(T)... expressions. Added two tests for those cases.

bonus question: will roslyn cover the whole signature service soon? it sounds like something that should be handled out-of-the-box

### DIFF
--- a/tests/OmniSharp.Roslyn.CSharp.Tests/SignatureHelpFacts.cs
+++ b/tests/OmniSharp.Roslyn.CSharp.Tests/SignatureHelpFacts.cs
@@ -747,7 +747,7 @@ class B : A
     }
 }";
             var actual = await GetSignatureHelp(filename, source);
-            Assert.NotEmpty(actual.Signatures);
+            Assert.Contains(actual.Signatures, signature => signature.Name == "ToString" && signature.Parameters.Count() == 0);
         }
 
         [Theory]
@@ -764,7 +764,7 @@ class B : A
     }
 }";
             var actual = await GetSignatureHelp(filename, source);
-            Assert.NotEmpty(actual.Signatures);
+            Assert.Contains(actual.Signatures, signature => signature.Name == "GetMembers" && signature.Parameters.Count() == 0);
         }
 
         [Fact]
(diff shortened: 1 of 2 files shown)


DustinCampbell:
> since Roslyn 2.9.0, Quick Info API is there dotnet/roslyn#20554

@filipw: yeah, I've been meaning to go and use that in OmniSharp. Still trying to find time to make a public Signature Help API.

DustinCampbell:
@SirIntruder: A public signature help API for Roslyn is on my plate: https://github.com/dotnet/roslyn/issues/3540

tests/OmniSharp.Roslyn.CSharp.Tests/SignatureHelpFacts.cs
```diff
@@ -733,6 +733,40 @@ void M1(int a,int b,int c)
...
+            const string source =
+@"class A
+{
+    static void M()
+    {
+        var members = typeof(A).GetMembers($$);
+    }
+}";
+            var actual = await GetSignatureHelp(filename, source);
+            Assert.NotEmpty(actual.Signatures);
```
david-driscoll: It would be great if this could also assert that one of the intended signatures exists.

tests/OmniSharp.Roslyn.CSharp.Tests/SignatureHelpFacts.cs
```diff
@@ -733,6 +733,40 @@ void M1(int a,int b,int c)
...
+            const string source =
+@"class A
+{
+    static void M()
+    {
+        string three = 3.ToString($$);
+    }
+}";
+            var actual = await GetSignatureHelp(filename, source);
+            Assert.NotEmpty(actual.Signatures);
```
david-driscoll: It would be great if this could also assert that one of the intended signatures exists.

david-driscoll: This way we can avoid potential signatures being removed when they shouldn't have been.

SirIntruder:
@david-driscoll New enhanced assert now checks if signature list contains one known valid signature (right name, zero parameters). Would this be enough?

PR #1293, opened 2018-09-11
Fix OmniSharp log filtering

DustinCampbell:
When the logging packages were updated from 1.1 to 2.1, we took a breaking change (https://github.com/aspnet/Announcements/issues/238) that caused the `--loglevel` switch to no longer work properly. So, passing `--loglevel debug` doesn't do anything.

This change uses the new APIs for configuring logging and centralizes the filter. Note that I didn't change the LSP implementation because it looks to me like LSP is using loggers differently.

### DIFF
--- a/src/OmniSharp.Host/CompositionHostBuilder.cs
+++ b/src/OmniSharp.Host/CompositionHostBuilder.cs
@@ -147,7 +147,7 @@ namespace OmniSharp
                         !category.Equals(workspaceInformationServiceName, StringComparison.OrdinalIgnoreCase) &&
                         !category.Equals(projectEventForwarder, StringComparison.OrdinalIgnoreCase));
 
-                configureLogging(builder);
+                configureLogging?.Invoke(builder);
             });
 
             return services.BuildServiceProvider();
--- a/src/OmniSharp.Host/CompositionHostBuilder.cs
+++ b/src/OmniSharp.Host/CompositionHostBuilder.cs
@@ -106,7 +106,12 @@ namespace OmniSharp
             }
         }
 
-        public static IServiceProvider CreateDefaultServiceProvider(IOmniSharpEnvironment environment, IConfigurationRoot configuration, IEventEmitter eventEmitter, IServiceCollection services = null)
+        public static IServiceProvider CreateDefaultServiceProvider(
+            IOmniSharpEnvironment environment,
+            IConfigurationRoot configuration,
+            IEventEmitter eventEmitter,
+            IServiceCollection services = null,
+            Action<ILoggingBuilder> configureLogging = null)
         {
             services = services ?? new ServiceCollection();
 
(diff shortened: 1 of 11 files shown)


src/OmniSharp.LanguageServerProtocol/LanguageServerLoggerFactory.cs
```diff
@@ -33,5 +34,30 @@ public void Dispose()
         {
             throw new NotImplementedException();
         }
+
+        private static bool LogFilter(string category, LogLevel level, IOmniSharpEnvironment environment)
```
DustinCampbell: I moved this here because LSP is the only one using this method now. It might be that this is unneded, but LSP's logging looked a bit fancier, so I left it.

src/OmniSharp.Host/CompositionHostBuilder.cs
```diff
@@ -130,7 +135,20 @@ public static IServiceProvider CreateDefaultServiceProvider(IOmniSharpEnvironmen
...
             services.AddSingleton(configuration);
 
-            services.AddLogging();
+            services.AddLogging(builder =>
+            {
+                var workspaceInformationServiceName = typeof(WorkspaceInformationService).FullName;
+                var projectEventForwarder = typeof(ProjectEventForwarder).FullName;
+
+                builder.AddFilter(
+                    (category, logLevel) =>
```
DustinCampbell: Note that all of the logic below is inverted from the original `HostHelpers.LogFilter(...)` method.

src/OmniSharp.Host/CompositionHostBuilder.cs
```diff
@@ -130,7 +135,20 @@ public static IServiceProvider CreateDefaultServiceProvider(IOmniSharpEnvironmen
             services.Configure<OmniSharpOptions>(configuration);
             services.AddSingleton(configuration);
 
-            services.AddLogging();
```
DustinCampbell: The code below is how logging is supposed to be configured these days. Filtering is a first-class feature for the logging APIs now.

rchande:
@DustinCampbell How hard would it be to add some tests for this?

DustinCampbell:
It'd take a bunch more refactoring. The problem is that we use a different `IServiceProvider` for tests and that's the component where logging is configured.

DustinCampbell:
I think I see a place where I can add a test. I'll give it a shot.

DustinCampbell:
Added a couple of tests

src/OmniSharp.LanguageServerProtocol/LanguageServerLoggerFactory.cs
david-driscoll: That should be fine, I'll see if I can take a look at lsp logging later.

PR #1302, opened 2018-09-25
Fixed issue with "go to implementation" and symbol overrides

filipw:
This PR fixes the issue where "find implementation" OmniSharp would not find symbol overrides in child types. We were missing the look up of overrides on the symbols returned from `SymbolFinder.FindImplementationsAsync`.

Consider the following:

```csharp
    interface IFoo
    {
        string Hi();
    }

    public class Bar : IFoo
    {
        public virtual string Hi() => "hi";
    }

    public class Baz : Bar
    {
        public override string Hi() => "hello";
    }
```

In Visual Studio, trying to go to implementation of `Hi()`, from the interface declaration, would find both `Bar.Hi()` and `Baz.Hi()`. However, OmniSharp would find only `Bar.Hi()`.

This is with the fix applied:

screenshot 2018-09-25 19 38 17

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All changes to the project will be documented in this file.
 
 ## [1.33.0] - Not yet released
 * Fixed issue where `/codestructure` endpoint did not return enum members. (PR: [#1285](https://github.com/OmniSharp/omnisharp-roslyn/pull/1285))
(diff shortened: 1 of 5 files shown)


filipw:
@rchande @david-driscoll @DustinCampbell  can you guys take a quick look? I'd like to get this in so that it doesn't lay around for ages 🎃

filipw:
thanks - there is another PR on the same file(s) so we need to start unclogging them 🙃

PR #1304, opened 2018-09-27
Use FindSourceDeclarationsWithPatternAsync to have nice fuzzy search

SirIntruder:
I've noticed symbol search seems much less powerful than VS's one, which always seems to return nice fuzzy searches. 

o# uses "SymbolFinder.FindSourceDeclarationsAsync" with a custom matching algorithm passed in, while there is "SymbolFinder.FindSourceDeclarationsWithPatternAsync", which seems to the right thing when the search filter is passed in. 

After this change "#ProjMana" is able to find **Proj**ect**Mana**ger

So far it works as I expected, but somebody who know this API better should check if this makes sense. 

btw, when request.filter is not specified, full symbol list should be returned, which used to be done with (x => true). It seems funny, but I couldn't find a better way.


DustinCampbell:
glad to have gotten this API in so OmniSharp can use it!

filipw:
thanks!

PR #1308, opened 2018-10-02
Fix o# getting stuck after undoing delete/rename/move transient files

SirIntruder:
1. Have simple MSBuild workspace with one .csproj
2. Add file to path "src/Foo.cs"
3. Delete file on path "src/Foo.cs"

Path "src/Foo.cs" is now "poisoned" because workspace removed the created document, but BufferManager didn't remove it from its internal collections of transient documents.

Adding anything to that path results in documend addition failing with "KeyAlreadyExists" exception, and from the user's perspective, there will be no language features on this path until omnisharp is restarted, and the rest of the workspace will treat contents on that path as non existing.

PR forces BufferManager.OnWorkspaceChanged to properly update its collections on DocumentRemove event.

I expanded existing test,  which was file creation only, to test create-delete-create. I could copy that test and add addition asserts, just didn't want to dent on build time with seting up new o# test host (somebody correct me if this is too paranoid :smile:). 

Create-delete-create is a pretty artificial case, but similar issue can happen in many situations, as in undoing renames, accidental drag'n'drops and switching branches etc.

### DIFF
--- a/src/OmniSharp.Roslyn/BufferManager.cs
+++ b/src/OmniSharp.Roslyn/BufferManager.cs
@@ -222,21 +222,29 @@ namespace OmniSharp.Roslyn
 
             lock (_lock)
             {
-                if (_transientDocumentIds.Contains(args.DocumentId))
-                {
-                    return;
-                }
+                bool replacedWithRealDocument =
+                    args.Kind == WorkspaceChangeKind.DocumentAdded && !_transientDocumentIds.Contains(args.DocumentId);
+                bool documentRemoved =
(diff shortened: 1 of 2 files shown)


SirIntruder:
By https://github.com/OmniSharp/omnisharp-roslyn/pull/1330, actually. This PR would have prevented dangling references to deleted documents, but with #1330 those dangling references are just overwritten over and no longer cause "KeyAlreadyExist" exceptions. 

I am finished with this PR, it fixes the underlying issue as it is. But since the underlying issue currently doesn't manifest itself, it's ok to close this just to avoid change noise.

Issue #953, opened 2017-08-28
Be a good citizen on Windows (don't dump config into %USERPROFILE%).


MicahZoltu:
On Windows, applications should _not_ be creating folders/files directly in `%USERPROFILE%`.  The proper location for configuration files is `%APPDATA%\<vendor>\<product>` and the proper location for cache files is `%LOCALAPPDATA%\<vendor>\<product>`.  Things should only be added to `%USERPROFILE%` if the user _explicitly_ requests it via a save dialog box.

Please move `%USERPROFILE%\.omnisharp` to `%APPDATA%\omnisharp` as the default location.  Also please provide an environment variable override as a temporary mitigation to this problem until a full migration can be completed (I recognize that changing storage locations can be a long process).

In the discussion that lead to this change this comment was made: https://github.com/OmniSharp/omnisharp-roslyn/pull/809#discussion_r108952877
> No idea. On my machine, I see the following under USERPROFILE:
> .dnx
> .node-gyp
> .nuget
> .templateengine
> .vscode
> .vscode-insiders
> .vscode-oss-dev
> I think we're in good company in picking USERPROFILE.

Unfortunately, this was just a pointer to how other people were peeing in the pool so omnisharp should join the party.  Also, I believe all of the above have since been fixed or are in the process of being fixed (working towards a long-term migration strategy) to follow Windows best practices.  The fact that configuration files should go in `%APPDATA%` is well documented in Microsoft's OS documentation in a number of places as well as MSDN.  Unfortunately, a lot of people are coming from Linux where `$HOME` is a dumping ground for whatever data any application wants to add, since prior to XDG, linux didn't really have a story for where files should go (this has since been resolved though most apps haven't caught up yet).

Also, in Windows folders should not be named with a `.` prefix as Windows explorer doesn't allow creation of folders with that name which leads to problems.

DustinCampbell:
Related: https://github.com/Microsoft/vscode/issues/17691

david-driscoll:
I don't have any strong feelings keeping it in user profile, we just did because it was all there already.  @DustinCampbell @filipw ?

filipw:
I have no problem moving this to `%APPDATA%`. While many other tools still don't allow overriding the path (i.e. yarn) I think we can definitely add something like `OMNISHARP_HOME` env variable where you can provide your own custom path.

[...]

PR #1317, opened 2018-10-16
allow setting a custom OmniSharp home directory

filipw:
At the moment, OmniSharp home directory is in user profile - `%USERPROFILE%\.omnisharp` or ` ~/.omnisharp`.
This PR introduces an `OMNISHARPHOME` environment variable which can be used to set a custom "home" path.

I mentioned this a while ago https://github.com/OmniSharp/omnisharp-roslyn/issues/953#issuecomment-325577768 and this should be enough to close that issue from OmniSharp server perspective (the remaining aspects are related to VS Code extension).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 All changes to the project will be documented in this file.
 
+## [1.32.7] - not yet released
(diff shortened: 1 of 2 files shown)


filipw:
ok merging 😀

PR #1318, opened 2018-10-23
Fix FindImplementationService not finding all implementations of the partial class

SirIntruder:
Currently, when you use "Find Implementations" on a partial class, it appears to randomly select one of the definitions instead of returning all of them (like VS does). Looks like this was unintentionally broken in order to handle partialMethod case (guess there is makes sense to take symbol.Locations.First() instead of symbol.Locations).

Should be easy to merge with @filipw fix for separate bug fix in the same service.

### DIFF
--- a/tests/OmniSharp.Roslyn.CSharp.Tests/FindImplementationFacts.cs
+++ b/tests/OmniSharp.Roslyn.CSharp.Tests/FindImplementationFacts.cs
@@ -152,7 +152,6 @@ namespace OmniSharp.Roslyn.CSharp.Tests
             Assert.Equal(SymbolKind.Method, implementation.Kind);
         }
 
-
         [Theory]
         [InlineData("dummy.cs")]
         [InlineData("dummy.csx")]
--- a/tests/OmniSharp.Roslyn.CSharp.Tests/FindImplementationFacts.cs
+++ b/tests/OmniSharp.Roslyn.CSharp.Tests/FindImplementationFacts.cs
@@ -95,13 +95,13 @@ namespace OmniSharp.Roslyn.CSharp.Tests
         public async Task CanFindTypeDeclaration(string filename)
         {
             const string code = @"
-                public class MyClass 
-                { 
+                public class MyClass
+                {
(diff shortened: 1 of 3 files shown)


filipw:
can you add a test illustrating what this is fixing?

SirIntruder:
Will push test in later today

SirIntruder:
Sorry for the delay. I added test for finding all implementations of the partial class and one test for partial methods (to make sure this functionality is kept intact).

filipw:
I merged https://github.com/OmniSharp/omnisharp-roslyn/pull/1302, could you please resolve the conflicts - thanks! Otherwise, looks good to me

SirIntruder:
@filipw done!

tests/OmniSharp.Roslyn.CSharp.Tests/FindImplementationFacts.cs
```diff
@@ -152,6 +152,38 @@ public class MyClass
...
+            const string code = @"
+                public partial class SomeClass { partial void Some$$Method(); }
+                public partial class SomeClass { partial void SomeMethod() { /* this is implementation of the partial method */ }";
+
+            var implementations = await FindImplementationsAsync(code, filename);
+
+            Assert.Single(implementations);
+
+            var implementation = implementations.First();
+            Assert.Equal("SomeMethod", implementation.Name);
```
rchande: Assert on the location of `implementation` to prove you found the implementation copy?

filipw: this is a good idea.

david-driscoll: @SirIntruder if you could fix this feedback, this looks good to me.

SirIntruder:
@rchande @filipw @david-driscoll Thanks a lot for the tip, I would actually regress partial method without it. I enhanced the test and changed handler code, looks like it works correctly now.

filipw:
looks good, indeed this is the behavior we want 👍

filipw:
thanks!

PR #1321, opened 2018-10-26
Fixed a regression that broke FindSymbols for CSX

filipw:
This PR fixes a regression introduced in https://github.com/OmniSharp/omnisharp-roslyn/pull/1282 that broke symbol search for CSX.
I also enabled the FindSymbolsFacts to specifically run twice - once for both regular C# and once for CSX to prevent this in the future (we have same pattern on most tests, but it wasn't there on this one though).

This is quite critical so it would be great to merge this ASAP.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All changes to the project will be documented in this file.
 ## [1.32.7] - not yet released
 * It's now possible to override the default location of OmniSharp's global folder (%USERPROFILE%\.omnisharp or ~/.omnisharp.) with an OMNISHARPHOME environment variable (PR: [#1317](https://github.com/OmniSharp/omnisharp-roslyn/pull/1317)) 
 * OmniSharp no longer searches for `config.json` in its source directory to load configuration (PR: [#1319](https://github.com/OmniSharp/omnisharp-roslyn/pull/1319))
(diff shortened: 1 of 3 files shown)


mholo65:
I remember mentioning this in the original PR here https://github.com/OmniSharp/omnisharp-roslyn/pull/1282#discussion_r215375984

filipw:
ah I was on holidays then so it slipped past me 😄

rchande:
@filipw re: shipping this--A big focus for the C# extension in 17.0 has been working with the Asp.NET team to ship an initial Razor experience in the extension. This has involved a lot of validation based on the 1.32.6 release of OmniSharp. Since 1.32.6, there have been other changes unrelated to this one, so we'd probably need to redo some of our validation if we inserted 1.32.7 to the extension. We first shipped the CSX find symbols regression in 1.16.0 in September, so it has been this way for some time. I would prefer to merge your fix and recommend that users who need it set the "omnisharp.version" option to "latest". We can then insert 1.32.7 into 17.1 and include it in betas. Thoughts?

filipw:
thanks, that's OK, let's do it that way

After 1.17.0 ships, would be great if you could publish the 1.17.1-beta1 fairly quickly though, so that there is a concrete extension build that can be installed as a workaround. 
The "latest" way works in the immediate term (as soon as this PR is merged) but longer term is a bit invasive, plus we had broken builds on master before 😀

filipw:
I will merge it then 👍

Issue #1316, opened 2018-10-14
FEATURE PROPOSAL: Smart/configurable projects loading when working with big C# codebases


dmgonch:
**PROBLEM DESCRIPTION**: One of the major factors contributing to VSCode success is its speed. For JavaScript for example users get rich development experience almost immediately after launching the IDE. When working with C# codebases though developers have to wait for ALL projects to be loaded until many code navigation features are available. For codebase containing more than few dozens of projects this delay is very significant. This experience gets even worse when a developer often switches b/w branches which is a common place when working with Git. After checking out a branch, Ominsharp would need to potentially reload many projects which can take minutes. And in bigger codebases a user typically only needs to only work with a small subset of the code, particularly if s/he is only browsing/studying it. The workarounds of opening multiple instances of VSCode rooted in particular subfolders or managing a list of designated folders in a workspace creates unnecessary friction in developer workflow particularly when s/he needs to quickly "jump" b/w various places in the codebase (which is often the case when doing codereviews).

**PROPOSED EXPERIENCE**: One approach for improving the experience described above is to allow user to control whether all projects loaded automatically. When automatic loading is disabled VSCOde could only start loading projects for files when they are opened in the IDE for editing (including all referenced projects recursively). This could be done for example by looking for a csproj file in the same folder where the opened for editing file is located or walk up the folder hierarchy until such project file is found. The PROS of this approach is that VSCode will load only projects for the source code that user is working on without wasting time on projects are not relevant at the moment. Should the user need to switch the current branch or restart VSCode, s/he can get back to the working environment much faster. The obvious CONS of this approach is that user might be getting incomplete semantic information for example when IDE shows references for symbols. Arguably though when reviewing code or working on a particular component in a bigger codebase this could be a good compromise comparing to always getting full list of references for the cost of many minute wait (and draining battery when on laptop) while this information might not be needed at all. Allowing user to make this choice seems like the right thing to do and the prototype of this approach that I was using with the codebase containing almost 800 C# projects for last few months confirms this for me.

**DESIGN**: Introduce new VSCode setting `omnisharp.EnableOnDemandMsBuildProjectsLoad` with default value 'false' which will ensure the existing projects loading behavior. When the setting is set to 'true' VSCode will behave as described above and will load only projects relevant to files that are currently opened in the IDE.
[...]

PR #1322, opened 2018-10-28
On demand projects loading when working with big C# codebases

dmgonch:
Implements feature https://github.com/OmniSharp/omnisharp-roslyn/issues/1316.


dmgonch:
@rchande, @akshita31, @DustinCampbell : appreciate if you could take a quick look and comment on the overall direction of the changes.

src/OmniSharp.MSBuild/ProjectManager.cs
```diff
@@ -93,7 +137,6 @@ protected override void DisposeCore(bool disposing)
 
         public void QueueProjectUpdate(string projectFilePath, bool allowAutoRestore)
         {
-            _logger.LogInformation($"Queue project update for '{projectFilePath}'");
```
rchande: >_logger.LogInformation($"Queue project update for '{projectFilePath}'"); [](start = 12, length = 72)

Why remove this?

src/OmniSharp.MSBuild/ProjectManager.cs
```diff
@@ -75,6 +92,33 @@ public ProjectManager(ILoggerFactory loggerFactory, IEventEmitter eventEmitter,
...
+            }
+        }
+
+        private void OnWorkspaceDocumentRequested(object sender, DocumentRequestedEventArgs args)
+        {
+            // Search and queue for loading C# projects that are likely to reference the requested file
+            string projectDir = Path.GetDirectoryName(args.DocumentPath);
+            while(projectDir != null && projectDir.StartsWith(_environment.TargetDirectory.TrimEnd(Path.DirectorySeparatorChar)))
+            {
+                List<string> csProjFiles = Directory.EnumerateFiles(projectDir, "*.csproj", SearchOption.TopDirectoryOnly).ToList();
```
rchande: >List<string> csProjFiles = Directory.EnumerateFiles(projectDir, "*.csproj", SearchOption.TopDirectoryOnly).ToList(); [](start = 16, length = 116)

If you turn this feature on, and someone has a project that contains files outside the subtree of the project file, will they be broken?

src/OmniSharp.MSBuild/ProjectManager.cs
dmgonch: accidental change - will restore

---
In reply to: [229800396](https://github.com/OmniSharp/omnisharp-roslyn/pull/1322#discussion_r229800396) [](ancestors = 229800396)

src/OmniSharp.MSBuild/ProjectManager.cs
dmgonch: effectively. do you have a better idea for how to search to "relevant" projects? This approach worked well with codebases I had to deal with so far. Would just adding logging a warning in that case be enough in your opinion?

---
In reply to: [229806732](https://github.com/OmniSharp/omnisharp-roslyn/pull/1322#discussion_r229806732) [](ancestors = 229806732)

dmgonch: @rchande : I kept thinking about a better general approach and couldn't come up with one. For C# projects it is very rare when C# source file is located NOT under project's directory. I could think of few customization for the search that could be applied but I do not believe it is worth to try solving the problem without having compelling cases in hand. I would be interested in your thoughts on this.

src/OmniSharp.Abstractions/Services/IProjectSystem.cs
```diff
@@ -29,5 +29,11 @@ public interface IProjectSystem
         /// <param name="filePath">The file path to the project to retrieve. Alternatively,
         /// a file path to a document within a proejct may be specified.</param>
         Task<object> GetProjectModelAsync(string filePath);
+
+        /// <summary>
+        /// This project system will try to locate projects that reference given file and will wait until the projects and their references are fully loaded.
+        /// </summary>
+        /// <param name="filePath">File for which to locate and load projects.</param>
+        Task WaitForProjectsToLoadForFileAsync(string filePath);
```
dmgonch: @rchande @akshita31 @DustinCampbell : myself and couple of folks were dogfooding the changes and so far everything works as expected (within the scenarios described in https://github.com/OmniSharp/omnisharp-roslyn/issues/1316). I'm moving towards beefing up UTs and appreciate if you could share any additional concerns you might have about the approach. Thanks!

dmgonch:
Added UTs so the changes are ready for review.

dmgonch:
@rchande @DustinCampbell @akshita31 : Folks, assuming you are still supporting proposal https://github.com/OmniSharp/omnisharp-roslyn/issues/1316, could you please comment on this implementation? Thanks!

src/OmniSharp.Roslyn.CSharp/Services/Structure/BlockStructureService.cs
```diff
@@ -50,7 +53,15 @@ public BlockStructureService(IAssemblyLoader loader, OmniSharpWorkspace workspac
 
         public async Task<BlockStructureResponse> Handle(BlockStructureRequest request)
         {
+            // Waiting until the document is fully formed in memory (for project systems that have this ability) 
+            // helps to reduce chances of returning invalid document block structure while compilation is still in progress.
+            await _projectSystems.WaitForAllProjectsToLoadForFileAsync(request.FileName);
```
rchande: This doesn't seem like the right abstraction. 
If I write something that uses OmniSharp workspace, but I don't import all the project systems to call this API, what happens?

If a requirement of GetDocument is that we wait for the project systems, we should move this into GetDocument (or some abstraction over the combination thereof)

rchande: Case in point: BlockStructure is a feature that only require the syntax tree for a document. It doesn't do any cross/document cross/project analysis...

---
In reply to: [232089084](https://github.com/OmniSharp/omnisharp-roslyn/pull/1322#discussion_r232089084) [](ancestors = 232089084)

rchande: The Visual Studio components of Roslyn have a term for this: you can request a document with "full" or "partial" semantics. Requesting partial semantics will result in less work being done to produce the document + project graph.

---
In reply to: [232089407](https://github.com/OmniSharp/omnisharp-roslyn/pull/1322#discussion_r232089407) [](ancestors = 232089407,232089084)

dmgonch: @rchande : Very good points! Indeed introducing the concept of a partial vs fully loaded semantic model would be a better abstraction. At the moment the behavior of existing OmniSharpWorkspace.GetDocumentXXXX methods IMO matches more to retrieving data from a partial semantic model. The alternative design that I would like to propose (to the one that is in this PR) is to add methods like the following. The methods will be waiting until all projects are loaded before returning the result:

```
        public DocumentId GetDocumentIdFromFullSemanticModel(string filePath)
        public IEnumerable<Document> GetDocumentsFromFullSemanticModel(string filePath)
        public Document GetDocumentFromFullSemanticModel(string filePath)
```

Regarding BlockStructure requests processing – it turned out that they do need to use the full semantic model because otherwise blocks like doc-comments or usings that might not be marked as collapsible in the editor.

rchande: @dmgonch I agree with the concept but think we can improve on the terminology at little. Instead of `SemanticModel`, maybe we should be explicit about how this waits for all the projects to finish? `GetDocument*FromFullProjectModel`?

dmgonch: @rchande : `GetDocument*FromFullProjectModel` sounds better - I will go with it. Thanks for the feedback!

src/OmniSharp.MSBuild/ProjectManager.cs
dmgonch: Added OnDemandProjectsLoadSearchStopsAt option to allow controlling the top folder to project searches. By default will stop at a git repo root.

src/OmniSharp.Roslyn.CSharp/Services/Structure/BlockStructureService.cs
dmgonch: @rchande Reworked and retested changes as discussed above. The new methods are Async because they need to perform wait while projects are being loaded and request handlers calling the methods are already async as well.

src/OmniSharp.MSBuild/ProjectManager.cs
rchande: @dmgonch I don't understand why `.git` is in the list by default. Isn't `.git` where git stores its internal data? No user files (should) be in it...

src/OmniSharp.Roslyn/BufferManager.cs
```diff
@@ -46,8 +56,12 @@ public async Task UpdateBufferAsync(Request request)
 
             var solution = _workspace.CurrentSolution;
 
-            var documentIds = solution.GetDocumentIdsWithFilePath(request.FileName);
-            if (!documentIds.IsEmpty)
+            // To properly update the buffer wait until all projects are loaded.
+            var documentIds = inFullProjectModel ? 
+                await _workspace.GetDocumentIdsWithFilePathFromFullProjectModelAsync(request.FileName) :
```
rchande: This is...concerning. What happens if the queue has a lot of stuff in it and buffer update requests start piling up?

src/OmniSharp.Roslyn.CSharp/Services/Structure/CodeStructureService.cs
```diff
@@ -32,7 +32,8 @@ public class CodeStructureService : IRequestHandler<CodeStructureRequest, CodeSt
 
         public async Task<CodeStructureResponse> Handle(CodeStructureRequest request)
         {
-            var document = _workspace.GetDocument(request.FileName);
+            // To provide complete code structure for the document wait until all projects are loaded.
+            var document = await _workspace.GetDocumentFromFullProjectModelAsync(request.FileName);
```
rchande: I noticed that Intellisense did not move onto the `FullProject` APIs. Because of that, you could open a file and start typing and get no intellisense.

src/OmniSharp.MSBuild/ProjectManager.cs
dmgonch: @rchande : this is just a name of a marker folder. When walking up a tree of dirs, once the one is reached containing one of such marker folders (or files) the search will stop. Indeed we will never search in that folder. Does this clarify things?

src/OmniSharp.Roslyn.CSharp/Services/Structure/CodeStructureService.cs
dmgonch: @rchande : the same is true currently as well - only after OnmiSharp loads all/some projects IntelliSense shows up. Or am I missing your question?

src/OmniSharp.Roslyn/BufferManager.cs
dmgonch: @rchande : the updates will wait. Without this I've seen lot of errors generated reported for a file while projects are still being loaded. Let me try to get a stable repro for this through and get back to you.

dmgonch: Reverted this part of the change because it turns out even without any of my changes the error repros. Opened [https://github.com/OmniSharp/omnisharp-roslyn/issues/1356](https://github.com/OmniSharp/omnisharp-roslyn/issues/1356)

---
In reply to: [239610163](https://github.com/OmniSharp/omnisharp-roslyn/pull/1322#discussion_r239610163) [](ancestors = 239610163)

src/OmniSharp.MSBuild/ProjectManager.cs
dmgonch: I believe this is resolved now particularly after adding OnDemandProjectsLoadSearchStopsAt

---
In reply to: [230495784](https://github.com/OmniSharp/omnisharp-roslyn/pull/1322#discussion_r230495784) [](ancestors = 230495784)

src/OmniSharp.Roslyn/BufferManager.cs
dmgonch: ... and confirmed that the fix for the bug works when projects are loaded on demand as well.

---
In reply to: [240308562](https://github.com/OmniSharp/omnisharp-roslyn/pull/1322#discussion_r240308562) [](ancestors = 240308562,239610163)

src/OmniSharp.MSBuild/ProjectManager.cs
```diff
@@ -75,6 +89,61 @@ public ProjectManager(ILoggerFactory loggerFactory, IEventEmitter eventEmitter,
...
+                        {
+                            QueueProjectUpdate(csProjFile, allowAutoRestore:true);
+                        }
+                    }
+
+                    break;
+                }
+
+                bool foundStopAtItem = false;
+                foreach (string stopAtItemName in _options.OnDemandProjectsLoadSearchStopsAt)
```
rchande: There's also IOmniSharpEnvironment.TargetDirectory, which is the directory that VS Code tells omnisharp it's opened against. Do you think stopping searches at TargetDirectory would sufficient? If so we could just get rid of this option.

dmgonch: @rchande : I tried and found out that IOmniSharpEnvironment.TargetDirectory is quite inconsistent. It looks like the code in VSCode looks for any *.sln files and if found then uses that dir as TargetDirectory. In few bigger codebases I worked in, there were various *.sln files checked in, but I expect if I type "code ." from the root of the repo, the project search would stop at that root, not what TargetDirectory might have set to. So I think unfortunately the answer is no, the property can't be used here.

rchande: Well...maybe. It would be perfectly reasonable to use TargetDirectory for a repo with only one sln file. Please remember that your repo with many solutions/thousands of projects/etc is an extreme outlier :)

src/OmniSharp.Roslyn/OmniSharpWorkspace.cs
```diff
@@ -25,6 +25,8 @@ public class OmniSharpWorkspace : Workspace
 
         private readonly ILogger<OmniSharpWorkspace> _logger;
 
+        private readonly ConcurrentBag<Func<string, Task>> _waitForProjectModelReadyEventHandlers = new ConcurrentBag<Func<string, Task>>();
```
rchande: Why ConcurrentBag? Do you actually add more than one of these?

src/OmniSharp.Roslyn/OmniSharpWorkspace.cs
```diff
@@ -37,6 +39,11 @@ public OmniSharpWorkspace(HostServicesAggregator aggregator, ILoggerFactory logg
 
         public override bool CanOpenDocuments => true;
 
+        public void AddWaitForProjectModelReadyEventHandler(Func<string, Task> eventHandler)
```
rchande: It's not really an EventHandler. Is there a better name?

src/OmniSharp.Roslyn.CSharp/Services/Structure/MembersAsTreeService.cs
```diff
@@ -25,7 +25,8 @@ public async Task<FileMemberTree> Handle(MembersTreeRequest request)
         {
             return new FileMemberTree()
             {
-                TopLevelTypeDefinitions = await StructureComputer.Compute(_workspace.GetDocuments(request.FileName), _discovers)
+                // To provide complete members tree for the document wait until all projects are loaded.
+                TopLevelTypeDefinitions = await StructureComputer.Compute(await _workspace.GetDocumentsFromFullProjectModelAsync(request.FileName), _discovers)
```
rchande: What happens if this just uses the single document overload?

tests/OmniSharp.MSBuild.Tests/OnDemandProjectsLoadTests.cs
```diff
@@ -0,0 +1,162 @@
...
+                Assert.Null(workspaceInfo.SolutionPath);
+                Assert.NotNull(workspaceInfo.Projects);
+                Assert.Equal(2, workspaceInfo.Projects.Count);
+                Assert.Equal("App.csproj", Path.GetFileName(workspaceInfo.Projects[0].Path));
+                Assert.Equal("Lib.csproj", Path.GetFileName(workspaceInfo.Projects[1].Path));
+            }
+        }
+
+        [Fact]
+        public async Task LoadProjectAndItsReferenceOnDemand()
```
rchande: Worth adding a test for transitives of transitives?

tests/OmniSharp.MSBuild.Tests/OnDemandProjectsLoadTests.cs
```diff
@@ -0,0 +1,162 @@
...
+                MSBuildWorkspaceInfo workspaceInfo = await host.RequestMSBuildWorkspaceInfoAsync();
+
+                // Expect empty workspace initially since no documents have been requested yet
+                Assert.Null(workspaceInfo.SolutionPath);
+                Assert.NotNull(workspaceInfo.Projects);
+                Assert.Equal(0, workspaceInfo.Projects.Count);
+
+                // Requesting library document should load only that project
+                GetCodeActionsService codeActionHandler = host.GetRequestHandler<GetCodeActionsService>(OmniSharpEndpoints.V2.GetCodeActions);
+                GetCodeActionsResponse codeActionResponse = await codeActionHandler.Handle(
```
rchande: I think both CodeActions and MembersAsTree use the GetDocuments flavor. Can you add a test that makes a request from the non-plural version?

src/OmniSharp.MSBuild/ProjectManager.cs
dmgonch: @rchande : this is definitely would be much simpler if OnDemandProjectsLoadSearchStopsAt is not there at all, but unfortunately I don't believe there is a way to find out where "code ." was launched from if TargetDirectory is pointing to somewhere deep in source tree. If the former path can be discovered somehow regardless to what TargetDirectory is set to then this would definitely be a good thing to switch to. Please advise.

rchande: @dmgonch Depends on what we think the feature is. If it's "open a solution the way you normally do, but projects load more optimally", then TargetDirectory is the right solution. I think you're proposing something different though?

dmgonch: @rchande : sorry for the confusion. I was referring to opening a folder, the repo root, in VSCode, which the equivalent of  executing command "code ." in the repo root folder from a command line. I  couldn't find a way to reliably get that folder path when there are *.sln files in the repo. In my testing IOmniSharpEnvironment.TargetDirectory sometimes would be set to the folder path but more often would be set of a solution file path found in the tree.

rchande: @dmgonch Ah. In the C# extension, users can choose the "launch target" by clicking the folder icon immediately right of the green flame in the VS Code status bar at the bottom of the window. There's also a C# extension option you can set to automatically select that launch target if it's available. Do you think instructing users to use that feature would be sufficient?

dmgonch: @rchande : Don't see the folder icon - what am I doing wrong? 
image
Are you referring to "omnisharp.defaultLaunchSolution" setting? That one doesn't seem to match the experience for big repos. The opposite of it would be useful i.e. where any checked in solution is ignored.
Additionally there should be a way IMO to open VSCode from VS Dev Command Prompt, particularly for Net Full projects.
So to tell the truth OnDemandProjectsLoadSearchStopsAt still seems the like the best option with '.git' as a default, though in many cases it doesn't needed to be explicitly defined.
I'm truly looking to find the best compromise here but I wouldn't want to sacrifice experience working with big repos.
Please advise if you have any other ideas.

src/OmniSharp.Roslyn/OmniSharpWorkspace.cs
dmgonch: @rchande Do you think "CallBack" or just "Hander" sound better?

tests/OmniSharp.MSBuild.Tests/OnDemandProjectsLoadTests.cs
dmgonch: @rchande Not sure I'm following. Could you please clarify what test you are proposing to be added?

src/OmniSharp.Roslyn.CSharp/Services/Structure/MembersAsTreeService.cs
dmgonch: @rchande StructureComputer.Compute at the moment takes "IEnumerable<Document>" I'm assuming for a reason. In general though I was trying hard to make zero changes to the extension's behavior when on-demand-projects-loading is not enabled. Please advise if you think that only a single document can be used in either mode.

tests/OmniSharp.MSBuild.Tests/OnDemandProjectsLoadTests.cs
rchande: `LoadProjectWithTransitiveReferenceCalledFooAndTransitiveReferencesOfFooAreLoaded`?

src/OmniSharp.Roslyn.CSharp/Services/Structure/MembersAsTreeService.cs
rchande: Ok, let's leave this for me.

src/OmniSharp.MSBuild/ProjectManager.cs
rchande: @dmgonch That's a good idea (adding on demand project mode to the extension). That + removing the "searchatthing" will make for a more cohesive experience overall. Let's work on that once this is merged.

src/OmniSharp.Roslyn/OmniSharpWorkspace.cs
rchande: Handler is fine.
I guess the overall concern is that there's now a circular dependency between ProjectLoader and OmniSharpWorkspace. I'm not sure of a better way to improve this right now though...

tests/OmniSharp.MSBuild.Tests/OnDemandProjectsLoadTests.cs
dmgonch: Actually Code Actions service does already request a single document: https://github.com/OmniSharp/omnisharp-roslyn/blob/24ab36e6e98beb21629f400e4905da53494e5a26/src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/BaseCodeActionService.cs#L66

src/OmniSharp.MSBuild/ProjectManager.cs
dmgonch: To enable few more scenarios (listed in a comment in the code) I ended simply removing OnDemandProjectsLoadSearchStopsAt and going to the original implementation I had. I'm still going to be adding omnisharp.enableMsBuildLoadProjectsOnDemand C# extension setting.

dmgonch:
@rchande I believe I addressed all outstanding feedback but please let me know if I missed anything. Thanks!

PR #1328, opened 2018-11-08
Improve MSBuild discovery for future scenarios

DustinCampbell:
This change fixes MSBuild discovery to handle scenarios where the MSBuild tools path is not included in a "15.0" folder. In future releases of Visual Studio, MSBuild will move to a folder named "Current" and OmniSharp should support that. In addition, I've added some extra logging during discovery to help indicate why a particular MSBuild instance isn't chosen. Finally, this change introduces a new option that causes OmniSharp to produce MSBuild binary logs when loading projects. This should be a big help for MSBuild debugging. It can be enabled in an omnisharp.json file like so:

```JSON
{
    "MSBuild": {
        "GenerateBinaryLogs": true
    }
}
```

### DIFF
--- a/src/OmniSharp.MSBuild/ProjectLoader.cs
+++ b/src/OmniSharp.MSBuild/ProjectLoader.cs
@@ -77,9 +77,27 @@ namespace OmniSharp.MSBuild
 
                 var projectInstance = evaluatedProject.CreateProjectInstance();
                 var msbuildLogger = new MSBuildLogger(_logger);
+
+                var loggers = new List<MSB.Framework.ILogger>()
+                {
+                    msbuildLogger
+                };
+
+                if (_options.GenerateBinaryLogs)
+                {
+                    var binlogPath = Path.ChangeExtension(projectInstance.FullPath, ".binlog");
+                    var binaryLogger = new MSB.Logging.BinaryLogger()
+                    {
+                        CollectProjectImports = MSB.Logging.BinaryLogger.ProjectImportsCollectionMode.Embed,
+                        Parameters = binlogPath
+                    };
+
+                    loggers.Add(binaryLogger);
+                }
+
                 var buildResult = projectInstance.Build(
                     targets: new string[] { TargetNames.Compile, TargetNames.CoreCompile },
-                    loggers: new[] { msbuildLogger });
+                    loggers);
 
                 var diagnostics = msbuildLogger.GetDiagnostics();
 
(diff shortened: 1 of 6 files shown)


DustinCampbell:
Linux is failing because Directory.Exists appears to be case-insensitive. Looking for a good solution...

filipw:
> In future releases of Visual Studio, MSBuild will move to a folder named "Current" and OmniSharp should support that.

oh wow, thanks for staying on top of that

src/OmniSharp.MSBuild/ProjectLoader.cs
```diff
@@ -77,9 +77,27 @@ public ProjectLoader(MSBuildOptions options, string solutionDirectory, Immutable
...
+                {
+                    msbuildLogger
+                };
+
+                if (_options.GenerateBinaryLogs)
+                {
+                    var binlogPath = Path.ChangeExtension(projectInstance.FullPath, ".binlog");
+                    var binaryLogger = new MSB.Logging.BinaryLogger()
+                    {
+                        CollectProjectImports = MSB.Logging.BinaryLogger.ProjectImportsCollectionMode.Embed,
```
filipw: I think this is actually the default behaviour so maybe we can skip explicitly specifying it

src/OmniSharp.MSBuild/ProjectLoader.cs
```diff
@@ -77,9 +77,27 @@ public ProjectLoader(MSBuildOptions options, string solutionDirectory, Immutable
...
+
+                var loggers = new List<MSB.Framework.ILogger>()
+                {
+                    msbuildLogger
+                };
+
+                if (_options.GenerateBinaryLogs)
+                {
+                    var binlogPath = Path.ChangeExtension(projectInstance.FullPath, ".binlog");
+                    var binaryLogger = new MSB.Logging.BinaryLogger()
```
filipw: this is a great addition, will definitely be useful in troubleshooting

src/OmniSharp.MSBuild/ProjectLoader.cs
DustinCampbell: I figured the default behavior was "None": https://github.com/Microsoft/msbuild/blob/master/src/Build/Logging/BinaryLogger/BinaryLogger.cs#L47.

DustinCampbell: Oh, I guess it is the default: https://github.com/Microsoft/msbuild/blob/master/src/Build/Logging/BinaryLogger/BinaryLogger.cs#L68

That's OK, let's keep the code specific so there's no question. I think our use case is definitely to Embed.

Issue #1113, opened 2018-02-15
[LSP] shutdown and exit commands throw an exception


LoneBoco:
I am trying to use omnisharp-roslyn in LSP mode and I am unable to shut down the server through normal means.

```
LSP:  --> shutdown
server: [dbug]: OmniSharp.Extensions.LanguageServer.LspRequestRouter
        Finding descriptor for shutdown
LSP:server: [crit]: OmniSharp.Extensions.LanguageServer.LspRequestRouter
        Failed to handle notification shutdown
System.Reflection.TargetParameterCountException: Parameter count mismatch.
   at System.Reflection.RuntimeMethodInfo.InvokeArgumentsCheck(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at OmniSharp.Extensions.JsonRpc.ReflectionRequestHandlers.HandleRequest(IHandlerDescriptor instance, CancellationToken token)
   at OmniSharp.Extensions.LanguageServer.LspRequestRouter.<RouteRequest>d__11.MoveNext()
```

I tried throwing /shutdownServer to see what would happen (it is currently listed as not supported), and I get this response:

```
LSP:  --> shutdownServer
server: [dbug]: OmniSharp.Extensions.LanguageServer.LspRequestRouter
        Finding descriptor for shutdownServer
server: [dbug]: OmniSharp.Extensions.LanguageServer.LspRequestRouter
[...]

PR #1345, opened 2018-11-27
Solve issue #1113.

LoneBoco:
Send cancellation token when the server shuts down so we terminate the process.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All changes to the project will be documented in this file.
 
 ## [1.32.10] - Not Yet Released
 * Updated to Roslyn 3.0 to match [VS 2019](https://docs.microsoft.com/en-us/visualstudio/releases/2019/release-notes-preview#VS2019_Preview2) (PR: [#1391](https://github.com/OmniSharp/omnisharp-roslyn/pull/1391))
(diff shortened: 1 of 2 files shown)


rchande:
@david-driscoll Does this look right to you?

mholo65:
I’d say we should look at updating LSP libs to latest instead. @david-driscoll, this is fixed there, right?

mholo65:
That said, I might have some time updating that in the end of this week.

LoneBoco:
I took a look at OmniSharp.Extensions.LanguageServer version 0.10.0.  I couldn't get it to work correctly because there has been a ton of changes since the 0.8.0 version I use in my local repository, but this would still be a problem with the latest version.

The issue is here:
https://github.com/OmniSharp/omnisharp-roslyn/blob/0496b7e914b70f18180d9c63b16eac7febc62708/src/OmniSharp.Stdio.Driver/Program.cs#L35-L43

OmniSharp.exe waits on the cancellation token.  The only things that trigger this cancellation token is closing the console or waiting for the host process to terminate.  If the host process sends an Exit request, the server will shut down, but the cancellation token for OmniSharp.exe will never get triggered, so the .exe keeps running.

If you want to use version 0.10.0, you have to do something like:
```C#
_server.Exit.Subscribe(_cancellationTokenSource.Token);
```

I think that would work.  I never tested it though as I couldn't get 0.10.0 to work.

rchande:
@mholo65 @david-driscoll I haven't been following along with LSP very much so I'm going to leave it to you to handle this PR.

filipw:
👍

PR #1347, opened 2018-11-27
Fix devconsole MSBuild discovery

rchande:
cc @dmgonch


src/OmniSharp.Host/MSBuild/Discovery/Providers/DevConsoleInstanceProvider.cs
```diff
@@ -26,7 +27,9 @@ public override ImmutableArray<MSBuildInstance> GetInstances()
                 return NoInstances;
             }
 
-            var toolsPath = FindMSBuildToolsPath(path);
+            var msbuildPath = Path.Combine(path, "MSBuild");
```
dmgonch: I would suggest adding UTs as well to make sure regressions like these are flagged right away in the future. Thanks for the fix!

PR #1349, opened 2018-11-27
[WIP] Add check for dotnet sdk resolvers dll

### LINKED ISSUE #1340: Failing to find the right MSBuild instance and there's no config to help the locator
Hello!

I'm trying to use VSCode in my solution and it keeps failing to build it because the `MSBuildLocator` keeps choosing the wrong instance. Is there any config that I can set to help it to select the second MSBuild instance?
```log
Starting OmniSharp server at 11/19/2018, 11:57:18 AM
    Target: c:\dev\repos\my-project\my-project.sln

OmniSharp server started.
    Path: C:\Users\JohnnySantos\.vscode\extensions\ms-vscode.csharp-1.17.1\.omnisharp\1.32.8\OmniSharp.exe
    PID: 13520

[info]: OmniSharp.Stdio.Host
        Starting OmniSharp on Windows 6.2.9200.0 (x64)
[info]: OmniSharp.Services.DotNetCliService
        DotNetPath set to dotnet
        Located 3 MSBuild instance(s)
            1: Visual Studio Build Tools 2017 15.9.28307.53 - "C:\Program Files (x86)\Microsoft Visual Studio\2017\BuildTools\MSBuild\15.0\Bin"
            2: Visual Studio Community 2017 15.9.28307.53 - "C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\MSBuild\15.0\Bin"
            3: StandAlone 15.0 - "C:\Users\JohnnySantos\.vscode\extensions\ms-vscode.csharp-1.17.1\.omnisharp\1.32.8\msbuild\15.0\Bin"
        Registered MSBuild instance: Visual Studio Build Tools 2017 15.9.28307.53 - "C:\Program Files (x86)\Microsoft Visual Studio\2017\BuildTools\MSBuild\15.0\Bin"
```
The first instance does not have the .net core tooling installed (it only has C++ tooling in it).
I would hope for some kind of configuration in the `omnisharp.json` file to make sure that this sure find an instance with the .net core tooling in it.

afbeelding

Comparing these two folders I found out that there's this folder `SdkResolvers` which keeps the tooling in it. So I guess that a verification for the resolvers in it would be beneficial to find the right msbuild instance.

johnnyasantoss:
Should fix #1340, but I need to know if this is the best way to do it, and if it is I'll add tests.
Pls help :)

### DIFF
--- a/src/OmniSharp.Host/MSBuild/Discovery/Extensions.cs
+++ b/src/OmniSharp.Host/MSBuild/Discovery/Extensions.cs
@@ -1,4 +1,5 @@
-﻿using Microsoft.Extensions.Logging;
+﻿using System.IO;
+using Microsoft.Extensions.Logging;
 
 namespace OmniSharp.MSBuild.Discovery
 {
@@ -18,7 +19,9 @@ namespace OmniSharp.MSBuild.Discovery
                 else
                 {
                     instanceToRegister = instance;
-                    break;
+
+                    if (instance.HasDotNetSdksResolvers())
+                        break;
                 }
             }
 
@@ -41,6 +44,19 @@ Try updating Visual Studio 2017 to the most recent release to enable better MSBu
             }
         }
 
+
+        public static bool HasDotNetSdksResolvers(this MSBuildInstance instance)
+        {
+            const string dotnetSdkResolver = "Microsoft.DotNet.MSBuildSdkResolver";
+            return File.Exists(
+                Path.Combine(
+                    instance.MSBuildPath,
+                    "SdkResolvers",
+                    dotnetSdkResolver,
+                    dotnetSdkResolver + ".dll"
+                )
+            );
+        }
 
         public static bool IsInvalidVisualStudio(this MSBuildInstance instance)
             // MSBuild from Visual Studio 2017 RTM cannot be used.


DustinCampbell:
I don't think this is the right fix. I'm not sure the .NET Core SDK resolvers are installed when the .NET Core workload isn't installed. If that's true, this would cause OmniSharp to stop working for .NET Framework projects.

johnnyasantoss:
Sorry, I'm failing to understand what do you mean with:

> I'm not sure the .NET Core SDK resolvers are installed when the .NET Core workload isn't installed

You are talking about the VS Installation?
btw: I don't think that this has something to do with the Full framework, from what I understand (and from what the name suggests) it's a dotnet core resolver only.

I would like some help to make sure that this works for both environments. Do you have any tests in mind?

PS: I've been using a local build with this fix for 6 days now on VSCode and it works for every solution I've tested so far

johnnyasantoss:
Also, sorry for taking so long to reply. GH is not the best at reminding you that something is pending. hehe 😅

DustinCampbell:
Yes, I meant the VS installation. I suspect that `Microsoft.DotNet.MSBuildSdkResolver.dll` may only be installed in Visual Studio when the .NET Core workload is installed. So, if we require for the presence of that binary, we'd break fully supported .NET scenarios that don't require .NET Core to be installed (such as .NET Framework).

DustinCampbell:
Unfortunately, the perfect solution isn't exactly possible because the projects can't correctly be analyzed without using MSBuild to analyze the projects, and this code is all about finding the MSBuild that will be used. It's a chicken-and-the-egg situation. :smile:

> the loop just tries to find a better (.net full + .net core) VS Installation

I don't think that's precisely accurate WRT to .net full + .net core. The loop will always prefer a VS installation with the .NET Core workload installed over .NET desktop workload. Note that these are different workloads in the VS installer and different installations of VS can have none, one, or both of them installed.

I don't think this change will break things in practice (honestly, it'll probably make things better), but I don't think it's the right approach long term. I'd rather see us add configuration rather than more magic.

johnnyasantoss:
Yeah definitely is a chicken-and-the-egg situation hehehe but can we just analyze the egg from the outside in this situation? 😆

What I mean is that we can try to read all the `.csproj` staticly just to get the sdk it uses:

`<Project Sdk="Microsoft.NET.Sdk">`  = dotnet core

```xml
<?xml version="1.0" encoding="utf-8"?>
<Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/19tooOld" ToolsVersion="old">
```
 = full framework
It might be the same thing that msBuild has to do somewhere..

Would that work? 
And if we do that we also need to have a check to know if the VS installation has the .NET Full framework there, right?

DustinCampbell:
@johnnyasantoss: That'd be an estimate, but there would likely be edge cases where it didn't work. Also, we should be careful not to conflate .NET Core and .NET Framework with SDK-style projects and older-style projects. They aren't equivalent. After all, an SDK-style project can be made to target a .NET Framework TFM.

> It might be the same thing that msBuild has to do somewhere

MSBuild doesn't have to do this at all, it just assumes that everything is installed to build any requested project. The SDK attribute is just a general MSBuild feature.

> Would that work?
> And if we do that we also need to have a check to know if the VS installation has the .NET Full framework there, right?

I wouldn't recommend any of that. It would almost certainly result hair-pulling, difficult to debug issues, not unlike the one you're trying to solve. :smile:

Let's just go with your tweak for now. I really think the ideal experience is to provide good diagnostics when something goes wrong and add the proper configuration support to fix it.

src/OmniSharp.Host/MSBuild/Discovery/Extensions.cs
```diff
@@ -18,7 +19,9 @@ public static void RegisterDefaultInstance(this IMSBuildLocator msbuildLocator,
                 else
                 {
                     instanceToRegister = instance;
-                    break;
+
+                    if (instance.HasDotNetSdksResolvers())
```
rchande: So the idea here is that if we find an MSBuild that did come with the .NET SDK, we should stop searching?

DustinCampbell: Correct. Now that I look a bit more closely, I guess the biggest difference is that if there aren't any VS instances that contain the MSBuildSdkResolver, we'll pick the last instance instead of the first instance. That should probably be fixed before merging to avoid any subtle breaking changes.

johnnyasantoss: Hey 👋 

So, I've rewrote it in a different manner. Which keeps the current behavior (preferring the first) but favors DotNet Core when comparing instances :)

I'll write more tests tomorrow and push it (If I have the time 😢)

johnnyasantoss: Can you please review it again? :)

src/OmniSharp.Host/MSBuild/Discovery/Extensions.cs
```diff
@@ -1,52 +1,104 @@
...
         }
 
+        public static bool HasDotNetSdksResolvers(this MSBuildInstance instance)
+        {
+            const string dotnetSdkResolver = "Microsoft.DotNet.MSBuildSdkResolver";
+
+            return File.Exists(
+                Path.Combine(
+                    instance.MSBuildPath
+                    , "SdkResolvers"
```
rchande: Commas on preceding line

src/OmniSharp.Host/MSBuild/Discovery/Extensions.cs
```diff
@@ -1,52 +1,104 @@
...
+        }
+
+        private static int GetInstanceFeatureScore(MSBuildInstance i)
+        {
+            var score = 0;
+
+            if (i.HasDotNetSdksResolvers())
+                score++;
+
+            if (i.IsInvalidVisualStudio())
```
rchande: It it's invalid, might as well return -1 or so, right? We don't ever want to pick this one

src/OmniSharp.Host/MSBuild/Discovery/Extensions.cs
```diff
@@ -1,52 +1,104 @@
...
+                && (instance.DiscoveryType == DiscoveryType.DeveloperConsole
+                    || instance.DiscoveryType == DiscoveryType.VisualStudioSetup);
+
+        public static MSBuildInstance GetBestInstance(this IMSBuildLocator msbuildLocator, out bool invalidVSFound)
+        {
+            invalidVSFound = false;
+            MSBuildInstance bestMatchInstance = null;
+            var bestMatchScore = -1;
+
+            // same as (without so many allocations)
```
rchande: Remove comment

tests/OmniSharp.MSBuild.Tests/ExtensionsTests.cs
```diff
@@ -0,0 +1,132 @@
+using System;
+using OmniSharp.MSBuild.Discovery;
+using TestUtility;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace OmniSharp.MSBuild.Tests
+{
+    public class ExtensionsTests : AbstractTestFixture
```
rchande: MSBuildSelectionTests?

src/OmniSharp.Host/MSBuild/Discovery/Extensions.cs
johnnyasantoss: Yeah, it will always be the one with the lowest score

tests/OmniSharp.MSBuild.Tests/ExtensionsTests.cs
johnnyasantoss: It was a temp name 😅

src/OmniSharp.Host/MSBuild/Discovery/Extensions.cs
johnnyasantoss: Fixed.

src/OmniSharp.Host/MSBuild/Discovery/Extensions.cs
johnnyasantoss: Done.

tests/OmniSharp.MSBuild.Tests/ExtensionsTests.cs
johnnyasantoss: Renamed

johnnyasantoss:
@rchande are you going to review this again? or is it okay to be merged? :)
Please let me know because I don't know how you guys handle pull requests in this project

src/OmniSharp.Host/MSBuild/Discovery/Extensions.cs
rchande: I was thinking that you could put this case first and return some value that you believe to be the lowest value, making it clear that invalid versions get the worse score.

rchande:
I pulled and built this locally and it appears to work.

johnnyasantoss:
Now it is ready to be merged 😄

johnnyasantoss:
Hey, someone 📡  any updates on this PR?

Issue #1356, opened 2018-12-10
"Ambiguity between '<Type> and '<Type>’" errors if a file is edited before projects are loaded


dmgonch:
PROBLEM: Editing a source file before the project referencing a file is fully loaded causes  "Ambiguity between '<Type> and '<Type>’ " errors.

REPRO STEPS:
1.	Launch VSCode with OmniSharp installed
2.	Clone https://github.com/OmniSharp/omnisharp-roslyn repo locally and open in VSCode
3.	Close all document if any are opened and open src\OmniSharp.Roslyn\OmniSharpWorkspace.cs
4.	Click menu View -> Output and select OmniSharp.log
5.	Close VSCode
6.	Launch VSCode again – it will automatically open omnisharp-roslyn folder and OmniSharpWorkspace.cs file
7.	Wait until you see any text showing up in OmniSharp.log of Output windows
8.	In OmniSharpWorkspace constructor quickly typing in the following: test <Enter> test <Enter> test <Enter>. Then save the file.
9.	Wait until OmniSharp finishes loading all projects.
10.	Hover over red squiggles in the constructor and observe the following errors
Ambiguity between 'OmniSharpWorkspace.BufferManager' and 'OmniSharpWorkspace.BufferManager'
Ambiguity between 'OmniSharpWorkspace._logger' and 'OmniSharpWorkspace._logger'
11.	Undo changes to OmniSharpWorkspace.cs made in step 8 and save the file. Note that ‘Ambiguity’ were not cleared.
12.	Close and reopen OmniSharpWorkspace.cs. Note that ‘Ambiguity’ errors are still there.

EXPECTED: No ‘Ambiguity’ error should be shown in steps 10-12.

ATTACHED:
- Full trace log [AmbiguityErrorsLog.zip](https://github.com/OmniSharp/omnisharp-roslyn/files/2664197/AmbiguityErrorsLog.zip)
- Screenshot for the error ![ambiguityerrors](https://user-images.githubusercontent.com/30739723/49749889-5223bf80-fc5e-11e8-8786-fd64c87eddd8.png)

SirIntruder:
Thanks for the report, I've seen this multiple times. I've submitted a PR that should fix this: #1357

I've tested this in a couple of repos and it's very easy to repro - 1) restart omnisharp 2) edit and save something while it's still loading 3) wait for the errors to appear.

dmgonch:
Thanks for the fix - tested and it is working great!

PR #1357, opened 2018-12-10
Fix "Ambiguity" errors appearing when files are modified during project load

SirIntruder:
fixes https://github.com/OmniSharp/omnisharp-roslyn/issues/1356

Context of what happens:

1) When you edit file before projects get loaded, o# puts them in "Misc" project, because that's the best it can do at the moment.
2) On project load, MSBuildProjectManager creates a real Document on the same path, which kills the "Misc" document
3) **Problem is**: there is a bit of code that tries to "promote" misc documents into the newly loaded project, and this happens before step 2.

This unfortunate timing means that both "TryPromote" and "ProjectLoad" create each their own version of non-misc document (and ambiguity errors follow).

Moving "TryPromote" after actual project load, lets MSBuild properly delete misc files, so unnecessary document creating in "TryPromote" is avoided

cc @akshita31 @rchande @DustinCampbell

On a sIde note: 
Its weird that MiscProject has its own version of "Try-to-put-file-into-exisitng-projects" - IMHO TryPromote should use some part of bufferManager's "TryAddTransientDocument".

### DIFF
--- a/src/OmniSharp.MSBuild/ProjectManager.cs
+++ b/src/OmniSharp.MSBuild/ProjectManager.cs
@@ -353,11 +353,11 @@ namespace OmniSharp.MSBuild
                 return;
             }
 
-            _workspace.TryPromoteMiscellaneousDocumentsToProject(project);
             UpdateSourceFiles(project, projectFileInfo.SourceFiles);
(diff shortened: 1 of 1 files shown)


rchande:
Thanks @SirIntruder!

glenlance:
I understand the reason of what this error happen, but I don't know how to resolve it. Could somebody can help me please?

glenlance:
Well, in some case, this error seem neither effect the compile process nor the app running normally.

PR #825, opened 2017-04-18
Get organized

DustinCampbell:
This PR is a work-in-progress to get things cleaned up a bit to prepare for https://github.com/OmniSharp/omnisharp-roslyn/issues/822.

This touches a lot of files, but it's mostly renaming and moving things around a bit (notably the V1 models).

Issue #822, opened 2017-04-13
Publish to NuGet/MyGet


mholo65:
As discussed with @david-driscoll on Slack. To get plugin development started, libraries should be published to NuGet/MyGet.

For a Cake plugin, we would at least need following:
* OmniSharp.Abstractions
* OmniSharp.Roslyn
* OmniSharp.Roslyn.CSharp

david-driscoll:
After my talk is ready for this Saturday, I'll be picking up the work to get the packages deployed.

@DustinCampbell do you think this would integrate directly into the C# plugin (through some sort of plugin provider) or would it make more sense to have a separate plugin that supported cake files?

My thinking here is if we break an interface between versions the cake plugin wouldn't work until it was updated.  However if we have a separate plugin, the user would be downloading OmniSharp twice... hmm.

@mholo65 thoughts on what kind of integration you guys are looking for?

DustinCampbell:
I think this is a great way to get the plugin functionality going. Shall we create a `#cake` channel on the OmniSharp slack for discussion of this effort?

DustinCampbell:
One thing to consider is what the actually public API for OmniSharp is. At the moment, things a real hodge-podge of public vs. internal.

david-driscoll:
Totally agreed.  I'm with releasing something soon but calling it "beta" in that we may break you, but we didn't do it on purpose. 😄

mholo65:
@david-driscoll as we need to translate linenumbers on both requests and responses and do CodeGen on Buffer updates/changes, I thought creating our own Cake RequestHandlers would be best option.

I managed to work out a solution which doesn't directly depend on OmniSharp.Roslyn.CSharp. Instead MEF will inject all handlers and we will find the corresponding C# handler ( see https://github.com/mholo65/omnisharp-roslyn/blob/wip/new-cake-plugin/src/Cake.OmniSharp/RequestHandlers/CakeRequestHandler.cs#L43)

So, as long as you wont break Abstractions and Roslyn we'll be fine :smile:

mholo65:
@DustinCampbell +1 for ```#cake``` channel

david-driscoll:
That is actually not to bad!

I know we will eventually move away from them toward LSP types as a breaking change in the future but we'll know far in advance when that is happening.  I'm also thinking we may keep the interfaces for some time, and instead have a mapping between the existing requests, and the new interfaces, so that we don't break all our consumers right away. (no promises however)

[...]

PR #1360, opened 2018-12-12
OmniSharp.Abstractions cleanup

filipw:
This is the first step towards publishing OmniSharp packages as NuGet packages (https://github.com/OmniSharp/omnisharp-roslyn/issues/822), as discussed during the last call.
I kept it small for now on purpose, to avoid making many invasive changes and cluttering the PR with too many changes.
It builds up on work done in #825 

 - moved all "utility" style code out of `OmniSharp.Abstractions` into a new utility project `OmniSharp.Shared`
 - removed `OmniSharp.Plugins` and moved the plugin code to `OmniSharp.Abstractions`. I don't really think there is a need to split them at this point yet.
 - only interfaces / models / constants + plugin code are left in `OmniSharp.Abstractions`
 - moved protocol specific types (such as `RequestPacket` etc) to `OmniSharp.Host`
 - moved shared services (such as `ISharedTextWriter`) out of the old (wrong) namespace `OmniSharp.Stdio.Services`

In the next steps we will start moving `public` types to `internal` to avoid publishing to NuGet an API surface that we are not comfortable with.

This change has no impact on any current functionalities. Technically speaking there are some small breaking changes due to namespace changes etc, but we don't publish the packages to Nuget yet so there is no impact.


filipw:
> Curious, is anything using `OmniSharp.Plugins` currently?

I don't think so. It was never published and Razor uses OmniSharp.MsBuild as submodule. 
If anything, it would be someone who compiled from source.

david-driscoll:
Changes look good... going to update branch and see if there are any breaks...

filipw:
@david-driscoll let's merge this and continue building on top of that?

Issue #1274, opened 2018-08-21
OmniSharp-Roslyn v1.32.2+ HTTP is refusing connections


axvr:
### The Problem

The OmniSharp-Roslyn HTTP (v1.32.2) server is refusing all HTTP connections.

The problem seems to have occurred between [v1.32.1 and v1.32.2](https://github.com/OmniSharp/omnisharp-roslyn/compare/v1.32.1...v1.32.2).

### Systems this occurs on

This issue has been reported to occur on these devices:

* Fedora Linux 28 (_OmniSharp-Roslyn HTTP Linux_)
* macOS High Sierra (_OmniSharp-Roslyn HTTP OSX Mono_)
* Arch Linux (_OmniSharp-Roslyn HTTP Linux_)
* Ubuntu 18.10 (_OmniSharp-Roslyn HTTP Linux_)

It does not occur on the Windows versions of the builds.

### Running v1.32.2 of the server from the command line

This was run with the `-v` flag, the previous version of the server printed thousands of lines of `debug:` data, whereas v1.32.2 did not.

<details>
<summary>Click to view output.</summary>
<pre>
[axvr@arctic Redacted.Auth master]$ ~/.omnisharp/omnisharp-roslyn/run -v -p 4242 -s "$(pwd)/Redacted.Auth.sln"
info: OmniSharp.Http.Startup[0]
      Starting OmniSharp on fedora 28.0 (x64)
info: OmniSharp.Cake.CakeProjectSystem[0]
      Detecting Cake files in '/home/axvr/Documents/Projects/work/Redacted/Redacted.Auth'.
info: OmniSharp.Cake.CakeProjectSystem[0]
      Could not find any Cake files
info: OmniSharp.Http.Startup[0]
      Project system 'OmniSharp.DotNet.DotNetProjectSystem' is disabled in the configuration.
info: OmniSharp.MSBuild.ProjectSystem[0]
      Detecting projects in '/home/axvr/Documents/Projects/work/Redacted/Redacted.Auth/Redacted.Auth.sln'.
info: OmniSharp.MSBuild.ProjectManager[0]
      Queue project update for '/home/axvr/Documents/Projects/work/Redacted/Redacted.Auth/Redacted.Auth.Api/Redacted.Auth.Api.csproj'
info: OmniSharp.MSBuild.ProjectManager[0]
[...]

PR #1361, opened 2018-12-12
Fix refusing HTTP connections ( #1274 )

JulioJu:
See especially https://github.com/OmniSharp/omnisharp-roslyn/issues/1274#issuecomment-446622795 and the work with @nickspoons


nickspoons:
I can confirm that copying the files listed in `DeleteUnnecessaryAssemblies` in `build.cake` from `bin/Debug/OmniSharp.Http.Driver/net461/` to `artifacts/publish/OmniSharp.Http.Driver/mono` results in a useable OmniSharp-roslyn server which responds to HTTP requests and can therefore be used by OmniSharp-vim.

JulioJu:
"Delete unnecessary files" are  created by commit d4264fbab789652e44c238ed22ced4ab2031bd06 by @DustinCampbell 

I'm going to check if some files could be deleted without problems.

ohjames:
Ouch this is still open :( Shame the devs don't prioritise us non-Windows users enough to make a simple fix to a totally broken project.

PR #1362, opened 2018-12-14
removed unused stdio flag

filipw:
The STDIO application has an unnecessary flag "stdio" exposed, which is not used for anything. This is a leftover from the days where both HTTP and STDIO could be launched from the same entry point.


filipw:
can we merge this?

filipw:
thanks 😀

PR #1368, opened 2018-12-19
Detect when files move to empty directories

rchande:
Augment the IFileSystemWatcher to allow watching a directory
recursively. This enables OmniSharp to watch a project's entire
directory tree (as oppposed to the specific subdirectories that contain
documents)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ mono: latest
 os:
 - linux
 - osx
-osx_image: xcode9.2
+osx_image: xcode9.4
 script:
 - git fetch --unshallow
 - git fetch origin '+refs/heads/*:refs/heads/*'
@@ -48,6 +48,7 @@ install:
     sudo ldconfig
   elif [ "$TRAVIS_OS_NAME" == "osx" ]; then
     # On OSX, we install libuv via Homebrew
+    brew update
     brew install libuv
   fi
 - export PATH="$PWD/.dotnet:$PATH"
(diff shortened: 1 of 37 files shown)


SirIntruder:
Glad to see this issue finally getting on the chopping block! Yet, I have questions 😆 

1. Wouldn't this result in double trigger of "OnDocumentChanged" when doing stuff in existing directories?

2. Seems there is an arbitrary limitation of only one "WatchRecursively" listener being able to be triggered per event?

Looking at this, I wonder can't MSBuildProjectmanager just subscribe to all "*.cs" on start instead of doing anything per directory? BufferManager will put new files into correct projects, and others will be forwarded to the misc project. If anything, it should make dispatching events in ManualSystemWatcher cheaper, as it would replace 100s of hooks with just one (note that ManualSystemWatcher is not a real system file watcher, it just re-distributes upstream events coming from /fileChanged endpoint. I was confused by this at first)

rchande:
@SirIntruder 

> I wonder can't MSBuildProjectmanager just subscribe to all "*.cs" on start instead of doing anything per directory?

Interesting idea. Let's see what the tests have to say about that...

rchande:
And of course, if that's a good solution, maybe we can get rid of the "FileSystemWatcher" concept altogether, since as you say, it's more of an abstraction for controlling what events from the "real" file system get handled.

rchande:
@filipw care to review?

PR #1379, opened 2019-01-09
update changelog

rchande:


filipw:
OSX on travis is having some weird problems, but that's not related to the PR.

PR #1387, opened 2019-01-21
Update LSP to 0.11.x

mholo65:
Ping @david-driscoll 

There were still some issues with the Cake script service and incremental changes. I've disabled it for now and will enable once it works.


src/OmniSharp.Host/CommandLineApplication.cs
```diff
@@ -80,7 +80,7 @@ private void DebugAttach()
         {
             if (Debug)
             {
-                Console.WriteLine($"Attach debugger to process {Process.GetCurrentProcess().Id} to continue...");
+                //Console.WriteLine($"Attach debugger to process {Process.GetCurrentProcess().Id} to continue...");
```
mholo65: Oops.. This should not be here.

BTW, @david-driscoll, we need a way to get this working with LSP? Currently VSCode starts yelling at me when non-json-RPC formatted content arrives on STDOUT.

david-driscoll:
I'll be getting back to this tomorrow, to figure out what is causing issues with ci.

src/OmniSharp.Host/CommandLineApplication.cs
david-driscoll: I'd be fine with it commented for now... worst case we can always have someone uncomment it for debugging purposes, if they haven't figure out which pid they need.

Issue #1382, opened 2019-01-13
When is next release?


mickaelistria:
The Language Server Protocol was broken for a while and recently got fixed.
When is next release available so we can upgrade Omnisharp in code consuming it over LSP?

mickaelistria:
cc @filipw @david-driscoll @DustinCampbell

filipw:
Recently we aligned released with a new version being needed for the VS Code extension, I am all for publishing a new release as soon as possible, especially as it comes with new Roslyn version. cc @rchande 

when you talk ab out the fix, which PR you refer to? Also there is a pending PR to update LSP #1387 which will be very useful

filipw:
Release 1.32.9 has been published today - thanks for opening this issue 
https://github.com/OmniSharp/omnisharp-roslyn/releases/tag/v1.32.9

mickaelistria:
Thank you!

Issue #1269, opened 2018-08-15
[LSP] `-lsp` option seems to be ignored


svenefftinge:
I want to start omnisharp-roslyn in LSP mode using stdio.
I have installed : https://github.com/OmniSharp/omnisharp-roslyn/releases/download/v1.32.2/omnisharp-linux-x86.tar.gz
When I run, I see output messages not conforming to LSP. 
```
$ ./omnisharp/run -stdio -lsp

{"Event":"log","Body":{"LogLevel":"INFORMATION","Name":"OmniSharp.Stdio.Host","Message":"Starting OmniSharp on ubuntu 18.4 (x64)"},"Seq":1,"Type":"event"}
{"Event":"log","Body":{"LogLevel":"INFORMATION","Name":"OmniSharp.Services.DotNetCliService","Message":"DotNetPath set to dotnet"},"Seq":2,"Type":"event"}
{"Event":"log","Body":{"LogLevel":"INFORMATION","Name":"OmniSharp.MSBuild.Discovery.MSBuildLocator","Message":"Located 1 MSBuild instance(s)\n        1: StandAlone 15.0 - \"/workspace/theia-csharp-extension/csharp/omnisharp/omnisharp/msbuild/15.0/Bin\""},"Seq":3,"Type":"event"}
{"Event":"log","Body":{"LogLevel":"INFORMATION","Name":"OmniSharp.MSBuild.Discovery.MSBuildLocator","Message":"MSBUILD_EXE_PATH environment variable set to '/workspace/theia-csharp-extension/csharp/omnisharp/omnisharp/msbuild/15.0/Bin/MSBuild.dll'"},"Seq":4,"Type":"event"}
...
```

Connecting it to my client (Theia) doesn't work either. The initialize request is sent but never answered. Seems like I am doing something wrong.

svenefftinge:
@mickaelistria I saw you are using the same mode in aCute, but an older version. Do you have an idea what's wrong with my approach?

david-driscoll:
@svenefftinge try without `-stdio` and just do `-lsp`

svenefftinge:
No difference:
```
$ ./omnisharp/run -lsp
{"Event":"log","Body":{"LogLevel":"INFORMATION","Name":"OmniSharp.Stdio.Host","Message":"Starting OmniSharp on ubuntu 18.4 (x64)"},"Seq":1,"Type":"event"}
[...]

PR #1388, opened 2019-01-22
added missing LSP note to CHANGELOG.md

filipw:
related to #1382 and #1269


filipw:
this should still go into into 1.32.9 release, I'll update the notes here too

PR #1391, opened 2019-01-24
Upgrade to 3.0 Roslyn packages to get better C# 8 support

rchande:


filipw:
can you add a note to release notes?

tests/OmniSharp.Roslyn.CSharp.Tests/IntellisenseFacts.cs
```diff
@@ -282,7 +282,7 @@ public MyClass2()
                 ";
 
             var completions = await FindCompletionsAsync(filename, source);
-            ContainsCompletions(completions.Select(c => c.CompletionText).Take(1), "text:");
+            ContainsCompletions(completions.Select(c => c.CompletionText).Take(1), "text");
```
mholo65: Shouldn’t the actual completion be `text:` here?

rchande: This test started failing after I updated Roslyn--I guess this is a behavior change on their side.

mholo65: Yes, I figured. Don’t know the internals of OmniSharp completion handler, but it feels like we need to figure out if this is a bug in Roslyn or if we are ”doing it wrong”. AFAIK, this is a named parameter, thus `:` should be required?

PR #1076, opened 2017-12-26
Roslyn analyzers and code fixes

savpek:
Question 1:
Actually this version is incorrect, analyzer results should likely be returned via event emitter since they may take while. Currently they are returned via CodeCheckService result which causes that analyses doesn't surface if changes aren't made on target projects documents. Easy to see on startup where problems doesn't exist until you make first change. Its also means that if you write change that causes new issue, it is possible that you dont see warning about it before you write another change somewhere.

At first version there wasn't caching on codecheckservice and it calculalated analysis for every document request, it basically anhilated cpu on any bigger solution (surprise surprise).

I tried to change so that event emitter surfaces analyzer results but it seems there are parts missing, i checked how CSharpDiagnosticService makes same thing but there are only QuickFixes currently available on message. This probably means that i need to update consuming applications (like omnisharp-vscode) that they can use those analyzer and surface results correctly. This is where i need guidelines and help how this should be made since it its omnisharp public api.

Question 2:
Should i be concerned about used language? Currently all diagnostic types are loaded, however it is same thing on refactorings and code fixes.

**TODO**
- [x] Implement global analyzers.
- [x] Implement project level analyzers.
- [x] Implement global code fixes.
- [x] Implement project level code fixes.
- [x] Basic support for RuleSets.

PR #1406, opened 2019-02-21
Codeactions with options/ui.

### LINKED ISSUE #1220: Support for code actions/refactorings that show UI in Visual Studio
There are a number of useful code actions in Visual Studio that, when invoked, show some UI that let the user specify the behavior of the code action. This includes, (but might not be limited to) "Extract Interface", "Generate Type", "Generate Equals/Hashcode", "Reorder Parameters", etc. Apart from "Reorder Parameters", it's likely that not showing a UI and simply generating all possible members could be useful.

savpek:
Resolves #1220

List:
- [x] ChangeSignatureCodeAction
  - Todo: blacklist this, i don't think this is possible to impelement without UI...
- [x] ExtractInterfaceCodeAction
- [x] GenerateConstructorWithDialogCodeAction
- [x] GenerateOverridesWithDialogCodeAction
- [x] GenerateEqualsAndGetHashCodeWithDialogCodeAction
- [x] GenerateTypeCodeActionWithOption
  - Blacklist. This don't give extra value over GenerateType (without options) which we have already. With default ui this ends up nearly identical class.
- [x] PullMemberUpWithDialogCodeAction
  - ~~This is bit trickier, may work OK if select first available parent interface as target.~~
  - Blacklist: Theres existing pull members up without dialog. Use that one and blacklist this, this is identical when used without UI.

There are some parts that are internal that are reguired to implement this feature. Pending issue on https://github.com/dotnet/roslyn/issues/33277. This is written with reflection/proxy combination which should be replaced once roslyn issue is addressed.

GenerateConstructorWithDialogCodeAction, GenerateOverridesWithDialogCodeAction, GenerateEqualsAndGetHashCodeWithDialogCodeAction:
code_actions_with_options

Extract interface:
code_actions_with_options_extract_interface

However extract interface isn't available before this code is merged with #1076 since it's action for one of analysis.


david-driscoll:
I'm going to focus on getting this and the related analyzers in right away.

src/OmniSharp.Roslyn/OmniSharp.Roslyn.csproj
```diff
@@ -16,6 +16,7 @@
     <PackageReference Include="Microsoft.Extensions.Caching.Memory" />
     <PackageReference Include="Microsoft.CodeAnalysis.Common" />
     <PackageReference Include="System.ComponentModel.Composition" />
+    <PackageReference Include="Castle.Core" Version="4.3.1" />
```
mholo65: We probably want to avoid a dependency for Castle.Core?

src/OmniSharp.Roslyn/WorkspaceServices/ExtractInterfaceWorkspaceService.cs
```diff
@@ -0,0 +1,47 @@
...
+using System.Reflection;
+using System.Threading.Tasks;
+using Castle.DynamicProxy;
+using Microsoft.CodeAnalysis;
+
+namespace OmniSharp
+{
+    public class ExtractInterfaceWorkspaceService : IInterceptor
+    {
+        public void Intercept(IInvocation invocation)
```
mholo65: To avoid a dependency on Castle.Core, you can have a look in [MetadataHelper](https://github.com/OmniSharp/omnisharp-roslyn/blob/c9d45c06a58d33e02c16328afb2a08f8f3db5dfe/src/OmniSharp.Roslyn/MetadataHelper.cs#L66) on how this could be done.

savpek:
In metadata helper concrete internal type is created (if i am not missing something completly 😄) but this is bit trickier since concrete implementation of internal interface is created without existing concrete implementation. Implementation for IPickMembersService for is in visual studio libraries and not available.

However it think i can write solution based on https://stackoverflow.com/questions/8606412/create-type-that-implements-internal-interface which looks something like example there:
```
public class InterfaceImplementer : RealProxy, IRemotingTypeInfo
{
    readonly Type _type;
    readonly Func<MethodInfo, object> _callback;

    public InterfaceImplementer(Type type, Func<MethodInfo, object> callback) : base(type)
    {
        _callback = callback;
        _type = type;
    }

    public override IMessage Invoke(IMessage msg)
    {
        var call = msg as IMethodCallMessage;

        if (call == null)
            throw new NotSupportedException();

        var method = (MethodInfo)call.MethodBase;

        return new ReturnMessage(_callback(method), null, 0, call.LogicalCallContext, call);
    }

    public bool CanCastTo(Type fromType, object o) => fromType == _type;

    public string TypeName { get; set; }
}
```

Which doesn't look too bad to write and should remove requirement for castle.core depency.

mholo65:
@david-driscoll @filipw and @rchande, any opions on the above suggestion by @savpek? One problem is that neither `RealProxy` nor `IRemotingTypeInfo` is part of netstandard (according to https://apisof.net/)

savpek:
I will try with `DispatchProxy` which seems to be part of net standard.

savpek:
@mholo65 can you check new version which is based on DispatchProxy.

src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/BaseCodeActionService.cs
```diff
@@ -90,13 +90,19 @@ protected async Task<IEnumerable<AvailableCodeAction>> GetAvailableCodeActions(I
...
 
-            return availableActions;
+            return FilterBlacklistedCodeActions(availableActions);
+        }
+
+        private static IEnumerable<AvailableCodeAction> FilterBlacklistedCodeActions(IEnumerable<AvailableCodeAction> codeActions)
+        {
+            // Most of actions with UI works fine with defaults, however there's few exceptions:
+            return codeActions.Where(x =>
+                x.CodeAction.Title != "Generate new type..." &&         // Blacklisted because doesn't give additional value over non UI generate type (when defaults used.)
```
filipw: let's do the blacklisting based on the original Roslyn type name instead of the title text

src/OmniSharp.Roslyn/OmniSharp.Roslyn.csproj
```diff
@@ -16,6 +16,7 @@
     <PackageReference Include="Microsoft.Extensions.Caching.Memory" />
     <PackageReference Include="Microsoft.CodeAnalysis.Common" />
     <PackageReference Include="System.ComponentModel.Composition" />
+    <PackageReference Include="System.Reflection.DispatchProxy" Version="4.5.1" />
```
filipw: please add the reference here https://github.com/OmniSharp/omnisharp-roslyn/blob/master/build/Packages.props and then use it without a version in the csproj

tests/OmniSharp.Roslyn.CSharp.Tests/CodeActionsWithOptionsFacts.cs
```diff
@@ -0,0 +1,323 @@
...
+            //
+            // In case it's exposed by roslyn team these services can be simplified.
+            // Steps are likely following:
+            // - Remove proxies
+            // - Replace ExportWorkspaceServiceFactoryWithAssemblyQualifiedName with Microsoft.CodeAnalysis.Host.Mef.ExportWorkspaceServiceAttribute
+            // - Fix proxy classes to implement IPickMembersService / IExtractInterfaceOptionsService ... instead of proxy and reflection.
+            // - Remove all factories using ExportWorkspaceServiceFactoryWithAssemblyQualifiedName and factory itself.
+            // Following issue may have additional information: https://github.com/dotnet/roslyn/issues/33277
+            var pickMemberServiceType = Assembly.Load("Microsoft.CodeAnalysis.Features").GetType("Microsoft.CodeAnalysis.PickMembers.IPickMembersService");
+            Assert.False(pickMemberServiceType.IsPublic);
```
filipw: this test is a very good idea 👏

tests/OmniSharp.Roslyn.CSharp.Tests/CodeActionsWithOptionsFacts.cs
```diff
@@ -0,0 +1,323 @@
...
+        {
+            Assert.Equal(TrimLines(expected), TrimLines(actual), false, true, true);
+        }
+
+        private static string TrimLines(string source)
+        {
+            return string.Join("", source.Split('\n').Select(s => s.Trim()));
+        }
+
+        private async Task<RunCodeActionResponse> RunRefactoringAsync(string code, string refactoringName, bool wantsChanges = false)
```
filipw: this code looks identical to the code in `CodeActionsV2Facts`.
Could you make a new `AbstractCodeActionsTestFixture`, move that code there and use it in both of these test classes so that it's not duplicated?

savpek:
@filipw review fixes ready. Can you check?

mholo65:
@filipw @david-driscoll can you kick Travis? I don't have enough privileges.

filipw:
> can you kick Travis?

restarted

savpek:
Ready for merge? @filipw @mholo65 🙂

mholo65:
@rchande before we merge, do you have any objections? This is a great addition to code actions!

src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/BaseCodeActionService.cs
```diff
@@ -90,13 +90,19 @@ protected async Task<IEnumerable<AvailableCodeAction>> GetAvailableCodeActions(I
...
 
-            return availableActions;
+            return FilterBlacklistedCodeActions(availableActions);
+        }
+
+        private static IEnumerable<AvailableCodeAction> FilterBlacklistedCodeActions(IEnumerable<AvailableCodeAction> codeActions)
+        {
+            // Most of actions with UI works fine with defaults, however there's few exceptions:
+            return codeActions.Where(x =>
+                x.CodeAction.GetType().Name != "GenerateTypeCodeActionWithOption" &&        // Blacklisted because doesn't give additional value over non UI generate type (when defaults used.)
```
rchande: Can we reduce this to one reflection invocation?

tests/OmniSharp.Roslyn.CSharp.Tests/CodeActionsWithOptionsFacts.cs
rchande: Well, unless they change the API shape before making it public :)

rchande:
@savpek Thanks for the PR! Generally I'm not crazy about adding more reflection but this seems like a reasonable change. 

@mholo65 @filipw If we update Roslyn and something has changed to make one or more of these code actions stop working, what do you propose we do? Seems like we might regress this pretty easily.

mholo65:
@rchande fair point. Until now, we have been lucky with the reflection we are already doing (goto metadata).

As someone with more insights than us, how likely is it to break? Personally, I think it’s ok to break between pre-release versions of Roslyn, but we should make sure it works between stable versions. Also, what’s the likelihood of getting these services public so that we can avoid reflection?

src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/BaseCodeActionService.cs
david-driscoll: I think Ravi was looking at only calling `x.CodeAction.GetType().Name` once instead of three times.

david-driscoll:
One minor nit, otherwise we're good to go.

savpek:
@david-driscoll can you point me to that minor nit, i don't find any comment or anything for it 🙂

src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/BaseCodeActionService.cs
savpek: image current version

david-driscoll:
Looks like you fixed it 🎉

tests/OmniSharp.Roslyn.CSharp.Tests/CodeActionsWithOptionsFacts.cs
filipw: we could add the check for the parameters too, although we don't do it for the current reflected APIs - I think we could handled that in one sweep separately

savpek:
@filipw @david-driscoll is anything holding this PR back at this point? Its bit burder to maintain PRs for extended period of time 🙂

filipw:
sorry for the delay - this is good to go - thanks a lot for your hard work 👍

PR #1410, opened 2019-02-24
Proposal to add roslyn 'tags' to diagnostic response

savpek:
Tags contains metadata from compiler like 'unused', 'telemetry', 'compiler' and so on. Interesting bit at this point is unused tag that can be used to implement 'fadeout' feature for unused code.

Feature like:
image

Does this api change looks good @rchande @filipw @david-driscoll @DustinCampbell ?

Theres multiple options to surface this on model, i think another option is to add "isFaded" flag or something. But personally i like this tag system better since it can support other scenarios as well. And it seems to be common way to deal this since both VSCode and roslyn implement same behavior.

To get actual value from this change this needs #1076. I don't know any CS diagnostics that hides anything, they seems to be IDE analyzers.


filipw:
I like this idea. Calling it tags also makes sense to me, as it's a known concept and already called like that in both Roslyn in VS Code.
I think, however, that it might add some unnecessary overhead to always report all of those [diagnostic tags](http://source.roslyn.io/#Microsoft.CodeAnalysis/Diagnostic/WellKnownDiagnosticTags.cs,12) for everything, especially as most of it doesn't add much value for now. 

I'd suggest we could include the tags property into the API already but start by just reporting `Unnecessary` values for now. We can opt-in more stuff later if needed or there is a use case.

Also, it would be good to add a test, for example here https://github.com/OmniSharp/omnisharp-roslyn/blob/master/tests/OmniSharp.Roslyn.CSharp.Tests/DiagnosticsFacts.cs to make sure we don't break this (since it's not typed and served as a string array it's susceptible to easy breaking).

savpek:
Good points. I added few tests and tag filtering which passes only Unnessery though. However without analyzers the 'happy path' with unnecessary tag is nonexistent. I added placeholder test for it to be implemented once analyzers are available.

One question: Should we add Unnesessary tag for few selected CS warning like unnesessary import? This isn't implemented by roslyn but it gives bit extra to users without analyzers enabled.

SirIntruder:
VSCode march update got a cool new feature to mark code fixes as preferred - preferred action is autoselected in the dropdown and can be triggered with a hotkey. Does roslyn provide a "preferred" tag or something that can be used to that effect?

savpek:
I think there isn't such tag in diagnostics results from Roslyn. However if there's fix all provider available for diagnostic, it could be reasonable assumption that it could work as preferred code action. This is something which could be prototyped to see if it works well 🤔 

Does visual studio impelement some kind of preferred code action feature? Roslyn has most of it's features implemented in visual studio IDE so if it has feature like VSCode then i think it's definetly doable in VScode too.

david-driscoll:
Blocked waiting on #1076

savpek:
This is ready for review now @filipw @david-driscoll @SirIntruder

savpek:
@filipw build ready (appvoyer was very slow for some reason) and passing. Could you merge this? I'd like to continue with https://github.com/OmniSharp/omnisharp-vscode/pull/2873 🙂

filipw:
🍻

PR #1415, opened 2019-02-28
updated changelog to mark 1.32.11 as released

filipw:


filipw:
nothing to review here so I will merge

PR #1411, opened 2019-02-25
Switch to use CodeStructureService in DocumentSymbolHandler.

mholo65:
This PR will get outline view working in VS Code when using LSP.
image

Needs https://github.com/OmniSharp/csharp-language-server-protocol/pull/123 (ping @david-driscoll)

Fixes #1407

PR #1423, opened 2019-03-19
Fix renaming in LSP

mholo65:
Blocked by #1411 

Still random issues when renaming to a shorter name (the change is too long). Seems to be related to the fact that rename changes are not applied in the server.

//cc @david-driscoll


mholo65:
Ready for review! Everything works ok now!

PR #1424, opened 2019-03-19
Use LTS version of .NET Core 2.1 SDK

rchande:
2.1.301 is pretty old

PR #1426, opened 2019-03-20
Publish windows on unix

rchande:
This will allow our Travis build to publish windows builds, thus fixing the issue where "latest" users who need a Windows build don't get it if the appveyor build hasn't completed.


filipw:
tests will still break, you need to include this change https://github.com/OmniSharp/omnisharp-roslyn/pull/1424#issuecomment-474580941

vim.exe.stackdump
```diff
@@ -0,0 +1,19 @@
+Stack trace:
```
filipw: this should be ignored

david-driscoll:
This is good in the interim, I'll be pushing for moving to azure pipelines shortly (now that I have it building successfully!)

PR #1427, opened 2019-03-20
Use BlockingCollection in SharedTextWriter

mholo65:
Did some profiling and noticed a lot of time was spent on `ShareTextWriter.WriteLine` and `SharedTextWriter.WriteLineAsync`.

Noticed the `lock` and figured we could use `SemaphoreSlim` instead.


src/OmniSharp.Host/Services/SharedTextWriter.cs
```diff
@@ -19,26 +19,33 @@ public SharedTextWriter(TextWriter writer)
         protected override void DisposeCore(bool disposing)
         {
             // Finish sending queued output to the writer.
-            _task.Wait();
+            _semaphoreSlim.WaitAsync().GetAwaiter().GetResult();
         }
 
         public void WriteLine(object value)
```
david-driscoll: Can we refactor this method away?

mholo65:
image

src/OmniSharp.Host/Services/SharedTextWriter.cs
david-driscoll: looking at the code it looks like it's called in other async contexts.

mholo65:
Could do some benchmarks to see if there's any actual performance boost.

src/OmniSharp.Host/Services/SharedTextWriter.cs
mholo65: Yeah, it's not used anymore. As we don't ship NuGets, we probably don't care about obsoleting, just remove it from the interface?

david-driscoll:
I just figure if we can avoid blocking a thread at all it would probably be better.

src/OmniSharp.Host/Services/SharedTextWriter.cs
david-driscoll: yeah, I say just nuke it before we start shipping nugets 😄 👍

rchande:
@mholo65 What do you think of implementing this with BlockingCollection<T>? https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.blockingcollection-1?view=netcore-3.0

david-driscoll:
@rchande I'd be fine with that.

@mholo65 lsp already has this with the output handler https://github.com/OmniSharp/csharp-language-server-protocol/blob/7d8567c09c88d27b4b4a2209959f689df8e9419d/src/JsonRpc/OutputHandler.cs

rchande:
Actually, blockingCollection blocks a thread, so this is probably better.

david-driscoll:
A blocking collection will only block it's own thread, not the calling task, assuming you just queue the item up.  For input / output it probably makes more sense to just have the dedicated thread, then we can queue data up for it as fast as we're allowed, but it'll write out at whatever rate it can.

As long as the order is guaranteed (which by default `BlockingCollection` uses `ConcurrentQueue`) I think dumping handling the writing in another thread might make sense.

Thoughts? @mholo65 @rchande  

As it is I'm fine with this change, but it might almost make more sense to refactor it entirely.

mholo65:
@david-driscoll @rchande I think that BlockingCollection is even better, then we'll dump the IO to another thread, and won't block main thread.

mholo65:
@david-driscoll @rchande @filipw I modified this PR to use `BlockingCollection` instead, also no need for the Async version anymore.

filipw:
thanks 👏

PR #1432, opened 2019-03-21
Added nuget package publishing

david-driscoll:


david-driscoll:
Finally this one is working it seems (at least on azure devops)

filipw:
where is the package version coming from? since we rely on beta dependencies (i.e. Roslyn) the packages must be pre release versions too

azure-pipelines.yml
```diff
@@ -155,19 +161,19 @@ jobs:
...
+      - task: DotNetCoreCLI@2
+        displayName: 'Push NuGet packages to myget.org'
+        inputs:
+          command: push
+          nuGetFeedType: external
+          packagesToPush: $(System.ArtifactsDirectory)/nuget/**/*.nupkg;!$(System.ArtifactsDirectory)/nuget/**/*.symbols.nupkg'
+          publishFeedCredentials: myget.org
+        condition: and(succeeded(), or(startsWith(variables['Build.SourceBranch'], 'refs/tags/v'), eq(variables['Build.SourceBranch'], 'refs/heads/master')))
+      - task: DotNetCoreCLI@2
+        displayName: 'Push NuGet packages to nuget.org'
```
filipw: unfortunately at the moment we can't really push to nuget.org because we have dependencies on private feeds https://github.com/OmniSharp/omnisharp-roslyn/blob/master/NuGet.Config

src/OmniSharp.MSBuild/OmniSharp.MSBuild.csproj
```diff
@@ -1,8 +1,10 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net472</TargetFramework>
+    <TargetFrameworks>net472;netcoreapp2.1</TargetFrameworks>
```
filipw: I am also a bit surprised by the appearance of `netcoreapp2.1` target 😀

david-driscoll: Me too, but https://www.nuget.org/packages/Microsoft.Build/ supports it, figured we might as well also.

azure-pipelines.yml
david-driscoll: why not?  They just need to have the private feed to use the package restore... 😈 

Thoughts on what to do?  Just not publish there yet?

david-driscoll:
Package version will come from our release version (if we issue a tag).  There isn't a hard and fast rule that we can't take a dependency on a beta package... it's just recommended.

We could always postfix the packages with `-beta` if you want.  Either way anyone that takes a dependency is going to have that beta dependency whether our version is a beta or not.

filipw:
> Package version will come from our release version (if we issue a tag). There isn't a hard and fast rule that we can't take a dependency on a beta package... it's just recommended.

actually you are right.. I remembered this being an error but [they changed it a while back](https://github.com/NuGet/NuGet.Client/pull/1073). Then it's all fine 👍

azure-pipelines.yml
filipw: I would start by just publishing to myget and not to nuget yet.
I am not so convinced that the API is in the best shape yet so starting with myget only sounds like a good way to "ease in" if you ask me.

PR #1440, opened 2019-03-31
Load project to hold ruleset compilation options

savpek:
Resolves one review issue from original PR #1076 / first two bullets from #1301 and #1458.

This version loads rulesets as part of project instead of joining them to project on every analysis routine.

End user gets:
- Dynamically loaded / modifiable rulesets instead without full restart on omnisharp after every change.
- Reanalyze updated projects.
- CS0000 diagnostics can be controlled by rulesets even when analyzers are not enabled.
- (Very) slightly faster analysis since project isn't updated every time.
- When project is restored -> it is re-analyzed with correct depencies.

Pending
- [x] Tests (fix existing + add new that tests for project loader).
- [x] Tweaks


savpek:
@rchande @filipw @mholo65 @david-driscoll ready for review.

filipw:
Marking as blocked until https://github.com/OmniSharp/omnisharp-roslyn/pull/1474 (so that this one goes first).

savpek:
This also resolves #1458

savpek:
@filipw i think blocked can be removed? I noticed that this also resolves #1458 because this causes re-analyzis when project changes and added test for that case.

Theres one problem that this doesn't solve: analyzers from nuget without omnisharp restart (one added test has to use workaround because of it where restore is made before host start). I will do it in another pr since its likely unreleated with these changes. Added it to #1301

tests/OmniSharp.MSBuild.Tests/ProjectWithAnalyzersTests.cs
```diff
@@ -17,30 +17,57 @@ public ProjectWithAnalyzersTests(ITestOutputHelper output)
...
-                    Assert.NotEmpty(diagnostics.QuickFixes);
 
                     // This should not throw when analyzers are shadow copied.
                     Directory.Delete(Path.Combine(testProject.Directory, "./nugets"), true);
                 }
             }
         }
 
+        [Fact]
+        public async Task LoadProjectWithNewCompilationOptions()
```
filipw: could you rename this test to a better name please?
Imagine if it fails one day, it doesn't say much what it is - it would take a while to grasp that it's about diagnostic options from a ruleset 😀

src/OmniSharp.Roslyn/OmniSharpWorkspace.cs
```diff
@@ -156,6 +157,15 @@ public void TryPromoteMiscellaneousDocumentsToProject(Project project)
             }
         }
 
+        public void UpdateRulesetsForProject(ProjectId projectId, ImmutableDictionary<string, ReportDiagnostic> newRules)
+        {
+            var project = this.CurrentSolution.GetProject(projectId);
+            var existingRules = project.CompilationOptions.SpecificDiagnosticOptions;
+            var distinctRulesWithProjectSpecificRules = newRules.Concat(existingRules.Where(x => !newRules.Keys.Contains(x.Key)));
```
rchande: >var distinctRulesWithProjectSpecificRules = newRules.Concat(existingRules.Where(x => !newRules.Keys.Contains(x.Key))); [](start = 12, length = 118)

Given that this adds both the new rules and the old rules that aren't in the new ones, how does someone *remove* a rule?

tests/OmniSharp.MSBuild.Tests/ProjectWithAnalyzersTests.cs
rchande: >sk LoadProjectWithNewCompilationOptions() [](start = 23, length = 41)

How is this different from the one in ProjectFileInfoTests?

rchande: +1

---
In reply to: [281243133](https://github.com/OmniSharp/omnisharp-roslyn/pull/1440#discussion_r281243133) [](ancestors = 281243133)

rchande:
>     public class ProjectWithAnalyzersTests : AbstractMSBuildTestFixture

Can you add a test for changing the ruleset file?

---
Refers to: tests/OmniSharp.MSBuild.Tests/ProjectWithAnalyzersTests.cs:13 in c02bfbb. [](commit_id = c02bfbbcf6d87bf2c359419fd0ada680c9adfafd, deletion_comment = False)

tests/OmniSharp.MSBuild.Tests/ProjectWithAnalyzersTests.cs
savpek: I think i copied `LoadProjectWithNewCompilationOptions` as test base and forgot to rename it 😄 I will update this to something more meaningfull.

src/OmniSharp.Roslyn/OmniSharpWorkspace.cs
savpek: Valid point: i think i need to find clean version of `SpecificDiagnosticOptions` and use that one as base instead of project own possibly already updated version.

savpek:
> > ```
> > public class ProjectWithAnalyzersTests : AbstractMSBuildTestFixture
> > ```
> 
> Can you add a test for changing the ruleset file?
> 
> Refers to: tests/OmniSharp.MSBuild.Tests/ProjectWithAnalyzersTests.cs:13 in [c02bfbb](https://github.com/OmniSharp/omnisharp-roslyn/commit/c02bfbbcf6d87bf2c359419fd0ada680c9adfafd). [](commit_id = [c02bfbb](https://github.com/OmniSharp/omnisharp-roslyn/commit/c02bfbbcf6d87bf2c359419fd0ada680c9adfafd), deletion_comment = False)

Sure thing, i will write test that updates rules and validate that they get automatically updated after that.

tests/OmniSharp.MSBuild.Tests/ProjectWithAnalyzersTests.cs
savpek: At `ProjectFileInfoTests` tested that csproj is loaded to project info. 

Since then test `LoadProjectWithNewCompilationOptions` was added which tests that rulesets get correctly loaded to project compilation options. I feel that test in `ProjectFileInfoTests` is obsolete duplicate because of that. 

Removed test from `ProjectFileInfoTests`. 

Renamed `LoadProjectWithNewCompilationOptions` -> `WhenProjectIsLoadedThenItContainsCustomRulesetsFromCsproj`.

savpek: Renamed LoadProjectWithNewCompilationOptions -> WhenProjectIsLoadedThenItContainsCustomRulesetsFromCsproj.

src/OmniSharp.Roslyn/OmniSharpWorkspace.cs
savpek: Fixed by https://github.com/OmniSharp/omnisharp-roslyn/pull/1440/commits/a7aa1fd617cee604073eb835b6328c1ce0a1f858

savpek:
Ready for re-review @rchande @filipw 

Current behavior:
updatable-rulesets

filipw:
@savpek if you don't mind, it should be squashed on merge

src/OmniSharp.Roslyn.CSharp/Workers/Diagnostics/CSharpDiagnosticWorkerWithAnalyzers.cs
```diff
@@ -153,46 +150,48 @@ private void QueueForAnalysis(ImmutableArray<DocumentId> documentIds)
...
-            {
-                _currentDiagnosticResults.TryRemove(changeEvent.DocumentId, out _);
+                case WorkspaceChangeKind.DocumentChanged:
+                case WorkspaceChangeKind.DocumentAdded:
+                case WorkspaceChangeKind.DocumentReloaded:
+                case WorkspaceChangeKind.DocumentInfoChanged:
+                    QueueForAnalysis(ImmutableArray.Create(changeEvent.DocumentId));
+                    break;
+                case WorkspaceChangeKind.DocumentRemoved:
+                    _currentDiagnosticResults.TryRemove(changeEvent.DocumentId, out _);
```
filipw: let's log removal failures too

savpek:
> @savpek if you don't mind, it should be squashed on merge

Squashing is fine 🙂

filipw:
@rchande @david-driscoll @mholo65 could you guys please have a look? imho should be good to go

src/OmniSharp.MSBuild/ProjectFile/ProjectFileInfoExtensions.cs
```diff
@@ -27,13 +28,13 @@ public static CSharpCompilationOptions CreateCompilationOptions(this ProjectFile
                 compilationOptions = compilationOptions.WithGeneralDiagnosticOption(ReportDiagnostic.Error);
             }
 
+            compilationOptions = compilationOptions.WithSpecificDiagnosticOptions(projectFileInfo.GetDiagnosticOptions());
+
```
mholo65: Does order make any difference (why was this moved above `NullableContextOptions`)? If not, consider moving this above all the `if`s for better readability.

src/OmniSharp.MSBuild/ProjectManager.cs
```diff
@@ -95,9 +93,8 @@ public ProjectToUpdate(string filePath, bool allowAutoRestore, ProjectIdInfo pro
             _queue = new BufferBlock<ProjectToUpdate>();
             _processLoopCancellation = new CancellationTokenSource();
             _processLoopTask = Task.Run(() => ProcessLoopAsync(_processLoopCancellation.Token));
-            _assemblyLoader = assemblyLoader;
+            _analyzerAssemblyLoader = assemblyLoader;
```
mholo65: Maybe rename `assemblyLoader` to `analyzerAssemblyLoader` while we are renaming.

filipw:
thanks!

filipw:
> Squashing is fine 🙂

forgot to do it 😄

PR #1441, opened 2019-04-01
try to fix issue with pipelines deployment

david-driscoll:


david-driscoll:
I need to bypass this merge, to test against master.  (There are no code changes)

PR #1444, opened 2019-04-03
Use mono 5.18

rchande:


rchande:
@filipw Some of my builds are having an issue similar to the sqlite thing you fixed a while ago. Any idea what we should do?

filipw:
it looks similar indeed, I will have a look

filipw:
although now it seems like you managed to move past it and it's now failing at `GetReferenceNearestTargetFrameworkTask`

filipw:
you still need to bump the Mono version in here https://github.com/OmniSharp/omnisharp-roslyn/blob/master/azure-pipelines.yml#L34
but great job making all the loose pieces work 😅

src/OmniSharp.Stdio.Driver/app.config
```diff
@@ -51,6 +51,10 @@
                 <assemblyIdentity name="System.Composition.TypedParts" publicKeyToken="b03f5f7f11d50a3a" culture="neutral"/>
                 <bindingRedirect oldVersion="0.0.0.0-1.0.31.0" newVersion="1.0.31.0"/>
             </dependentAssembly>
+            <dependentAssembly>
```
filipw: this file (and other binding redirects files) have MsBuild redirects to 15.1.0.0, I think they should be updated too?

rchande:
@filipw It's actually weirder than that. The Azure Devops Mac machines only have mono 5.16. If  I change it to 5.18, they can't find any mono (it actually works by just concatenating your mono version with a mono base path) and they don't have 5.18. I'm inclined to just no-ope the Mac builds in order to get this unblocked...

rchande:
@filipw Actually, I think it's simple enough to just install mono...maybe

src/OmniSharp.Stdio.Driver/app.config
rchande: Tried updating this and binding failed; opened assemblies from msbuild 16 and the version is still 15.1...

filipw: oh.. of course it is... 🤯

PR #1404, opened 2019-02-20
Support for NullableContextOptions in csproj

filipw:
Correctly propagate `NullableContextOptions` from csproj into `CSharpCompilationOptions`
Fixes #1396

PR #1448, opened 2019-04-07
Updated to Roslyn 3.1

filipw:
We can take this in already, or we can wait a bit until there is an official dev16.1 preview, however I wanted to do the necessary changes to make use 3.1 compatible.
There is some reflection that goes away, and a few changes we needed to make to adapt our existing reflection 😀

### DIFF
--- a/src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/RunCodeActionService.cs
+++ b/src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/RunCodeActionService.cs
@@ -171,7 +171,7 @@ namespace OmniSharp.Roslyn.CSharp.Services.Refactoring.V2
                             var newFilePath = GetNewFilePath(newDocument.Name, oldDocument.FilePath);
                             var text = await oldDocument.GetTextAsync();
                             var temp = solution.RemoveDocument(documentId);
-                            solution = temp.AddDocument(documentId, newDocument.Name, text, oldDocument.Folders, newFilePath);
+                            solution = temp.AddDocument(DocumentId.CreateNewId(oldDocument.Project.Id, newDocument.Name), newDocument.Name, text, oldDocument.Folders, newFilePath);
 
                             filePathToResponseMap[filePath] = new RenamedFileResponse(oldDocument.FilePath, newFilePath);
                             filePathToResponseMap[newFilePath] = new OpenFileResponse(newFilePath);
--- a/src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/RunCodeActionService.cs
+++ b/src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/RunCodeActionService.cs
@@ -67,10 +67,10 @@ namespace OmniSharp.Roslyn.CSharp.Services.Refactoring.V2
             {
                 if (o is ApplyChangesOperation applyChangesOperation)
                 {
-                    var fileChanges = await GetFileChangesAsync(applyChangesOperation.ChangedSolution, solution, directory, request.WantsTextChanges, request.WantsAllCodeActionOperations);
+                    var fileChangesResult = await GetFileChangesAsync(applyChangesOperation.ChangedSolution, solution, directory, request.WantsTextChanges, request.WantsAllCodeActionOperations);
 
-                    changes.AddRange(fileChanges);
-                    solution = this.Workspace.CurrentSolution;
+                    changes.AddRange(fileChangesResult.FileChanges);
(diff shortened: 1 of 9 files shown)


src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/RunCodeActionService.cs
```diff
@@ -197,32 +160,57 @@ private async Task<IEnumerable<ModifiedFileResponse>> GetFileChangesAsync(Soluti
                 foreach (var documentId in projectChange.GetChangedDocuments())
                 {
                     var newDocument = newSolution.GetDocument(documentId);
+                    var oldDocument = oldSolution.GetDocument(documentId);
                     var filePath = newDocument.FilePath;
 
-                    if (!filePathToResponseMap.TryGetValue(filePath, out var modifiedFileResponse))
+                    // file rename
```
filipw: `RenameDocumentOperation` doesn't exist in Roslyn 3.1 anymore, renames are normal `ApplyChangesOperation` so I needed to update this code a bit to make this work. On the upside, there is no reflection anymore

src/OmniSharp.Roslyn/MetadataHelper.cs
```diff
@@ -95,7 +95,7 @@ public async Task<(Document metadataDocument, string documentPath)> GetAndAddDoc
                 var service = _csharpMetadataAsSourceService.CreateInstance(temporaryDocument.Project.LanguageServices);
                 var method = _csharpMetadataAsSourceService.GetMethod(AddSourceToAsync);
 
-                var documentTask = method.Invoke<Task<Document>>(service, new object[] { temporaryDocument, topLevelSymbol, default(CancellationToken) });
+                var documentTask = method.Invoke<Task<Document>>(service, new object[] { temporaryDocument, await metadataProject.GetCompilationAsync(), topLevelSymbol, default(CancellationToken) });
```
filipw: breaking change on the internal API

src/OmniSharp.Script/ScriptProjectProvider.cs
```diff
@@ -40,7 +40,7 @@ public class ScriptProjectProvider
             "System.Threading.Tasks"
         };
 
-        private static readonly CSharpParseOptions ParseOptions = new CSharpParseOptions(LanguageVersion.Latest, DocumentationMode.Parse, SourceCodeKind.Script);
+        private static readonly CSharpParseOptions ParseOptions = new CSharpParseOptions(LanguageVersion.CSharp8, DocumentationMode.Parse, SourceCodeKind.Script);
```
filipw: Latest doesn't resolve to C# 8 yet

tests/OmniSharp.Roslyn.CSharp.Tests/CodeActionsV2Facts.cs
```diff
@@ -256,6 +256,8 @@ public async Task Can_send_rename_and_fileOpen_responses_when_codeAction_renames
                 Assert.Equal(2, changes.Length);
                 Assert.Equal(FileModificationType.Renamed, changes[0].ModificationType);
                 Assert.Contains("Class1.cs", ((RenamedFileResponse)changes[0]).NewFileName);
+                Assert.False(File.Exists(((RenamedFileResponse)changes[0]).FileName), "The old renamed file exists - even though it should not.");
```
filipw: this is an assert that was missing and helps in refactoring. In the initial run I had all the tests passing, but the rename code action was actually still broken when running in the editor because of the problems with file lifecycles. This is the assert that would have failed

rchande:
@filipw Can you try this out locally and verify that using the "rename file to match type" refactoring still works? Thanks!

filipw:
> Can you try this out locally and verify that using the "rename file to match type" refactoring still works? Thanks!

yes it does (on my machine ™️). That's also the reason why I added an extra assert in this PR to the rename rename file/open code action test. 
Initially I had the tests green, but the feature didn't work in the editor. This assert would have caught it in the first place.

rchande:
@filipw As an aside, how do you feel about working in forks vs working in branches within omnisharp-roslyn? Right now, our builds are configured so that you have a lot more CI builds to run if you work on omnisharp-roslyn vs in your fork :)

filipw:
thanks for the merge 👍I normally do use the fork (see i.e. https://github.com/OmniSharp/omnisharp-roslyn/pull/1404), I push branches to this repo only if I want to have the CI run *before* I open the PR, just to make sure when I open the PR everything is in order already. but yeah you are right, it is a bit "overhead-y" 😄

PR #1449, opened 2019-04-09
make "dotnet build" work on non-Windows platforms

filipw:
This follows a similar solution from the F# compiler https://github.com/Microsoft/visualfsharp/pull/4391

We now import Mono reference assemblies to allow commands like `dotnet build` or `dotnet test` to work outside of Windows. 
This simplifies and helps development as you can quickly build without having to use the build script and it's bootstrapping features.

At the moment trying to run `dotnet build` on non-Windows against omnisharp-roslyn results in a flood of errors like these:

```
The reference assemblies for framework 
".NETFramework,Version=v4.7.2" were not found. To resolve this, install the
 SDK or Targeting Pack for this framework version or retarget your application to a version of the framework for which you have the SDK or
 Targeting Pack installed. Note that assemblies will be resolved from the
 Global Assembly Cache (GAC) and will be used in place of reference 
assemblies. Therefore your assembly may not be correctly targeted for the 
framework you intend.
```


tests/OmniSharp.Tests/OmniSharp.Tests.csproj
```diff
@@ -6,6 +6,8 @@
     <GenerateRuntimeConfigurationFiles>true</GenerateRuntimeConfigurationFiles>
   </PropertyGroup>
 
+  <Import Project="../../build/netfx.props" />
```
david-driscoll: Can these be brought in via `Directory.build.props` or `Directory.build.targets`?

Seems silly to have to add it to all project files if we could avoid it.

filipw: I tried that but it doesn't get loaded at all then. Seems like `<Import Project="..."  />` doesn't work via those extensibility points.

filipw:
I have removed the original quite "hacky" approach and used the new `Microsoft.NETFramework.ReferenceAssemblies` (see https://github.com/dotnet/designs/pull/33#issuecomment-489264196). This is now a clean approach and should be good to merge.

PR #1451, opened 2019-04-10
Update changelog

rchande:


filipw:
@rchande I added some additional entries, but since I am not allowed to push to your fork, I branched from yours and opened a new PR here #1452. Hope it makes sense - it should be good to merge now.

PR #1454, opened 2019-04-11
Do not hash target frameworks and rename ProjectGuid->ProjectId

akshita31:


akshita31:
On the extension we would like to align it with the VS behavior so we need to call it "ProjectId" instead of "ProjectGuid"

PR #1459, opened 2019-04-14
 added support for TreatWarningsAsErrors in csproj

filipw:
At the moment we do not respect `<TreatWarningsAsErrors>` setting in csproj file and even if build correctly reports errors, we still show warnings only in the editor.
This is a bit ironic, because [OmniSharp itself](https://github.com/OmniSharp/omnisharp-roslyn/blob/master/build/Settings.props#L6) uses `<TreatWarningsAsErrors>true</TreatWarningsAsErrors>` - and we do not catch our own warnings 😀

This PR fixes that.


tests/OmniSharp.MSBuild.Tests/ProjectFileInfoTests.cs
```diff
@@ -57,8 +57,12 @@ public async Task HelloWorld_has_correct_property_values()
                 Assert.Equal("obj/Debug/netcoreapp2.1/", projectFileInfo.IntermediateOutputPath.EnsureForwardSlashes());
```
rchande: Probably would also be good to add a test for the actual diagnostics we report.

PR #1467, opened 2019-04-17
Update gitversion and tweak config to avoid version increments from merge commits

mholo65:
Note, only updating GitVersion solved the issue, but after discussing with @gep13 the changes I made to the config also makes sense. Only updating config didn't work (seems to be a bug in v4 of GitVersion regarding merge-commits).

### DIFF
--- a/GitVersion.yml
+++ b/GitVersion.yml
@@ -1,6 +1,10 @@
 assembly-versioning-scheme: MajorMinor
 mode: ContinuousDeployment
-branches: {}
+branches:
+  master:
+    prevent-increment-of-merged-branch-version: true
+    track-merge-target: false
 ignore:
   sha: []
 continuous-delivery-fallback-tag: beta
+commit-message-incrementing: Disabled
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -11,6 +11,7 @@ variables:
   DotNetVersion: "2.1.505"
   CakeVersion: "0.32.1"
   NuGetVersion: "4.9.2"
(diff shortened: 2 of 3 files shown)


mholo65:
One easy workaround would be to just ignore 4ffabfa8d8279cd5afa17912143cec9612f77020, but that would not prevent this happening again.

filipw:
which version does it now resolve to?

mholo65:
`1.32.19-beta6` and with this commit it should be `beta7` also tested creating a branch named `2.5.4` and then it resolved to `2.5.4`

mholo65:
You can also see from AppVeyor history that it works https://ci.appveyor.com/project/david-driscoll/omnisharp-roslyn/history

filipw:
ok excellent! 
Any idea what caused the 3.5.x here https://dev.azure.com/omnisharp/Builds/_build/results?buildId=319? since some builds resolved to 3.1.x and some to 3.5.x

mholo65:
We run `/diag` on ADO with GitVersion, here's the commit:
```
Update build tools')
  INFO [04/15/19 17:52:48:94] Base version used: Merge message 'Merge pull request #699 from DustinCampbell/update-nuget-to-3.5.0
```

Also noticed we download another GitVersion version as ADO build step, will need to fix that as well.

mholo65:
@filipw you can now see that ADO is also fixed [here](https://dev.azure.com/omnisharp/Builds/_build/results?buildId=327) Check the `Run GitVersion` task in `GitVersion Job`.

rchande:
@mholo65 Thanks a lot!

PR #1470, opened 2019-04-17
only upload from azure pipelines

david-driscoll:
This will only ever upload omnisharp binaries to the storage account from azure pipelines, which will only ever happen when all 3 builds succeed.

cc @rchande @filipw


filipw:
I think it's time to delete travis and appveyor configurations and disable the hooks too.

PR #1471, opened 2019-04-17
ignore shas creating incorrect version numbers

david-driscoll:
@rchande @mholo65 @filipw this should get us back to a sane version number


mholo65:
@david-driscoll I think this isn’t needed anymore since updating GitVersion in https://github.com/OmniSharp/omnisharp-roslyn/pull/1467

filipw:
let's merge this nevertheless then we can be more flexible about gitversion versions if needed

Issue #1458, opened 2019-04-14
Analyzers enabled: re-analyze when project is restored.


savpek:
Restore doesn't trigger any event that cause documents to be re-analyzed. However they should be, now they are not analyzed correctly before something else (like edit) triggers update.

What it looks like in practise after restore + triggering by whitespace:
![analyzers-after-restore-issue](https://user-images.githubusercontent.com/481896/56088564-58d07680-5e8c-11e9-9d78-ed8547061ee0.gif)

Fixing this requires event/message that causes project to be re-analyzed after succesfull restore.

This is blocker to correctly implement integration tests against roslyn-analyzer based things like in https://github.com/OmniSharp/omnisharp-vscode/pull/2873

savpek:
Fixed by #1440

PR #1474, opened 2019-04-19
Shadow copying analyzer assembly loader.

### LINKED ISSUE #1465: Omnisharp locking analyzer DLLs
I can't share the project, but when I dotnet build a .net core/xunit project that I've opened in VS Code, I sometimes get:

```
C:\Program Files\dotnet\sdk\3.0.100-preview3-010431\NuGet.targets(119,5): error : Access to the path 'xunit.analyzers.dll' is denied. [C:\MLS-Monitoring\Monitoring.sln]
```

Killing omnisharp seems to make the build succeed.
cc @savpek

savpek:
To resolve #1465 

Stripped down version from roslyn code base.


savpek:
Ready for review @rchande @filipw @david-driscoll @mholo65

src/OmniSharp.Host/Services/AnalyzerAssemblyLoader.cs
```diff
@@ -0,0 +1,265 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
+using OmniSharp.Utilities;
+
+namespace OmniSharp.Host.Services
+{
+    // This is shadow copying loader. Makes sure that analyzer assemblies are not locked
+    // on disk during analysis.
+    public class AnalyzerAssemblyLoader : IAnalyzerAssemblyLoader
```
filipw: since this is strongly based on the Roslyn implementation, I would recommend to include the original copyright header too https://github.com/dotnet/roslyn/blob/master/src/Compilers/Shared/ShadowCopyAnalyzerAssemblyLoader.cs#L1 with a note that it's derived

filipw:
do we have any known repro steps for a case that could be included as a test?
perhaps something like trying to use an analyzer from a given path and then deleting that path from the test?

savpek:
Only thing comes to mind is to configure nuget restore folder to local folder next of project and test that i can remove/move nuget packages after startup. That should fail if analyzer dll:s are not loaded as shadow copies. Probably `RestorePackagesPath` will do the trick in csproj.

@filipw can you check pr #1440 because it contains test-assets/test-projects/ProjectWithRulesets/, which could be renamed -> test-assets/test-projects/ProjectWithAnalyzers/ -> can be used part of this test + other thing related analyzers when full e2e is needed.

filipw:
yes the test infra there looks good there, and is very welcome.

I'd personally prefer the order of merging - first this PR, and only then the other. 
Would it be a problem for you to pluck that test infra from that other PR and use it here with a custom `RestorePackagesPath` as discussed? thanks a lot

savpek:
Started to implement tests, there's definetly something that i don't understand about analyzer loading from package references yet 😄 They are not loaded as analyzers (item property type doesn't match) but they magically appears anyway. This escalates in tests because they implement something differently and project scoped analyzers doesn't get loaded. It's likely that something has been implemented incorrectly in current version and it's pure luck that it works. I will debug deeper later this week ...

savpek:
image

Test idea seems legit (poc made without shadow copying loader).

Problem behind tests was #1458 -> restore must currently be done before host is started on tests, it should reload analyzers when ever restore is done but currently it doesn't behave that way. I will fix that in another pr and write bit dummier version of test in this where restore is made beforehand.

src/OmniSharp.Host/Services/AnalyzerAssemblyLoader.cs
savpek: Fixed at https://github.com/OmniSharp/omnisharp-roslyn/pull/1474/commits/b33d245f018e14a66b7346d1fd6aa1dde76c33c1

savpek:
Review fixes ready @filipw

filipw:
Any pending thoughts @mholo65 @rchande @david-driscoll?

savpek:
@mholo65 i noticed that project has analyzer refecenes available at workspace -> made it lazy. This was  refactoring, way it originally should be done if i had better knowledge at that time. By making it lazy with cache it moves loading from startup which makes startup faster and makes startup more robust. Original version failed fatally (and silently, i think few of current issues at load up may be even caused by this) if there was something wrong with project analyzers like locked or missing dll. Noticed that behavior when i tried to replicate locked / invalid dlls for assembly loader.

PR #1479, opened 2019-05-01
Update Roslyn build

rchande:


filipw:
The test failures are real.

filipw:
Looks like the root cause is that `Symbols` and `Provider` properties are no longer available on the `CompletionItem`. We used them to find/decode matching symbols here https://github.com/OmniSharp/omnisharp-roslyn/blob/6836fadb9c35a88d4695276d14302336a460b841/src/OmniSharp.Roslyn.CSharp/Services/Intellisense/CompletionItemExtensions.cs#L42-L64

rchande:
Looking...

src/OmniSharp.Roslyn.CSharp/Services/Intellisense/CompletionItemExtensions.cs
```diff
@@ -18,55 +19,54 @@ internal static class CompletionItemExtensions
...
-            if (properties.TryGetValue(Provider, out var provider) && provider == SymbolCompletionProvider)
+            var providerName = (string)_getProviderName.GetValue(completionItem);
+            if (providerName == SymbolCompletionProvider)
             {
                 return recommendedSymbols.Where(x => x.Name == properties[SymbolName] && (int)x.Kind == int.Parse(properties[SymbolKind])).Distinct();
             }
 
             // if the completion provider encoded symbols into Properties, we can return them
-            if (properties.ContainsKey(Symbols))
+            if (properties.ContainsKey(SymbolName))
```
filipw: I thought most (all?) completion items will be created via `CreateWithSymbolId`? Roslyn doesn't encode the symbol into the completion item anymore?

but if that's the case, we can make this return the same as above - `return recommendedSymbols.Where(x => x.Name == properties[SymbolName] && (int)x.Kind == int.Parse(properties[SymbolKind])).Distinct();`

src/OmniSharp.Roslyn.CSharp/Services/Intellisense/CompletionItemExtensions.cs
```diff
@@ -18,56 +19,60 @@ internal static class CompletionItemExtensions
...
                 if (decodedSymbolsTask != null)
                 {
                     return await decodedSymbolsTask;
                 }
             }
 
+            var properties = completionItem.Properties;
+
+            // if the completion provider encoded symbols into Properties, we can return them
+            if (properties.ContainsKey(SymbolName) && properties.ContainsKey(SymbolName))
```
SteamUpdate: I read changelog and saw that line. `properties.ContainsKey(SymbolName)` - why this code is duplicated? Maybe second check must be `properties.ContainsKey(SymbolKind)` ?

filipw: yes, it's a typo thanks for finding this 👍I opened the fix here https://github.com/OmniSharp/omnisharp-roslyn/pull/1491

Issue #1496, opened 2019-05-15
Roslynator.Analyzers 


timjroberts:
The latest build of Omnisharp Roslyn fails to use the referenced `Roslynator.Analyzers` package for a given project, giving the following error repeated multiple times:
```
Could not load signature of Roslynator.CSharp.Analysis.UnconstrainedTypeParameterCheckedForNullAnalyzer:Analyze due to:
Could not load file or assembly 'Roslynator.CSharp, Version=1.0.0.8, Culture=neutral, PublicKeyToken=390be46f77b79f52' or one of its dependencies.
```
This has been working but appears to have begun failing since `1.32.19`. 

**dotnet --info**
```
.NET Core SDK (reflecting any global.json):
 Version:   2.2.203
 Commit:    e5bab63eca

Runtime Environment:
 OS Name:     Mac OS X
 OS Version:  10.13
 OS Platform: Darwin
 RID:         osx.10.13-x64
 Base Path:   /usr/local/share/dotnet/sdk/2.2.203/

Host (useful for support):
  Version: 2.2.4
  Commit:  f95848e524

.NET Core SDKs installed:
  2.1.202 [/usr/local/share/dotnet/sdk]
  2.1.403 [/usr/local/share/dotnet/sdk]
  2.2.101 [/usr/local/share/dotnet/sdk]
  2.2.203 [/usr/local/share/dotnet/sdk]

.NET Core runtimes installed:
  Microsoft.AspNetCore.All 2.1.5 [/usr/local/share/dotnet/shared/Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.2.0 [/usr/local/share/dotnet/shared/Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.2.4 [/usr/local/share/dotnet/shared/Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.App 2.1.5 [/usr/local/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 2.2.0 [/usr/local/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 2.2.4 [/usr/local/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 2.0.9 [/usr/local/share/dotnet/shared/Microsoft.NETCore.App]
[...]

PR #1497, opened 2019-05-15
Fixed mistake on shadow copying assembly loader.

savpek:
Double checked implementation based on original from roslyn and found important bit that was missing, this may fix issue #1496


filipw:
it should fix it I think, the linked issue looks like a bindings redirect issue indeed

filipw:
I have pulled this one locally and tested, and indeed it fixes https://github.com/OmniSharp/omnisharp-roslyn/issues/1496 
since this is code directly from Roslyn, I think it should be safe to merge- thanks

PR #1504, opened 2019-05-23
updated to Roslyn 3.1.0 beta4-19266-03 to match dev16.1

filipw:
see https://github.com/dotnet/versions/tree/master/build-info/dotnet/roslyn/dev16.1


filipw:
I can tell you, after we merged #1449 it's such a pleasure to work on a Mac, once you run build script at least once (so that the necessary assets are pulled down) it's just `dotnet test` to test 😀

PR #1507, opened 2019-05-26
Background analysis support

savpek:
Implements analysis with foreground - background queues.
- Foreground: file you are usually currently editing, meant to be as instant as possible.
- Backgound: optimized for project sized analysis work, meant to analysis project / hole solution in backgound.

Added event of project analyzed which can be used to notify that theres more background results available. VScode current fetch is based on editor actions -> they really don't know when more data is available.

Gives faster feedback during (initial) analysis (omnisharp-roslyn codes during startup):

Old:
before-bg-analysis-speed

New:
with-bg-analysis-speed2

Add re-analyze api for omnisharp that can be used to trigger either project or all scoped analysis:

image

When ever diagnostics are not based "calculate everything for every request (which is expensive if theres anything more than simple semantic diagnostics)" we face following problem: how to know when we need to diagnose other files that may have references to current directly or indirectly, for example in case of manual rename of interface (this is true on current LSP implementation too i think). Adds manual option to work with that problem. Even after some sophisticated logic is added theres very tricky cases like change git branches that can modify things in very crazy way and make things offsync -> this manual option is likely needed anyway.

reanalyze-project

Do anyone know does roslyn expose api to figure those indirect refences out? cc: @rchande @filipw @david-driscoll @mholo65 

Current perf: 
Original diagnostic solution without analyzer get diagnostics from all files every time, for that reason it linearly scales up when solution is bigger. Heres perf comparison with omnisharp code base:
perf-without-analyzers2

With analyzers (no background work running):
perf-with-analyzers

With version (analyzers enabled) that doesn't cause pressure of diagnostic fetch api (updated on vscode side) on every change, instead if is based on project events:
perf-with-analyzers2

With further optimizations analyzer enabled version are getting close to get par with original version.

Perf is currently better with analyzers enabled when project get large. In VScode this is currently worked around by flag that limits fetching all diagnostics if there is over 1000 files (configurable). However with that you cannot get errors from all files. I think similar flag should be added to analyzer too and what it does is that it doesn't trigger initial full solution analysis on startup, however it can be invoked manually -> get good from both worlds when working with larger projects.

Ready for review @rchande @filipw @mholo65 @david-driscoll


loligans:
Is there something that is preventing this PR from going through? It seems like a really hot feature to have.

loligans:
Is there a way I can take this for a test @savpek? I'm using VSCode.

savpek:
You can setup vscode with https://github.com/OmniSharp/omnisharp-vscode/pull/3089 (build vsix plugin from branch and install it to vscode) and configure "omnisharp.path" to omnisharp-roslyn version built from this PR.

One of new tests has been broken in mac for some reason for several builds, have to check it out before this is ready. Otherwise this PR is good to go in my mind, but theres new API and events in this PR so feedback/reviewers are really needed 😄 

Vscode side requires tests, otherwise it is ready i think.

david-driscoll:
@savpek you have a failing test on mac/linux (doesn't happen on windows)

```
OmniSharp.Roslyn.CSharp.Tests.DiagnosticsFacts.CheckAllFilesWithAnalyzersWillEventuallyReturnAllResults [FAIL]
      Assert.Contains() Failure
      Not found: (filter expression)
      In value:  WhereEnumerableIterator<DiagnosticLocation> [Make field readonly (IDE0044) (0:15) - (0:16), Private member 'C2.n' can be removed as the value assigned to it is never read. (IDE0052) (0:15) - (0:16), Accessibility modifiers required (IDE0040) (0:15) - (0:16), Accessibility modifiers required (IDE0040) (0:6) - (0:8), Cannot implicitly convert type 'bool' to 'int' (CS0029) (0:19) - (0:23)]
```

See the azure devops build for more info!

savpek:
@david-driscoll fixed test

src/OmniSharp.Roslyn.CSharp/Workers/Diagnostics/AnalyzerWorkQueue.cs
```diff
@@ -12,98 +12,118 @@ namespace OmniSharp.Roslyn.CSharp.Workers.Diagnostics
 {
     public class AnalyzerWorkQueue
```
david-driscoll: Is this something that a [`BlockingCollection`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.blockingcollection-1) might make a little more straight forward?  

I've used this in the past to deal with spinning and waiting for things https://github.com/OmniSharp/csharp-language-server-protocol/blob/9fa60ffdaf7762ef29f8badea6182f272d0e5979/src/JsonRpc/ProcessScheduler.cs#L73

david-driscoll: The big win is you can avoid using locks because a Blocking collection will handle it for you.

src/OmniSharp.Abstractions/Models/v1/ReAnalyze/ReanalyzeRequest.cs
```diff
@@ -0,0 +1,13 @@
+using OmniSharp.Mef;
+using OmniSharp.Models;
+
+namespace OmniSharp.Abstractions.Models.V1.ReAnalyze
+{
+    [OmniSharpEndpoint(OmniSharpEndpoints.ReAnalyze, typeof(ReAnalyzeRequest), typeof(ReanalyzeResponse))]
+    public class ReAnalyzeRequest: IRequest
```
filipw: shouldn't this just be an empty class inheriting from `SimpleFileRequest`? then the filename will have consistent naming to other endpoints, plus it will get the same directory separator semantics

src/OmniSharp.Roslyn.CSharp/Services/Diagnostics/ReAnalyzeService.cs
```diff
@@ -0,0 +1,81 @@
...
+                currentOpenFilePathAsContext,
+                StringComparison.InvariantCultureIgnoreCase) == 0;
+        }
+
+        private static ImmutableArray<ProjectId> GetDocumentsFromProjectWhereDocumentIs(string currentOpenFilePathAsContext, Solution currentSolution)
+        {
+            return currentSolution
+                .GetDocumentIdsWithFilePath(currentOpenFilePathAsContext)
+                .Select(docId => currentSolution.GetDocument(docId).Project.Id)
+                .ToImmutableArray();
```
filipw: shouldn't you deduplicate here using Distinct()?

src/OmniSharp.Roslyn.CSharp/Workers/Diagnostics/CSharpDiagnosticWorker.cs
```diff
@@ -186,7 +186,14 @@ private static async Task<ImmutableArray<Diagnostic>> GetDiagnosticsForDocument(
 
         public ImmutableArray<DocumentId> QueueAllDocumentsForDiagnostics()
         {
-            var documents = _workspace.CurrentSolution.Projects.SelectMany(x => x.Documents).ToImmutableArray();
+            var documents = _workspace.CurrentSolution.Projects.SelectMany(x => x.Documents);
+            QueueForDiagnosis(documents.Select(x => x.FilePath).ToImmutableArray());
+            return documents.Select(x => x.Id).ToImmutableArray();
+        }
+
+        public ImmutableArray<DocumentId> QueueDocumentsOfProjectsForDiagnostics(ImmutableArray<ProjectId> projectIds)
```
filipw: it's probably nitpicking but I don't think you need to express again in the name what is already expressed in the argument. It could be simply an overload of the other method - `QueueAllDocumentsForDiagnostics(ImmutableArray<ProjectId> projectIds)`

src/OmniSharp.Roslyn.CSharp/Services/Diagnostics/ReAnalyzeService.cs
```diff
@@ -0,0 +1,81 @@
...
+
+        private static bool CompareProjectPath(string currentOpenFilePathAsContext, Project x)
+        {
+            return String.Compare(
+                x.FilePath,
+                currentOpenFilePathAsContext,
+                StringComparison.InvariantCultureIgnoreCase) == 0;
+        }
+
+        private static ImmutableArray<ProjectId> GetDocumentsFromProjectWhereDocumentIs(string currentOpenFilePathAsContext, Solution currentSolution)
```
filipw: this method name seems weird. it says `GetDocumentsFromProjectWhereDocumentIs` whereas in reality it is something more of a `GetProjectIdsFromDocumentFilePaths`

src/OmniSharp.Roslyn.CSharp/Workers/Diagnostics/AnalyzerWorkQueue.cs
```diff
@@ -12,98 +12,118 @@ namespace OmniSharp.Roslyn.CSharp.Workers.Diagnostics
...
+        private readonly Dictionary<AnalyzerWorkType, Queue> _queues = null;
 
+        private readonly ILogger<AnalyzerWorkQueue> _logger;
         private readonly Func<DateTime> _utcNow;
         private readonly int _maximumDelayWhenWaitingForResults;
-        private readonly ILogger<AnalyzerWorkQueue> _logger;
+        private readonly object _queueLock = new Object();
 
-        public AnalyzerWorkQueue(ILoggerFactory loggerFactory, Func<DateTime> utcNow = null, int timeoutForPendingWorkMs = 15*1000)
+        public AnalyzerWorkQueue(ILoggerFactory loggerFactory, Func<DateTime> utcNow = null, int timeoutForPendingWorkMs = 15 * 1000)
```
filipw: should this timeout be configurable in omnisharp.json?

src/OmniSharp.Roslyn.CSharp/Workers/Diagnostics/CSharpDiagnosticWorker.cs
filipw: or both can be called `QueueDocumentsForDiagnostics`

src/OmniSharp.Roslyn.CSharp/Workers/Diagnostics/AnalyzerWorkQueue.cs
savpek: In previous version i think it could, but now in this version it only applies to request for diagnostics from single file. It's basically just failsafe to prevent api hang in this version. However i am not sure is this needed any more at all since theres timeout for file analysis too on worker side which should prevent that situation.

This timeout could be valid to remove altogether. Timeout for single file analysis in worker could indeed be configurable and that way it's implemented only on single location.

savpek:
Thanks for feedback. On trip for week now but i will fix raised issues after that 😄

src/OmniSharp.Roslyn.CSharp/Workers/Diagnostics/AnalyzerWorkQueue.cs
savpek: Tried to implement with BlockinCollection but got to dead end because theres throttling and timeout mechamism (it's assertion after this PR but anyway) that still require locking. Another problematic part was that in large workspaces there may be tens of thousands of files and on those situation cases like promote mechanism needs to loop all files through to find is there match or not in collection, which is reason why hashset is used originally.

RX could be good fit to this since theres time based throttling and producer/consumer patterns combined 🤔 But i could say lets leave it for further refactoring since this PR already simplifies queue mechanism?

savpek:
Added new configuration `DocumentAnalysisTimeoutMs` which controls how long is timeout of single document analysis. Now queue timeout which is assertion after this PR is that * 3. And this gives possiblity to control timeout for analysis instead of hard coded 10 seconds.

savpek:
Ready for re-review @david-driscoll @filipw

src/OmniSharp.Roslyn.CSharp/Workers/Diagnostics/AnalyzerWorkQueue.cs
david-driscoll: Yeah, I'm fine with bringing in RX later to help with that.  I was just wondering it might help with the scenario or not. 👍

PR #1516, opened 2019-06-04
changelog updates for 1.32.20

filipw:
The usual housekeeping


filipw:
> Sorry for not remembering to update this 👍

actually it's easier to do it in one sweep after the release than having conflicts on all PRs so that's no problem

PR #1520, opened 2019-06-05
Fixed declaration name completion regression

filipw:
With one of the recent Roslyn betas, we have (or rather Roslyn has) regressed on declaration name completion - it doesn't work for us anymore at all. 
There was no test for that, so it went unnoticed - the test is added in this PR now.

The root cause is here http://source.roslyn.io/#Microsoft.CodeAnalysis.CSharp.Features/Completion/CompletionProviders/DeclarationNameCompletionProvider.cs,70 Roslyn actually swallows all exceptions for `DeclarationNameCompletionProvider`. With the current beta, Roslyn actually throws as it can't find a reference to `Humanizer`. It is a bug in Roslyn since this package is required at runtime but not listed as dependency of the Roslyn Nuget. 

Adding the package directly restores our old behavior and the declaration name completion works again (I tested in the editor too). I will open this as bug in Roslyn but until it's resolved, we should take in this fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 All changes to the project will be documented in this file.
 
+## [1.32.21] - not yet released
+* Fixed a regression on declaration name completion (PR: [#1520](https://github.com/OmniSharp/omnisharp-roslyn/pull/1520))
+
 ## [1.32.20] - 2019-06-03
 * Added support for `TreatWarningsAsErrors` in csproj files (PR: [#1459](https://github.com/OmniSharp/omnisharp-roslyn/pull/1459))
 * Updated to Roslyn `3.2.0-beta3-19281-01` to match VS dev16.2p2 (PR: [#1511](https://github.com/OmniSharp/omnisharp-roslyn/pull/1511))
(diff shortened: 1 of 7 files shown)


filipw:
> Wait, what? Roslyn has a dependency on Humanizer?

yes, for pluralization https://github.com/dotnet/roslyn/blob/master/src/Features/CSharp/Portable/Completion/CompletionProviders/DeclarationNameCompletionProvider.NameGenerator.cs#L5
I think @rchande even implemented it at some point? 😀

mholo65:
Did you @filipw open an issue in Roslyn so that they’ll list Humanizer as a dependency?

mholo65:
> I think @rchande even implemented it at some point? 😀

Yes, saw that. It was back in 2017. How come we haven’t seen any issue with missing references before? Was it because Roslyn was swallowing exceptions?

filipw:
> Yes, saw that. It was back in 2017. How come we haven’t seen any issue with missing references before? Was it because Roslyn was swallowing exceptions?

it used to work. 
Look at the older version - https://www.nuget.org/packages/Microsoft.CodeAnalysis.CSharp.Features/2.10.0 the Humanizer  dependency is there
However it's missing in the 3.x.x versions and it doesn't work anymore. It doesn't crash completion since Roslyn simply swallows the exception. 

as I wrote in the PR, I'll open a Roslyn bug for this, but for now this gives an immediate fix - and resolves the regression that we suffered when we moved to Roslyn 3.x

rchande:
@filipw Yeah, sounds like they messed up the package authoring somehow. Thanks for working around/filing the bug.

PR #1540, opened 2019-07-02
Enable LSIF indexer in Azure Pipelines

mholo65:
cc @david-driscoll @arjun27

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -104,6 +104,11 @@ jobs:
       - template: ./.pipelines/artifacts.yml
         parameters:
           Artifacts: $(Artifacts)
+      - task: RichCodeNavIndexer@0
+        inputs:
+          serviceConnection: 'omnisharp-rich-code-nav'
+          githubServiceConnection: 'github'
+          languages: csharp
   - job: Release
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -104,6 +104,11 @@ jobs:
       - template: ./.pipelines/artifacts.yml
         parameters:
           Artifacts: $(Artifacts)
+      - task: RichCodeNavIndexer@0
+        inputs:
+          serviceConnection: 'omnisharp-rich-code-nav'
+          githubServiceConnection: 'github'
+          languages: csharp
```
david-driscoll: just had a thought, can we add `continueOnError: true` just below `task:` but before `inputs:`?

That way this can't break builds if it has an issue internally.

arjun27: Yup, that's a great idea. I'll also update the docs to recommend this

mholo65:
Closing. Let's reopen once we figure out the details with authenticating with the LSIF service from a PR build.

arjun27:
Thanks @mholo65, we are working on this fix and I will get back to you.

Issue #1521, opened 2019-06-07
Early request to /highlight endpoint throws NullReferenceException


nickspoons:
I am trying to work out when the server is ready to receive requests from OmniSharp-vim, using the Stdio version of OmniSharp-roslyn.

My latest attempt is to listen to log events for "Queue project ..." followed by "Successfully loaded project ...", and consider the server "loaded" once all "queued" projects are loaded. At this time we request diagnostics and highlighting.

In a tiny demo project this works correctly. However, in a larger project such as OmniSharp-roslyn, each of these requests fails in a different way.

The "/highlight" request receives a `NullReferenceException` in the response "Message":

```
System.NullReferenceException: Object reference not set to an instance of an object.
   at OmniSharp.Roslyn.CSharp.Services.Highlighting.HighlightingService.<Handle>d__1.MoveNext() in C:\code\external\omnisharp-roslyn\src\OmniSharp.Roslyn.CSharp\Services\Highlighting\HighlightingService.cs:line 45
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at OmniSharp.Endpoint.EndpointHandler`2.<HandleSingleRequest>d__18.MoveNext() in C:\code\external\omnisharp-roslyn\src\OmniSharp.Host\Endpoint\EndpointHandler.cs:line 156
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at OmniSharp.Endpoint.EndpointHandler`2.<Process>d__16.MoveNext() in C:\code\external\omnisharp-roslyn\src\OmniSharp.Host\Endpoint\EndpointHandler.cs:line 127
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[...]

PR #1544, opened 2019-07-04
Rewrite project event forwarder.

mholo65:
- The previous implementation had one lock for all events. This implementation only locks for same event.
- The previous implementation had a 500ms delay before getting project information request. Don't think that's necessary anymore. Especially not after removing support for project.json. MSBuild project system will make sure all projects in the queue are processed before returning project information.

Quickly tested, and it felt like large code bases loaded faster.

Here be dragons...

### DIFF
--- a/src/OmniSharp.Roslyn/ProjectEventForwarder.cs
+++ b/src/OmniSharp.Roslyn/ProjectEventForwarder.cs
@@ -9,6 +9,7 @@ using OmniSharp.Models.Events;
 using OmniSharp.Models.ProjectInformation;
 using OmniSharp.Services;
 using System.Threading;
+using OmniSharp.Utilities;
 
 namespace OmniSharp.Roslyn
 {
@@ -17,7 +18,7 @@ namespace OmniSharp.Roslyn
     {
         private readonly OmniSharpWorkspace _workspace;
         private readonly IEventEmitter _emitter;
-        private readonly ConcurrentDictionary<SimpleWorkspaceEvent, object> _eventLocks;
+        private readonly ConcurrentDictionary<SimpleWorkspaceEvent, AsyncLock> _eventLocks;
         private readonly IEnumerable<IProjectSystem> _projectSystems;
 
         [ImportingConstructor]
@@ -29,7 +30,7 @@ namespace OmniSharp.Roslyn
             _projectSystems = projectSystems;
             _workspace = workspace;
             _emitter = emitter;
-            _eventLocks = new ConcurrentDictionary<SimpleWorkspaceEvent, object>();
+            _eventLocks = new ConcurrentDictionary<SimpleWorkspaceEvent, AsyncLock>();
         }
 
         public void Initialize()
@@ -37,7 +38,7 @@ namespace OmniSharp.Roslyn
             _workspace.WorkspaceChanged += OnWorkspaceChanged;
         }
 
-        private void OnWorkspaceChanged(object source, WorkspaceChangeEventArgs args)
+        private async void OnWorkspaceChanged(object source, WorkspaceChangeEventArgs args)
         {
             SimpleWorkspaceEvent workspaceEvent = null;
 
@@ -58,7 +59,7 @@ namespace OmniSharp.Roslyn
             }
 
             var added = false;
-            lock (_eventLocks.GetOrAdd(workspaceEvent, (_) => { added = true; return new object(); }))
+            using (await _eventLocks.GetOrAdd(workspaceEvent, _ => { added = true; return new AsyncLock(); }).LockAsync())
             {
                 // We are already processing a similar event, no need send it again
                 if (added)
@@ -70,9 +71,7 @@ namespace OmniSharp.Roslyn
                         if (workspaceEvent.EventType != EventTypes.ProjectRemoved)
                         {
                             // Project information should be up-to-date so there's no need to wait.
-                            // TODO: Use AsyncLock (or SemaphoreSlim) instead of lock to allow async/await.
-                            payload = GetProjectInformationAsync(workspaceEvent.FileName)
-                                .GetAwaiter().GetResult();
+                            payload = await GetProjectInformationAsync(workspaceEvent.FileName);
                         }
 
                         _emitter.Emit(workspaceEvent.EventType, payload);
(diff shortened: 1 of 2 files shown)


src/OmniSharp.Roslyn/ProjectEventForwarder.cs
```diff
@@ -36,46 +38,47 @@ public void Initialize()
...
-                            await Task.Delay(500);
-
-                            object payload = null;
-                            if (e.EventType != EventTypes.ProjectRemoved)
-                            {
-                                payload = await GetProjectInformationAsync(e.FileName);
-                            }
+                            // Project information should be up-to-date so there's no need to wait.
+                            // TODO: Use AsyncLock (or SemaphoreSlim) instead of lock to allow async/await.
+                            payload = GetProjectInformationAsync(workspaceEvent.FileName)
```
filipw: maybe update to SemaphoreSlim(1) straight away, the changeset is not that big yet

mholo65: Yes, I was just concerned with the disposal of it. When and who should dispose? If one is waiting on the ”outside”, we can’t dispose it on the ”inside”.

david-driscoll: If you use a `SemaphoreSlim` instead of an `object` you can just call `semaphore.Wait()` to hold until it releases and then you don't have to remove the event lock, just call `semaphore.Release(1);`

mholo65: Yes, but `SemaphoreSlim` implements `IDisposable`, thus needs to be disposed, orherwise we’ll get a memory leak here. And I’m not sure it’s safe to dispose it after it’s removed from the dictionary, and released, as there might be another thread waiting on the semaphore above.

mholo65: I think I’ll just copy over the `AsyncLock` implementation from Orleans https://github.com/dotnet/orleans/blob/master/src/Orleans.Core/Async/AsyncLock.cs

mholo65:
@filipw is there a way to know that a project system is enabled but not yet initialized?
https://github.com/OmniSharp/omnisharp-roslyn/pull/1544/files#diff-e6170c657dc5cf9a062216bde7c8deb9R91

I think it would be wise in startup to wait for the project system to initialize before pushing project information.

filipw:
yeah at the moment, no. the `Initialized` flag is false both for disabled and enabled but not yet initialized projects. I guess we could expose another flag, `Enabled`?

nickspoons:
Would these initialisation flags by any chance help with #1521?

mholo65:
@nickspoons yes, `Initialized` flag introduced in #1543 should help with that. Problem is that we should probably wait for `Enabled` project system to get initialized, otherwise we might miss a request, or like in this PR, don’t forward a project event.

src/OmniSharp.Roslyn/ProjectEventForwarder.cs
```diff
@@ -28,55 +30,55 @@ public class ProjectEventForwarder
...
-                    e = new SimpleWorkspaceEvent(args.OldSolution.GetProject(args.ProjectId).FilePath, EventTypes.ProjectRemoved);
+                    workspaceEvent = new SimpleWorkspaceEvent(args.OldSolution.GetProject(args.ProjectId).FilePath, EventTypes.ProjectRemoved);
                     break;
+                default:
+                    return;
             }
 
-            if (e != null)
+            var added = false;
+            using (await _eventLocks.GetOrAdd(workspaceEvent, _ => { added = true; return new AsyncLock(); }).LockAsync())
```
rchande: I think what this code is trying to say is:
* For a given event type, attempt to add a lock over that workspace type
* If my valueCreator is called, I am the one who added the lock
* Because of that I can lock my lock and do my processing

However, per the remarks [here](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.getoradd?view=netcore-3.0):

>  However, the valueFactory delegate is called outside the locks to avoid the problems that can arise from executing unknown code under a lock. Therefore, GetOrAdd is not atomic with regards to all other operations on the ConcurrentDictionary<TKey,TValue> class.

Since a key/value can be inserted by another thread while valueFactory is generating a value, you cannot trust that just because valueFactory executed, its produced value will be inserted into the dictionary and returned. 

So I don't think this is going to have the desired effect.

mholo65:
@rchande you are absolutely correct! Back to double locking (or do we even need locks?). Problem is that project system might be in a completely different state, so that was probably why it was double locking before, and it had a 500ms delay. The 500ms delay should not be needed as getting project info from MSBuild project system empties the processing queue before responding. The lock would then just assure we don’t send multiple events of the same form within a specific time. What are these events used for on VSCode side?

rchande:
@mholo65 C# extension sends these events in response to file change notifications from VS Code. I think we just need to make sure whatever we come up with maintains the same "correctness" we had before. IMO, it might be nice to make some more concrete statements about the gains than "Quickly tested, and it felt like large code bases loaded faster." :)

src/OmniSharp.Roslyn/ProjectEventForwarder.cs
david-driscoll: That's where something like `BlockingCollection` might work better in this scenario, we're just locking because `HashSet` is not thread safe, but in reality we want the data to push out asap, the delay doesn't need to be there for sure.

`OnWorkspaceChanged` adds an item to the queue synchronously.  Then we spin up a dedicated thread that just sits and does `TryTake(out var item, TimeSpan.Infinity)` and just writes out to the emitter.  This lets us queue these operations up, and dequeue them as quickly as possible.

Another option is to use `System.Reactive` and buffer the requests so that we pipe out all events every <500 ms (or whatever)>.

The blocking collection works well for LSP, so it should probably work okay here.

filipw:
what is the status of this? 😀

mholo65:
@filipw 
> 
> what is the status of this? 😀

Probably not going to happen in current form. Without any tests around "how it should work", we'll probably just end up breaking things 😄 I must however say that the `await Task.Delay(500);` in the original code annoys me very much.

filipw:
@mholo65 I will close this PR then, because they start to pile up - please open a new one when you feel like it's ready 😀

PR #1557, opened 2019-07-14
Escape markdown to fix plaintext returns.

### LINKED ISSUE #1556: [LSP] Server is returning plaintext inside MarkedString
I'm seeing this a lot with hover text.  When I hover over a Linq Where clause, the server returns this:

```JSON
{
    "contents": [
        "IEnumerable<KeyValuePair<int, string>> IEnumerable<KeyValuePair<int, string>>.Where<KeyValuePair<int, string>>(Func<KeyValuePair<int, string>, bool> predicate)",
        "Filters a sequence of values based on a predicate.\nsource: An System.Collections.Generic.IEnumerable`1 to filter.\npredicate: A function to test each element for a condition.The type of the elements of source .\nReturns: An System.Collections.Generic.IEnumerable`1 that contains elements from the input sequence that satisfy the condition.\nSystem.ArgumentNullException: source or predicate is null."
    ]
}
```

As you can see, the contents includes ``System.Collections.Generic.IEnumerable`1`` many times.  Back quotes are markdown that inserts inline code blocks.  This causes that text to be rendered like this:

Filters a sequence of values based on a predicate.
source: An System.Collections.Generic.IEnumerable`1 to filter.
predicate: A function to test each element for a condition.The type of the elements of source .
Returns: An System.Collections.Generic.IEnumerable`1 that contains elements from the input sequence that satisfy the condition.
System.ArgumentNullException: source or predicate is null.

The returned text should escape the back quote: ``System.Collections.Generic.IEnumerable\`1``

source: An System.Collections.Generic.IEnumerable\`1 to filter.
predicate: A function to test each element for a condition.The type of the elements of source .
Returns: An System.Collections.Generic.IEnumerable\`1 that contains elements from the input sequence that satisfy the condition.
System.ArgumentNullException: source or predicate is null.

LoneBoco:
Fixes #1556.  Omnisharp is returning plaintext and assigning to MarkedString, which, according to the LSP spec, is a markdown formatted string.  Add a function to escape the text to prevent parsing errors.

### DIFF
--- a/src/OmniSharp.LanguageServerProtocol/Handlers/HoverHandler.cs
+++ b/src/OmniSharp.LanguageServerProtocol/Handlers/HoverHandler.cs
@@ -54,7 +54,7 @@ namespace OmniSharp.LanguageServerProtocol.Handlers
             {
                 // TODO: Range?  We don't currently have that!
                 // Range =
-                Contents = new MarkedStringsOrMarkupContent(new MarkedStringContainer(omnisharpResponse.Type, omnisharpResponse.Documentation))
+                Contents = new MarkedStringsOrMarkupContent(new MarkedStringContainer(Helpers.EscapeMarkdown(omnisharpResponse.Type), Helpers.EscapeMarkdown(omnisharpResponse.Documentation)))
             };
         }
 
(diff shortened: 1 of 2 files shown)


LoneBoco:
I'm not sure if you would be happy with using Regex to fix this, but it should handle all the possible escaped Markdown characters as shown [in the Markdown guide](https://www.markdownguide.org/basic-syntax/#escaping-characters).

david-driscoll:
Just making sure the errors were transient with the build.  Thanks @LoneBoco for the contribution!

PR #1563, opened 2019-07-19
Added support for "sync namespace" refactoring

### LINKED ISSUE #1475: Make new SyncNamespace refactoring work with omnisharp-roslyn
3.x roslyn introduced sync namespace refactoring https://github.com/dotnet/roslyn/issues/14341

However it doesn't show up in omnisharp-roslyn with current version. Reason behind this is project DefaultNamespace that is null, it should be set up during project creation/updates.

Method to make it is currently internal http://source.roslyn.io/#Microsoft.CodeAnalysis.Workspaces/Workspace/Solution/Solution.cs,e596842b39b5f5f9

Heres proof of concept that after invoking that method via reflection it works as expected
image

Changes required for previous setup https://github.com/savpek/omnisharp-roslyn/commit/614d2e03f2ca51a81a2cded8605de0cdbd367ded

Roslyn issue: https://github.com/dotnet/roslyn/issues/35173

filipw:
Fixes #1475 

Demo (first from within a folder, and then from the root of the project):

tjh31xd8tC

We have actually PR-ed into Roslyn a change that opened up the necessary APIs to public https://github.com/dotnet/roslyn/pull/35486 to make it work.
This PR introduces a bump to Roslyn `3.3.0-beta1-19365-07` to have access to them - this is a build for the `dev16.3` release train. 

We also now track "folders" of the documents added to the workspace (the refactoring works off that) - which is basically a relative path.
Default namespace is, by default, the assembly name, unless overwritten by `<RootNamespace>` MsBuild property. A change in MsBuild project file is respected by us in real time (no need to restart OmniSharp), by updating the workspace accordingly. Unfortunately there is still a bug in Roslyn where the default namespace change doesn't propagate correctly via `TryApplyChanges` on the solution, so we have a small reflection workaround for that. I will PR the necessary change into Roslyn and once it's available we can remove the workaround.

cc @savpek

### DIFF
--- a/src/OmniSharp.Roslyn/OmniSharpWorkspace.cs
+++ b/src/OmniSharp.Roslyn/OmniSharpWorkspace.cs
@@ -5,6 +5,7 @@ using System.Collections.Immutable;
 using System.Composition;
 using System.IO;
 using System.Linq;
+using System.Reflection;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
--- a/src/OmniSharp.Roslyn/OmniSharpWorkspace.cs
+++ b/src/OmniSharp.Roslyn/OmniSharpWorkspace.cs
@@ -473,5 +473,23 @@ namespace OmniSharp
         {
             OnAdditionalDocumentRemoved(documentId);
         }
+
+        protected override void ApplyProjectChanges(ProjectChanges projectChanges)
+        {
+            // since Roslyn currently doesn't handle DefaultNamespace changes via ApplyProjectChanges
+            // and OnDefaultNamespaceChanged is internal, we use reflection for now
+            if (projectChanges.NewProject.DefaultNamespace != projectChanges.OldProject.DefaultNamespace)
+            {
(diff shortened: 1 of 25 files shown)


src/OmniSharp.Roslyn/OmniSharpWorkspace.cs
```diff
@@ -427,5 +474,23 @@ public void RemoveAdditionalDocument(DocumentId documentId)
...
+            {
+                var onDefaultNamespaceChanged = this.GetType().GetMethod("OnDefaultNamespaceChanged", BindingFlags.Instance | BindingFlags.NonPublic);
+                if (onDefaultNamespaceChanged != null)
+                {
+                    onDefaultNamespaceChanged.Invoke(this, new object[] { projectChanges.ProjectId, projectChanges.NewProject.DefaultNamespace });
+                }
+            }
+            else
+            {
+                base.ApplyProjectChanges(projectChanges);
```
david-driscoll: Should we call the base `ApplyProjectChanges` all the time or only if the namespace didn't change?  Would this possibly miss some changes if they were batched together with the namespace change?

filipw: thanks @david-driscoll this is a great catch - change accordingly 👍

PR #1566, opened 2019-07-21
Fixed analyzer timeout issue.

kurobon-jp:
Hello.  
I think that the cause of this [timeout issue](https://github.com/OmniSharp/omnisharp-roslyn/issues/1552) is in [GetAnalyzerSemanticDiagnosticsAsync](https://github.com/OmniSharp/omnisharp-roslyn/blob/61d7b82b79058d6250afe4d4db0d9e6bf7ffeac7/src/OmniSharp.Roslyn.CSharp/Workers/Diagnostics/CSharpDiagnosticWorkerWithAnalyzers.cs#L240).  
This function may not return a response.  
I checked the difference with [AnalyzerRunner](https://github.com/dotnet/roslyn/blob/version-3.1.0/src/Tools/AnalyzerRunner/Program.cs#L211) and made a correction.  
And it seems like it worked.

I'm glad if this information helps a little.

### DIFF
--- a/src/OmniSharp.Roslyn.CSharp/Workers/Diagnostics/CSharpDiagnosticWorkerWithAnalyzers.cs
+++ b/src/OmniSharp.Roslyn.CSharp/Workers/Diagnostics/CSharpDiagnosticWorkerWithAnalyzers.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Collections.Immutable;
@@ -235,12 +235,12 @@ namespace OmniSharp.Roslyn.CSharp.Services.Diagnostics
                 }
                 else if (allAnalyzers.Any()) // Analyzers cannot be called with empty analyzer list.
                 {
-                    var semanticDiagnosticsWithAnalyzers = await compiled
-                        .WithAnalyzers(allAnalyzers, workspaceAnalyzerOptions)
+                    var compilationWithAnalyzers = compiled.WithAnalyzers(allAnalyzers, new CompilationWithAnalyzersOptions(workspaceAnalyzerOptions, null, false, false, false));
+
+                    var semanticDiagnosticsWithAnalyzers = await compilationWithAnalyzers
                         .GetAnalyzerSemanticDiagnosticsAsync(documentSemanticModel, filterSpan: null, perDocumentTimeout.Token);
 
-                    var syntaxDiagnosticsWithAnalyzers = await compiled
-                        .WithAnalyzers(allAnalyzers, workspaceAnalyzerOptions)
+                    var syntaxDiagnosticsWithAnalyzers = await compilationWithAnalyzers
                         .GetAnalyzerSyntaxDiagnosticsAsync(documentSemanticModel.SyntaxTree, perDocumentTimeout.Token);
 
                     diagnostics = semanticDiagnosticsWithAnalyzers
--- a/src/OmniSharp.Roslyn.CSharp/Workers/Diagnostics/CSharpDiagnosticWorkerWithAnalyzers.cs
+++ b/src/OmniSharp.Roslyn.CSharp/Workers/Diagnostics/CSharpDiagnosticWorkerWithAnalyzers.cs
@@ -235,7 +235,11 @@ namespace OmniSharp.Roslyn.CSharp.Services.Diagnostics
                 }
                 else if (allAnalyzers.Any()) // Analyzers cannot be called with empty analyzer list.
                 {
-                    var compilationWithAnalyzers = compiled.WithAnalyzers(allAnalyzers, new CompilationWithAnalyzersOptions(workspaceAnalyzerOptions, null, false, false, false));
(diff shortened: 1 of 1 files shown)


JamieBriersSageCom:
I have built this from source, and pointed my VS Code at the .exe.  This appears to have solved the FxCop timeout issues I was facing for what it's worth.

david-driscoll:
This being the constructor, this makes me feel like we might need to configure some of those false values.

```csharp
public CompilationWithAnalyzersOptions (Microsoft.CodeAnalysis.Diagnostics.AnalyzerOptions options, Action<Exception,Microsoft.CodeAnalysis.Diagnostics.DiagnosticAnalyzer,Microsoft.CodeAnalysis.Diagnostic> onAnalyzerException, bool concurrentAnalysis, bool logAnalyzerExecutionTime, bool reportSuppressedDiagnostics);
```

@savpek Do you have any thoughts here?

@filipw also?

src/OmniSharp.Roslyn.CSharp/Workers/Diagnostics/CSharpDiagnosticWorkerWithAnalyzers.cs
```diff
@@ -235,12 +235,12 @@ private async Task AnalyzeDocument(Project project, ImmutableArray<DiagnosticAna
                 }
                 else if (allAnalyzers.Any()) // Analyzers cannot be called with empty analyzer list.
                 {
-                    var semanticDiagnosticsWithAnalyzers = await compiled
-                        .WithAnalyzers(allAnalyzers, workspaceAnalyzerOptions)
+                    var compilationWithAnalyzers = compiled.WithAnalyzers(allAnalyzers, new CompilationWithAnalyzersOptions(workspaceAnalyzerOptions, null, false, false, false));
```
savpek: Sorry for delay

It could be good to add named parameters since its much easier to understand how analyzers are configured with them, like reportSuppressedDiagnostics: false instead of "false" :thinking: 

About error logging, personally i feel that it could be good idea to pass logger function instead of null. Situations what i am thinking are mainly cases like missing services from workspaces etc because there truly are differences between VS and omnisharp in certain cases + configurations may differ which can trigger error situations. That way instead of issue "missing diagnostic X" becomes "missing diagnostic X and heres these errors on logs". Trace could be good place to start as severity since its common to ask trace level logs on issue and then real error is catch up.

filipw:
this is a great fix thanks a lot. I also pulled it locally and it indeed fixes #1552 

Screenshot 2019-07-23 12 03 36

src/OmniSharp.Roslyn.CSharp/Workers/Diagnostics/CSharpDiagnosticWorkerWithAnalyzers.cs
```diff
@@ -261,6 +265,14 @@ private async Task AnalyzeDocument(Project project, ImmutableArray<DiagnosticAna
             }
         }
 
+        private void OnAnalyzerException(Exception ex, DiagnosticAnalyzer analyzer, Diagnostic diagnostic)
+        {
+            _logger.LogError($"Exception in diagnostic analyzer." +
```
filipw: as suggested by @savpek let's change it to lower level i.e. `Debug`.

kurobon-jp:
capture
What is this going on?

filipw:
it was a transient build issue

src/OmniSharp.Roslyn.CSharp/Workers/Diagnostics/CSharpDiagnosticWorkerWithAnalyzers.cs
```diff
@@ -235,12 +235,16 @@ private async Task AnalyzeDocument(Project project, ImmutableArray<DiagnosticAna
                 }
                 else if (allAnalyzers.Any()) // Analyzers cannot be called with empty analyzer list.
                 {
-                    var semanticDiagnosticsWithAnalyzers = await compiled
-                        .WithAnalyzers(allAnalyzers, workspaceAnalyzerOptions)
+                    var compilationWithAnalyzers = compiled.WithAnalyzers(allAnalyzers, new CompilationWithAnalyzersOptions(workspaceAnalyzerOptions,
+                        onAnalyzerException: OnAnalyzerException,
+                        concurrentAnalysis: false,
```
filipw: actually, this is the reason why it works now.
when we use the old code `CompilationWithAnalyzersOptions` gets initialized to the default values and by default it sets `concurrentAnalysis: true`. Looks like then something deadlocks internally and it never completes because of that.

Ultimately I'd like us to be able to recognize why this is happening and how we can avoid that problem but short term this is OK to do it this way as with the current set up it doesn't work at all.

filipw: actually I found an answer myself http://source.roslyn.io/#Microsoft.CodeAnalysis.Features/Diagnostics/AnalyzerHelper.cs,365
looks like VS also switches off concurrency

```
                // in IDE, we always set concurrentAnalysis == false otherwise, we can get into thread starvation due to
                // async being used with syncronous blocking concurrency.
```

filipw:
this should be good to go - I will merge it, thanks @kurobon-jp! 👏

PR #1568, opened 2019-07-23
Unit test support in non-English countries.

kurobon-jp:
Hello.
Some of unit tests don't work in my area.
capture

So we added the function to set the default culture.
However, this may not be the best as not all test classes inherit AbstractTestFixture.
Are there any other good ideas?

### DIFF
--- a/tests/OmniSharp.Script.Tests/WorkspaceInformationTests.cs
+++ b/tests/OmniSharp.Script.Tests/WorkspaceInformationTests.cs
@@ -77,44 +77,35 @@ namespace OmniSharp.Script.Tests
         public async Task CsiScriptWithFileCreatedAfterStartingServer()
         {
             using (var testProject = TestAssets.Instance.GetTestScript("EmptyScript"))
+            using (var host = CreateOmniSharpHost(testProject.Directory))
             {
-                using (var host = CreateOmniSharpHost(testProject.Directory))
-                {
-                    var workspaceInfo = await GetWorkspaceInfoAsync(host);
-                    // ScriptProjectSystem is not initialized here.
-                    // That is workspaceInfo is null.
-                    Assert.Null(workspaceInfo);
-                }
+                var workspaceInfo = await GetWorkspaceInfoAsync(host);
+                Assert.Empty(workspaceInfo.Projects);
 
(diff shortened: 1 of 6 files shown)


filipw:
I think if you added it into the static constructor, it should also work https://github.com/OmniSharp/omnisharp-roslyn/blob/master/tests/TestUtility/SharedOmniSharpHostFixture.cs

kurobon-jp:
Thank you for the advice.

tests/OmniSharp.Script.Tests/WorkspaceInformationTests.cs
```diff
@@ -77,35 +77,44 @@ public async Task SingleCsiScriptWithCustomRspNamespacesAndReferences()
         public async Task CsiScriptWithFileCreatedAfterStartingServer()
         {
             using (var testProject = TestAssets.Instance.GetTestScript("EmptyScript"))
-            using (var host = CreateOmniSharpHost(testProject.Directory))
             {
-                var workspaceInfo = await GetWorkspaceInfoAsync(host);
-                Assert.Empty(workspaceInfo.Projects);
+                using (var host = CreateOmniSharpHost(testProject.Directory))
```
filipw: why is this test changed?

filipw: it actually changes the intent of the test, the intent was:
1) initialize workspace to some directory
2) verify no script files are there
3) create a file *after* the host is running (now with the change, the file is created *before* the host is running)
4) verify that the script file was picked up by a running host

kurobon-jp: In my environment, the original CsiScriptWithFileCreatedAfterStartingServer did not pass.
capture1
capture2

filipw: thanks for finding this. see https://github.com/OmniSharp/omnisharp-roslyn/pull/1570
once that is merged, you can rebase on top of latest master and everything should be fine without any need to change this test anymore.

kurobon-jp: Thank you @filipw.

PR #1570, opened 2019-07-24
 run script tests as part of the build script

filipw:
This is discovered as investigation of https://github.com/OmniSharp/omnisharp-roslyn/pull/1568#discussion_r306826127

Turns out the build script has never executed scripting tests 🙈😂

One test was broken since https://github.com/OmniSharp/omnisharp-roslyn/pull/1543 - but it never showed up in the build script, cause that skipped scripting tests...
This PR adds scripting tests to the builds script and fixes the initialization logic that made the test fail.

### DIFF
--- a/build.json
+++ b/build.json
@@ -23,7 +23,8 @@
     "OmniSharp.Stdio.Tests",
     "OmniSharp.DotNetTest.Tests",
     "OmniSharp.Tests",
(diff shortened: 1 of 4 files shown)


filipw:
I need to investigate the failing test - it passes on my Mac

filipw:
it's fine now, should be good to go!

PR #1571, opened 2019-07-26
Fix invalid xml in XmlDocComment of PackageVersionRequest.cs

dannyvv:


dannyvv:
I had omnisharp under the debugger with break on exceptions and noticed an exception

PR #1573, opened 2019-07-31
renamed NullableContextOptions to Nullable

filipw:
since vs 16.2p1 the msbuild property to enable nullable reference types is called `Nullable` instead of `NullableContextOptions` (this is LDM decision).

see related PRs here: https://github.com/dotnet/roslyn/pull/35516 and here https://github.com/dotnet/docs/pull/12497

issue reported by @shoe-diamente - thanks!


filipw:
resolved conflicts - this should be good go

filipw:
@david-driscoll @mholo65 @rchande can we get this in? at the moment nullable reference types is broken in omnisharp in the latest .net core 3.0 previews

PR #1577, opened 2019-08-01
Enhanced override completion.

kurobon-jp:
Hello.
I am not satisfied with the omnisharp override completion feature.
So i added the funcution to provide method implementations from the override completion list.
I would like to get your feedback on this.
override


filipw:
thanks a lot for the PR - however we will be implementing this differently. 

The way to go forward will be to use completion service from Roslyn fully. Once the completion service is properly consumed on the server side, we'll be able to offer a new general purpose resolve endpoint that will take care of not only override improvements but many other things VS does now like importing extra namespaces on completion etc. This was initiated in #884 and I also have a private branch I've been working on to move this forward.

Thanks again for your suggestion though.

filipw:
cc @david-driscoll

kurobon-jp:
Thank you for telling me.
I hope it will be released soon.

PR #1578, opened 2019-08-05
changelog updates for 1.34.1

filipw:
The existing release notes were not complete.
I will self-merge it, because there is not much to review here 😀


filipw:
FWIW the build failure was transient - I re-run it again and it was green, but the status was not reported back https://dev.azure.com/omnisharp/Builds/_build/results?buildId=723&view=results

Issue #1519, opened 2019-06-04
Make it possible to "auto-fix" documents


RononDex:
Currently, omnisharp is not supporting "auto-fix" functionality to VSCode.

Implementing this functionality would lead to a very efficient way to produce clean code.

One possibility would be to add an action "omnisharp: fix document", have a setting in the vscode user json file like
```
"omnisharp.autofixactions": [
    "RemoveUnusedUsings",
    "SimplifyNames",
   ...
]
```

filipw:
This will be possible once #1526 is merged. We'll then be able to add a fix endpoint that fixes everything according to `.editorconfig` settings.

RononDex:
Perfect, thanks!

savpek:
Currently working on this, may be few weeks before i push PR though.

filipw:
fixed by https://github.com/OmniSharp/omnisharp-roslyn/pull/1581

PR #1581, opened 2019-08-11
Fix all providers support

savpek:
Fix all providers implementation.

Fix one or all issues in document, project and solution scope.

Todo:
- [x] Fixes, tweaks, additional tests etc.
- [x] Using handling seems to require special case, it is very common thing that wanted to be auto fixed so it should be done before initial release.
- [x] Roslynator fixes doesn't work for some reason. Underlaying issue probably affects many custom providers and not only roslynator.
- [x] Fix flaky behavior on fix all issues on document.

Supports:
 - Fix one category (by ID) to fix in document, project and solution scope.
 - Fix all autofixable issues in document scope.

Document scope
fix_onebyone_document

Project scope
fix_one_by_one_project

Solution scope:
solution_scope_fix

Fix all issues on document when saved at vscode:

autofix_on_save


tests/OmniSharp.Roslyn.CSharp.Tests/FixAllFacts.cs
```diff
@@ -48,14 +42,21 @@ internal class C { }
...
-                var changesFromOnlyDocument = response.Changes.Single().Changes.Single();
-
                 AssertUtils.AssertIgnoringIndent(textAfterFix, expectedText);
-                Assert.Equal("\ninternal class C { }\n", changesFromOnlyDocument.NewText);
-                Assert.Equal(0, changesFromOnlyDocument.StartLine);
-                Assert.Equal(0, changesFromOnlyDocument.StartColumn);
-                Assert.Equal(2, changesFromOnlyDocument.EndLine);
-                Assert.Equal(16, changesFromOnlyDocument.EndColumn);
+
+                var internaalClassChange = response.Changes.Single().Changes.Single(x => x.NewText == "internal ");
```
seesharper: small typo internaal => internal

savpek:
Ready for review @filipw @david-driscoll @mholo65  @seesharper

david-driscoll:
@savpek awesome as always.  I do have one question, would it possible to also provide these into the normal quick fix menu (perhaps as a setting)?   I may be the only one, but my muscle memory is to `Ctrl+.` or `Alt+Enter` my way to happiness.  Having it as something opt-in would be wonderful (this could however be done as a separate PR)

src/OmniSharp.Roslyn.CSharp/Workers/Diagnostics/ICsDiagnosticWorker.cs
```diff
@@ -1,13 +1,14 @@
...
 using Microsoft.CodeAnalysis;
+using OmniSharp.Roslyn.CSharp.Services.Diagnostics;
 
 namespace OmniSharp.Roslyn.CSharp.Workers.Diagnostics
 {
     public interface ICsDiagnosticWorker
     {
-        Task<ImmutableArray<(string projectName, Diagnostic diagnostic)>> GetDiagnostics(ImmutableArray<string> documentPaths);
-        Task<ImmutableArray<(string projectName, Diagnostic diagnostic)>> GetAllDiagnosticsAsync();
+        Task<ImmutableArray<DocumentDiagnostics>> GetDiagnostics(ImmutableArray<string> documentPaths);
```
david-driscoll: This is technically a binary breaking change now that we ship nuget packages. ( https://www.nuget.org/packages/OmniSharp.Roslyn.CSharp/ )

@filipw @savpek 
a) Should we pull this into a different interface?
b) Mark this is a breaking change and v2 OmniSharp?

savpek: Ah didn't notice that its published with package 🤔 This feature can be implemented without that change however the current `(string projectName, Diagnostic diagnostic)` will face breaking change sooner or later since it's bad fit for public API in extensibility perspective. Another approach may be to add that another interface (..diag..V2 etc) that has new model. 

Personally i vote for breaking change (2.0.x omnisharp). Since at this point there isn't really that many consumers (?) that have to change but adding another supported interface will add maintenance burden and makes public API of omnisharp more complicated 🙂

savpek:
> @savpek awesome as always. I do have one question, would it possible to also provide these into the normal quick fix menu (perhaps as a setting)? I may be the only one, but my muscle memory is to `Ctrl+.` or `Alt+Enter` my way to happiness. Having it as something opt-in would be wonderful (this could however be done as a separate PR)

I think this can be made on client side with multiple aproaches that pop on my mind (vscode as example): 
- When vscode requests code actions it will also request fix all actions and if there is any -> show them with actions or maybe show only "fix everything in current document" action. This probably must be opt-in because it adds clutter to action list.
- Add hotkey option to vscode to execute fix everything on current document.
- Add hotkey to apply latest code action made to document/project/solution.
- Add opt-in feature where when you fix any code issue from ctrl+. menu it will try to fix all similar issues in document if any automatically.
- Something else?

Current prototype of vscode side has "fix automatically all issues on save document option". However i forgot latest version on laptop at work so i cannot push PR of vscode side this weekend 😄

savpek:
Another way to implement get side is to add "hasFixAll" flag to diagnostic results which could make get fix all actions obsolete 🤔 However consuming it like it currently is as list is harder then. Or maybe both should exist.

src/OmniSharp.Roslyn.CSharp/Workers/Diagnostics/ICsDiagnosticWorker.cs
david-driscoll: I'm fine with it @filipw ?

david-driscoll:
@savpek I'm mainly thinking about something like LSP that won't be calling these new endpoints at all, as there is no mirror for this, having the ability for them show up as code actions would be great.

Surfacing through code actions also allows us to use some of the vscode features around that. Like this extension.
https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker
image

Something to think about.  Waiting on any input on the version thing.  We could
A) Ignore it
B) bump to 2.0

filipw:
> Waiting on any input on the version thing. We could
> A) Ignore it
> B) bump to 2.0

I would ignore it. Yeah, it's not great, but I'd swallow it, instead of moving to 2.0 - otherwise we end up with version 67.0 soon 😀
We didn't bump to 2.0 when we added massive features - analyzers, editorconfig, .net core 3.0 support etc. it would be silly to move because of this one change (that likely has 0 consumers).

I would use the time towards 2.0 to properly design API surface and hide most things (I hid a few things into internal earlier but we should do it on a larger scale).

src/OmniSharp.Roslyn.CSharp/Services/Refactoring/RunFixAllCodeActionService.cs
```diff
@@ -0,0 +1,144 @@
...
+                    if (fixes == null)
+                        continue;
+
+                    var operations = await fixes.GetOperationsAsync(CancellationToken.None);
+
+                    foreach (var o in operations)
+                    {
+                        _logger.LogInformation($"Applying operation {o.ToString()} from fix all with fix provider {singleFixableProviderWithDocument.CodeFixProvider} to workspace document {document.FilePath}.");
+
+                        if (o is ApplyChangesOperation applyChangesOperation)
```
filipw: there is a lot of overlap between what this endpoint does and what the existing run code action does - however the logic there seems a lot more dilligent https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/RunCodeActionService.cs#L100 as it handles added documents too etc. it seems like this code could be reconciled together for better maintenance and correctness?

also, I am not sure the changes should be applied to the workspace here? I think it should be controlled via `ApplyTextChanges` like in the other endpoint

src/OmniSharp.Roslyn.CSharp/Workers/Diagnostics/CSharpDiagnosticWorkerWithAnalyzers.cs
```diff
@@ -183,6 +180,12 @@ private void OnWorkspaceChanged(object sender, WorkspaceChangeEventArgs changeEv
                     var projectDocumentIds = _workspace.CurrentSolution.GetProject(changeEvent.ProjectId).Documents.Select(x => x.Id).ToImmutableArray();
                     QueueForAnalysis(projectDocumentIds, AnalyzerWorkType.Background);
                     break;
+                case WorkspaceChangeKind.SolutionAdded:
```
filipw: this is the type of change that ideally should go in a separate PR - to make this one a bit more manageable. would it be a problem to extract it into a different one? thanks

src/OmniSharp.Roslyn/OmniSharpWorkspace.cs
```diff
@@ -24,14 +24,27 @@ namespace OmniSharp
...
+            get { return isInitialized; }
+            set
+            {
+                if (isInitialized == value) return;
+                isInitialized = value;
+                Initialized(isInitialized);
+            }
+        }
+
+        public event Action<bool> Initialized = delegate { };
```
filipw: let's not make a breaking change here - the old property could stay named the same way and just adopt your new logic, while the event could be named differently

tests/TestUtility/AssertUtils.cs
```diff
@@ -0,0 +1,18 @@
+using System.Linq;
+using Xunit;
+
+namespace TestUtility
+{
+    public static class AssertUtils
```
filipw: good idea 👍

src/OmniSharp.Roslyn.CSharp/Services/Refactoring/RunFixAllCodeActionService.cs
```diff
@@ -0,0 +1,144 @@
...
+                        _fixAllDiagnosticProvider,
+                        CancellationToken.None
+                    );
+
+                    var fixes = await singleFixableProviderWithDocument.FixAllProvider.GetFixAsync(fixAllContext);
+
+                    if (fixes == null)
+                        continue;
+
+                    var operations = await fixes.GetOperationsAsync(CancellationToken.None);
```
filipw: an approach with a request-supplied timeout would probably be better? https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp.Roslyn.CSharp/Services/Navigation/GotoDefinitionService.cs#L78

src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/CachingCodeFixProviderForProjects.cs
```diff
@@ -84,12 +85,15 @@ private ImmutableArray<CodeFixProvider> LoadFrom(Project project)
                         return null;
                     }
                 })
-                .Where(x => x != null)
-                .ToImmutableArray();
+                .Where(x => x != null);
+
+            var builtInCodeFixes = _providers.SelectMany(provider => provider.CodeFixProviders);
```
filipw: this is another example of something that could be handled in a separate PR 🙈

filipw:
thanks @savpek ! this is a great feature 👏
i have some concerns and suggestions, but hopefully we could work towards getting this out soon!

I would also love to be able to take some of the smaller tweaks and improvements in separate PRs so that we can ship them faster, if that's OK with you

src/OmniSharp.Roslyn.CSharp/Workers/Diagnostics/CSharpDiagnosticWorkerWithAnalyzers.cs
savpek: I think other changes except diagnostic api change (it affects many parts) are easy to extract in additional pr:s that can be merged before this change. There was lots of trial and error before this version and some refactorings originate from them even after they ended up as dead ends 😄 I will check what parts can be extracted out from this PR.

src/OmniSharp.Roslyn.CSharp/Services/Refactoring/RunFixAllCodeActionService.cs
savpek: @filipw what WantsAllCodeActionOperations stands for?

src/OmniSharp.Roslyn/OmniSharpWorkspace.cs
savpek: Reverted names: Initialized and OnInitialized (new event)

savpek:
Fixes #1519

src/OmniSharp.Roslyn.CSharp/Services/Refactoring/RunFixAllCodeActionService.cs
savpek: Added timeout parameter

src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/CachingCodeFixProviderForProjects.cs
savpek: Yeah i got bit sidetracked 😄

savpek: Yeah got bit sidetracked 😄

src/OmniSharp.Roslyn.CSharp/Services/Refactoring/RunFixAllCodeActionService.cs
savpek: Refactored fixAll use same code to resolve file changes as code action service.

src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/BaseCodeActionService.cs
```diff
@@ -211,5 +210,153 @@ private IEnumerable<AvailableCodeAction> ConvertToAvailableCodeAction(IEnumerabl
...
+                    var documentsInProject = Workspace.GetDocument(fileName).Project.Documents.Select(x => x.FilePath).ToImmutableArray();
+                    return await diagnostics.GetDiagnostics(documentsInProject);
+                case FixAllScope.Document:
+                    return await diagnostics.GetDiagnostics(ImmutableArray.Create(fileName));
+                default:
+                    throw new NotImplementedException();
+            }
+        }
+
+        protected async Task<(Solution Solution, IEnumerable<FileOperationResponse> FileChanges)> GetFileChangesAsync(Solution newSolution, Solution oldSolution, string directory, bool wantTextChanges, bool wantsAllCodeActionOperations)
```
savpek: Refactored this down from CodeActionService so that FixAllService can use it too.

src/OmniSharp.Roslyn.CSharp/Services/Refactoring/RunFixAllCodeActionService.cs
savpek: Supporting `ApplyTextChanges` seems to be tricky, i think routine that currently generates result must be written ground up to get it work. If i use same kind of solution as in code actions with it batch fixing sometimes works, sometimes not and sometimes it results inconsistent changes. That only occurs when i try "fix all issues in current document", it seems to work ok(ish) when only one type of issue is fixed at time. However fix all issues on save/hotkey is very nice feature to support initially 🤔

I believe reason behind for those problems is that after every type new dianostics are used, however diagnostics are based on workspace state, not solution currently half-updated between routines which messes up things (diagnostics can point incorrect locations at file etc). That is pretty tricky thing to implement correctly. Is it ok if `ApplyTextChanges` support is postponed until someone truly likes to use it? I think i will return with to this thing later on once i have heureka moment how to actually do it correctly. That solution probably also gives possibility to support "fix everything at project" and "fix everything at solution" that had similar kind of issues and didn't suppor them for that reason at first version.

savpek:
> I would also love to be able to take some of the smaller tweaks and improvements in separate PRs so that we can ship them faster, if that's OK with you

Sure thing 🙂 

@filipw I think review fixes are ready now

filipw:
LGTM!

can you please resolve the conflicts?

savpek:
@filipw build and mergefixes done

tests/TestUtility/AssertUtils.cs
```diff
@@ -12,7 +12,7 @@ public static void AssertIgnoringIndent(string expected, string actual)
 
         private static string TrimLines(string source)
         {
-            return string.Join("\n", source.Split('\n').Select(s => s.Trim()));
+            return string.Join("", source.Split('\n').Select(s => s.Trim()));
```
mholo65: Any reason for this change? I'd argue readability is better if line breaks are kept in place.

tests/TestUtility/AbstractCodeActionsTestFixture.cs
```diff
@@ -44,9 +44,9 @@ protected OmniSharpTestHost CreateOmniSharpHost(TestFile[] testFiles, IEnumerabl
         protected async Task<RunCodeActionResponse> RunRefactoringAsync(string code, string refactoringName, bool wantsChanges = false, bool isAnalyzersEnabled = true)
         {
             var refactorings = await FindRefactoringsAsync(code, configurationData: TestHelpers.GetConfigurationDataWithAnalyzerConfig(isAnalyzersEnabled));
-            Assert.Contains(refactoringName, refactorings.Select(a => a.Name));
+            Assert.Contains(refactoringName.ToLower(), refactorings.Select(a => a.Name.ToLower()));
 
-            var identifier = refactorings.First(action => action.Name.Equals(refactoringName)).Identifier;
+            var identifier = refactorings.First(action => action.Name.ToLower().Equals(refactoringName.ToLower())).Identifier;
             return await RunRefactoringsAsync(code, identifier, wantsChanges);
```
mholo65: Use `StringComparision.OrdinalIgnoreCase` instead.

tests/TestUtility/AbstractCodeActionsTestFixture.cs
```diff
@@ -44,9 +44,9 @@ protected OmniSharpTestHost CreateOmniSharpHost(TestFile[] testFiles, IEnumerabl
         protected async Task<RunCodeActionResponse> RunRefactoringAsync(string code, string refactoringName, bool wantsChanges = false, bool isAnalyzersEnabled = true)
         {
             var refactorings = await FindRefactoringsAsync(code, configurationData: TestHelpers.GetConfigurationDataWithAnalyzerConfig(isAnalyzersEnabled));
-            Assert.Contains(refactoringName, refactorings.Select(a => a.Name));
+            Assert.Contains(refactoringName.ToLower(), refactorings.Select(a => a.Name.ToLower()));
```
mholo65: Use StringComparer

tests/OmniSharp.Cake.Tests/CodeActionsV2Facts.cs
```diff
@@ -120,9 +120,9 @@ public void Whatever()
         private async Task<RunCodeActionResponse> RunRefactoringAsync(string code, string refactoringName)
         {
             var refactorings = await FindRefactoringsAsync(code);
-            Assert.Contains(refactoringName, refactorings.Select(a => a.Name));
+            Assert.Contains(refactoringName.ToLower(), refactorings.Select(a => a.Name.ToLower()));
 
-            var identifier = refactorings.First(action => action.Name.Equals(refactoringName)).Identifier;
+            var identifier = refactorings.First(action => action.Name.ToLower().Equals(refactoringName.ToLower())).Identifier;
```
mholo65: Use StringComparision

tests/OmniSharp.Cake.Tests/CodeActionsV2Facts.cs
```diff
@@ -120,9 +120,9 @@ public void Whatever()
         private async Task<RunCodeActionResponse> RunRefactoringAsync(string code, string refactoringName)
         {
             var refactorings = await FindRefactoringsAsync(code);
-            Assert.Contains(refactoringName, refactorings.Select(a => a.Name));
+            Assert.Contains(refactoringName.ToLower(), refactorings.Select(a => a.Name.ToLower()));
```
mholo65: Use StringComparer

tests/TestUtility/AssertUtils.cs
savpek: There were multiple versions of method with same name, however there are some tests that require that new lines are ignored also. I ended up to add another utility method for those cases `AssertIgnoringIndentAndNewlines` which is now used for those few cases where new lines were important.

filipw:
thanks!!

savpek:
@filipw any progress with this? Could be nice to get vscode side implemented after this is merged 😄

loligans:
Curious if this is waiting on anything to merge?

loligans:
@savpek Curious if this is waiting on anything to merge?

david-driscoll:
@savpek I think we just forgot about it... let me get these conflicts resolved and I we can :shipit:

david-driscoll:
The windows test failure has be confused honestly.  I clearly works on mac/linux and it fails somewhere deep in the bowles of roslyn.

david-driscoll:
As always thanks @savpek sorry for the delay!

Issue #1383, opened 2019-01-13
FEATURE PROPOSAL: Navigate massive C# codebases using symbol information provided by Azure DevOps Code Search


dmgonch:
[Sorry for so lengthy description but there are a lot things to unpack here]

**USER EXPERIENCE:**
This proposal makes it easier to navigate massive C# codebases where loading of all projects takes too long or isn't even feasible. This proposal builds on previous work of limiting the number of symbols returned by "Go to Symbol in Workspace" (https://github.com/OmniSharp/omnisharp-roslyn/issues/1243) and loading projects on demand (https://github.com/OmniSharp/omnisharp-roslyn/issues/1316). The experience that all these improvements are seeking to enable is where a developer can open the whole repo folder, with potentially hundreds or thousands of C# projects, and almost immediately be able to navigate the whole codebase by using 'Go To Symbols in Workspace', 'Go To Definition' and 'Find References' operations. As user locates and opens source files, the relevant C# projects are loaded on demand allowing for even richer development experience.

Azure DevOps Services offer free Code Search extension (https://docs.microsoft.com/en-us/azure/devops/project/search/code-search?view=vsts&tabs=new-nav) that turns on indexing of repos and unlocks the ability to query for branch symbol information (API: https://docs.microsoft.com/en-us/rest/api/azure/devops/search/?view=azure-devops-rest-5.0; .NET client: https://www.nuget.org/packages/Microsoft.VisualStudio.Services.Search.Client). 

Here are few examples of Code Search requests, using a clone of omnisharp-roslyn repo code:
- Find symbols with names starting with 'msbuild': https://dev.azure.com/productive-dev/public-projects/_search?text=ext%3Acs%20def%3Amsbuild*&type=code&lp=code-Project&filters=ProjectFilters%7Bpublic-projects%7DRepositoryFilters%7BCodeSearchEnabled%7D&_a=contents&result=DefaultCollection%2Fpublic-projects%2FCodeSearchEnabled%2FGBmaster%2F%2Ftests%2FOmniSharp.Tests%2FFileSystemHelperFacts.cs
[...]

PR #1582, opened 2019-08-11
Fixes bugs when there are multiple handlers are defined for a language for a given request

dmgonch:
OmniSharp supports plugins via “--plugin” command like option. A plugin may expose handlers for operations like “Go to definition” or “Find Symbols”. These changes fix bugs in how handers are invoked in this case and how their results are merged. Unit tests added demonstrate which scenarios are fixed with these changes.


dmgonch:
@david-driscoll, @filipw  appreciate your guys feedback

filipw:
thanks a lot, while I think I understand technically what is going on here, could you provide a real world example of a situation this addresses? it would be helpful 😀

dmgonch:
> thanks a lot, while I think I understand technically what is going on here, could you provide a real world example of a situation this addresses? it would be helpful 😀

Take a look at https://github.com/OmniSharp/omnisharp-roslyn/issues/1383. I'm thinking of a new extension that plugs into OmniSharp and provides additional implementation for few services to allow code navigation while projects are still being loaded (or their load is done lazily when omnisharp.enableMsBuildLoadProjectsOnDemand setting is set). The plugin will inspect Roslyn workspace and provide symbol information only for projects that haven't been loaded yet thus achieving "smart" merging of results.

dmgonch:
@filipw Please let me know if you have any concerns about the change or have a better idea how the scenarios I outlined can be enabled.

dmgonch:
@david-driscoll, @filipw curious what you guys think about the proposed fix

src/OmniSharp.Host/Endpoint/EndpointHandler.cs
```diff
@@ -142,18 +146,75 @@ private Task<object> HandleLanguageRequest(string language, TRequest request, Re
...
             return HandleAllRequest(request, packet);
         }
 
-        private async Task<object> HandleSingleRequest(string language, TRequest request, RequestPacket packet)
+        private async Task<IAggregateResponse> AggregateResponsesFromLanguageHandlers(ExportHandler<TRequest, TResponse>[] handlers, TRequest request)
+        {
+            IAggregateResponse aggregateResponse = null;
+
+            var responses = new List<Task<TResponse>>();
+            foreach (var handler in handlers)
```
filipw: we should probably short circuit here in case there is only 1? so that we don't create an unnecessary list or run Task.WhenAll

filipw:
I don't have anything against, I just worry about adding additional overhead, other than that looks OK to me.
I'll suggest @david-driscoll has a look as he initially bootstrapped that part of the code

src/OmniSharp.Abstractions/Models/v1/GotoDefinition/GotoDefinitionResponse.cs
```diff
@@ -3,13 +3,14 @@
 
 namespace OmniSharp.Models.GotoDefinition
 {
-    public class GotoDefinitionResponse
+    public class GotoDefinitionResponse : ICanBeEmptyResponse
```
filipw: it's too late to solve this one, but fundamentally I believe the mistake was returning an uninitialized object in case no match was made. we should have used better semantics to express this but it would be a breaking change now

dmgonch: @filipw Hmm, could you please clarify the scenario where you think this change would be cause a break?

src/OmniSharp.Abstractions/Models/ICanBeEmptyResponse.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace OmniSharp.Models
+{
+    public interface ICanBeEmptyResponse
```
david-driscoll: Could we mark this interface as `internal`? The outside world doesn't need to know about this (technically). This way we can avoid a major breaking change. (well kind of anyway)

src/OmniSharp.Abstractions/Models/v1/GotoDefinition/GotoDefinitionResponse.cs
david-driscoll: We could make `IsEmpty` internal for now (along with it's interface noted above) and work with the interface.

david-driscoll: @dmgonch keep in mind we now publish nuget packages for Abstractions, hence we have to be careful what changes we make.

david-driscoll: Now that I think about it... technically speaking GotoDefinition could return mutliple results, this happens as has been adapted into LSP iirc.  This may be something we should consider making first class (returning a single definition or a list of definitions)

Reference: https://microsoft.github.io/language-server-protocol/specification#textDocument_definition

src/OmniSharp.Abstractions/Models/ICanBeEmptyResponse.cs
dmgonch: Made ICanBeEmptyResponse interface internal.

src/OmniSharp.Abstractions/Models/v1/GotoDefinition/GotoDefinitionResponse.cs
dmgonch: I'm assuming you wouldn't want to bundle this change with this PR. But I made ICanBeEmptyResponse interface internal as per the other comment.

src/OmniSharp.Host/Endpoint/EndpointHandler.cs
dmgonch: Added special handling for the case when there is only a single handler for Endpoint/language combination. Interestingly enough I had to add a cast to IAggregateResponse while else case below doesn't need it.

filipw:
yeah looks like the merge of latest master didn't trigger the build correctly, can you try to rebase this on top of master and force push instead?

dmgonch:
@filipw @david-driscoll Folks, please advise if anything is needed on my side before these changes can be merged.

filipw:
nothing - looks good to me  👍

dmgonch:
@filipw @david-driscoll Unless you guys see anything else that needs to be done for this change, I appreciate if you guys could merge it into the master.

dmgonch:
@filipw @david-driscoll Appreciate if you guys could comment.

filipw:
thanks, as I said, it looks good to me and I gave my approval already 😀
 however, I'd like @david-driscoll to have a second look since he wrote that code originally - hope this makes sense

dmgonch:
@david-driscoll Do you see anything else that needs to be done for this change? Thanks!

dmgonch:
@david-driscoll Appreciate if you could comment. Thanks!

dmgonch:
@filipw Do you think you will be comfortable with pulling these changes into master? @david-driscoll I hope you don't mind - but I would like to move forward with these changes unless you see more issues to address. Appreciate your help folks.

david-driscoll:
@dmgonch sorry for the delay I just saw something ✨ shiny ✨ .  Updating the branch and will check back in today.  It looks like all my concerns were hit.

david-driscoll:
@dmgonch Thanks for the contribution!  Sorry for the delay however. 💯

PR #1583, opened 2019-08-12
test for #1566 

filipw:
Adds a test for #1566 
This test fails without the changes from that PR and passes now. Will help us not regress on this.


filipw:
there is some weird stuff happening in the CI - it passes locally for me on both Windows and Mac, need to have a look

filipw:
should be ok now 😅

PR #1539, opened 2019-07-02
End of life for .NET Core 1.0 / 1.1 (project.json)

filipw:
.NET Core 1.0/1.1 reached end of life and went out of support on [27 June 2019](https://devblogs.microsoft.com/dotnet/net-core-1-0-and-1-1-will-reach-end-of-life-on-june-27-2019/).

We discussed this a while back and we have disabled the `project.json` features by default already last summer (they were opt-in), and we said we will wait with complete removal until EOL is reached.
I think now is a good time to pull the trigger.

This PR:
 - removes `project.json` project system support
 - removes `project.json` test manager support
 - removes `project.json` test assets
 - adjusts the build script accordingly

@rchande @akshita31 if you have any concerns based one the telemetry of the C# extension for VS Code please shout 😀

PR #1584, opened 2019-08-13
retarget test projects to .NET Core 2.1 (LTS)

filipw:
We dropped .NET Core 1.x support a while ago (#1539) however some of the test projects were still targeting 1.x (!).
The result was the test suite was testing something we didn't support 😂

This PR moves test projects to 2.1 LTS and updates the build accordingly (build should be faster now too).

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -118,14 +118,6 @@ Task("InstallDotNetCoreSdk")
             version: buildPlan.DotNetVersion,
             installFolder: env.Folders.DotNetSdk);
 
-        foreach (var runtimeVersion in buildPlan.DotNetSharedRuntimeVersions)
-        {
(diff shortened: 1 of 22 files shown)


filipw:
@mholo65 I am not too familiar with that, how do I generate Cake file for different .NET Core version then?

mholo65:
@filipw today it includes manual work to get those generated. The changes you made actually comes from the documentation of the Cake aliases (which the builds the DSL). This is where it came from: https://github.com/cake-build/cake/blob/2669dac25c388967f78c3536228c34a0e9b01a94/src/Cake.Common/Tools/DotNetCore/DotNetCoreAliases.cs#L238

Noticed that these are over 2 years old. However the change to those files should not matter test execution whatsoever.

PR #1587, opened 2019-08-18
 added support for CheckForOverflowUnderflow in csproj

filipw:
We now respect `<CheckForOverflowUnderflow>` setting when constructing compilation options.
This is equivalent to passing `/checked` switch to csc.


filipw:
resolved conflicts - should be good to go

PR #1596, opened 2019-09-08
updated to Roslyn 3.3.1 beta3

filipw:
We are always very early adopters of Roslyn but this time we are a bit late 3.3.1 beta3 is already even on public Nuget feed.


filipw:
superseded by #1597

PR #1607, opened 2019-09-25
copy MsBuild 16.3 on Mono too

filipw:
This is a follow up to #1606 

We now copy the MsBuild 16.3 from Nuget for use on Mono, rather than the custom package we had in the past.
I tested this on my Mac and seems to have resolved the problem of .NET Core 3.0 not working.


filipw:
ugh looks like it causes some builds to hang - need to investigate.. will close this until it's resolved

PR #1609, opened 2019-09-25
Bump Dotnet.Script.DependencyModel to version 0.50.0

seesharper:
This PR bumps the `Dotnet.Script.DependencyModel` and `Dotnet.Script.DependencyModel.NuGet` packages to version `0.50.0` adding support for .Net Core 3.0 based scripts.

### DIFF
--- a/build/Packages.props
+++ b/build/Packages.props
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Project ToolsVersion="14.0"
+  xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
   <PropertyGroup>
     <MSBuildPackageVersion>16.0.461</MSBuildPackageVersion>
@@ -11,8 +12,8 @@
   <ItemGroup>
     <PackageReference Update="Cake.Scripting.Transport" Version="0.3.0" />
 
-    <PackageReference Update="Dotnet.Script.DependencyModel" Version="0.6.1" />
-    <PackageReference Update="Dotnet.Script.DependencyModel.NuGet" Version="0.6.2" />
+    <PackageReference Update="Dotnet.Script.DependencyModel" Version="0.50.0" />
+    <PackageReference Update="Dotnet.Script.DependencyModel.NuGet" Version="0.50.0" />
 
     <PackageReference Update="Humanizer" Version="2.2.0" />
     <PackageReference Update="McMaster.Extensions.CommandLineUtils" Version="2.2.4" />
--- a/build/Packages.props
+++ b/build/Packages.props
@@ -1,90 +1,90 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="14.0"
-    xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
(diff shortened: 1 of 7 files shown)


build/Packages.props
```diff
@@ -1,89 +1,90 @@
...
-
-    <PackageReference Update="System.Reflection.DispatchProxy" Version="4.5.1" />
-
-    <PackageReference Update="xunit.runner.visualstudio" Version="$(XunitPackageVersion)" />
-    <PackageReference Update="xunit" Version="$(XunitPackageVersion)" />
-  </ItemGroup>
-  <ItemGroup>
-    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0-beta2-18618-05" PrivateAssets="All" />
-  </ItemGroup>
+<Project ToolsVersion="14.0"
```
filipw: could you please reset the changes to this file so that only the relevant lines are updated? we often look here at the blame to identify when a specific dependency was introduced - and this will break it

src/OmniSharp.Script/ScriptContextProvider.cs
```diff
@@ -66,8 +69,9 @@ public ScriptContext CreateScriptContext(ScriptOptions scriptOptions)
             CompilationDependency[] compilationDependencies = null;
             try
             {
+                var allCsxFiles = _fileSystemHelper.GetFiles("**/*.csx").ToArray();
```
filipw: the same files are already discovered in ScriptProjectSystem; since this is potentially an expensive operation at start up, we should not do it twice. can we pass a list of files here instead?

src/OmniSharp.Script/ScriptProjectSystem.cs
```diff
@@ -54,15 +54,18 @@ public void Initalize(IConfiguration configuration)
         {
             if (Initialized) return;
 
+            _logger.LogInformation($"Detecting CSX files in '{_env.TargetDirectory}'.");
+
+            // Nothing to do if there are no CSX files
+            var allCsxFiles = _fileSystemHelper.GetFiles("**/*.csx").ToArray();
+
             _scriptOptions = new ScriptOptions();
+            _scriptOptions.CsxFiles = allCsxFiles;
```
filipw: actually it would be better to just add an extra argument to `CreateScriptContext(...)`.
`ScriptOptions` is bound from the `omnisharp.json` so it's not good idea to use it for just transporting objects or data from one place to another

seesharper:
. Net Core 3.0 compilation deps needs more work

filipw:
@mholo65 any thoughts here?
@david-driscoll you OK with bumping Nuget deps to 5.2.0?

src/OmniSharp.Script/ScriptContextProvider.cs
```diff
@@ -92,7 +111,13 @@ public ScriptContext CreateScriptContext(ScriptOptions scriptOptions)
                 isDesktopClr = false;
                 HashSet<string> loadedFiles = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
 
-                foreach (var compilationAssembly in compilationDependencies.SelectMany(cd => cd.AssemblyPaths).Distinct())
+                // Pick the highest version
+                var resolvedAssemblyPaths = compilationDependencies.SelectMany(cd => cd.AssemblyPaths)
```
filipw: "one liner" 😂

david-driscoll:
@filipw yeah I'm fine with nuget 5.2, we need to stay more up to date on these things anyway.

Issue #1569, opened 2019-07-24
Remove AppVeyor and Travis integrations


filipw:
Since we now use Azure DevOps for everything, the checks on AppVeyor and Travis seem redundant and can be removed. Especially on Travis the builds are very unstable and it all adds maintenance overhead.

cc @david-driscoll

david-driscoll:
I have disabled the integrations and removed them as required for pushing to master.

filipw:
awesome, thanks!

david-driscoll:
This is done... and hopefully we don't run into any problems.  Worst case we can turn them on as I didn't remove the config files yet.

filipw:
this is resolved - thanks!

PR #1612, opened 2019-09-26
use msbuild from Mono 6.4.0 on linux/macos

filipw:
This is a follow up to #1606

We now copy the MsBuild 16.3 from Mono 6.4.0, rather than the custom package we had in the past.

### DIFF
--- a/build.json
+++ b/build.json
@@ -2,7 +2,7 @@
   "DotNetInstallScriptURL": "https://dot.net/v1",
   "DotNetChannel": "preview",
   "DotNetVersion": "2.1.505",
-  "RequiredMonoVersion": "5.18.0.0",
+  "RequiredMonoVersion": "6.0.0",
   "DownloadURL": "https://roslynomnisharp.blob.core.windows.net/ext",
   "MonoRuntimeMacOS": "mono.macOS-5.18.1.0.zip",
   "MonoRuntimeLinux32": "mono.linux-x86-5.18.1.0.zip",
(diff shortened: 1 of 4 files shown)


build.json
```diff
@@ -2,7 +2,7 @@
   "DotNetInstallScriptURL": "https://dot.net/v1",
   "DotNetChannel": "preview",
   "DotNetVersion": "2.1.505",
-  "RequiredMonoVersion": "5.18.0.0",
+  "RequiredMonoVersion": "6.0.0",
```
filipw: a short explanation why this isn't 6.4.0 here - there is no 6.4.0 for travis yet (unless we customize the build and install manually which I want to avoid) so the build would fail.

It doesn't really matter since on ADO we explicitly set 6.4.0 and build with that version, and this is where the real "publishing" happens. Plus we have #1569 pending and at that point we can stop worrying about travis too.

As soon as #1569 is done or travis has 6.4.0, we can raise the version in this file too.

david-driscoll: So what you're saying is you want me to disconnect travis and appveyor.

filipw: exactly, I think it adds maintenance overhead unnecessarily - and we really can get (and we do get) everything done via Azure DevOps already anyway

filipw:
> This may increase the size of the output assets

we only copy core MSBuild libs (no MSBuild runtime) here (which we already did anyway, it's just we take it from Mono 6.4.0 now) so the size is more or less the same

build.json
david-driscoll: > Assuming this passes CI :shipit:
> 
> This may increase the size of the output assets, so we should review with that @rchande and @akshita31 to make sure it's too much larger when consumed from the vscode plugin.

I checked the outputs from the current master and this change... and the was no real large change in overall zip size.  It looked to be about 50KB at most, so

PR #1615, opened 2019-09-29
Support plugins configuration in omnisharp.json

dmgonch:
Locations for OminSharp’s plugins can currently be configured only via –plugin command line argument. But when OminSharp is activated as a part of ms-vscode.csharp extension, other VSCode extensions cannot use this approach to supply their own plugins for OmniSharp. The only exception are code actions which can be configured in omnisharp.json via RoslynExtensionsOptions. The proposed changes enable OmniSharp to discover any plugin using the same technique via new Plugins section in omnisharp.json, i.e.:

    "Plugins": {
        "LocationPaths": [ "path_to_plugin1.dll", "path_to_plugin2.dll" ]
    }

An example of VSCode extension that provides plugins via omnisharp.json is Roslynator which does it via RoslynExtensionsOptions.


filipw:
thanks - overall this is a good idea and a welcome addition.
I am not convinced that this is the best implementation though. I'd rather see types under `OmniSharp.Options` to be used in a consistent, strongly typed way rather than using the raw `IConfigurationRoot` to bind to them multiple times in multiple places.

This would mean that we need to refactor the code a bit to perhaps delay the decision about plugin discovery or move the options bootstrapping up.
I will have a closer look when I have some time.

david-driscoll:
I'm with @filipw I like the idea but the solution feels wrong somehow.  I'm going to have to circle back to this, this weekend after my local code camp.

dmgonch:
Good feedback! Switched to using strongly typed object for declaring locations for plugins.

dmgonch:
@filipw @david-driscoll Curious what you guys think of the latest changes.

filipw:
thanks a lot - it will be reviewed carefully as soon as there is some free time.
but after https://github.com/OmniSharp/omnisharp-roslyn/pull/1625 we don't want to rush any changes without thoroughly testing and pulling locally.

src/OmniSharp.Shared/Options/RoslynExtensionsOptions.cs
```diff
@@ -5,10 +5,14 @@
...
 {
-    public class RoslynExtensionsOptions
+    public class RoslynExtensionsOptions : OmniSharpExtensionsOptions
     {
         public bool EnableAnalyzersSupport { get; set; }
         public int DocumentAnalysisTimeoutMs { get; set; } = 10 * 1000;
+    }
+
+    public class OmniSharpExtensionsOptions
+    {
```
TheCubicNoobik: Remove the excessive blank line

src/OmniSharp.Http/Startup.cs
```diff
@@ -45,8 +52,15 @@ public IServiceProvider ConfigureServices(IServiceCollection services)
                                 !category.Equals(projectEventForwarder, StringComparison.OrdinalIgnoreCase)));
                 });
 
+            var options = serviceProvider.GetRequiredService<IOptionsMonitor<OmniSharpOptions>>();
+            var plugins = new PluginAssemblies(_commandLinePlugins.Concat(options.CurrentValue.Plugins.GetNormalizedLocationPaths(_environment)));
```
TheCubicNoobik: This line is too long

src/OmniSharp.Http/Startup.cs
```diff
@@ -45,8 +52,15 @@ public IServiceProvider ConfigureServices(IServiceCollection services)
...
+
+            var loggerFactory = serviceProvider.GetRequiredService<ILoggerFactory>();
+            var logger = loggerFactory.CreateLogger<Startup>();
+            var assemblyLoader = serviceProvider.GetRequiredService<IAssemblyLoader>();
             _compositionHost = new CompositionHostBuilder(serviceProvider)
                 .WithOmniSharpAssemblies()
+                .WithAssemblies(assemblyLoader.LoadByAssemblyNameOrPath(logger, plugins.AssemblyNames).ToArray())
```
TheCubicNoobik: `
var assemblies = assemblyLoader.LoadByAssemblyNameOrPath(logger, plugins.AssemblyNames).ToArray();
`
Then:
`
    .WithAssemblies(assemblies);
`

dmgonch: Mimicked existing order for the statements as in src\OmniSharp.LanguageServerProtocol\LanguageServerHost.cs and src\OmniSharp.Stdio.Driver\Program.cs

src/OmniSharp.Shared/Options/RoslynExtensionsOptions.cs
dmgonch: removed

src/OmniSharp.Http/Startup.cs
dmgonch: I would like to push back in this one. The line's length AFAICS is comparable or even shorter many other lines in the same file and in the repo and breaking in up IMO will hurt readability.

---
In reply to: [341832182](https://github.com/OmniSharp/omnisharp-roslyn/pull/1615#discussion_r341832182) [](ancestors = 341832182)

dmgonch:
@filipw a gentle ping since it's been few weeks. thanks

filipw:
sorry! I will do my best to review in the coming days! 🙏

filipw:
any thoughts? @david-driscoll @mholo65 @JoeRobich 

if not IMHO this is good to merge and a very valuable feature

src/OmniSharp.Abstractions/Services/IAssemblyLoader.cs
```diff
@@ -54,11 +55,22 @@ public static IEnumerable<Assembly> Load(this IAssemblyLoader loader, params str
             }
         }
 
-        public static IEnumerable<Assembly> LoadByAssemblyNameOrPath(this IAssemblyLoader loader, IEnumerable<string>  assemblyNames)
+        public static IEnumerable<Assembly> LoadByAssemblyNameOrPath(this IAssemblyLoader loader, ILogger logger, IEnumerable<string>  assemblyNames)
```
mholo65: Can we inject a `LoggerFactory` into `AssemvlyLoader` and create logger and log there instead? Feels wrong to pass the logger around.

dmgonch: @mholo65 To clarify, are you proposing to pass ILoggerFactory instead of ILogger into LoadByAssemblyNameOrPath?

src/OmniSharp.Http/Host.cs
```diff
@@ -44,7 +45,7 @@ public void Start()
                     serviceCollection.AddSingleton(_environment);
                     serviceCollection.AddSingleton(_sharedTextWriter);
                     serviceCollection.AddSingleton(NullEventEmitter.Instance);
-                    serviceCollection.AddSingleton(_pluginAssemblies);
+                    serviceCollection.AddSingleton(_commandLinePlugins);
```
JoeRobich: Registering an `IEnumerable<string>` as a service seems like a code smell to me.

dmgonch: @JoeRobich , are you proposing having a dedicated class containing the list?

JoeRobich: I think registering a `PluginAssemblies` here to represent the Plugins the environment was initialized with would make sense.

dmgonch: @JoeRobich Reworked this part - please take another look. Thanks!

src/OmniSharp.Abstractions/Services/IAssemblyLoader.cs
dmgonch: @mholo65 Switched to passing in ILoggerFactory - please take another look. Thanks!

mholo65: No, what I meant was injecting `ILoggerFactory` to `AssemblyLoader`. But it seems like we are already doing this (see [AssemblyLoader.cs](https://github.com/OmniSharp/omnisharp-roslyn/blob/a3e7dbf7e8b6cc14e0c43879c53d940e5910a892/src/OmniSharp.Host/Services/AssemblyLoader.cs)). Why do we need additional logging in the extension method, isn't the logging in the implementation enough? It looks like we already catching and logging all exceptions. So this `catch` will never hit.

dmgonch: @mholo65 Below is the callstack where the exception gets thrown if plugin assembly cannot be loaded. With the way the code is structured regardless of my change, the plugins loading code path is going through extension methods, not throw the actual AssemblyLoader implementation. Because added exception catching and logging attempts to address the existing problem with plugins loading I can leave it out of this change. Please advise.

System.IO.FileLoadException
  HResult=0x80131047
  Message=The given assembly name or codebase was invalid. (Exception from HRESULT: 0x80131047)
  Source=mscorlib
  StackTrace:
   at System.Reflection.AssemblyName.nInit(RuntimeAssembly& assembly, Boolean forIntrospection, Boolean raiseResolveEvent)
   at System.Reflection.AssemblyName..ctor(String assemblyName)
   at OmniSharp.Services.IAssemblyLoaderExtensions.Load(IAssemblyLoader loader, String name) in C:\src\omnisharp-roslyn\src\OmniSharp.Abstractions\Services\IAssemblyLoader.cs:line 33

mholo65: Ah, so this is where it actually throws https://github.com/OmniSharp/omnisharp-roslyn/blob/6d3ce87c502264453507f1669314c412a44acc1b/src/OmniSharp.Abstractions/Services/IAssemblyLoader.cs#L33

The provided path is neither a file that exists nor a valid assembly name. Sorry, don't have a quick solution for providing valuable information about the assembly that it failed to load. Maybe passing the `ILogger` is the best option for now.

dmgonch: @mholo65 Thanks - I reverted the last commit to go back to passing in ILogger. Please let me know if you have any other concerns.

david-driscoll:
LGTM

filipw:
thanks!

PR #1616, opened 2019-09-30
Make sure we include msbuild 16.3 in Linux distro

filipw:
This should finally allow us to close the famous issue https://github.com/OmniSharp/omnisharp-vscode/issues/3290

This is a follow up to #1606 which fixed the problem on Windows (confirmed by users) and follow up to #1607 which fixed the problem on Mac (also confirmed by users.

We still had the issue on Linux even though I wasn't able to initially reproduce it locally. Turns out, when we run on Azure DevOps we have Mono 6.0.0 (MSBuild 16.3) and Mono 6.4.0 (MSBuild 16.3) side by side and we copied from a wrong place... 😇 Also changed the distro from stable-xenial to vs-xenial.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -62,7 +62,7 @@ jobs:
 
   - job: Linux
     pool:
-      vmImage: "Ubuntu-18.04"
+      vmImage: "Ubuntu-16.04"
     dependsOn: GitVersion
     variables:
       MONO_VERSION: 6.4.0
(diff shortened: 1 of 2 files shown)


filipw:
I also asked the users to test these changes and they confirm it fixes things on Linux 🤞 https://github.com/OmniSharp/omnisharp-vscode/issues/3290#issuecomment-536309629

Issue #1622, opened 2019-10-03
Pass Roslyn diagnostic id as VSCode diagnostic code


henryju:
Hi,

A bit of context: SonarSource is providing a C# analyzer based on Roslyn. We already have a VSCode extension (SonarLint) with our own language server, that is providing analysis for other languages.

For C#, we plan to rely on you :) we will ensure our analyzer is used as part of the build (similar to what Roslynator is doing). Still we would like to provide some extra features, like some code actions. The problem we see is that currently, diagnostics created by omnisharp-vscode have no source nor code attribute set. This is a problem for us to differentiate diagnostics that are coming from our analyzer, from other one (compiler, or other third party analyzer).

I noticed that the diagnostic id is in fact appended to the message very early: https://github.com/OmniSharp/omnisharp-roslyn/blob/57e8052bc905a77a74c5f7623f6715df4aac4e47/src/OmniSharp.Roslyn.CSharp/Helpers/DiagnosticExtensions.cs#L25

Is there any reason for that? Would you be open to a pullrequest where the diagnostic id is instead passed through the VSCode diagnostic code?

Regarding the VSCode diagnostic source, it seems Roslyn itself is not exposing the information to know from which analyzer a diagnostic is coming (it would have been great to have the analyzerId). So maybe we could hardcode "csharp", "roslyn", or "omnisharp", just to avoid mixing diagnostics you create with something else. I can also provide a PR if you like.

filipw:
thanks for the issue. I am not sure I understand what you are trying to do.

Diagnostic IDs should be globally unique so you should always be able to understand which diagnostics come from where i.e. there are *IDExxx* ones (which, I guess, are the "roslyn" diagnostics you mentioned), *CSxxx* ones (which, I guess are the "csharp" diagnostics you mentioned).OmniSharp uses no other built-in analyzers. Now, if you have Roslynator installed, that uses *RCSxxx* and so on. 

wouldn't your diagnostics simply have different prefix?

Also, I am not sure I follow how you want to integrate - I moved this to omnisharp-roslyn, since it appears to me you want to integrated the analyzers into the server side, since I guess you'd want to it to invoke your code fixes too, but I don't know if that is the correct assumption.

henryju:
> However, I am not sure I understand what you are trying to do.

[...]

PR #1623, opened 2019-10-03
Remove diagnostic id from message

henryju:
Relates to https://github.com/OmniSharp/omnisharp-roslyn/issues/1622

In VSCode, the diagnostic id should better be passed in the `code` attribute. Having it also in the message lead to duplication in the UI.

### DIFF
--- a/src/OmniSharp.Roslyn.CSharp/Helpers/DiagnosticExtensions.cs
+++ b/src/OmniSharp.Roslyn.CSharp/Helpers/DiagnosticExtensions.cs
@@ -22,7 +22,7 @@ namespace OmniSharp.Helpers
                 Column = span.StartLinePosition.Character,
                 EndLine = span.EndLinePosition.Line,
                 EndColumn = span.EndLinePosition.Character,
-                Text = $"{diagnostic.GetMessage()} ({diagnostic.Id})",
+                Text = diagnostic.GetMessage(),
                 LogLevel = diagnostic.Severity.ToString(),
                 Tags = diagnostic
                     .Descriptor.CustomTags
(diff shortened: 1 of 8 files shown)


filipw:
I have no problem with this change however there are other consumers of these APIs than just VS Code, and it might be that removing the ID from the message could "break" them.

@razzmatazz @nickspoons is this change OK to you - you'd need to pick up the diagnostic ID from a dedicated property as it will no longer be in the message text

also, are you using https://github.com/OmniSharp/csharp-language-server-protocol?

henryju:
> also, are you using https://github.com/OmniSharp/csharp-language-server-protocol?

If the question was for me, then no, I'm not using it. We are currently only targeting VSCode integration, and from what I saw, the VSCode C# extension are calling directly the omnisharp server, without relying on the LSP.

nickspoons:
@filipw thanks for checking. This no problem from OmniSharp-vim's side, we use the `Id` property already. If users decide they miss seeing the Id in the message, we can just add it on our end.

I also am unsure who the LSP question was to - we don't use it. We use the stdio version without LSP, and still also use the HTTP version for older clients that can't use stdio.

nickspoons:
Not necessarily at the same time - it just has to be merged _after_ the VSCode PR. This issue is just removing the code from the message - it is already available in the `Id` property so the VSCode PR can go first.

filipw:
sorry I meant "this needs to be merged too" - English is my 3rd language

razzmatazz:
Sorry, I cannot check the code for the moment as I’m on vacation. Will update the code if it actually does use the Id field after your update.

However, no, omnisharp-emacs does not use lsp.

—Saulius Menkevičius

david-driscoll:
The change seems reasonable to me.  LSP supports the code as well, so that should map through.  https://microsoft.github.io/language-server-protocol/specifications/specification-3-14/#Diagnostic

PR #1633, opened 2019-10-15
Moved the build process to .NET Core 3.0 SDK

filipw:
The PR only affects our build script/pipeline no runtime features are affected.
As a bonus I also added a small netcoreapp3.0 test which would have failed on the old msbuild version.


.pipelines/init.yml
```diff
@@ -17,7 +17,7 @@ steps:
     displayName: 'Deserialize GitVersion'
```
Delpbrad, line 17: Brad

PR #1640, opened 2019-10-24
Use mono 6.4.0

JoeRobich:


build.json
```diff
@@ -4,11 +4,11 @@
...
-  "MonoRuntimeMacOS": "mono.macOS-5.18.1.0.zip",
-  "MonoRuntimeLinux32": "mono.linux-x86-5.18.1.0.zip",
-  "MonoRuntimeLinux64": "mono.linux-x86_64-5.18.1.0.zip",
-  "MonoMSBuildRuntime": "Microsoft.Build.Runtime.Mono-5.18.1.0.zip",
-  "MonoMSBuildLib": "Microsoft.Build.Lib.Mono-5.18.1.0.zip",
+  "MonoRuntimeMacOS": "mono.macOS-6.4.0.198.zip",
+  "MonoRuntimeLinux32": "mono.linux-x86-6.4.0.198.zip",
+  "MonoRuntimeLinux64": "mono.linux-x86_64-6.4.0.198.zip",
+  "MonoMSBuildRuntime": "Microsoft.Build.Runtime.Mono-6.4.0.198.zip",
+  "MonoMSBuildLib": "Microsoft.Build.Lib.Mono-6.4.0.198.zip",
```
filipw: maybe we can drop this completely. at the moment, the way build script has been changed recently - it doesn't take msbuild from these ZIPs anyway, it copies it from the global Mono on the build agent (since those are configured to use Mono 6.4.0).

filipw: the embedded mono, on the contrary, is still taken from here so that part is still relevant

JoeRobich: Ah, so a minimal install of mono is now built into and installed with omnisharp instead of omnisharp installing these artifacts?

JoeRobich: Nevermind my assumption, apparently github only loaded the first of your comments.

filipw: yeah no problem 😀

it's not immediately obvious as the build script is quite convoluted, but basically "both" (I quote "both" since there is more of them, but there are two variants - mono and msbuild) these ZIPs are downloaded at build time and embedded into the produced packages. however, since recent changes to move to msbuild 16.3, the msbuild binaries are actually copied from build agent's mono so the msbuild zip here is probably redundant (or we go back to the original approach and stop copying it from build agent's mono).

JoeRobich: Any reason to not go "all in" on the build script and let it package mono as well?

filipw: that would make sense to me 👍 especially as it would also make it easier to move to newer versions of mono 

in practice it would mean running the mono packaging script instead of this code https://github.com/OmniSharp/omnisharp-roslyn/blob/master/build.cake#L141-L164 at build time

JoeRobich:
I agree. I'll try to find some time to look into the build changes.

Issue #1650, opened 2019-11-12
InvalidProjectFileException: The imported project "/.../Microsoft.Common.props" was not found


ccidral:
My projects are not loading anymore after Omnisharp automatically self-updated this morning.  Below is an excerpt of the error message:

```
[info]: OmniSharp.MSBuild.ProjectManager
        Loading project: /Users/ccidral/lusa/proj/my-project/src/MyProject/MyProject.csproj
[warn]: OmniSharp.MSBuild.ProjectManager
        Failed to load project file '/Users/ccidral/lusa/proj/my-project/src/MyProject/MyProject.csproj'.
/Users/ccidral/lusa/proj/my-project/src/MyProject/MyProject.csproj(1,1)
Microsoft.Build.Exceptions.InvalidProjectFileException: The imported project "/Library/Frameworks/Mono.framework/Versions/5.18.1/lib/mono/xbuild/Current/Microsoft.Common.props" was not found. Also, tried to find "Current/Microsoft.Common.props" in the fallback search path(s) for $(MSBuildExtensionsPath) - "/Library/Frameworks/Mono.framework/External/xbuild/" . These search paths are defined in "/Users/ccidral/.vscode/extensions/ms-vscode.csharp-1.21.7/.omnisharp/1.34.7/omnisharp/OmniSharp.exe.config". Confirm that the path in the <Import> declaration is correct, and that the file exists on disk in one of the search paths.  /usr/local/share/dotnet/sdk/2.2.402/Sdks/Microsoft.NET.Sdk/Sdk/Sdk.props
  at Microsoft.Build.Shared.ProjectErrorUtilities.ThrowInvalidProject (System.String errorSubCategoryResourceName, Microsoft.Build.Shared.IElementLocation elementLocation, System.String resourceName, System.Object[] args) [0x00040] in <10b73e4bf233421cae439ef765e97ef0>:0 
  at Microsoft.Build.Shared.ProjectErrorUtilities.ThrowInvalidProject (Microsoft.Build.Shared.IElementLocation elementLocation, System.String resourceName, System.Object[] args) [0x00000] in <10b73e4bf233421cae439ef765e97ef0>:0 
[...]

PR #1656, opened 2019-11-22
do not set mono paths in standalone mode

filipw:
Since we now run in stand alone mode on top of mono 6.4.0 and raised the global mono minimum version to 6.4.0, we should not need to set mono paths in stand alone mode anymore. 

This fixes https://github.com/OmniSharp/omnisharp-vscode/issues/3410 and fixes https://github.com/OmniSharp/omnisharp-vscode/issues/3340 and a couple of other issues that I need to dig up (also lots of errors were reported as comments under old closed issues). In short the problem was, if you have old Mono (say 5.4.0) installed, we don't pick it as global Mono as it's < 6.4.0 and run in standalone mode. However we still set:

```
TargetFrameworkRootPath = /Library/Frameworks/Mono.framework/Versions/5.4.0/lib/mono/xbuild-frameworks
MSBuildExtensionsPath = /Library/Frameworks/Mono.framework/Versions/5.4.0/lib/mono/xbuild
```

which causes .NET Core 3.0 to fail.

By the way, the tests explicitly avoided to set any mono paths, so any problem with this, and users reported plenty, only ever manifested itself only once running OmniSharp server in real world.

### DIFF
--- a/src/OmniSharp.Host/MSBuild/Discovery/MSBuildLocator.cs
+++ b/src/OmniSharp.Host/MSBuild/Discovery/MSBuildLocator.cs
@@ -47,13 +47,13 @@ namespace OmniSharp.MSBuild.Discovery
                     new DevConsoleInstanceProvider(loggerFactory),
                     new VisualStudioInstanceProvider(loggerFactory),
                     new MonoInstanceProvider(loggerFactory),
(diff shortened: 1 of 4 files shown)


filipw:
fixes https://github.com/OmniSharp/omnisharp-roslyn/issues/1650 too

mholo65:
Thanks!

PR #1660, opened 2019-11-29
Support #line directive in find references

### LINKED ISSUE #1649: Support line pragmas for find all references
In Razor scenarios where you have a front level file such as 

```Razor
<!-- Counter.razor -->
<p>@MyType.Method()</p>
```

It generates virtual C# documents under the path `Counter.razor__virtual.cs` that looks like:
```C#
...
#line 456 "C:/path/to/Index.razor"
__o = MyType.Method();
#line default
```

When you do find all references on the corresponding C# MyType.Method member you get:
image

If find all references (FAR) respected line pragmas and then assumed column 0 (line pragmas don't give column info) this would work without having to actually read the Razor file 😄

filipw:
Fixes #1649 
If the mapping goes to an external file, we assume column 0,0

### DIFF
--- a/src/OmniSharp.Roslyn.CSharp/Helpers/LocationExtensions.cs
+++ b/src/OmniSharp.Roslyn.CSharp/Helpers/LocationExtensions.cs
@@ -25,9 +25,9 @@ namespace OmniSharp.Helpers
                 Text = text.Trim(),
                 FileName = path,
                 Line = line,
-                Column = lineSpan.StartLinePosition.Character,
+                Column = lineSpan.HasMappedPath ? 0 : lineSpan.StartLinePosition.Character, // when a #line directive maps into a separate file, assume columns (0,0)
                 EndLine = lineSpan.EndLinePosition.Line,
-                EndColumn = lineSpan.EndLinePosition.Character,
+                EndColumn = lineSpan.HasMappedPath ? 0 : lineSpan.EndLinePosition.Character,
                 Projects = documents.Select(document => document.Project.Name).ToArray()
             };
         }
--- a/src/OmniSharp.Roslyn.CSharp/Helpers/LocationExtensions.cs
+++ b/src/OmniSharp.Roslyn.CSharp/Helpers/LocationExtensions.cs
@@ -13,7 +13,7 @@ namespace OmniSharp.Helpers
             if (!location.IsInSource)
                 throw new Exception("Location is not in the source tree");
 
-            var lineSpan = location.GetLineSpan();
+            var lineSpan = location.GetMappedLineSpan();
             var path = lineSpan.Path;
             var documents = workspace.GetDocuments(path);
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 All changes to the project will be documented in this file.
 
 ## [1.34.9] - not yet released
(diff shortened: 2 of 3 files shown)


filipw:
@mholo65 Cake tests fail with this, should I exclude Cake from this span mapping logic?

mholo65:
@filipw let me have a look at the Cake, and see if I can fix it. Cake, behind the scenes uses a lot of `#line` mappings, as we basically merge all `#load`ed files + DSL into one giant `csx` with custom host which we pass to Roslyn.

filipw:
> @filipw let me have a look at the Cake, and see if I can fix it. Cake, behind the scenes uses a lot of `#line` mappings, as we basically merge all `#load`ed files + DSL into one giant `csx` with custom host which we pass to Roslyn.

did you get a chance to look at this?

mholo65:
> 
> 
> > @filipw let me have a look at the Cake, and see if I can fix it. Cake, behind the scenes uses a lot of `#line` mappings, as we basically merge all `#load`ed files + DSL into one giant `csx` with custom host which we pass to Roslyn.
> 
> did you get a chance to look at this?

I'm looking now.

src/OmniSharp.Roslyn.CSharp/Helpers/LocationExtensions.cs
```diff
@@ -13,7 +13,7 @@ public static QuickFix GetQuickFix(this Location location, OmniSharpWorkspace wo
             if (!location.IsInSource)
                 throw new Exception("Location is not in the source tree");
 
-            var lineSpan = location.GetLineSpan();
+            var lineSpan = location.GetMappedLineSpan();
```
mholo65: @filipw So, I tried looking at it. But honestly don't know why Roslyn returns so weird line spans when using `GetMappedLineSpan()` with the Cake file.

I debugged the `ShouldIncludeUsagesFromLoadedFiles()` fact from Cake's `FindUsagesFacts` and flipping this back to `GetLineSpan()` results in this:
image
While the new change gives this:
image

The whole `build.cake` file can be seen [here](https://gist.github.com/mholo65/9e55c33b4d63a7b90c3e37cc183eee32).  The loaded `foo.cake` is basically the same, but from line 8481 onwards, it's
```csharp
#line 1 "C:/src/gh/omnisharp-roslyn/test-assets/test-projects/CakeProject/foo.cake"
Information("Hello World");
```

Could the issue be that we the contents of `foo.cake` is available in both `build.cake` (src included via `#load` directive) and in the document itself. Quickest fix, would be as you already mentioned, to exclude `.cake` files from this mapping, and just do it the old way, but I'd really like to avoid having Cake specific code inside `OmniSharp.Roslyn.CSharp`.

mholo65: Ok, so now I see. The line spans are actually correct (we'll just need to change Cake stuff to not translate it). But the text it fetches here:
https://github.com/OmniSharp/omnisharp-roslyn/blob/8b913fe547f97535c55db64490cbe362a7f4ea6b/src/OmniSharp.Roslyn.CSharp/Helpers/LocationExtensions.cs#L21 doesn't do line span mapping, which results in such weird text being returned. The line span mapping in Cake, I can remove. But how to get the correct text, I don't know.

mholo65: Sent you a PR for the quickest fix here
https://github.com/filipw/omnisharp-roslyn/pull/1

filipw:
thanks @mholo65 

should be good to go

PR #1662, opened 2019-12-09
updated to msbuild 16.4.0 and Microsoft.NETFramework.ReferenceAssemblies 1.0.0

filipw:
stable releases are public so we can drop the pre-release references

please note that on *nix we copy msbuild from Mono so until 6.6.0 is released, we still use 6.4.0 which has msbuild 16.3


filipw:
the mac build hung for no apparent reason but after restart everything is back to normal

PR #1668, opened 2019-12-19
Check null

mattn:
Some Language Server Client null for contentChanges when SyncKind = 0.
Also signatureparams.


filipw:
is there any test that can verify this?

david-driscoll:
We don't have a lot of tests around the language server stuff yet.  This looks okay.

filipw:
thanks

PR #1669, opened 2019-12-19
Updated build process to use Mono 6.6.0

filipw:
Mono 6.6.0 is now stable and it comes with MSBuild 16.4.0
Since we copy MsBuild from global Mono at build time, updated the build steps to use Mono 6.6.0.

This follows up to #1662 which moved us to MsBuild 16.4.0 on Windows,  but on *nix we stayed on 16.3 since we copied it at package time from Mono 6.4.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,11 @@
 # Changelog
 All changes to the project will be documented in this file.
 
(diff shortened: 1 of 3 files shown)


mholo65:
Thanks @filipw :+1:

PR #1670, opened 2019-12-20
fixed symbol decoding from completion items

filipw:
This got broken a while ago - around these changes https://github.com/OmniSharp/omnisharp-roslyn/commit/1f10506972481e5db2b44caa117e94f54e7f17be. It was fragile, as it's reflection.
 
It impacts the quality of some of the completion item driven completions, which we don't use a lot but nevertheless - i.e. object initializer, which were missing symbol information.

fixes https://github.com/OmniSharp/omnisharp-vscode/issues/3465

### DIFF
--- a/src/OmniSharp.Roslyn.CSharp/Services/Intellisense/CompletionItemExtensions.cs
+++ b/src/OmniSharp.Roslyn.CSharp/Services/Intellisense/CompletionItemExtensions.cs
@@ -21,9 +21,9 @@ namespace OmniSharp.Roslyn.CSharp.Services.Intellisense
         private const string ParitalMethodCompletionProvider = "Microsoft.CodeAnalysis.CSharp.Completion.Providers.PartialMethodCompletionProvider";
         private const string ProviderName = nameof(ProviderName);
         private const string SymbolCompletionItem = "Microsoft.CodeAnalysis.Completion.Providers.SymbolCompletionItem";
-        private const string SymbolCompletionProvider = "Microsoft.CodeAnalysis.CSharp.Completion.Providers.SymbolCompletionProvider";
         private const string SymbolKind = nameof(SymbolKind);
         private const string SymbolName = nameof(SymbolName);
+        private const string SymbolsKey = nameof(Symbols);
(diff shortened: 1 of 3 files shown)


tests/OmniSharp.Roslyn.CSharp.Tests/IntellisenseFacts.cs
```diff
@@ -259,6 +259,7 @@ public MyClass2()
 
             var completions = await FindCompletionsAsync(filename, source);
             ContainsCompletions(completions.Select(c => c.CompletionText), "Foo");
+            ContainsCompletions(completions.Select(c => c.ReturnType), "string");
```
filipw: without the changes from this PR, this assert fails (for example on current master). It would have prevented the regression too.

src/OmniSharp.Roslyn.CSharp/Services/Intellisense/CompletionItemExtensions.cs
```diff
@@ -21,9 +21,9 @@ internal static class CompletionItemExtensions
         private const string ParitalMethodCompletionProvider = "Microsoft.CodeAnalysis.CSharp.Completion.Providers.PartialMethodCompletionProvider";
         private const string ProviderName = nameof(ProviderName);
         private const string SymbolCompletionItem = "Microsoft.CodeAnalysis.Completion.Providers.SymbolCompletionItem";
-        private const string SymbolCompletionProvider = "Microsoft.CodeAnalysis.CSharp.Completion.Providers.SymbolCompletionProvider";
         private const string SymbolKind = nameof(SymbolKind);
         private const string SymbolName = nameof(SymbolName);
+        private const string SymbolsKey = nameof(Symbols);
```
mholo65: We already have `Symbols = nameof(Symbols)` below, and it seems that it's not used. Should you use that const instead (or remove it). Both `SymbolsKey` and `Symbols` have the same value.

filipw: good point - i consolidated them

PR #1495, opened 2019-05-14
Added support for Roslyn rename options

filipw:
This PR introduces support (via `omnisharp.json` and the rest of [OmniSharp configuration model](https://github.com/OmniSharp/omnisharp-roslyn/wiki/Configuration-Options)) for Roslyn **renaming options**.

```
{
    "RenameOptions": {
        "RenameInComments": true,
        "RenameOverloads": true,
        "RenameInStrings": true
   }
}
```

The solutions builds upon the original workspace options extensibility created for `FormattingOptions`. The options are respected in real time - changes are propagated to the workspace without the need to restart OmniSharp.

Following options are supported (all of them are switched off by default and need to be enabled explicitly):

### RenameOverloads

rename2

### RenameInComments

rename

### RenameInStrings

rename3

I will update the documentation https://github.com/OmniSharp/omnisharp-roslyn/wiki/Configuration-Options once this is merged.

PR #1672, opened 2019-12-29
added implement type options

filipw:
Similar to #1495 except for implement type options.
This allows users to set i.e. property generation style (see https://github.com/OmniSharp/omnisharp-vscode/issues/2392).

This can be configured in Visual Studio so makes sense that we expose it too. The defaults remain like in Roslyn.


filipw:
any thoughts? @JoeRobich @david-driscoll @mholo65

mholo65:
Thanks @filipw!

PR #1677, opened 2020-01-06
Do not discover abstract types as candidates for code fix providers

filipw:
Also, do not report code fix providers that are missing `ExportCodeFixProviderAttribute` as INFO - this leads to confusing messages in the logs like:

```
[info]: OmniSharp.Roslyn.CSharp.Services.Refactoring.V2.CachingCodeFixProviderForProjects
        Skipping code fix provider 'Microsoft.CodeQuality.Analyzers.QualityGuidelines.MarkMembersAsStaticFixer' because its language doesn't match 'C#'.
```

This happens for all base types used by code fixes, which are not intended for discovery in the first place.
From now on, we should only attempt to instantiate non-abstract code fixes marked with `ExportCodeFixProviderAttribute`.

cc @savpek


mholo65:
Thanks @filipw!

PR #1686, opened 2020-01-22
Added support for organizing imports in formatting

filipw:
The organize imports formatting feature is now public in Roslyn.
Added as a flag since we have an approach to make all these things configurable and opt-in.

### DIFF
--- a/src/OmniSharp.Roslyn.CSharp/Workers/Formatting/FormattingWorker.cs
+++ b/src/OmniSharp.Roslyn.CSharp/Workers/Formatting/FormattingWorker.cs
@@ -89,6 +89,11 @@ namespace OmniSharp.Roslyn.CSharp.Workers.Formatting
                 : document.Project.Solution.Workspace.Options;
             var newDocument = await Formatter.FormatAsync(document, TextSpan.FromBounds(start, end), optionSet);
 
+            if (omnisharpOptions.FormattingOptions.OrganizeImports)
+            {
+                newDocument = await Formatter.OrganizeImportsAsync(document);
+            }
+
             return await TextChanges.GetAsync(newDocument, document);
         }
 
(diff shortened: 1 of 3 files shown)


JoeRobich:
Would it be good to have this functionality split the way VS does, Format vs Code Cleanup? The format command defaulting to just being whitespace formatting. Then a new code cleanup command which brings in other types of changes such as organize imports. 

image

If the user always wanted the code cleanup behavior, then maybe there is an omnisharp option to run code cleanup during format.

filipw:
yes we could have another endpoint for this and then build it up with other capabilities to fix i.e. the IDExxxx diagnostics and so on.

the reason why I added it to formatting endpoint here is that in LSP there is only the "formatting" concept, no "cleanup" concept, so I thought maybe that would be cleaner to not create something completely new.

filipw:
@JoeRobich thanks for the approve. I really think it makes sense this way.

in the "clean up" endpoint we could implement a more sophisticated handling like the current fix usings does already where it also removes unused stuff. It really also aligns with the feature discussed here #1581

by the way, do you currently call/plan to call this `Formatter.OrganizeImportsAsync` in `dotnet format`?

JoeRobich:
> do you currently call/plan to call this `Formatter.OrganizeImportsAsync` in `dotnet format`?

We do. We plan to make it part of the "style" formatting which would also include running analyzers and code fixes. This work is hinging on getting changes into Roslyn for it to setup workspaces to automatically make use of the compiler's EditorConfig support and for Analyzers to pull their configuration Options from the AnalyzerConfigOptions instead of relying on a Workspace OptionSet.

filipw:
OK very cool, then we will do the same in the future 😃

we need to change EditorConfig implementation to get support for dotnet_diagnostic severity control too.

jmihalich:
Hi,

Is this actually live in 1.21.10 as it says in the release notes?  Is there documentation for this feature anywhere, cuz i have a few questions:

1) looking at the code change it looks like the relevant setting in omnisharp.json is FormattingOptions.OrganizeImports, and that's a boolean value?

2) the auto fill is not showing an OrganizeImports option when i try to add it to the json file

3) is it possible to specify the order of organization?  or is it fixed based on an internal sort algorithm?

Thanks,
Joe

PR #1695, opened 2020-02-03
updated to Cake 0.37.0

filipw:
Fixes the build failure on linux.
I came up with the fix on my own 😀

The old version doesn't work anymore on Linux build agents and all builds fail.


mholo65:
@filipw thank you for your contribution! **Your** changes have been merged...

PR #1696, opened 2020-02-04
Add NuGetSdkResolver to minimal msbuild

JoeRobich:
Fixes https://github.com/OmniSharp/omnisharp-roslyn/issues/1678

Tested with dotnet/format repo which relies on the 'Microsoft.DotNet.Arcade.Sdk'


filipw:
do these nuget packages need a special feed?

Unable to find version '5.4.0-rtm.6292' of package 'NuGet.Common'.

mholo65:
Can't we just use 5.4.0 which is already on NuGet? https://www.nuget.org/packages/NuGet.Common/5.4.0

JoeRobich:
> Can't we just use 5.4.0 which is already on NuGet? 

Unfortunately NuGetSdkResolver isn't on NuGet.org.

Issue #1693, opened 2020-01-31
Omnisharp's mono tries to use system mono lib instead of bundled mono


benknoble:
I get this output when OmniSharp tries to start up in vim:

```
channel 0 open: 'dyld: Library not loaded: /Library/Frameworks/Mono.framework/Versions/6.6.0/lib/libintl.8.dylib'
channel 0 open: '  Referenced from: /Users/Knoble/.cache/omnisharp-vim/omnisharp-roslyn/bin/mono'
channel 0 open: '  Reason: image not found'
channel 0 open: '/Users/Knoble/.cache/omnisharp-vim/omnisharp-roslyn/run : ligne 28 : 54423 Abort trap: 6           "${mono_cmd}" "${omnisharp_cmd}" "$@"'
```

I can confirm that running the `run` script and the `bin/mono` fail, so this may be more for OmniSharp/omnisharp-roslyn.

I have an (old?) mono install floating around:

```
+λ ls /Library/Frameworks/Mono.framework/Versions/5.4.1/                              
Commands/   bin/        etc/        lib/        share/                                
VERSION     docs/       include/    man/        updateinfo
```

Regardless, I'm not sure why this mono is searching `/Library/Frameworks` if it was supposed to be fully bundled?

After installing an up-to-date mono with `brew`, the bundled mono still looks in `/Library/Frameworks` where the new version is not.

(From OmniSharp/omnisharp-vim#556.)

filipw:
the minimum required Mono version is 6.6.0.
OmniSharp looks at global mono if you have it installed on your machine because it would prefer to use that msbuild instead of the bundled one.

benknoble:
Right, but I definitely don't have mono 6.6.0 (or, I didn't...). Surely this should be easily detectable, and omnisharp could use the bundled mono?

DwaynesWorld:
Yea, I am running in to the same issue. It happens when I am running version 1.21.10 of the C# extension in vscode (ms-vscode.csharp). If I downgrade everything seems to be fine.

filipw:
@JoeRobich isn't there something missing in the bundle here? https://github.com/OmniSharp/omnisharp-roslyn/pull/1679/files

libintl.8.dylib?

traegerp:
I am having the same issue as @DwaynesWorld with vs code: 

dyld: Library not loaded: /Library/Frameworks/Mono.framework/Versions/6.6.0/lib/libintl.8.dylib

[...]

PR #1697, opened 2020-02-04
Update Mono to 6.8.0

JoeRobich:
Mono 6.6.0 added a dependency on libintl that ships in the Mono pkg (https://github.com/mono/mono/pull/18531). This dependency has been removed from later builds of 6.8.0.

Tested locally with 6.8.0.96 and saw failures. The stable MacOS release of 6.8.0.104 contains the fix. Linux builds use 6.8.0.96 as that is the released stable build.

Resolves https://github.com/OmniSharp/omnisharp-roslyn/issues/1693
Resolves https://github.com/OmniSharp/omnisharp-vscode/issues/3545


filipw:
during the .NET Core 3.0 debacle with msbuild 16.3 the build script was updated to copy MSBuild from global mono at the build agent, instead of these zips. probably time to revert this 😁

JoeRobich:
> probably time to revert this

You mean rely purely on the MSBuild zip files?

filipw:
I am talking about this change https://github.com/OmniSharp/omnisharp-roslyn/pull/1612

would you mind reverting it as part of this PR?

otherwise we have older MSBuild in the released packages...

JoeRobich:
Ah, gotcha. Sure thing.

filipw:
the alternative is to remove them from the JSON completely as they are not used at the moment and update build agents to 6.8.0 too so that it's in line with the bundled runtime when it comes to MSBuild

filipw:
I somehow don't have access to that blob storage and when we needed to urgently update to MSBuild 16.3 due to the breaking changes in .NET SDK, and Ravi was unreachable, the best and quickest idea I had was to copy from the global Mono at build time instead 😅 that's the reason behind that change in the first place, in case you were wondering... 

(sorry for slow responses, typing on mobile)

JoeRobich:
@filipw I think building it from the build script is probably the way to go. It looks like the build script was already augmenting the msbuild zips anyways,

filipw:
OK makes sense to me. maybe let's just align the versions here https://github.com/OmniSharp/omnisharp-roslyn/blob/master/azure-pipelines.yml#L59 and https://github.com/OmniSharp/omnisharp-roslyn/blob/master/azure-pipelines.yml#L93 with the embedded runtime ones

JoeRobich:
In light of the way the Linux and MacOS legs failed. We may want to update this to the 6.8.0.105 version that just dropped.

ifuseandprint:
Duplicate of #

PR #1701, opened 2020-02-07
Adds support for aliases on both ProjectReference and Reference

mholo65:
The previous implementation (alias on `Reference`) may or may not have worked previously. It's hard to know as the tests only checks that the values are present in the `ProjectFileInfo`. It currently lacks tests in `ProjectManager`, which is where the logic resides for actually adding the alias to the reference (Project or Metadata).

This PR also adds support for aliases on `ProjectReference`.

I don't think we have any tests on `ProjectManager` at this point, but should I just att a simple test that verifies that there are no errors in the test-projects that currently use aliases for references? That's a quick way to ensure we don't break this in the future.


filipw:
> I don't think we have any tests on `ProjectManager` at this point, but should I just att a simple test that verifies that there are no errors in the test-projects that currently use aliases for references? That's a quick way to ensure we don't break this in the future.

maybe you could test via `WorkspaceInformationTests`?

mholo65:
Added tests (which verifies from workspace). `WorkspaceInformation` itself doesn't contain information needed. But I can fetch it from the `Workspace` directly.

ping @filipw

Issue #1589, opened 2019-08-20
CS0579 (Duplicate Attribute) for project with multiple target frameworks.


seesharper:
There seem to be a problem related to `AssemblyInfo.cs` when having a project with multiple target frameworks.

Here is my csproj file 

```
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <TargetFrameworks>netstandard2.0;netstandard1.6</TargetFrameworks>
  </PropertyGroup>
</Project>
```

When building from VS Code I get the following output in the problems panel. 

![image](https://user-images.githubusercontent.com/1034073/63357083-2a4a3700-c369-11e9-81bd-c1584c6d1f55.png)

Note that I DON'T have a `AssemblyInfo.cs` file in my project. 

I have also created a repro project here 

https://github.com/seesharper/AssemblyInfoRepro

I have also tried to open the same project in VS for Mac and there are no problems there. 

`<NoWarn>$(NoWarn);CS0579</NoWarn>` has no effect, but I guess that is because this is an error and not a warning :)

Environment

```
.NET Core SDK (reflecting any global.json):
 Version:   3.0.100-preview6-012264
 Commit:    be3f0c1a03

Runtime Environment:
 OS Name:     Mac OS X
 OS Version:  10.14
 OS Platform: Darwin
 RID:         osx.10.14-x64
 Base Path:   /usr/local/share/dotnet/sdk/3.0.100-preview6-012264/

Host (useful for support):
  Version: 3.0.0-preview6-27804-01
  Commit:  fdf81c6faf

.NET Core SDKs installed:
  1.1.14 [/usr/local/share/dotnet/sdk]
  2.1.301 [/usr/local/share/dotnet/sdk]
  2.1.402 [/usr/local/share/dotnet/sdk]
  2.1.505 [/usr/local/share/dotnet/sdk]
  2.2.100 [/usr/local/share/dotnet/sdk]
  2.2.300 [/usr/local/share/dotnet/sdk]
  3.0.100-preview-010184 [/usr/local/share/dotnet/sdk]
  3.0.100-preview3-010431 [/usr/local/share/dotnet/sdk]
  3.0.100-preview5-011568 [/usr/local/share/dotnet/sdk]
  3.0.100-preview6-012264 [/usr/local/share/dotnet/sdk]

.NET Core runtimes installed:
[...]

PR #1704, opened 2020-02-09
Prevents project manager from updating default excluded source files

leandroslc:
Prevents OmniSharp project manager of updating source files described in the _MSBuild_ property `<DefaultItemExcludes>`.

This should fix the following issues:
- #1589
- OmniSharp/omnisharp-vscode#3249
- OmniSharp/omnisharp-vscode#3082
- OmniSharp/omnisharp-vscode#3207

Thank you for OmniSharp. It is an honor to contribute 😄

### DIFF
--- a/src/OmniSharp.MSBuild/ProjectFile/ProjectFileInfo.ProjectData.cs
+++ b/src/OmniSharp.MSBuild/ProjectFile/ProjectFileInfo.ProjectData.cs
@@ -56,6 +56,7 @@ namespace OmniSharp.MSBuild.ProjectFile
             public bool RunAnalyzers { get; }
             public bool RunAnalyzersDuringLiveAnalysis { get; }
             public string DefaultNamespace { get; }
+            public ImmutableArray<string> DefaultItemExcludes { get; }
 
             private ProjectData()
             {
@@ -72,6 +73,7 @@ namespace OmniSharp.MSBuild.ProjectFile
                 AdditionalFiles = ImmutableArray<string>.Empty;
                 ReferenceAliases = ImmutableDictionary<string, string>.Empty;
                 ProjectReferenceAliases = ImmutableDictionary<string, string>.Empty;
+                DefaultItemExcludes = ImmutableArray<string>.Empty;
             }
 
             private ProjectData(
@@ -160,7 +162,8 @@ namespace OmniSharp.MSBuild.ProjectFile
                 bool runAnalyzersDuringLiveAnalysis,
                 RuleSet ruleset,
                 ImmutableDictionary<string, string> referenceAliases,
-                ImmutableDictionary<string, string> projectReferenceAliases)
+                ImmutableDictionary<string, string> projectReferenceAliases,
+                ImmutableArray<string> defaultItemExcludes)
                 : this(guid, name, assemblyName, targetPath, outputPath, intermediateOutputPath, projectAssetsFile,
                       configuration, platform, targetFramework, targetFrameworks, outputKind, languageVersion, nullableContextOptions, allowUnsafeCode, checkForOverflowUnderflow,
                       documentationFile, preprocessorSymbolNames, suppressedDiagnosticIds, signAssembly, assemblyOriginatorKeyFile, treatWarningsAsErrors, defaultNamespace, runAnalyzers, runAnalyzersDuringLiveAnalysis, ruleset)
@@ -173,6 +176,7 @@ namespace OmniSharp.MSBuild.ProjectFile
                 AdditionalFiles = additionalFiles.EmptyIfDefault();
                 ReferenceAliases = referenceAliases;
                 ProjectReferenceAliases = projectReferenceAliases;
+                DefaultItemExcludes = defaultItemExcludes.EmptyIfDefault();
             }
 
             public static ProjectData Create(MSB.Evaluation.Project project)
@@ -257,6 +261,8 @@ namespace OmniSharp.MSBuild.ProjectFile
 
                 var ruleset = ResolveRulesetIfAny(projectInstance);
 
+                var defaultItemExcludes = PropertyConverter.SplitList(projectInstance.GetPropertyValue(PropertyNames.DefaultItemExcludes), ';').ToImmutableArray();
+
                 var sourceFiles = GetFullPaths(
                     projectInstance.GetItems(ItemNames.Compile), filter: FileNameIsNotGenerated);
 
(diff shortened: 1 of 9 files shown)


filipw:
Thanks a lot for the PR 👍
There already is a `Microsoft.NET.Sdk.DefaultItems.targets` which automatically runs for all SDK projects and will automatically be respected.

For example, if you add `<DefaultItemExcludes>Foo.cs</DefaultItemExcludes>` to the project, then this line here (the compile items) https://github.com/OmniSharp/omnisharp-roslyn/blob/aeb1b50de3528189141fd4b73f2bddaa8dd78bed/src/OmniSharp.MSBuild/ProjectFile/ProjectFileInfo.ProjectData.cs#L243-L244 automatically will exclude `Foo.cs` from being returned as compile items and we will ignore it. Therefore I believe there is no need to manually check anything.

Do you have some specific repro scenario that you are addressing?

leandroslc:
Thank you for the comment 😄. Sorry, I didn't give you more details before.
I have included the `DefaultItemExcludes` in ProjectData to allow Omnisharp to check if an compile item is excluded before adding the document to the workspace.

For example, I have a project that has two `TargetFrameworks`:
`<TargetFrameworks>netstandard2.0;net472</TargetFrameworks>`

Assuming `GenerateAssemblyInfo` is set to true by default, at startup Omnisharp seens to get both the `obj/Debug/net472/AssemblyInfo.cs` and `obj/Debug/netstandard2.0/AssemblyInfo.cs` as source files, even though DefaultItemExcludes has `obj/**`. But in this case, no error is generated because Omnisharp gets only the first `TargetFramework`, and just `obj/Debug/netstandard2.0/AssemblyInfo.cs` is loaded into the workspace.

Now, if we clean and build the project, both the AssemblyInfos are regenerated and ProjectManager tries to add them into workspace because of the OnFileChanged request. Hence, these erros appears:
assemblyinfoerror

What I thought was checking if the source file is not one of the files listed in DefaultItemExcludes
before adding the document in:
https://github.com/OmniSharp/omnisharp-roslyn/blob/aeb1b50de3528189141fd4b73f2bddaa8dd78bed/src/OmniSharp.MSBuild/ProjectManager.cs#L487

And before updating the BufferManager in:
https://github.com/OmniSharp/omnisharp-roslyn/blob/aeb1b50de3528189141fd4b73f2bddaa8dd78bed/src/OmniSharp.MSBuild/ProjectManager.cs#L520

Am I on the right track?
Let me know what you think 😄

david-driscoll:
I have seen this issue myself, and never really could figure out where it came from... I will look into this PR and see if there might be a different way that makes sense but at first glance it looks okay.

david-driscoll:
@leandroslc I was thinking we could update the file system watcher... but then this value is project specific, so this change kind of makes sense tbh.

I can't see too many holes at the moment.  @filipw @mholo65  thoughts?

filipw:
the `UpdateSourceFiles` is always invoked as direct effect of the full MSBuild project processing, so the check should not be required anymore, as MSBuild would have respected `DefaultItemExcludes` already.

Without analyzing this deeper, I believe the culprit code path would be the buffer update via `OnDirectoryFileChanged`?

leandroslc:
Yes, the `OnDirectoryFileChanged` is the culprid in this case. But if you have a look at the `UpdateSourceFiles` it also loads the exluded file, so I made a check for it too just to ensure.

filipw:
> But if you have a look at the `UpdateSourceFiles` it also loads the exluded file, so I made a check for it too just to ensure.

right, but that's what I was saying - `UpdateSourceFiles` is (or should be) only ever called as a callback after full MSBuild processing. At that point the `Microsoft.NET.Sdk.DefaultItems.targets` should exclude the excluded files and not surface them anymore, so we should not do manual processing. It would only ignore them when dealing with non-SDK projects, but that's fine because there the default items exclusion is not supported anyway.

I have not looked closely at the repro steps but to me it sounds like the entire fix could be based only on applying the filtering on the buffer watcher (which listens to new files that are reported by the editor).

leandroslc:
Hello, sorry for the late reply. I could not find a better solution, so I removed the check I made on the project load and just left the verification on `OnDirectoryFileChanged`, before entering the BufferManager. What do you think?

SirIntruder:
@leandroslc First, thanks for tackling this 😄  If I understood correctly, your check in the end are applied only in `OnDirectoryFileChanged()`, am I right? This is relevant for files that are changed on fileSystem by outside processes (commonly, git, and honestly this is the main vector for this bug). 

Most buffer update requests are done through the API by the editor, no fileSystem changes involved. So having a check one step down, in the BufferManager, would be a complete solution.

I think the correct way to insert this check might be here, in BufferManager:

```
      private IEnumerable<Project> FindProjectsByFileName(string fileName)
      {
            var fileInfo = new FileInfo(fileName);
            var dirInfo = fileInfo.Directory;
            var candidates = _workspace.CurrentSolution.Projects
                .Where(project => !String.IsNullOrWhiteSpace (project.FilePath))
                .GroupBy(project => new FileInfo(project.FilePath).Directory.FullName)
                .ToDictionary(grouping => grouping.Key, grouping => grouping.ToList());

            List<Project> projects = null;
            while (dirInfo != null)
            {
                if (candidates.TryGetValue(dirInfo.FullName, out projects))
                {
                    return projects;
                }

                dirInfo = dirInfo.Parent;
            }

            return Array.Empty<Project>();
        }
```

This is method BufferManager invokes whenever ir encounters a change on a c# file which is in no project. It does some simple heuristics, and the (un)lucky winner Project(s) gets to keep the new Document. If returned enumeration is empty, c# file is put into a "Misc Documents" bin, which is what is expected for the excluded files. 

I think replacing:

` .Where(project => !String.IsNullOrWhiteSpace (project.FilePath))`

with:

` .Where(project => _workspace.FileBelongsToProject(filename, project))`

and fixing `_workspace.FileBelongsToProject()` to respect default excludes, would cover all of the cases. 

It would also remove the need for a separate heuristic you had to do in order to retrieve `ProjectFileInfo` from document fileName.

leandroslc:
Thank you for the sugestion @SirIntruder 😄. I made some tests but with no success. My first solution was to check the excluded file before entering the BufferManager, because it was MSBuild specific. I am thinking if maybe ignoring auto generated files could be a better solution.

Currently Roslyn has this [Language Service](https://github.com/dotnet/roslyn/blob/master/src/Workspaces/SharedUtilitiesAndExtensions/Workspace/Core/LanguageServices/GeneratedCodeRecognition/AbstractGeneratedCodeRecognitionService.cs) that more specifically uses this [utility code](https://github.com/dotnet/roslyn/blob/master/src/Compilers/Core/Portable/InternalUtilities/GeneratedCodeUtilities.cs). However both implementations are internal. I checked that the service works for DiagnosticAnalyzers, but I could not find a way to make it work with the Omnisharp workspace. What do you think?

SirIntruder:
Ah, so the issue is that ProjectFileInfo and friends are in MSBuild project, which OmnisharpWorkspace is agnostic about? Since Roslyn Project class has nothing inbuilt, guess the best solution would be if Workspace allowed project systems to supply "Inclusion condtition" per "ProjectId", so any project system could hook their own heuristic. Workspace could store conditions in Dictionary<ProjectId,> and invoke in "FileBelongsToProject".

I am not sure about the AutoGeneration - some autogenerated files should actually be included in projects, and some not-auto-generated could be in excluded folders. So it may solve some problems, but it would be spotty.

PR #1707, opened 2020-02-10
fixed out of bounds in line mapping

filipw:
Fixes https://github.com/OmniSharp/omnisharp-vscode/issues/3485
Also added more comprehensive tests.

Here is the summary of the logic:
 - correct mapping within a file - then we return the mapped line and mapped text
 - incorrect mapping within a file - then we return the mapped line (even though it's incorrect) and the original (unmapped) text since there is no way to fetch the mapped text as it's out of bounds
 - correct mapping into a file that exists in a workspace - then we return the mapped line and mapped text from that file
 - incorrect mapping into a file that exists in a workspace - then we return the mapped line and the original (unmapped) text since there is no way to fetch the mapped text as it's out of bounds
 - mapping into a file that does not exist in a workspace - then we return the mapped line and the original (unmapped) text since there is no way to fetch the mapped text as we don't have access to it (Razor case). The editor still ignores it anyway and shows the correct file in this case. 

cc @NTaylorMullen as this bug affected Razor 🙈

### DIFF
--- a/src/OmniSharp.Roslyn.CSharp/Helpers/LocationExtensions.cs
+++ b/src/OmniSharp.Roslyn.CSharp/Helpers/LocationExtensions.cs
@@ -21,8 +21,8 @@ namespace OmniSharp.Helpers
                 : location.GetMappedLineSpan();
 
             var documents = workspace.GetDocuments(lineSpan.Path);
-            (var sourceText, var isMappedDocument) = GetSourceText(location, documents, lineSpan.HasMappedPath);
-            var text = GetText(location, sourceText, lineSpan.StartLinePosition.Line, isMappedDocument);
+            var sourceText = GetSourceText(location, documents, lineSpan.HasMappedPath);
+            var text = GetText(location, sourceText, lineSpan.StartLinePosition.Line);
 
             return new QuickFix
             {
@@ -35,25 +35,31 @@ namespace OmniSharp.Helpers
                 Projects = documents.Select(document => document.Project.Name).ToArray()
             };
 
-            static (SourceText sourceText, bool isMappedDocument) GetSourceText(Location location, IEnumerable<Document> documents, bool hasMappedPath)
+            static SourceText GetSourceText(Location location, IEnumerable<Document> documents, bool hasMappedPath)
             {
                 // if we have a mapped linespan and we found a corresponding document, pick that one
                 // otherwise use the SourceText of current location
-                if (hasMappedPath && documents != null && documents.Any())
+                if (hasMappedPath)
                 {
-                    if (documents.First().TryGetText(out SourceText sourceText))
+                    if (documents != null && documents.Any() && documents.First().TryGetText(out SourceText sourceText))
                     {
-                        return (sourceText, true);
+                        // we have a mapped document that exists in workspace
+                        return sourceText;
                     }
+
+                    // we have a mapped document that doesn't exist in workspace
+                    // in that case we have to always fall back to original linespan
+                    return null;
                 }
 
-                return (location.SourceTree.GetText(), false);
+                // unmapped document so just continue with current SourceText
+                return location.SourceTree.GetText();
             }
 
-            static string GetText(Location location, SourceText sourceText, int startLine, bool isMappedDocument)
+            static string GetText(Location location, SourceText sourceText, int startLine)
             {
                 // bounds check in case the mapping is incorrect, since user can put whatever line they want
-                if (isMappedDocument && sourceText.Lines.Count > startLine)
+                if (sourceText != null && sourceText.Lines.Count > startLine)
                 {
                     return sourceText.Lines[startLine].ToString();
                 }
--- a/src/OmniSharp.Roslyn.CSharp/Helpers/LocationExtensions.cs
+++ b/src/OmniSharp.Roslyn.CSharp/Helpers/LocationExtensions.cs
@@ -22,11 +22,11 @@ namespace OmniSharp.Helpers
 
             var documents = workspace.GetDocuments(lineSpan.Path);
             var sourceText = GetSourceText(location, documents, lineSpan.HasMappedPath);
-            var text = GetText(location, sourceText, lineSpan.StartLinePosition.Line);
+            var text = GetLineText(location, sourceText, lineSpan.StartLinePosition.Line);
 
             return new QuickFix
             {
-                Text = text,
+                Text = text.Trim(),
                 FileName = lineSpan.Path,
                 Line = lineSpan.StartLinePosition.Line,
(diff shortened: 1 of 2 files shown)


src/OmniSharp.Roslyn.CSharp/Helpers/LocationExtensions.cs
```diff
@@ -17,22 +19,55 @@ public static QuickFix GetQuickFix(this Location location, OmniSharpWorkspace wo
...
                 Projects = documents.Select(document => document.Project.Name).ToArray()
             };
+
+            static SourceText GetSourceText(Location location, IEnumerable<Document> documents, bool hasMappedPath)
+            {
+                // if we have a mapped linespan and we found a corresponding document, pick that one
+                // otherwise use the SourceText of current location
+                if (hasMappedPath)
+                {
+                    if (documents != null && documents.Any() && documents.First().TryGetText(out SourceText sourceText))
```
NTaylorMullen: Would probably make more sense to grab the first document that is *able* to have their source text retrieved. Aka `FirstOrDefault(d => d.TryGetText(...)) != null`

src/OmniSharp.Roslyn.CSharp/Helpers/LocationExtensions.cs
```diff
@@ -17,22 +19,55 @@ public static QuickFix GetQuickFix(this Location location, OmniSharpWorkspace wo
...
+                }
+
+                // unmapped document so just continue with current SourceText
+                return location.SourceTree.GetText();
+            }
+
+            static string GetText(Location location, SourceText sourceText, int startLine)
+            {
+                // bounds check in case the mapping is incorrect, since user can put whatever line they want
+                if (sourceText != null && sourceText.Lines.Count > startLine)
```
NTaylorMullen: Can SourceText actually be `null` here?

Can the user put a negative line mapping accidentally?

src/OmniSharp.Roslyn.CSharp/Helpers/LocationExtensions.cs
```diff
@@ -17,22 +19,55 @@ public static QuickFix GetQuickFix(this Location location, OmniSharpWorkspace wo
...
+            var documents = workspace.GetDocuments(lineSpan.Path);
+            var sourceText = GetSourceText(location, documents, lineSpan.HasMappedPath);
+            var text = GetText(location, sourceText, lineSpan.StartLinePosition.Line);
 
             return new QuickFix
             {
-                Text = text.Trim(),
-                FileName = path,
-                Line = line,
+                Text = text,
```
NTaylorMullen: Curious: Should this be the entire text encompassed by the line mapping or just the first line? If it's the latter it might make sense to change the `GetText` method to be `GetLineText` to make it more clear

src/OmniSharp.Roslyn.CSharp/Helpers/LocationExtensions.cs
```diff
@@ -17,22 +19,55 @@ public static QuickFix GetQuickFix(this Location location, OmniSharpWorkspace wo
...
+                FileName = lineSpan.Path,
+                Line = lineSpan.StartLinePosition.Line,
                 Column = lineSpan.HasMappedPath ? 0 : lineSpan.StartLinePosition.Character, // when a #line directive maps into a separate file, assume columns (0,0)
                 EndLine = lineSpan.EndLinePosition.Line,
                 EndColumn = lineSpan.HasMappedPath ? 0 : lineSpan.EndLinePosition.Character,
                 Projects = documents.Select(document => document.Project.Name).ToArray()
             };
+
+            static SourceText GetSourceText(Location location, IEnumerable<Document> documents, bool hasMappedPath)
```
NTaylorMullen: Would probably make more sense if this method just took in the workspace and line soan instead of the documents and the "IsMapped" flag since it's purpose is to locate the source text for a given location/mapping

src/OmniSharp.Roslyn.CSharp/Helpers/LocationExtensions.cs
filipw: it should be the line, I changed it accordingly 👍

src/OmniSharp.Roslyn.CSharp/Helpers/LocationExtensions.cs
filipw: this is a very good idea

src/OmniSharp.Roslyn.CSharp/Helpers/LocationExtensions.cs
filipw: 1. yes if we can't find the mapped document in the workspace
2. yes - then the mapping doesn't resolve at in Roslyn and we will return the current file and current line number instead. I added an extra test for this so that the behavior is captured and won't change

src/OmniSharp.Roslyn.CSharp/Helpers/LocationExtensions.cs
filipw: the documents we need outside again for the response and we don't want to fetch them twice, so we'd need to make a tuple return or switch to non static local functions and capture variables. also the location is needed to fetch the SourceText

NTaylorMullen: Ah missed that you need them outside as well

filipw:
@JoeRobich @david-driscoll any objections? can this be merged?

src/OmniSharp.Roslyn.CSharp/Helpers/LocationExtensions.cs
```diff
@@ -17,22 +19,56 @@ public static QuickFix GetQuickFix(this Location location, OmniSharpWorkspace wo
...
+            {
+                // if we have a mapped linespan and we found a corresponding document, pick that one
+                // otherwise use the SourceText of current location
+                if (hasMappedPath)
+                {
+                    SourceText source = null;
+                    if (documents != null && documents.Any() && documents.FirstOrDefault(d => d.TryGetText(out source)) != null)
```
JoeRobich: I'm not as familiar with line mappings. Would it be more correct to loop through the documents where you can get text and see if the mapping applies?

JoeRobich: `documents.Any()` isn't necessary now that you are using `FirstOrDefault`

filipw: I think this can only ever return more than 1 in case of multi targeting so the document text would be identical

david-driscoll:
:shipit:

PR #1710, opened 2020-02-12
Add ability to use custom runsettings for tests

Samirat:
Currently, the RunSettings are sent to Omnisharp as a a path, and then loaded. Alternatively, we could load the runsettings in the extension, and then send XML with the request.

Related PR to omnisharp-vscode: https://github.com/OmniSharp/omnisharp-vscode/pull/3573

### DIFF
--- a/src/OmniSharp.DotNetTest/VSTestManager.cs
+++ b/src/OmniSharp.DotNetTest/VSTestManager.cs
@@ -30,7 +30,18 @@ namespace OmniSharp.DotNetTest
         {
             if (runSettingsPath != null)
             {
-                return File.ReadAllText(runSettingsPath);
+                try
+                {
+                    return File.ReadAllText(runSettingsPath);
+                }
+                catch (FileNotFoundException)
+                {
+                    EmitTestMessage(TestMessageLevel.Warning, $"RunSettings file {runSettingsPath} not found. Continuing with default settings...");
+                }
+                catch (Exception e)
+                {
+                    EmitTestMessage(TestMessageLevel.Warning, $"There was an error loading runsettings at {runSettingsPath}: {e}. Continuing with default settings...");
+                }
             }
 
             if (!string.IsNullOrWhiteSpace(targetFrameworkVersion))
--- a/src/OmniSharp.DotNetTest/VSTestManager.cs
+++ b/src/OmniSharp.DotNetTest/VSTestManager.cs
@@ -26,8 +26,13 @@ namespace OmniSharp.DotNetTest
         {
         }
 
-        private object GetDefaultRunSettings(string targetFrameworkVersion)
+        private object LoadRunSettingsOrDefault(string runSettingsPath, string targetFrameworkVersion)
         {
+            if (runSettingsPath != null)
+            {
+                return File.ReadAllText(runSettingsPath);
+            }
+
             if (!string.IsNullOrWhiteSpace(targetFrameworkVersion))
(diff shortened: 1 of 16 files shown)


src/OmniSharp.DotNetTest/VSTestManager.cs
```diff
@@ -101,18 +106,18 @@ private static void VerifyTestFramework(string testFrameworkName)
             }
         }
 
-        public override GetTestStartInfoResponse GetTestStartInfo(string methodName, string testFrameworkName, string targetFrameworkVersion)
+        public override GetTestStartInfoResponse GetTestStartInfo(string methodName, string runSettings, string testFrameworkName, string targetFrameworkVersion)
```
filipw: normally we do not break  the public APIs but this particular package is no published on nuget, so it's OK

src/OmniSharp.DotNetTest/VSTestManager.cs
```diff
@@ -26,8 +26,13 @@ public VSTestManager(Project project, string workingDirectory, IDotNetCliService
         {
         }
 
-        private object GetDefaultRunSettings(string targetFrameworkVersion)
+        private object LoadRunSettingsOrDefault(string runSettingsPath, string targetFrameworkVersion)
         {
+            if (runSettingsPath != null)
+            {
+                return File.ReadAllText(runSettingsPath);
```
filipw: you should check if the file exists, if not, I'd fallback to defaults to not crash?

tests/OmniSharp.DotNetTest.Tests/RunTestFacts.cs
```diff
@@ -223,5 +223,27 @@ public async Task RunMSTestStandardOutputIsReturned()
...
+                methodName: "Main.Test.MainTest.CheckRunSettings",
+                testFramework: "mstest",
+                shouldPass: true,
+                useRunSettings: true);
+        }
+
+        [Fact]
+        public async Task RunMSTestWithoutRunSettings()
+        {
+            await RunDotNetTestAsync(
```
filipw: if this is supposed to fail, could you assert on some failure conditions to distinguish a "failure by design" from "failure because something got broken"?

filipw:
thanks a lot for the PR - looks very good, just left some small comments

src/OmniSharp.DotNetTest/VSTestManager.cs
Samirat: Added an error message if runsettings load fails, and it will continue with default settings.

tests/OmniSharp.DotNetTest.Tests/RunTestFacts.cs
Samirat: Added some additional assertions.

src/OmniSharp.DotNetTest/VSTestManager.cs
Samirat: Even though these functions are public, TestManager and VSTestManager are actually internal classes, so there probably isn't any worry.

Samirat: I was more concerned about the changes to the over-the-wire interface. At least the stdio version of the server seems to treat arguments as optional, in which case I think everything remains compatible.

filipw:
@JoeRobich @mholo65 @david-driscoll  any objections on this?

PR #1715, opened 2020-02-19
Fixed a bug where organizing usings clashed with other formatting settings

filipw:
when organize usings was enabled, it prevented the regular formatting from working since it passed in the original document... 🙈

@JoeRobich after this is merged, let's release again so that you don't ship the buggy version in VS Code


filipw:
thanks! 😀💪

PR #1734, opened 2020-03-15
Add SemanticHighlight Endpoint

JoeRobich:
Implementation of a semantic highlight endpoint. The data that will eventually be returned is different enough from the Highlight endpoint that I thought a new one was justified. This is a simple implementation that supports classification and the static modifier.

As requested in https://github.com/OmniSharp/omnisharp-vscode/issues/3565


tests/OmniSharp.Roslyn.CSharp.Tests/SemanticHighlightFacts.cs
```diff
@@ -0,0 +1,191 @@
...
+using OmniSharp.Models.SemanticHighlight;
+using OmniSharp.Models.V2;
+using OmniSharp.Roslyn.CSharp.Services.SemanticHighlight;
+using TestUtility;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace OmniSharp.Roslyn.CSharp.Tests
+{
+    public class SemanticHighlightFacts : AbstractSingleRequestHandlerTestFixture<SemanticHighlightService>
```
filipw: can we have one more test with a static that checks that `AdditiveTypeNames` have been excluded?

src/OmniSharp.Abstractions/Models/v2/SemanticHighlight/SemanticHighlightModifier.cs
```diff
@@ -0,0 +1,6 @@
+namespace OmniSharp.Models.SemanticHighlight
+{
+    public enum SemanticHighlightModifier
```
filipw: this doesn't seem to be used?

filipw:
looks good, thanks for adding this 👍
I just left some small comments

tests/OmniSharp.Roslyn.CSharp.Tests/SemanticHighlightFacts.cs
```diff
@@ -0,0 +1,191 @@
...
+namespace N1
+{
+    class C1 { int n = true; }
+}
+");
+
+            var line = 3;
+            var highlights = await GetSemanticHighlightsForLineAsync(testFile, line);
+
+            Console.WriteLine(string.Join("\n", highlights.Select(span => span.Type.ToString())));
```
filipw: it's better to use `ITestOutputHelper`, it's on the base class

src/OmniSharp.Abstractions/OmniSharpEndpoints.cs
```diff
@@ -60,6 +60,8 @@ public static class V2
 
             public const string BlockStructure = "/v2/blockstructure";
             public const string CodeStructure = "/v2/codestructure";
+
+            public const string SemanticHighlight = "/v2/semantichighlight";
```
filipw: I think we can just call it `/v2/highlight`, as in an improved version of the existing classification
it's safe to assume that there won't be a v2 of the existing highlighting endpoint - we will only have this one

src/OmniSharp.Abstractions/Models/v2/SemanticHighlight/SemanticHighlightModifier.cs
JoeRobich: Although the work isn't planned, we can expect that Rolsyn will introduce a classification API which provides this additional information about classified spans. I wanted the API to reflect it so that consumers could expect it being returned. I could probably process the classifications and return a static modifier...

tests/OmniSharp.Roslyn.CSharp.Tests/SemanticHighlightFacts.cs
JoeRobich: Thanks, I actually should have removed that once I worked out why ranges weren't returning.

src/OmniSharp.Abstractions/OmniSharpEndpoints.cs
JoeRobich: Sure thing

tests/OmniSharp.Roslyn.CSharp.Tests/SemanticHighlightFacts.cs
JoeRobich: Better yet. I added a test to ensure it gets used =)

NTaylorMullen:
@ryanbrandenburg since you own our semantic classification you should take a look at this as well. It will also enable us to call through to Roslyn for semantic colorization as well 😄

src/OmniSharp.Roslyn.CSharp/Services/SemanticHighlight/SemanticHighlightService.cs
```diff
@@ -0,0 +1,174 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Classification;
+using Microsoft.CodeAnalysis.Text;
+using OmniSharp.Mef;
+using OmniSharp.Models.SemanticHighlight;
+
+namespace OmniSharp.Roslyn.CSharp.Services.SemanticHighlight
+{
+    [OmniSharpHandler(OmniSharpEndpoints.V2.Highlight, LanguageNames.CSharp)]
+    public class SemanticHighlightService : IRequestHandler<SemanticHighlightRequest, SemanticHighlightResponse>
```
NTaylorMullen: Just to clarify, this current impl doesn't take advantage of the partial semantic classification tech that's available right? 

Oh and "partial" in the sense of only updating a single lines classification if only that line changed.

I ask because for C# files this might be a huge performance hit.

JoeRobich: The highlight request has an optional range property so that only that TextSpan is classified.

NTaylorMullen: I thought VSCode re-requested the entire document pretty frequently though to account for external file changes, project changes etc. Maybe I'm wrong though, my information is coming from design discussions during the features creation; not out of personal trial and error.

JoeRobich: Ah, perhaps you are referring to the [`provideDocumentSemanticTokensEdits` api](https://github.com/microsoft/vscode/blob/d618f85bee6cfa2ce1b8eb50ca369dca58664e89/src/vs/vscode.proposed.d.ts#L292-L345)

NTaylorMullen: Yup exactly!

JoeRobich: After thinking about that API for a bit, I'm not sure the Edits api is super useful. It seems mostly to benefit requesting changes from a remote LSP where bandwidth is a concern. It also requires the LSP to hold on to previous state in order to generate the Edits diff as well as fully classify the new state of the document.

NTaylorMullen: I'm not entirely sure how the client interprets all of the data in a response but I imagine it'd also optimize the clients processing of the data

JoeRobich: Yeah, I feel like that api would basically be faked by the editor assuming they have a good method of generating the diffs from two sets of SyntaxTokens. Since they already have the before state which the LSP really shouldn't be expected to hold on to.

PR #1746, opened 2020-03-24
Update the minimal MSBuild to better support .NET 5 Previews

JoeRobich:
Update the msbuild components to match 16.5 releases. Update the nuget components to a preview that supports the net5.0 tfm.

Tested the minimal MSBuild with a .NET 5.0 Preview 2 app on both Windows and MacOS (Mono 6.8.0).

Resolves https://github.com/OmniSharp/omnisharp-vscode/issues/3653


filipw:
you should update here too https://github.com/OmniSharp/omnisharp-roslyn/blob/master/build/Packages.props#L6-L7

JoeRobich:
> you should update here too https://github.com/OmniSharp/omnisharp-roslyn/blob/master/build/Packages.props#L6-L7

OK. I've learned a lot today trying to update the NuGet version in Packages.props. It effectively becomes the low version of which SDKs are supported. The current version of 5.2.0 is good because that lines up with the 2.1.x SDK which is LTS.

Setting the version above 5.3.0 (3.0.x SDK) causes MissingMethod exceptions during design time builds similar to https://github.com/dotnet/core/issues/3402. This makes sense because the global.json used sets the SDK to 3.0.100 which uses 5.3.0. 

So for the time being, I left the NuGet version as is. Wonder if we effectively need a NuGetLocator ( ala MSBuildLocator ) that loads the NuGet dependencies from the MSBuild path.

filipw:
I see, thanks a lot.

stijnherreman:
It is still listed as 16.4 (while it actually is 16.5), I think a change similar to this commit needs to be done?

https://github.com/OmniSharp/omnisharp-roslyn/commit/77c4cd357e7001e037ab04485b9bf8fc949941a4

PR #1748, opened 2020-03-26
Include Region blocks in code folding

JoeRobich:
For Roslyn to tag Region blocks, options in BlockStructureOptions need to be set to true. We also need to convert from the type PreprocessorRegion to Region in the BlockStructureService.

Resolves https://github.com/OmniSharp/omnisharp-vscode/issues/2621

### DIFF
--- a/src/OmniSharp.Roslyn.CSharp/Services/BlockStructureWorkspaceOptionsProvider.cs
+++ b/src/OmniSharp.Roslyn.CSharp/Services/BlockStructureWorkspaceOptionsProvider.cs
@@ -1,0 +1,45 @@
+using System;
+using System.Composition;
+using System.Reflection;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Options;
+using OmniSharp.Options;
+using OmniSharp.Roslyn.Options;
+using OmniSharp.Services;
+using OmniSharp.Utilities;
+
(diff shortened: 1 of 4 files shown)


src/OmniSharp.Roslyn.CSharp/Services/BlockStructureWorkspaceOptionsProvider.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        [ImportingConstructor]
+        public BlockStructureWorkspaceOptionsProvider(IAssemblyLoader assemblyLoader)
+        {
+            _assemblyLoader = assemblyLoader;
+            _csharpFeatureAssembly = _assemblyLoader.LazyLoad(Configuration.RoslynFeatures);
+            _blockStructureOptions = _csharpFeatureAssembly.LazyGetType("Microsoft.CodeAnalysis.Structure.BlockStructureOptions");
+        }
+
+        public int Order => 100;
```
filipw: could you give it a different (unique) order? at the moment we have values 0, 100, 110, 200 already on various providers

the goal was to have them unique so that the applying of options is deterministic

JoeRobich: Sure thing!

PR #1751, opened 2020-04-01
Added support for decompilation based on ILSpy (similar to VS)

filipw:
This is implemented as a variation on the /gotodefinition and /metadata endpoints.
Setting the flag `RoslynExtensionsOptions:EnableDecompilationSupport` to `true` means the user opts into decompilation instead of metadata.

The changeset is relatively big because I needed to refactor the existing metadata service and then matched the new decompilation logic to it, this way metadata services and decompilation are fully interchangeable.

No client changes needed and you can navigate from decompiled code to another symbol that needs to be decompiled (just like we support in metadata at the moment)

preview: http://g.recordit.co/9XaNkEE4lt.gif


src/OmniSharp.Host/OmniSharp.Host.csproj
```diff
@@ -11,6 +11,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\OmniSharp.Abstractions\OmniSharp.Abstractions.csproj" />
+    <ProjectReference Include="..\OmniSharp.Roslyn.CSharp\OmniSharp.Roslyn.CSharp.csproj" />
```
mholo65, line 14: Is this needed?

src/OmniSharp.Roslyn.CSharp/Services/Decompilation/AssemblyResolver.cs
```diff
@@ -0,0 +1,126 @@
...
+        private readonly Dictionary<string, List<IAssemblySymbol>> _cache = new Dictionary<string, List<IAssemblySymbol>>();
+        private readonly ILogger _logger;
+
+        public AssemblyResolver(Compilation parentCompilation, ILoggerFactory loggerFactory)
+        {
+            _parentCompilation = parentCompilation;
+            _logger = loggerFactory.CreateLogger<AssemblyResolver>();
+            BuildReferenceCache();
+
+            void BuildReferenceCache()
```
mholo65, line 27: nit, local function doesn't bring any value here.

src/OmniSharp.Roslyn.CSharp/Services/ExternalSourceServiceFactory.cs
```diff
@@ -0,0 +1,77 @@
...
+            _loggerFactory = loggerFactory;
+        }
+
+        public IExternalSourceService Create(OmniSharpOptions omniSharpOptions, HostLanguageServices hostLanguageServices)
+        {
+            var enableDecompilationSupport = omniSharpOptions.RoslynExtensionsOptions.EnableDecompilationSupport;
+
+            if (enableDecompilationSupport)
+            {
+                if (_decompilationExternalSourceService == null)
```
mholo65, line 38: Can we solve this by having a `Lazy<DecompilationExternalSourceService>` instead. Create the delegate in constructor where we have assembly loader and logger factory.

src/OmniSharp.Roslyn.CSharp/Services/ExternalSourceServiceFactory.cs
```diff
@@ -0,0 +1,77 @@
...
+                        {
+                            _decompilationExternalSourceService = new DecompilationExternalSourceService(_assemblyLoader, _loggerFactory, hostLanguageServices);
+                        }
+                    }
+                }
+
+                return _decompilationExternalSourceService;
+            }
+
+            if (_metadataExternalSourceService == null)
```
mholo65, line 52: Can we solve this by having a `Lazy<MetadataExternalSourceService>` instead. Create the delegate in constructor where we have assembly loader and logger factory.

src/OmniSharp.Roslyn.CSharp/Services/Decompilation/AssemblyResolver.cs
filipw: this is internalized from here http://sourceroslyn.io/#Microsoft.CodeAnalysis.CSharp.EditorFeatures/DecompiledSource/AssemblyResolver.cs,29

I would like to keep all the code intact  (except the one place we needed a change)

src/OmniSharp.Host/OmniSharp.Host.csproj
filipw: no, left over from a  different approach

src/OmniSharp.Roslyn.CSharp/Services/ExternalSourceServiceFactory.cs
filipw: I actually changed it all to singletons managed by MEF now.
it used to be so that it required request data to first initialize, but turns out that's not needed

src/OmniSharp.Roslyn.CSharp/Services/Decompilation/OmniSharpCSharpDecompiledSourceService.cs
```diff
@@ -0,0 +1,191 @@
...
+            // Get the name of the type the symbol is in
+            var containingOrThis = symbol.GetContainingTypeOrThis();
+            var fullName = GetFullReflectionName(containingOrThis);
+
+            string assemblyLocation = null;
+            var isReferenceAssembly = symbol.ContainingAssembly.GetAttributes().Any(attribute => attribute.AttributeClass.Name == nameof(ReferenceAssemblyAttribute)
+                && attribute.AttributeClass.ToNameDisplayString() == typeof(ReferenceAssemblyAttribute).FullName);
+            if (isReferenceAssembly)
+            {
+                try
```
baronfel, line 76: Does this logic let you resolve the contents of ref assemblies to their runtime equivalents?  If so I'll port this over to ionide, because we currently have this same problem with resolving ref assemblies instead of implementation assemblies.

filipw: it will only work when the implementation is in the GAC (this is the same as in VS)

baronfel: Gotcha, so for .Net Core tooing you'd need to find the matching non-ref assembly locations and decompile those instead. About what I figured :) Thanks for clarifying.

src/OmniSharp.Roslyn.CSharp/Services/Decompilation/OmniSharpCSharpDecompiledSourceService.cs
```diff
@@ -0,0 +1,191 @@
...
+                }
+            }
+
+            // Decompile
+            document = PerformDecompilation(document, fullName, symbolCompilation, assemblyLocation);
+
+            document = await AddAssemblyInfoRegionAsync(document, symbol, cancellationToken).ConfigureAwait(false);
+
+            // Convert XML doc comments to regular comments, just like MAS
+            //var docCommentFormattingService = document.GetLanguageService<IDocumentationCommentFormattingService>();
```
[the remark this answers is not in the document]
filipw: no ,forgot this, thanks

src/OmniSharp.Roslyn.CSharp/Services/ExternalFeaturesHostServicesProvider.cs
```diff
@@ -2,6 +2,9 @@
 using System.Composition;
 using System.Linq;
 using System.Reflection;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Host;
+using Microsoft.CodeAnalysis.Host.Mef;
```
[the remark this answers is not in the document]
filipw: I probably started working on something inside that file and then moved it out 😂

filipw:
ok, approved then I'm merging this

PR #1761, opened 2020-04-10
Optimize Intellisense performance

SirIntruder:
https://github.com/OmniSharp/omnisharp-vscode/issues/3421

Auto complete popup always seemed to have a spotty performance and random delays, so I drilled the problem in VS profiler. Luckily, seems that a lot of it is very low hanging fruit. 

Main culprit seems to be `CompletionItemExtensions`, which did a complex search on `recommendedSymbols` times every `completionItem` in the dropdown. RecommendedSymbols seems like something that should be a very short list, but this still throttles the CPU. 

Lesser issue was a collection of string extensions methods used for sorting the resulting competition list. VS (or Rolsynator?) immediately started yelling in red squiggles about using StringComparison everywhere, so I did. Luckily `StringExtensions` class was only used by Intellisense service, so there are no worries about breaking something else.

<details>
  <summary>Before vs After</summary>
 
bad
good

  "Before" gif doesn't capture the "cpu melting noise" part of the experience
</details>


src/OmniSharp.Roslyn.CSharp/Services/Intellisense/CompletionItemExtensions.cs
```diff
@@ -60,9 +60,13 @@ public static async Task<IEnumerable<ISymbol>> GetCompletionSymbolsAsync(this Co
             }
 
             // if the completion provider encoded symbols into Properties, we can return them
-            if (properties.ContainsKey(SymbolName) && properties.ContainsKey(SymbolKind))
+            if (properties.TryGetValue(SymbolName, out string symbolNameValue)
+                && properties.TryGetValue(SymbolKind, out string symbolKindValue)
```
SirIntruder, line 64: Calls to the ImmutableDictionary.GetValue<> were the main CPU-hog, now they are cached outside of the loop.

src/OmniSharp.Roslyn.CSharp/Services/Intellisense/IntellisenseService.cs
```diff
@@ -51,7 +51,7 @@ public async Task<IEnumerable<AutoCompleteResponse>> Handle(AutoCompleteRequest
 
                     // get recommended symbols to match them up later with SymbolCompletionProvider
                     var semanticModel = await document.GetSemanticModelAsync();
-                    var recommendedSymbols = await Recommender.GetRecommendedSymbolsAtPositionAsync(semanticModel, position, _workspace);
+                    var recommendedSymbols = (await Recommender.GetRecommendedSymbolsAtPositionAsync(semanticModel, position, _workspace)).ToArray();
```
SirIntruder, line 54: Collecting recommendedSymbols into an array, instead of passing Rolsyn's opaque collection type, pulled a lot of weight here. Note that this gets the expensive LINQ call inside the BigN loop (number of elements in the auto complete dropdown).

Actually tried to test with this change only, and it does help a lot, but only does half of the job.

PR #1767, opened 2020-04-14
removed any attempt to handle ref assemblies via GAC when decompiling

filipw:
See the discussion here https://github.com/icsharpcode/ilspy-vscode/issues/45#issuecomment-613473929

Long story short, searching for ref assemblies in GAC results in no-definition found - Not Supported exception on Mono. This means that we can't decompile into BCL types on Mono. On Windows it "works" - as in, it doesn't crash, but it doesn't find implementation assemblies anyway.

This PR removes any attempt to handle ref assemblies; this means things should always decompile fine, but we won't find implementation assemblies. This is consistent with VS experience where ref assemblies are decompiled instead of implementation assemblies.

It's a super complex topic, and we can improve on that gradually in the future (or maybe VS implementation will improve).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 All changes to the project will be documented in this file.
 
+## [1.35.1] - not yet released
+* Fixed not supported exception when trying to decompile a BCL assembly on Mono (PR: [#1767](https://github.com/OmniSharp/omnisharp-roslyn/pull/1767))
+
 ## [1.35.0] - 2020-04-10
 * Support for `<RunAnalyzers />` and `<RunAnalyzersDuringLiveAnalysis />` (PR: [#1739](https://github.com/OmniSharp/omnisharp-roslyn/pull/1739))
 * Add `typeparam` documentation comments to text description ([omnisharp-vscode#3516](https://github.com/OmniSharp/omnisharp-vscode/issues/3516), PR: [#1749](https://github.com/OmniSharp/omnisharp-roslyn/pull/1749))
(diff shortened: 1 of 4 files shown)


SirIntruder:
Thanks for this! Decompilation of all System types were failing in Unity projects (that target .NetStandard API profile, which is default), I built the branch locally and it got `System.String` and friends decompiling again.

btw you mentioned it didn't crash, but I did find these exceptions on Windows previously (with logging set to `debug`).

`  "Request_seq": 261,
  "Command": "/gotoDefinition",
  "Running": true,
  "Success": false,
  "Message": "\"System.NotSupportedException: Decompiling types that are not part of the main module is not supported.\\r\\n   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.DecompileType(FullTypeName fullTypeName)\\r\\n   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.DecompileTypeAsString(FullTypeName fullTypeName)\\r\\n   at OmniSharp.Roslyn.CSharp.Services.Decompilation.OmniSharpCSharpDecompiledSourceService.PerformDecompilation(Document document, String fullName, Compilation compilation, String assemblyLocation) `

filipw:
yeah it's the same exception. It would happen when the GAC entry for the implementation assembly was indeed found but at different version, which technically can happen on Windows too (although I have not seen it). the problem for finding the runtime assembly, especially from a different runtime is really complex and for now we should keep it simple. 
also the original code was internalized from Roslyn (the code editor assembly which is Windows only) and it was never intended to be used on Mono so that only makes the problem worse

with this change it should just continue to the ref assembly, which is of limited value but it is a step in the right direction

Issue #1657, opened 2019-11-22
Use Roslyn's native .editorconfig support


JoeRobich:
EditorConfigDocumentOptionsProvider pulls options from Workspace.AnalyzerDocuments. Need to register this OptionsProvider with the OptionService during Workspace initialization.

coder2000:
Is this supported now?

filipw:
not yet

coder2000:
Any timeline for implementation? It gets kind of annoying having 800 warnings that are not supposed to be there.

filipw:
> Any timeline for implementation?

no timeline, hopefully soon

333fred:
Also looking for this, I'm assuming my issues with analyzer settings not being respected are due to this.

filipw:
yes correct, when we implemented editorconfig there was no native Roslyn support yet so we had to use the VS conventions package like dotnet format did at the time

PR #1771, opened 2020-04-18
Use Roslyn's editorconfig support

JoeRobich:
Resolves https://github.com/OmniSharp/omnisharp-roslyn/issues/1657


src/OmniSharp.Roslyn.CSharp/Workers/Formatting/FormattingWorker.cs
```diff
@@ -104,8 +100,8 @@ public static async Task<IEnumerable<LinePositionSpanTextChange>> GetFormattedTe
         private static async Task<Document> FormatDocument(Document document, OmniSharpOptions omnisharpOptions, ILoggerFactory loggerFactory, TextSpan? textSpan = null)
         {
             var optionSet = omnisharpOptions.FormattingOptions.EnableEditorConfigSupport
-                ? await document.Project.Solution.Workspace.Options.WithEditorConfigOptions(document.FilePath, loggerFactory)
-                : document.Project.Solution.Workspace.Options;
+                ? await document.GetOptionsAsync()
```
filipw, line 103: oh I see, so this will inlcude `EditorConfigDocumentOptionsProviderFactory` in the options computation?

src/OmniSharp.MSBuild/ProjectManager.cs
```diff
@@ -483,10 +484,27 @@ private void UpdateAdditionalFiles(Project project, IList<string> additionalFile
...
+                {
                     _workspace.AddAdditionalDocument(project.Id, file);
-                 }
+                }
+            }
+        }
+
+        private void UpdateAnalyzerConfigFiles(Project project, IList<string> analyzerConfigFiles)
+        {
+            var currentAnalyzerConfigDocuments = project.AnalyzerConfigDocuments;
```
filipw, line 496: what's the value of `AnalyzerConfigDocuments` if it's not set in the project file explicitly? Should you not fallback here to `EditorConfigFinder`?

filipw:
thanks a lot, this is a very welcome change!

I left some small comments and just have a few questions:
 - should we not add the discovered editorconfig to the watcher here https://github.com/OmniSharp/omnisharp-roslyn/blob/3163c234bc0cb2ece67624791a66d735b295b907/src/OmniSharp.MSBuild/ProjectManager.cs#L371 ? so far it was not needed because the options were reapplied before formatting
 - it would be good to have some extra test similar to this one https://github.com/OmniSharp/omnisharp-roslyn/blob/3163c234bc0cb2ece67624791a66d735b295b907/tests/OmniSharp.MSBuild.Tests/ProjectWithAnalyzersTests.cs#L67 that verifies that analyzer diagnostic options from editorconfig are correctly loaded
 - I wasn't aware of the `EditorConfigFiles` msbuild item - in which versions of VS is this supported? (I think in the past it wasn't possible to provide your own custom file path)

src/OmniSharp.MSBuild/ProjectManager.cs
```diff
@@ -483,10 +484,27 @@ private void UpdateAdditionalFiles(Project project, IList<string> additionalFile
...
-                 {
+                if (File.Exists(file))
+                {
                     _workspace.AddAdditionalDocument(project.Id, file);
-                 }
+                }
+            }
+        }
+
+        private void UpdateAnalyzerConfigFiles(Project project, IList<string> analyzerConfigFiles)
```
filipw, line 494: should this code be conditioned on editorconfig being enabled in OmniSharp?
otherwise, if you disable it, it would be respected on formatting, but the analyzer options would still apply

src/OmniSharp.Cake/CakeProjectSystem.cs
```diff
@@ -292,7 +306,8 @@ private ProjectInfo GetProject(CakeScript cakeScript, string filePath)
                 metadataReferences: GetMetadataReferences(cakeScript.References),
                 // TODO: projectReferences?
                 isSubmission: true,
-                hostObjectType: hostObjectType);
+                hostObjectType: hostObjectType)
+                .WithAnalyzerConfigDocuments(analyzerConfigDocuments);
```
filipw, line 310: I think the same should be applied here https://github.com/OmniSharp/omnisharp-roslyn/blob/3163c234bc0cb2ece67624791a66d735b295b907/src/OmniSharp.MSBuild/ProjectFile/ProjectFileInfoExtensions.cs#L61 as well?

src/OmniSharp.Roslyn.CSharp/Workers/Formatting/FormattingWorker.cs
JoeRobich: Right, since I opened the issue, we have moved the OptionsProvider registration into the base Workspace constructor.

src/OmniSharp.MSBuild/ProjectManager.cs
JoeRobich: The AnalyzerConfigDocuments property is computed by the build targets. It is present in build tools for 16.3 and higher, so should be present based on the supported msbuild versions.

JoeRobich:
> * it would be good to have some extra test similar to this one

Yep, I agree

> should this code be conditioned on editorconfig being enabled in OmniSharp?

Yeah, there are a few places to update if we want to disable the analyzer part. With it being under FormattingOptions I wasn't sure the expected behavior.

I'll update soon.

src/OmniSharp.MSBuild/ProjectManager.cs
filipw: ah good

src/OmniSharp.MSBuild/ProjectManager.cs
```diff
@@ -377,6 +377,11 @@ private void WatchProjectFiles(ProjectFileInfo projectFileInfo)
                 QueueProjectUpdate(projectFileInfo.FilePath, allowAutoRestore: true, projectFileInfo.ProjectIdInfo);
             });
 
+            _fileSystemWatcher.Watch(".editorconfig", (file, changeType) =>
```
filipw, line 380: should this watch on `projectFileInfo.AnalyzerConfigFiles` instead?

tests/OmniSharp.MSBuild.Tests/ProjectWithAnalyzersTests.cs
```diff
@@ -75,6 +75,54 @@ public async Task WhenProjectIsLoadedThenItContainsCustomRulesetsFromCsproj()
             }
         }
 
+        [Fact]
+        public async Task WhenProjectIsLoadedThenItContainsAnalyzerConfigurationFromEditorConfig()
```
filipw, line 79: 👍👍

src/OmniSharp.MSBuild/ProjectManager.cs
JoeRobich: Would it still need a catch-all to handle new .editorconfigs being added to subfolders? I expected this would be easy way of handling Add and Update.

src/OmniSharp.Host/WorkspaceInitializer.cs
```diff
@@ -29,6 +29,8 @@ public static void Initialize(IServiceProvider serviceProvider, CompositionHost
             projectEventForwarder.Initialize();
             var projectSystems = compositionHost.GetExports<IProjectSystem>();
 
+            workspace.EditorConfigEnabled = options.CurrentValue.FormattingOptions.EnableEditorConfigSupport;
```
filipw, line 32: it would be good to also reset this flag here https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp.Host/WorkspaceInitializer.cs#L62-L65
this way it should be effective in real time when options change (since they are watched)

JoeRobich: This would be useful if there were a way to force all the projects to update. Is this option worth all this extra complexity? I would assume the presence of an .editorconfig would be the indicator of whether a user wanted .editorconfig support.

filipw: you are right we can solve this systematically separately as there are other areas that would benefit from that

PR #1772, opened 2020-04-21
Watch .editorconfig files in folders above the project folder.

JoeRobich:
As [mentioned here](https://github.com/OmniSharp/omnisharp-roslyn/pull/1771#discussion_r411989013), we should also be looking for changes to .editorconfig files in folders above the project folder.


JoeRobich:
@filipw I should have thought through your suggestion more completely. Didn't think about .editorconfigs outside the project folder.

PR #1774, opened 2020-04-26
Fully update CompilationOptions when project files change.

SirIntruder:
When `.csproj` files change, o# gets fully updated `CSharpCompilationOptions` from MSBuild, but it doesn't apply them to the workspace. Because of this,  o# completely ignores changes to `AllowUnsafeCode`, `Nullable`, `DefineConstants`, `LangVersion` etc. after the initial project load.

Fortunately, o# was updating `CSharpCompilationOptions` already, for the purpose of updating analyzer rulesets. So fix amounted to passing 

 `projectFileInfo.CreateCompilationOptions()` 

instead of 

 `project.CompilationOptions.WithSpecificDiagnosticOptions(projectFileInfo.GetDiagnosticOptions())`

Problem with the old code was that `project.CompilationOptions` was essentially old compiler options. `ProjectFileInfo` is fresh data, and with this PR the whole CompilerOptions object will be fresh, not just the DiagnosticsOptions part.


SirIntruder:
I will make a test when I catch some more time this week, need to check how FileChanges are handled in the testHost.

filipw:
thanks a lot this very helpful.

> Because of this, o# completely ignores changes to `AllowUnsafeCode`, `Nullable`, `DefineConstants`, `LangVersion` etc. after the initial project load.

this is not entirely true, because OmniSharp already automatically updates parse options (like `DefineConstants`, `LangVersion`). Only compilation options are ignored.

JoeRobich:
@SirIntruder Thanks!

PR #1776, opened 2020-04-28
removed invalid comment

filipw:
removed misleading comment that was not removed during refactoring


filipw:
there is nothing to review here 😀

PR #1782, opened 2020-05-07
Add RunTestsInContext Command

333fred:
This command tries to replicate the `CTRL+R, T` experience in full VS. ~~I'm planning on sending a followup PR hopefully this weekend to add DebugInContext~~ (now in this PR), and then a PR to the vscode extension to enable the commands (https://github.com/OmniSharp/omnisharp-vscode/pull/3772). While Omnisharp does have an existing `TestCommandService` that does some of this, it doesn't quite replicate the experience: it will only run the test method that it's invoked in, not potentially the entire test class or whatever other context it's in. Additionally, it's not unified with the test running framework the test of the V2 test commands use, so I opted to introduce an entirely new command, RunTestsInContext, which will walk up the syntax tree for a given location to find the nearest test method or class, if one exists, and either run the method itself, or run the entire class.

Commit 1 moves Omnisharp forward to the 3.1 LTS SDK, from 3.0.
Commit 2 does some refactoring of the existing test commands to remove `.Result` usage and plumb async all the way through.
Commit 3 Implements RunTestsInContext.
Commit 4 is some cleanup.
Commit 5 Implements DebugTestsInContextGetStartInfo.
Commits 6-13 are some cleanup.


test-assets/test-projects/MSTestProject/TestProgram.cs
```diff
@@ -12,7 +12,7 @@ public class MainTest
         public void Test()
         {
             Assert.IsTrue(true);
-        }
+        }
```
333fred, line 15: This space (and the ones in the other test projects) are not a typo. They're there for testing the boundary cases.

src/OmniSharp.DotNetTest/TestFrameworks/TestFramework.cs
```diff
@@ -7,36 +7,39 @@ namespace OmniSharp.DotNetTest.TestFrameworks
 {
     internal abstract class TestFramework
     {
-        private static readonly ImmutableDictionary<string, TestFramework> s_frameworks;
+        private static readonly ImmutableArray<TestFramework> s_frameworks;
```
333fred, line 10: It's a neat idea to make this extensible, but with the amount of effort that would need to be put in to add another framework makes this a pointless optimization. It causes `GetFrameworks()` to allocate an iterator on every invocation (which is often), and even in the worst case is `GetFramework(string)` implemented as a foreach loop over an array will be twice as fast as a dictionary lookup: https://gist.github.com/333fred/2c259fff1ca0e074adead95f01db70ba.

333fred:
I've now added `DebugTestsInContextGetStartInfo`. I don't see any existing tests for the debugging commands: am I just missing something? If not, how should I go about testing this command?

src/OmniSharp.DotNetTest/Services/DebugTestsInContextService.cs
```diff
@@ -0,0 +1,45 @@
...
+            if (testManager.IsConnected)
+            {
+                _debugSessionManager.StartSession(testManager);
+                return _debugSessionManager.DebugGetStartInfoAsync(
+                    request.Line, request.Column, Workspace.GetDocument(request.FileName),
+                    request.RunSettings, request.TargetFrameworkVersion,
+                    CancellationToken.None);
+            }
+
+            throw new InvalidOperationException("The debugger could not be started");
```
333fred, line 42: These seem unfortunate, especially for the `InContext` version of the command, but the existing `DebugTestGetStartInfoResponse` doesn't have a place for a failure and failure message. I would have modified that to add a field, but I'm unsure if that will break existing clients. Is that a safe change to make, or should I add a new response that adds a failure/reason to it?

src/OmniSharp.DotNetTest/VSTestManager.cs
```diff
@@ -309,6 +317,32 @@ public override async Task<RunTestResponse> RunTestsInContextAsync(int line, int
...
+            return (methodNames, testFramework?.Name);
+
+            static bool isTestMethod(IMethodSymbol methodSymbol, ref TestFramework? framework)
+            {
+                if (framework is object)
+                {
+                    return framework.IsTestMethod(methodSymbol);
+                }
+
+                foreach (var f in TestFramework.Frameworks)
```
333fred, line 329: Technically, this means that you can't combine multiple testing framework types in a single file and have them run all at once. I think that's probably fine.

src/OmniSharp.DotNetTest/Services/DebugTestsInContextService.cs
JoeRobich: I'm not really sure whether this would be a problem for other clients. Maybe @filipw knows definitively.

333fred:
@JoeRobich I've implemented the vscode side here: https://github.com/OmniSharp/omnisharp-vscode/pull/3772

src/OmniSharp.DotNetTest/Services/DebugTestsInContextService.cs
filipw: it should be fine - additive changes are generally fine since it's json based.
moreover, the debug tests protocl is - I am pretty sure - only used by VS Code extension since the debugger is proprietary

PR #1802, opened 2020-05-18
Added asnychronous test completed events

Samirat:
This PR adds events whenever a test is completed. Without it, you have to wait for all the results to be available before you get any feedback. When running long suites of tests, this makes it a lot more usable.

All `TestResult` objects are already processed as events by the VS Code extension, so I don't think any changes are necessary to the vscode extension. This does mean it'll see them twice, once from the asynchronous event and once from the RunTestResponse, but I don't think this has any bad effects.


Samirat:
@JoeRobich I've had this one hanging for a while, can you review or mention someone to review

src/OmniSharp.DotNetTest/VSTestManager.cs
```diff
@@ -400,43 +401,57 @@ public override async Task<RunTestResponse> RunTestAsync(string[] methodNames, s
 
                         case MessageType.TestRunStatsChange:
                             var testRunChange = message.DeserializePayload<TestRunChangedEventArgs>();
-                            testResults.AddRange(testRunChange.NewTestResults);
+                            var newResults = ConvertResults(testRunChange.NewTestResults);
+                            passed = passed &&  testRunChange.NewTestResults.Any(o => o.Outcome == TestOutcome.Failed);
```
JoeRobich, line 405: I think passed should be true if previous runs were passing and none of the new results are failing.

```suggestion
                            passed = passed && !testRunChange.NewTestResults.Any(o => o.Outcome == TestOutcome.Failed);
```

src/OmniSharp.DotNetTest/VSTestManager.cs
```diff
@@ -400,43 +401,57 @@ public override async Task<RunTestResponse> RunTestAsync(string[] methodNames, s
...
+                            }
                             break;
 
                         case MessageType.ExecutionComplete:
                             var payload = message.DeserializePayload<TestRunCompletePayload>();
                             if (payload.LastRunTests != null && payload.LastRunTests.NewTestResults != null)
                             {
-                                testResults.AddRange(payload.LastRunTests.NewTestResults);
+                                var lastRunResults = ConvertResults(payload.LastRunTests.NewTestResults);
+                                passed = passed && payload.LastRunTests.NewTestResults.Any(o => o.Outcome == TestOutcome.Failed);
```
JoeRobich, line 418: Same as above.
```suggestion
                                passed = passed && !payload.LastRunTests.NewTestResults.Any(o => o.Outcome == TestOutcome.Failed);
```

src/OmniSharp.DotNetTest/VSTestManager.cs
Samirat: Huh, I've got this change in my fork, why is it not showing up here?
https://github.com/Samirat/omnisharp-roslyn/blob/master/src/OmniSharp.DotNetTest/VSTestManager.cs

Samirat: Oh this is from a branch, let me make sure this is all up to date.

Samirat: Nice, I had only fixed one of these

JoeRobich:
@david-driscoll @mholo65 @filipw Anyone have any additional thoughts?

Samirat:
@JoeRobich Doesn't look like I can merge this myself.

JoeRobich:
@Samirat Thanks

PR #1815, opened 2020-05-31
Upgrade to the latest lsp library

david-driscoll:
* Upgrade to the latest lsp library
* Added interop support with existing handler endpoints
  * using the namespace `o#/`
  * eg `o#/projects` or `o#/gotofile`
  * This allows consumers to transition to LSP and still use features that exist outside of the lsp spec
  * Events are emitted using the same prefix
* Fix bug with default logging level


src/OmniSharp.Host/CommandLineApplication.cs
```diff
@@ -98,7 +98,7 @@ private void DebugAttach()
         {
             if (Debug)
             {
-                Console.WriteLine($"Attach debugger to process {Process.GetCurrentProcess().Id} to continue...");
+                // Console.WriteLine($"Attach debugger to process {Process.GetCurrentProcess().Id} to continue...");
                 while (!Debugger.IsAttached)
```
mholo65, line 102: We need to do something about this.

src/OmniSharp.Roslyn/ProjectEventForwarder.cs
```diff
@@ -63,8 +63,6 @@ private void OnWorkspaceChanged(object source, WorkspaceChangeEventArgs args)
                     {
                         Task.Factory.StartNew(async () =>
                         {
-                            await Task.Delay(500);
```
mholo65, line 66: 😬

src/OmniSharp.Shared/Utilities/PhysicalFileProviderExtensions.PollingFileChangeToken.cs
```diff
@@ -118,10 +118,10 @@ public bool ActiveChangeCallbacks
 
             public IDisposable RegisterChangeCallback(Action<object> callback, object state)
             {
-                if (state != null)
-                {
-                    throw new ArgumentException($"Stateful callbacks are not supported for {nameof(PollingFileChangeToken)}", nameof(state));
-                }
+                // if (state != null)
```
mholo65, line 121: I saw this exception in my Cake LSP server when pulling in O#. Can't remember how I fixed it though.

tests/OmniSharp.Roslyn.CSharp.Tests/DiagnosticsV2Facts.cs
```diff
@@ -71,10 +71,10 @@ public async Task CheckAllFiles(string filename1, string filename2)
...
 
-        [Theory(Skip = "Test needs to be updated for service changes")]
-        [InlineData("a.cs", "b.cs")]
-        [InlineData("a.csx", "b.csx")]
-        public async Task EnablesWhenEndPointIsHit(string filename1, string filename2)
+            [Theory(Skip = "Test needs to be updated for service changes")]
+            [InlineData("a.cs", "b.cs")]
+            [InlineData("a.csx", "b.csx")]
+            public async Task EnablesWhenEndPointIsHit(string filename1, string filename2)
         {
```
mholo65, line 78: No indent

src/OmniSharp.Roslyn/ProjectEventForwarder.cs
david-driscoll: I don't remember if this was here for a good reason, but that kind of wait is killer when we're trying to make things more responsive.

It would be one thing if we were buffering and sending out the changes grouped together, but in this case we're literally just deferring each event by 500ms.

mholo65: Yes, I've stared at this delay a couple of times before, buy since there are no tests around this, we might easily break VS Code integration 😅 Nobody seems to know why it's there, and why it's 500ms.

src/OmniSharp.Host/CommandLineApplication.cs
david-driscoll: Yes, it breaks LSP (over stdio) sadly :(

filipw:
ooh I like the CI improvements

azure-pipelines.yml
```diff
@@ -46,63 +53,94 @@ jobs:
...
-          ./build.sh --configuration Release --verbosity Verbose --target All --publish-all --archive
+          ./build.sh --configuration Release --verbosity Verbose --target CI --publish-all --archive
         displayName: 'Build'
       - template: ./.pipelines/artifacts.yml
         parameters:
           Artifacts: $(Artifacts)
 
+  - job: Linux_tests
+    pool:
+      vmImage: "ubuntu-latest"
```
filipw, line 121: I don't think we should use ubuntu-latest -at the moment it is still `ubuntu-18.04` but maybe one day it will change to `20.04`. i think it's better to pin to `ubuntu-18.04` for now to have deterministic CI builds

src/OmniSharp.Host/CommandLineApplication.cs
filipw: i would just remove the line completely.
if you run with the feature to wait a debugger enabled
a) you should know what you are doing
b) you should be able to find the correct process yourself without the PID

filipw: or, alternatively, check if LSP flag is enabled, and don't print it, otherwise print it

src/OmniSharp.Host/CompositionHostBuilder.cs
```diff
@@ -116,7 +122,11 @@ private static IEnumerable<Type> SafeGetTypes(Assembly a)
             IServiceCollection services = null,
             Action<ILoggingBuilder> configureLogging = null)
         {
-            services = services ?? new ServiceCollection();
+            services ??= new ServiceCollection();
+
+            services.TryAddSingleton(_ => new ManualFileSystemWatcher());
+            services.TryAddSingleton<IFileSystemNotifier>(_ => _.GetRequiredService<ManualFileSystemWatcher>());
+            services.TryAddSingleton<IFileSystemWatcher>(_ => _.GetRequiredService<ManualFileSystemWatcher>());
```
filipw, line 129: this is an odd way to use a discard. why not a regular variable name e.g. `s => s.GetRequiredService`?

david-driscoll:
I will attempt to correct some of the formatting, I have muscle memory and use `ctrl+k,d` very regularly 🤕

src/OmniSharp.Host/CompositionHostBuilder.cs
david-driscoll: I dunno, I probably copied it from some old aspnet core code?  I've been doing that for a long while, but I can change it here.

david-driscoll:
I've updated some of the formatting changes.

filipw:
@david-driscoll it might be good to extract the build improvements out of this PR into a separate one - we'd benefit from them immediately, while it might still take a while to get this full LSP upgrade featureset to be merged, what do you think?

david-driscoll:
No feedback needed from the PS team, I just wanted to make sure you guys had seen it.

Unless there are other changes this one is good to go in my opinion.

david-driscoll:
Oh... I noticed the PR comment, that was from a while ago.  The powershell team has been using the latest version for a while now!

azure-pipelines.yml
david-driscoll: fixed

filipw:
ok perfect good to go then 👏

mholo65:
Yay!

PR #1818, opened 2020-06-02
updated dotnet script packages to 0.53.0

filipw:
We released 0.52 over a month ago to better support .net core 3.1 scripts https://github.com/filipw/dotnet-script/releases/tag/0.52.0


filipw:
Looks like this was caused by Nuge.ProjectModel incompatibility. 0.53.0 should resolve this

PR #1821, opened 2020-06-05
Fix folder deletion with VSCode

savpek:
VSCode has issue (however by design at current version) where removed folders doesn't send any notifications of removed files because file watcher has some kind of performance optimizations to ignore them.

However watching folders removal is possible. This PR adds support for omnishar-roslyn side so clients can notify removed directory instead of single files if needed -> which can be impelemented properly with current VSCode file watcher.

This fixes one of most common issue that requires omnisharp restart on any directory rename/move/... operation.

Vscode side implementation: https://github.com/OmniSharp/omnisharp-vscode/pull/3829

### DIFF
--- a/src/OmniSharp.Host/FileWatching/ManualFileSystemWatcher.cs
+++ b/src/OmniSharp.Host/FileWatching/ManualFileSystemWatcher.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 
 namespace OmniSharp.FileWatching
 {
@@ -8,6 +9,7 @@ namespace OmniSharp.FileWatching
     {
         private readonly object _gate = new object();
         private readonly Dictionary<string, Callbacks> _callbacksMap;
+        private readonly Callbacks _folderCallbacks = new Callbacks();
 
         public ManualFileSystemWatcher()
         {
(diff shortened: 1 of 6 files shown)


savpek:
Ready for review @filipw @JoeRobich @david-driscoll

src/OmniSharp.Host/FileWatching/ManualFileSystemWatcher.cs
```diff
@@ -18,6 +20,16 @@ public void Notify(string filePath, FileChangeType changeType = FileChangeType.U
         {
             lock (_gate)
             {
+                if(changeType == FileChangeType.DirectoryDelete)
+                {
+                    foreach(var matchingCallback in _callbacksMap.AsEnumerable().Where(x => x.Key.StartsWith(filePath + Path.DirectorySeparatorChar, StringComparison.OrdinalIgnoreCase)))
```
filipw, line 25: isn't this going to end up calling OnDocumentRemoved twice on a file? once from here as a file and then again through the directory?

savpek: True, i will remove that part as it's unnecessary. 

Original idea were to hide directory removal specifics from subscribers of file watcher, however since it uses other than full paths (extensions etc) too it were not possible and this was left behind from that version.

Removed at https://github.com/OmniSharp/omnisharp-roslyn/pull/1821/commits/e90ea72d72e6b107785d2b11459dda1b75de9936

PR #1823, opened 2020-06-06
Allow client to specify symbol filter for FindSymbols Endpoint.

juliuszint:
FindSymbols Endpoint now exposes a `SymbolFilter` Property. This enables the Client to specify what Symbols he is interested in. Previously `SymbolFilter.TypeAndMember` was used by default. Especially in larger projects this results in poor performance.

### DIFF
--- a/src/OmniSharp.Abstractions/Models/v1/FindSymbols/OmniSharpSymbolFilter.cs
+++ b/src/OmniSharp.Abstractions/Models/v1/FindSymbols/OmniSharpSymbolFilter.cs
@@ -9,7 +9,7 @@ namespace OmniSharp.Models.FindSymbols
         Namespace = 1,
         Type = 2,
         Member = 4,
-        TypeAndMember = 6,
-        All = 7,
+        TypeAndMember = Type | Member,
+        All = Type | Member | Namespace,
     }
 }
--- a/src/OmniSharp.Abstractions/Models/v1/FindSymbols/OmniSharpSymbolFilter.cs
+++ b/src/OmniSharp.Abstractions/Models/v1/FindSymbols/OmniSharpSymbolFilter.cs
@@ -1,0 +1,15 @@
+using System;
+
+namespace OmniSharp.Models.FindSymbols
+{
+    [Flags]
+    public enum OmniSharpSymbolFilter
+    {
+        None = 0,
+        Namespace = 1,
+        Type = 2,
+        Member = 4,
+        TypeAndMember = 6,
+        All = 7,
+    }
+}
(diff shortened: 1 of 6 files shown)


src/OmniSharp.Abstractions/Models/v1/FindSymbols/OmniSharpSymbolFilter.cs
```diff
@@ -0,0 +1,15 @@
...
+namespace OmniSharp.Models.FindSymbols
+{
+    [Flags]
+    public enum OmniSharpSymbolFilter
+    {
+        None = 0,
+        Namespace = 1,
+        Type = 2,
+        Member = 4,
+        TypeAndMember = 6,
```
filipw, line 12: you can use thee same syntax as in Roslyn e.g. `TypeAndMember = Type | Member` or `All = Namespace | Type | Member` to make it more readable

filipw:
this is a good idea, thanks. however, which client currently needs symbol filter? AFAIK all clients use "go to anything" type of functionality like VS or R# has, that's why the default behavior was what it was.

Also, you need to add some tests here https://github.com/OmniSharp/omnisharp-roslyn/blob/master/tests/OmniSharp.Roslyn.CSharp.Tests/FindSymbolsFacts.cs

src/OmniSharp.Cake/Services/RequestHandlers/Navigation/FindSymbolsHandler.cs
```diff
@@ -26,8 +26,9 @@ public override Task<QuickFixResponse> Handle(FindSymbolsRequest request)
                 return Task.FromResult(new QuickFixResponse { QuickFixes = Array.Empty<QuickFix>() });
             }
 
+            var symbolFilter = (Microsoft.CodeAnalysis.SymbolFilter)(request?.SymbolFilter ?? OmniSharpSymbolFilter.TypeAndMember);
```
filipw, line 29: you should use the same pattern as below `(request?.SymbolFilter).GetValueOrDefault(OmniSharpSymbolFilter.TypeAndMember)` for consistency

juliuszint: When i use `using Microsoft.CodeAnalysis;` there will be a Type ambigutity. I could change it to: `using SymbolFilter = Microsoft.CodeAnalysis.SymbolFilter` if you prefer this.

src/OmniSharp.Abstractions/Models/v1/FindSymbols/OmniSharpSymbolFilter.cs
juliuszint: Will do so. Thanks for the feedback.

juliuszint:
> this is a good idea, thanks. however, which client currently needs symbol filter? 

I would implement it in omnisharp-vim. I use it on a daily basis with a project where the `OmniSharpFindSymbols` command would take somewhere around 6 seconds and with the filter set to `Type` only it returns in less than a second.

filipw:
@JoeRobich do you see any use for this in the VS Code extension? you could add "go to type" alongside the "go to anything". in either case, the change is non-breaking so existing use cases are not affected

filipw:
thank you for the contribution

PR #1832, opened 2020-06-11
Update Roslyn to 3.7.0-4.20311.4

JoeRobich:


filipw:
the method https://github.com/OmniSharp/omnisharp-roslyn/blob/01d6f693b2a1135ed3ad42e2da9aa8fcfc5fc9e0/src/OmniSharp.Roslyn/BaseExternalSourceService.cs#L58 is now public

JoeRobich:
Yeah, one day I will learn to run the tests locally before I push. =p I need to rebuild my MacOS environment first because they fail in different unexpected reasons.

PR #1841, opened 2020-06-21
Respect MSBuild globs after initial project load

SirIntruder:
Currently, Omnisharp often glitches randomly when new documents are added, or projects get restored (which triggers "FileAdded" events under the hood)

The core issue is that the way MSBuild resolves "which path belongs to which project" can get much more complicated than what omnisharp's crude heuristic can handle. So everything is fine on initial project load, but subsequential changes cause errors.

https://github.com/OmniSharp/omnisharp-roslyn/pull/1704 lists a lot of connected issues. I've tried to pull this PR and test it, but it turns out:
- MSBuild globs are very specific and complicated, so strings in "DefaultExcludes" can't be trivially used to check file paths.
- MSBuild allows for many different ways to modify inclusion rules, besides default excludes

What we really need is to have a way to ask MSBuild "Should this file be in this project?". The closest thing I've found to that is MSBuild's GetAllGlobs() method. So this PR proposes storing a list of MSBuild glob objects directly from MSBuild project, and using them to match new files, instead of using any heuristics.


leandroslc:
Great 😄. I didn't know about GetAllGlobs(). It is really a much better solution than mine. In the meantime I was trying to hook an event to trigger a design time build to just update the source files, but I was unsure of how to do it properly.

I tested your solution and it worked fine for me. If this is accepted I can close my PR.

src/OmniSharp.MSBuild/ProjectFile/ProjectFileInfo.ProjectData.cs
```diff
@@ -187,6 +192,7 @@ private ProjectData()
                 AnalyzerConfigFiles = analyzerConfigFiles.EmptyIfDefault();
                 ReferenceAliases = referenceAliases;
                 ProjectReferenceAliases = projectReferenceAliases;
+                FileInclusionGlobs = fileInclusionGlobs;
             }
 
             public static ProjectData Create(MSB.Evaluation.Project project)
```
david-driscoll, line 198: Would there be any benefit in including the globs as part of the data returned by this method as well?

JoeRobich:
> What we really need is to have a way to ask MSBuild "Should this file be in this project?". The closest thing I've found to that is MSBuild's GetAllGlobs() method.

We use `GetAllGlobs` for this same purpose. =)

filipw:
@SirIntruder would you mind resolving conflicts - looks like a merge caused it to no longer build

david-driscoll:
I may have broken this one too trying to resolve conflicts.  I think I'll pull this down tonight and get it working.

david-driscoll:
I've fixed the build error, I'll swing back and check to make sure the build passed later

david-driscoll:
Thanks @SirIntruder

src/OmniSharp.MSBuild/ProjectFile/ProjectFileInfo.ProjectData.cs
SirIntruder: Sorry for the late reply! I've investigated a bit, and it seems that this codepath is used when solution contains projects outside the target workspace folder. Since they do get added to the workspace, it's probably worth a fix

PR #1820, opened 2020-06-05
Add more ProjectConfiguration information

JoeRobich:
The server-side work for https://github.com/OmniSharp/omnisharp-vscode/issues/3577

PR #1851, opened 2020-07-09
Optional telemetry gathering

### LINKED ISSUE #1844: Can no longer run under WSL
Under 1.53.3, OmniSharp-roslyn can no longer load solutions containing more than a single project under WSL. That is to say, from the WSL "Linux" filesystem, executing a Windows OmniSharp-roslyn build, which is extremely useful functionality provided by OmniSharp-vim.

I have bisected this to c88ed904 introduced in #1820 

The example solution I've been testing on is this repo, using a server built in Visual Studio.

```sh
cd /mnt/c/code/external/omnisharp-roslyn
bin/Debug/OmniSharp.Stdio.Driver/net472/OmniSharp.exe -v > output
```

Log from commit c88ed904: https://pastebin.com/raw/GnfrtYrn

Log from previous commit 319d9cbf: https://pastebin.com/raw/dD5RgMsi

The important difference is that in c88ed904, the server only loads a single project and then stops.

Running `bin/Debug/OmniSharp.Stdio.Driver/net472/OmniSharp.exe -v` by itself results in a full server output - all projects are loaded as expected. However, when stdout is written to a file (`> outfile`) or when the server is started as a stdio job with outputs captured by the editor then the above failure occurs.

nickspoons:
This PR adds command-line flag `-wt`/`--want-telemetry-info` to request calling `dotnet --info` for telemetry purposes, and to otherwise _not_ gather this information. Fetching this data slows OmniSharp-roslyn startup, and fails in certain circumstances (see #1844).

The essence of the PR is that the `--want-telemetry-info` flag is included in `OmniSharpEnvironment`, and when it is not set, the `_dotNetCli` service is not passed in to [ProjectManager](https://github.com/OmniSharp/omnisharp-roslyn/compare/master...nickspoons:optional-telemetry-gathering?expand=1#diff-63fe9836b6472be5b5ab510254cf5bb5R105-R106). This essentially undoes the effects of PR #1820, which is where the `_dotNetCli` service began being passed in to `ProjectManager`.

cc @filipw (commenter on #1844) and @JoeRobich (author of #1820) - does this look like the right fix? Would you rather we look at doing it a different way?

~~Note that this new flag is to opt-**out** of extra info being collected. We discussed opting-**in** in #1844 but that is more complicated as it involves changes on the VScode side, which I'm not familiar with. (This PR requires changes on the Vim side, (which I can make) - as well as any other editors which choose to opt out).~~

**Edit**: the `-nt`/`--no-telemetry-info`  opt-out flags have now been changed to opt-in flags, `-wt`/`--want-telemetry-info`

Fixes #1844

### DIFF
--- a/src/OmniSharp.Abstractions/IOmniSharpEnvironment.cs
+++ b/src/OmniSharp.Abstractions/IOmniSharpEnvironment.cs
@@ -10,6 +10,6 @@ namespace OmniSharp
         string SolutionFilePath { get; }
         string SharedDirectory { get; }
         string[] AdditionalArguments { get; }
-        bool NoTelemetryInfo { get; }
+        bool WantTelemetryInfo { get; }
     }
 }
--- a/src/OmniSharp.Abstractions/IOmniSharpEnvironment.cs
+++ b/src/OmniSharp.Abstractions/IOmniSharpEnvironment.cs
@@ -10,5 +10,6 @@ namespace OmniSharp
         string SolutionFilePath { get; }
         string SharedDirectory { get; }
         string[] AdditionalArguments { get; }
+        bool NoTelemetryInfo { get; }
     }
 }
--- a/src/OmniSharp.Host/CommandLineApplication.cs
+++ b/src/OmniSharp.Host/CommandLineApplication.cs
@@ -21,6 +21,7 @@ namespace OmniSharp
         private readonly CommandOption _logLevel;
         private readonly CommandOption _applicationRoot;
         private readonly CommandOption _debug;
+        private readonly CommandOption _noTelemetryInfo;
 
         public CommandLineApplication()
         {
@@ -34,6 +35,7 @@ namespace OmniSharp
             _zeroBasedIndices = Application.Option("-z | --zero-based-indices", "Use zero based indices in request/responses (defaults to 'false').", CommandOptionType.NoValue);
             _plugin = Application.Option("-pl | --plugin", "Plugin name(s).", CommandOptionType.MultipleValue);
             _debug = Application.Option("-d | --debug", "Wait for debugger to attach", CommandOptionType.NoValue);
+            _noTelemetryInfo = Application.Option("-nt | --no-telemetry-info", "Do not gather extra telemetry information", CommandOptionType.NoValue);
         }
 
         public int Execute(string[] args)
@@ -105,5 +107,7 @@ namespace OmniSharp
                 }
             }
         }
+
+        public bool NoTelemetryInfo => _noTelemetryInfo.HasValue();
     }
 }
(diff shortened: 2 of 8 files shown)


nickspoons:
There are failed tests but I can't see what the problem is - are these related to my code changes?

filipw:
thanks a lot

> Note that this new flag is to opt-**out** of extra info being collected. We discussed opting-**in** in #1844 but that is more complicated as it involves changes on the VScode side, which I'm not familiar with. (This PR requires changes on the Vim side, (which I can make) - as well as any other editors which choose to opt out).

IMHO if we are adding the flag it should make the telemetry metadata **opt-in**. New OmniSharp versions have to be manually added to VS Code extension anyway, so during that PR it would just be a matter of adding the extra startup flag, I wouldn't see that as very invasive.

Not to mention that, since this is improving perf quite considerably, it is in-line with the overall philosophy of Omnisharp of having a lean core and features on top of it that need to be enabled.

I am sure @david-driscoll and @JoeRobich would agree

nickspoons:
Thanks @filipw, in that case I'll change the flag to opt-in some time this weekend.

nickspoons:
I have replaced the ` -nt`/`--no-telemetry-info` opt-out flags with `-wt`/`--want-telemetry-info` opt-in flags

nickspoons:
Any chance you could give this a review, @JoeRobich, since it looks like you're getting ready for a release?

src/OmniSharp.Abstractions/IOmniSharpEnvironment.cs
```diff
@@ -10,5 +10,6 @@ public interface IOmniSharpEnvironment
         string SolutionFilePath { get; }
         string SharedDirectory { get; }
         string[] AdditionalArguments { get; }
+        bool WantTelemetryInfo { get; }
```
JoeRobich, line 13: nit: I know names are hard and this isn't a blocker but maybe `IncludeTelemetryInfo` instead

nickspoons: I'm happy to make that change. What about the flag names?  `--include-telemetry-info` and `-it` ?

nickspoons:
I've just built #1857 and works well for me. I tried it in this repo and did a few restarts, it generally is loading all projects on my machine in 21-23 seconds. Comparing that (my local build) with v1.35.2 (from the github release) it seems very similar - maybe a second slower but my testing is not at all robust enough to confirm that - and it may also be due to me having debug symbols in my local build.

I will therefore close this PR in favour of #1857 - thanks very much @JoeRobich (and @filipw for the helpful feedback).

PR #1854, opened 2020-07-19
upgraded to Mono 6.10.0.105, msbuild 16.6 and added missing targets

filipw:
Mono 6.10.0 is stable now. MsBuild 16.6 is also stable and the missing targets fix the currently broken build on Linux.
@JoeRobich after this is merged could you please also upgrade the embedded mono ZIPs that get downloaded at build time as dependencies? thanks


david-driscoll:
Looks like this fixes the linux issue, I wasn't able to get much help on wsl2 I think I had my git configured with wrong line endings, lol

PR #1855, opened 2020-07-21
Updated .NET 5 tests to use 5.0.100-preview.7.20366.6

mholo65:
It was released a couple of hours ago, we should be up to date


filipw:
thanks

PR #1860, opened 2020-07-24
Introduces a new hover provider, under V2 of the protocol, that uses Roslyn's QuickInfoService

333fred:
The existing provider uses a custom handler, which this replaces. Among other benefits, this brings nullability display when available, and ensures that any new additions to roslyn's info get propagated to users of this service. Unfortunately, however, TaggedText in VS is significantly more powerful than vscode's hover renderer: that simply uses markdown. Their implementation does not support any extensions to enable C# formatting of code inline, I created a poor-man's substitute: for the description line, we treat the whole line as C#. It does mean there can be a bit of odd formatting with `(parameter)` or similar, but this exactly mirrors what typescript does so I don't think it's a big deal. For other sections, I picked sections that looked ok when formatted as C# code, and I otherwise did a simple conversion, as best I could, from tagged text to inline markdown.

I've submitted a parallel PR to move vscode to use this provider, and provided some samples of what it looks like there: https://github.com/OmniSharp/omnisharp-vscode/pull/3928.

Fixes https://github.com/OmniSharp/omnisharp-roslyn/pull/1860.


filipw:
thanks a lot! I had it on my TODO since it became public in Roslyn 😀

src/OmniSharp.Abstractions/OmniSharpEndpoints.cs
```diff
@@ -65,6 +65,8 @@ public static class V2
             public const string CodeStructure = "/v2/codestructure";
 
             public const string Highlight = "/v2/highlight";
+
+            public const string QuickInfo = "/v2/quickinfo";
```
filipw, line 69: I think it doesn't need to be v2 since there never was a quickinfo v1

src/OmniSharp.Roslyn.CSharp/Services/QuickInfoProvider.cs
```diff
@@ -0,0 +1,197 @@
...
+using OmniSharp.Options;
+
+#nullable enable
+
+namespace OmniSharp.Roslyn.CSharp.Services
+{
+    [OmniSharpHandler(OmniSharpEndpoints.V2.QuickInfo, LanguageNames.CSharp)]
+    public class QuickInfoProvider : IRequestHandler<QuickInfoRequest, QuickInfoResponse>
+    {
+        // Based on https://github.com/dotnet/roslyn/blob/master/src/Features/LanguageServer/Protocol/Handler/Hover/HoverHandler.cs
```
filipw, line 20: maybe could you link to a commit instead of head since the head link might break when the Roslyn code is rearranged

src/OmniSharp.Roslyn.CSharp/Services/QuickInfoProvider.cs
```diff
@@ -0,0 +1,197 @@
...
+        {
+            var document = _workspace.GetDocument(request.FileName);
+            var response = new QuickInfoResponse();
+
+            if (document is null)
+            {
+                return response;
+            }
+
+            var quickInfoService = QuickInfoService.GetService(document);
```
filipw, line 58: when would this ever return null?

src/OmniSharp.Roslyn.CSharp/Services/QuickInfoProvider.cs
david-driscoll: ProTip: If you go to the url in github and hit `y` on your keyboard it will "Expand URL to its canonical form" (you can see the shortcut in the github help menu `?`)

src/OmniSharp.Abstractions/Models/v2/QuickInfoResponse.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        /// <summary>
+        /// Description of the symbol under the cursor. This is expected to be rendered as a C# codeblock
+        /// </summary>
+        public string? Description { get; set; }
+
+        /// <summary>
+        /// Documentation of the symbol under the cursor, if present. It is expected to be rendered as markdown.
+        /// </summary>
+        public string? Summary { get; set; }
```
david-driscoll, line 14: I'm not sure `Description` and `Summary` are really required here, and we can just reduce it down to a list of sections.

When looking at the [apis for vscode](https://code.visualstudio.com/api/references/vscode-api#Hover) or the [lsp spec](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_hover) they either take in one markdown block or a set of them.

There is no specific ordering, so Summary and description should be the first blocks of course.

src/OmniSharp.Roslyn.CSharp/Services/QuickInfoProvider.cs
```diff
@@ -0,0 +1,197 @@
...
+                                                  .Where(s => s.Kind != QuickInfoSectionKinds.Description && s.Kind != QuickInfoSectionKinds.DocumentationComments)
+                                                  .Select(s =>
+                                                  {
+                                                      switch (s.Kind)
+                                                      {
+                                                          case QuickInfoSectionKinds.AnonymousTypes:
+                                                          case QuickInfoSectionKinds.TypeParameters:
+                                                              return new QuickInfoResponseSection { IsCSharpCode = true, Text = s.Text };
+
+                                                          default:
```
david-driscoll, line 93: Should we special case any `example` as well and emit them as full markdown code blocks?

src/OmniSharp.Roslyn.CSharp/Services/QuickInfoProvider.cs
```diff
@@ -0,0 +1,197 @@
...
+                                                              return new QuickInfoResponseSection { IsCSharpCode = false, Text = getMarkdown(s.TaggedParts) };
+                                                      }
+                                                  })
+                                                  .ToArray();
+
+            return response;
+
+            string getMarkdown(ImmutableArray<TaggedText> taggedTexts)
```
david-driscoll, line 101: I'd mark this static and pass in the `StringBuilder` to avoid an extra closure.

src/OmniSharp.Abstractions/Models/v2/QuickInfoResponse.cs
david-driscoll: @filipw @JoeRobich @mholo65 thoughts?

src/OmniSharp.Abstractions/OmniSharpEndpoints.cs
333fred: I considered the existing TypeInfo v1. I can change it though.

src/OmniSharp.Roslyn.CSharp/Services/QuickInfoProvider.cs
333fred: I'm unsure. That's what it's annotated as though.

src/OmniSharp.Roslyn.CSharp/Services/QuickInfoProvider.cs
333fred: I don't believe there is an examples section in quickinfo.

src/OmniSharp.Roslyn.CSharp/Services/QuickInfoProvider.cs
filipw: maybe in that case we should log here to help troubleshooting

src/OmniSharp.Abstractions/Models/v2/QuickInfoResponse.cs
filipw: simpler is better indeed, would make it more straightforward to expose via the LSP too

mholo65: Agree, since VS Code and LSP API expects a single block, or block of text, there's no need for separate Summary and Description. Just build everything into the same string.

src/OmniSharp.Abstractions/Models/v2/QuickInfoResponse.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// <summary>
+        /// If true, the text should be rendered as C# code. If false, the text should be rendered as markdown.
+        /// </summary>
+        public bool IsCSharpCode { get; set; }
+        public string Text { get; set; }
+
+        public override string ToString()
+        {
+            return $@"{{ IsCSharpCode = {IsCSharpCode}, Text = ""{Text}"" }}";
+        }
```
mholo65, line 33: Don't think we need this?

333fred: It's useful for debugging, particularly when tests fail.

src/OmniSharp.Abstractions/Models/v2/QuickInfoResponse.cs
333fred: It can't be a single string. I can do this as a series of sections, but vscode's markdown string doesn't actually respect triple-backtick with a language. You have to actually call `addCodeblock` with a specific language.

src/OmniSharp.Roslyn.CSharp/Services/QuickInfoProvider.cs
333fred: Well, it does when you don't have vimium enabled :)

333fred:
@filipw @david-driscoll @mholo65 addressed feedback, and updated the vscode side as well.

src/OmniSharp.Abstractions/Models/v2/QuickInfoResponse.cs
333fred: I just realized that I was making some beginner javascript mistakes when I was testing the single-string support last night, and I can do better. I'll rewrite this to be a single string.

333fred: Updated this to send just a pre-formatted markdown string.

mholo65:
Thanks! I've kicked CI since Mac build stalled.

JoeRobich:
Thanks @333Fred!

PR #1865, opened 2020-07-27
added latest changes to changelog

filipw:
to indicate the significance of `/quickinfo`, I would suggest a bump to 1.36.0


JoeRobich:
Thanks!

PR #1867, opened 2020-07-27
Update Roslyn version and tools to match .NET 5 Preview8

JoeRobich:
Not sure whether we would want the included build tools to match the .NET SDK tools or the latest VS preview tools.


JoeRobich:
Versions taken from https://github.com/dotnet/sdk/blob/release/5.0.1xx-preview8/eng/Versions.props, although it can be a moving target.

JoeRobich:
I will need to update the SDK to a published version.

JoeRobich:
Going to revert changing the SDK we use in out tests. Will perform manual validation of this build with NET 5 preview. We can update the version after it is released instead of relying on a nightly.

JoeRobich:
Had to revert to an earlier build of Roslyn. These tests caught a regression in `RuntimeMetadataReferenceResolver` on Mono platforms.

filipw:
good test then 🥁

Issue #1806, opened 2020-05-21
LSP - Support inheritdoc


sebasmonia:
The server should give the LSP client the inherited docs in this case.

Let me know if you think this is something the client should do, but it seems the server has the context to resolve from whom to pick up the documents.

Example: 
ITheContract defines DoYourThing() with `/// <summary>` etc. etc. etc.
Then Concrete : ITheContract implements DoYourThing() and documents with `/// <inheritdoc />`

PR #1870, opened 2020-07-29
Move the Hover LSP provider to use the new quickinfo service

333fred:
@david-driscoll this does what it says on the tin. Hopefully this will be another thing that we can get working in vscode. I'm unsure what the `Range` in LSP represents here, though: is it supposed to be full range of the thing being hovered over? If so, I'm not sure if we actually _want_ to implement that: I'm used to highlighting either getting every instance of a thing, or not highlighting it at all. If we do want to implement that, though, it would be very straightforward to add, as QuickInfo already has the info.

Fixes https://github.com/OmniSharp/omnisharp-roslyn/issues/1806.

### DIFF
--- a/src/OmniSharp.LanguageServerProtocol/Handlers/OmniSharpHoverHandler.cs
+++ b/src/OmniSharp.LanguageServerProtocol/Handlers/OmniSharpHoverHandler.cs
@@ -3,11 +3,9 @@ using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
 using OmniSharp.Extensions.JsonRpc;
-using OmniSharp.Extensions.LanguageServer.Protocol.Client.Capabilities;
 using OmniSharp.Extensions.LanguageServer.Protocol.Document;
 using OmniSharp.Extensions.LanguageServer.Protocol.Models;
-using OmniSharp.Extensions.LanguageServer.Protocol.Server;
-using OmniSharp.Models.TypeLookup;
+using OmniSharp.Models;
 
 namespace OmniSharp.LanguageServerProtocol.Handlers
 {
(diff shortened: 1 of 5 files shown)


src/OmniSharp.LanguageServerProtocol/Handlers/OmniSharpHoverHandler.cs
```diff
@@ -48,7 +45,7 @@ public async override Task<Hover> Handle(HoverParams request, CancellationToken
             {
                 // TODO: Range?  We don't currently have that!
                 // Range =
-                Contents = new MarkedStringsOrMarkupContent(new Container<MarkedString>(Helpers.EscapeMarkdown(omnisharpResponse.Type), Helpers.EscapeMarkdown(omnisharpResponse.Documentation)))
+                Contents = new MarkedStringsOrMarkupContent(new MarkupContent() { Value = omnisharpResponse.Markdown })
```
333fred, line 48: This does imply a bug in the implementation of the QuickInfo provider, as it's not escaping any markdown in the raw text that it appends. The escape helper isn't available at that project level though: is there any other helper available there for escaping markdown?

filipw: `Regex.Replace(markdown, @"([\\`\*_\{\}\[\]\(\)#+\-\.!])", @"\$1")`? 😄

david-driscoll: https://microsoft.github.io/language-server-protocol/specifications/specification-3-16/#textDocument_hover

The request description says: `An optional range is a range inside a text document that is used to visualize a hover, e.g. by changing the background color.`

So likely this should be the range of the symbol that's under the cursor.

david-driscoll: Whats the behavior with this change, it might not need to be escaped (but I'm not sure).

I can probably test this tonight.

333fred: If there are underscores or stars in the text, they'll appear as markdown. It'll happen here or if you use the standard io version of the provider. I'll make a quick update tonight and add a test.

david-driscoll: Feel free to move / copy the helper if needed to the QuickInfo side!

david-driscoll:
Thanks @333fred

bstaletic:
I'm migrating my client to use LSP with OmniSharp-Roslyn.
I would have expected `textDocument/hover` to provide a docstring and type if available.
That used to be the case with the `TypeLookupRequest`.
Now, with the `QuickInfoRequest`, I am only getting the type information.
How are LSP clients supposed to get docstrings?

PR #1871, opened 2020-07-30
added support for EmbeddedLanguageCompletionProvider

filipw:
This is quite new in Roslyn. It gives completion results for e.g. datetime string literals and regex patterns

Bildschirmfoto 2020-07-30 um 21 29 06
Bildschirmfoto 2020-07-30 um 21 41 02


david-driscoll:
Does this work with Guids and other types that have well-known formats?

filipw:
no only these for now

PR #1875, opened 2020-08-06
Check for MSBuild version is higher than required by the .NET SDK

JoeRobich:
Provide a warning when the discovered MSBuild version is lower than the minimumMSBuildVersion supported by the configured SDK.

### DIFF
--- a/src/OmniSharp.Host/MSBuild/Discovery/Extensions.cs
+++ b/src/OmniSharp.Host/MSBuild/Discovery/Extensions.cs
@@ -1,13 +1,20 @@
-﻿using System.IO;
+﻿using System;
+using System.IO;
 using Microsoft.Extensions.Logging;
+using OmniSharp.Services;
 
 namespace OmniSharp.MSBuild.Discovery
 {
     internal static class Extensions
     {
-        public static void RegisterDefaultInstance(this IMSBuildLocator msbuildLocator, ILogger logger)
+        private static readonly Version s_minimumMSBuildVersion = new Version(16, 3);
+
+        public static void RegisterDefaultInstance(this IMSBuildLocator msbuildLocator, ILogger logger, DotNetInfo dotNetInfo = null)
         {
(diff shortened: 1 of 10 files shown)


src/OmniSharp.Host/MSBuild/Discovery/Extensions.cs
```diff
@@ -55,12 +87,12 @@ public static bool HasDotNetSdksResolvers(this MSBuildInstance instance)
         /// Checks if it is MSBuild from Visual Studio 2017 RTM that cannot be used.
```
filipw, line 87: this comment is long time incorrect 😁

src/OmniSharp.Host/MSBuild/Discovery/Extensions.cs
```diff
@@ -112,5 +144,28 @@ private static int GetInstanceFeatureScore(MSBuildInstance i)
...
+            if (dotNetInfo is null
+                || string.IsNullOrWhiteSpace(dotNetInfo.SdksPath)
+                || dotNetInfo.SdkVersion is null)
+            {
+                return s_minimumMSBuildVersion;
+            }
+
+            var version = dotNetInfo.SdkVersion;
+            var sdksPath = dotNetInfo.SdksPath;
+            var minimumVersionPath = Path.Combine(sdksPath, version.ToNormalizedString(), "minimumMSBuildVersion");
```
filipw, line 159: this is really there?! that's new to me!

src/OmniSharp.Host/MSBuild/Discovery/Extensions.cs
JoeRobich: Updated

src/OmniSharp.Host/MSBuild/Discovery/Extensions.cs
JoeRobich: I was happy to find it =)

Issue #1809, opened 2020-05-24
How to make override autocomplete fill snippet on enter


Thaina:
It was a habit of me to just type override select autocomplete like this

![image](https://user-images.githubusercontent.com/1042507/82750872-7d13bc00-9ddd-11ea-8511-b8b4eb0912c9.png)

I expect that if I press enter it would completely emit type and modifier from the base class like this

![image](https://user-images.githubusercontent.com/1042507/82750884-9f0d3e80-9ddd-11ea-96ce-3eb06c616206.png)

But currently now it only emit just this line

![image](https://user-images.githubusercontent.com/1042507/82750879-89981480-9ddd-11ea-8ba4-487f391fe891.png)

I am not sure I have making some mistake in omnisharp config, or it just not support in omnisharp vscode? If so may I request this feature?

filipw:
this is currently not supported - it will be supported once we move to the CompletionService fully, including the ability to resolve completion item by the client

Thaina:
@filipw Thank you for your information

filipw:
it is currently an Omnisharp server limitation

Issue #1725, opened 2020-02-27
`override` completion doesn't insert full signature


333fred:
## Issue Description ##

Override completion does not insert the full signature.

## Steps to Reproduce ##

1. Create a C# project with a class like this:
```cs
class C1
{
	public virtual void M() {}
}
class C2 : C1
{
    // Type here
}
```
2. At `// Type here`, replace the comment with `override` `<space>` `<ctrl+space>`. Valid override method completion should appear.
3. Select `M` to override. 

## Expected Behavior ##

The full signature of `M` is inserted, so you end up with `public override void M()` on the current line.

## Actual Behavior ##

Just `M()` is inserted, missing the accessibility modifier and the return type.

### OmniSharp log ###
<details>Post the output from Output-->OmniSharp log here</details>

### C# log ###
<details>Post the output from Output-->C# here</details>

## Environment information ##

**VSCode version**: 1.42.1
**C# Extension**: 1.21.12

<details><summary>Dotnet Information</summary>
.NET Core SDK (reflecting any global.json):
 Version:   3.1.200-preview-014977
 Commit:    ca5522cead

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.18363
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\3.1.200-preview-014977\

Host (useful for support):
  Version: 3.1.1
  Commit:  a1388f194c

.NET Core SDKs installed:
  3.0.100-preview8-013437 [C:\Program Files\dotnet\sdk]
  3.1.100 [C:\Program Files\dotnet\sdk]
  3.1.200-preview-014977 [C:\Program Files\dotnet\sdk]

.NET Core runtimes installed:
  Microsoft.AspNetCore.All 2.1.12 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.1.15 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.2.6 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.App 2.1.12 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 2.1.15 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
[...]

Issue #78, opened 2015-02-13
Use new Roslyn Completion List API for IntelliSense


nosami:
Internal, but much better. Handles keywords, snippets, named parameters, object initializer etc.

https://github.com/dotnet/roslyn/blob/47ff630cc838b56248022c1d494299f35daec474/src/Features/CSharp/Completion/CSharpCompletionService.cs

``` cs
var completionService = new CSharpCompletionService();
var t = await completionService.GetGroupsAsync(document, position, new mca.Completion.CompletionTriggerInfo(), completionService.GetDefaultCompletionProviders(), new CancellationToken());
```

davidfowl:
epic

jrieken:
:+1: reflection ftw

stirno:
I have a working impl of this in our codebase but it throws under Mono presently with a not-very-helpful error. Haven't had a chance to dig through that yet.

Also theres an internal extension method on document to help create the service, might be better to use that in case it gets more MEFy:

```
var completionService = document.GetLanguageService<ICompletionService>();
```

Pilchie:
Nice!  I would like to make something like that public at some point after I make something like ICompletionProvider public (my current side project).

nosami:
Seems that the features assembly doesn't yet have a nuget package - https://github.com/dotnet/roslyn/issues/572

aelij:
@Pilchie, while you're doing that, how about moving `ISignatureHelpProvider` to the Features assembly and making it public as well? :)
It has no dependency on the VS SDK, so it should be easy (I know because I've [done it](https://github.com/aelij/roslynpad/)).

Pilchie:
Tagging @dustincampbell on this too.

DustinCampbell:
I'm going to tracking this over at https://github.com/dotnet/roslyn/issues/3538

DustinCampbell:
And here: dotnet/roslyn#3540

khellang:
:+1:

DustinCampbell:
And if anyone's interested, the branch I'm working out of is at https://github.com/DustinCampbell/roslyn/tree/completion-api.

tpetrina:
Hi @DustinCampbell, can you please give an example on how to use the completion API in your branch?

DustinCampbell:
Patience. :smile: I'm actively integrating it back into Roslyn proper as an internal API. Once that happens, I'll run it through a formal API design process and we'll make it public.

[...]

PR #1877, opened 2020-08-10
Start moving omnisharp to directly using Roslyn's completion service, with a separate resolve step

333fred:
Today, Omnisharp uses Roslyn's completion service to gather completions, but doesn't then use the other completion helpers to fill in the rest of the detail of the completion. Additionally, Omnisharp does a bunch of work when creating and populating completions, and this can cause slow and inconsistent behavior when invoking completions in the middle of a pre-typed word. Some current issues:

* vscode fuzzy-matching changes based on whether you invoke completion after typing a few characters.
* Completions with large numbers of elements can take a bit of time to load, as Omnisharp does filtering and documentation building for all of them.
* Completion is not well setup to play nicely with LSP, which uses a much different interface.
* Completion is not capable of providing unimported types, particularly because providing the edits would be extremely expensive if not done only on request.

To solve these issues, this adds a new completion service that mirrors the structure of LSP completion and completion/resolve:

* On request, only the completion item and initial edit are provided.
* A new response service is added, which looks up documentation when a particular item is selected in the completion window.
    * This response service will additionally be able to provide extra edits, such as inserting the rest of an override statement or adding imports for unimported types, but these are not yet implemented.

An example of what this looks like in vscode:
image
image

Open questions/things still todo:

* [x] The first line of the documentation can likely be unconditionally formatted as C#, much like we do in the new quickinfo provider.
* [x] Do output diffing to insert other text. For example, in VS selecting a thing to override will fill in the rest of the signature. We might be able to use `additionalTextEdits` to do the same thing here, though the restrictions around the current insertion point are concerning and may be a technical limitation to work out.
* [ ] I'm planning on using the right side of the completion window (the `detail` section of a `CompletionItem`) to show the unimported namespace, like VS. However, is there something else we want to put in there?
* [x] The implementation is currently triggered by space characters. This isn't something that VS does, except for `override` and possibly a few other scenarios. This is because the Roslyn APIs for `ShouldTriggerCompletion` returns true for space characters. Is this something that we actually want to have happen? It seems to me like this might be a bug, at which point we should likely ask Roslyn to fix the bug, not try to work around it ourselves.

@filipw @david-driscoll @JoeRobich for a first look. I'm putting it in draft for now while I try and figure out the output diffing bit, but I wanted to at least get eyes on what I've implemented so far.

This supersedes https://github.com/OmniSharp/omnisharp-roslyn/pull/884.

Fixes https://github.com/OmniSharp/omnisharp-roslyn/issues/1809.
Fixes https://github.com/OmniSharp/omnisharp-roslyn/issues/1725.
I think this fixes https://github.com/OmniSharp/omnisharp-roslyn/issues/78?


tests/OmniSharp.Roslyn.CSharp.Tests/CompletionFacts.cs
```diff
@@ -0,0 +1,715 @@
...
+            var completions = await FindCompletionsAsync(filename, source);
+            Assert.Equal(new[] { "myClass", "my", "@class", "MyClass", "My", "Class", "GetMyClass", "GetMy", "GetClass" },
+                         completions.Items.Select(c => c.Label));
+        }
+
+
+        [Theory]
+        [InlineData("dummy.cs")]
+        [InlineData("dummy.csx")]
+        public async Task Returns_override_signatures(string filename)
```
filipw, line 354: these old override tests were only aiming to cover the rudimentary support we had for overrides in the past.
now with proper usage of completion service it would be  good to extend them

src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionService.cs
```diff
@@ -0,0 +1,272 @@
...
+            }
+
+            var typedSpan = completionService.GetDefaultCompletionListSpan(sourceText, position);
+            string typedText = sourceText.GetSubText(typedSpan).ToString();
+
+            ImmutableArray<string> filteredItems = typedText != string.Empty
+                ? completionService.FilterItems(document, completions.Items, typedText).SelectAsArray(i => i.DisplayText)
+                : ImmutableArray<string>.Empty;
+            _logger.LogTrace("Completions filled in");
+            _lastCompletion = (completions, request.FileName, position);
```
filipw, line 130: since this is a server I would say this must be locked

src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionService.cs
```diff
@@ -0,0 +1,272 @@
...
+        private readonly FormattingOptions _formattingOptions;
+        private readonly ILogger _logger;
+
+        private (CSharpCompletionList Completions, string FileName, int Position)? _lastCompletion = null;
+
+        [ImportingConstructor]
+        public CompletionService(OmniSharpWorkspace workspace, FormattingOptions formattingOptions, ILoggerFactory loggerFactory)
+        {
+            _workspace = workspace;
+            _formattingOptions = formattingOptions;
```
filipw, line 82: I now realized that the usage of `FormattingOptions` isn't optimal here because it wouldn't observe editorconfig, but this is something that we broke when we moved to Roslyns' editorcconfig so unrelated to this PR

src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionService.cs
```diff
@@ -0,0 +1,272 @@
...
+            if (document is null)
+            {
+                _logger.LogInformation("Could not find document for file {0}", request.FileName);
+                return new CompletionResponse { Items = ImmutableArray<CompletionItem>.Empty };
+            }
+
+            var sourceText = await document.GetTextAsync();
+            var position = sourceText.GetTextPosition(request);
+
+            var completionService = CSharpCompletionService.GetService(document);
```
filipw, line 101: do you want to activate ShowItemsFromUnimportedNamespaces here?

filipw:
thank you so much! this makes me extremely happy 👏

> The first line of the documentation can likely be unconditionally formatted as C#, much like we do in the new quickinfo provider.

yes that is a good idea

>  Do output diffing to insert other text. For example, in VS selecting a thing to override will fill in the rest of the signature. We might be able to use additionalTextEdits to do the same thing here, though the restrictions around the current insertion point are concerning and may be a technical limitation to work out.

that was my fear too. is there some diffing service in roslyn that we can just use, or do you need to roll out something by hand?

>  The implementation is currently triggered by space characters. This isn't something that VS does, except for override and possibly a few other scenarios. This is because the Roslyn APIs for ShouldTriggerCompletion returns true for space characters. Is this something that we actually want to have happen? It seems to me like this might be a bug, at which point we should likely ask Roslyn to fix the bug, not try to work around it ourselves.

IMHO this is also a bug, and overloads the server unnecessarily. But shouldn't this be rather fixed client side? e.g. at the moment VS Code would send `"WordToComplete": ""` and `"TriggerCharacter": " "` as you are typing code and use space

src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionService.cs
filipw: sorry, I read the PR description again and you explicitly mentioned it's not implemented yet.

IMHO this is a good idea to leave these extra things out and keep this PR manageable and merge, and only afterwards add extra features in separate PRs, if it makes sense

src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionService.cs
333fred: Should I be using something else? It's only used for formatting the markdown popups, so observing editorconfig isn't particularly important for this bit.

src/OmniSharp.Abstractions/Models/v1/Completion/CompletionItem.cs
```diff
@@ -0,0 +1,177 @@
...
+        [JsonProperty("data")]
+        public int Data { get; set; }
+
+        public override string ToString()
+        {
+            return $"{{ {nameof(Label)} = {Label}, {nameof(CompletionItemKind)} = {Kind} }}";
+        }
+    }
+
+    public struct TextEdit
```
333fred, line 110: I went and implemented new versions of these to line up with the lsp definitions here: it made my vscode development experience simpler. The reason is that `Point` and `Range` as they exist in omnisharp today are 1-based and pascalCased, whereas the lsp definitions are 0-based and camelCased. I need to take these and do a real mapping on the vscode side and get rid of them here, but I haven't yet.

333fred:
Here's what override completion looks like:

override-completion

I still need to adjust the documentation stuff. I can drastically simplify the implementation and clean up code in both the new quickinfo provider and the completion service around it with a couple of small changes, but for this iteration I was focused on override completion :D

tests/OmniSharp.Roslyn.CSharp.Tests/CompletionFacts.cs
333fred: I've added several more tests, I'm feeling decent about the coverage now.

filipw:
this is ⭐️⭐️⭐️!

src/OmniSharp.Roslyn.CSharp/Helpers/SourceTextExtensions.cs
```diff
@@ -0,0 +1,13 @@
...
+
+using Microsoft.CodeAnalysis.Text;
+using OmniSharp.Models;
+
+namespace OmniSharp.Roslyn.CSharp.Helpers
+{
+    internal static class SourceTextExtensions
+    {
+        public static int GetTextPosition(this SourceText sourceText, Request request)
+            => sourceText.Lines.GetPosition(new LinePosition(request.Line, request.Column));
```
filipw, line 11: I think we have a line like `sourceText.Lines.GetPosition(new LinePosition(request.Line, request.Column));` at least 10 times in different places, since you added this nice extension method, could you replace those usages too? 🙂

src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionService.cs
filipw: ah I now see only newline is used, then it doesn't matter anyway.
it used to be so that `FormattingOptions` was also populated by editorconfig settings, but not anymore. it would be more important if we ran the code through the C# formatter/code clean up before putting it in markdown but I don't think even VS does that

333fred:
> IMHO this is also a bug, and overloads the server unnecessarily. But shouldn't this be rather fixed client side? e.g. at the moment VS Code would send "WordToComplete": "" and "TriggerCharacter": " " as you are typing code and use space

We do send the trigger character, and whether the completion was triggered explicitly or by typing. It's just that Roslyn thinks "This was triggered by typing ` `" should always actually trigger completion.

333fred:
@filipw I believe this is now in a good state.

333fred:
@filipw can you confirm whether this fixes #78? I think it does, but I'm not 100%.

filipw:
yes it does 👍

xcap2000:
Is it available in the current stable version? How do I complete using "override "?

333fred:
Yes, it's available. You just invoke completion and it should work.

xcap2000:
I am trying to use on vscode using C# extension and when I type "override+space" it shows nothing, is there any configuration for this?

333fred:
Did you trigger completion, such as by hitting ctrl+space after typing override?

xcap2000:
I just tried that with and without space after the override, did not work. latest stable omnisharp and vscode.

333fred:
Please open an issue and include your omnisharp log.

PR #1883, opened 2020-08-12
Fix MSBuild version mismatch with new SDKs

JoeRobich:
The 3.1.401 SDK's minimumMSBuildVersion is 16.7.0. The latest stable and preview versions of Mono ship with 16.6.0. This causes the DotNet.MSBuildSDKResolver to no longer find the sdks needed to load projects.

Possibilities for working around this:
- Enable LegacySDKResolver by default
- Rewrite the LegacySDKResolver as a true SDKResolver to be consumed by the included MSBuild, but set the priority such that is only used as a last resort.
- Ship the MSBuild from the tools/packages.json
- Ship the MSBuild from Mono 6.13.x

Since we already ship the MSBuild from tools/packages on the Windows platform, this may be the easiest path to take.

Fixes https://github.com/OmniSharp/omnisharp-vscode/issues/3951


build.cake
```diff
@@ -206,7 +206,11 @@ Task("CreateMSBuildFolder")
...
         "Microsoft.Build.Tasks.Core",
-        "Microsoft.Build.Utilities.Core",
+        "Microsoft.Build.Utilities.Core"
+    };
+
+    var msbuildRefLibraries = new []
+    {
         "Microsoft.Build.Tasks.v4.0",
         "Microsoft.Build.Tasks.v12.0",
         "Microsoft.Build.Utilities.v4.0",
```
JoeRobich, line 216: There are no NuGet packages for these but we can still copy them from the Mono MSBuild folder.

build/Packages.props
```diff
@@ -57,10 +57,10 @@
...
     <PackageReference Update="Newtonsoft.Json" Version="12.0.3" />
 
-    <PackageReference Update="Nuget.Packaging" Version="$(NuGetPackageVersion)" />
-    <PackageReference Update="Nuget.Packaging.Core" Version="$(NuGetPackageVersion)" />
-    <PackageReference Update="Nuget.ProjectModel" Version="$(NuGetPackageVersion)" />
-    <PackageReference Update="Nuget.Versioning" Version="$(NuGetPackageVersion)" />
+    <PackageReference Update="NuGet.Packaging" Version="$(NuGetPackageVersion)" />
+    <PackageReference Update="NuGet.Packaging.Core" Version="$(NuGetPackageVersion)" />
+    <PackageReference Update="NuGet.ProjectModel" Version="$(NuGetPackageVersion)" />
+    <PackageReference Update="NuGet.Versioning" Version="$(NuGetPackageVersion)" />
```
JoeRobich, line 63: This only seemed to be an issue in the Linux build surprisingly

src/OmniSharp.DotNetTest/VSTestManager.cs
```diff
@@ -148,15 +148,6 @@ public override async Task<GetTestStartInfoResponse> GetTestStartInfoAsync(strin
...
 
             var testCases = await DiscoverTestsAsync(new string[] { methodName }, runSettings, targetFrameworkVersion, cancellationToken);
-
-            SendMessage(MessageType.GetTestRunnerProcessStartInfoForRunSelected,
-                new
-                {
-                    TestCases = testCases,
-                    DebuggingEnabled = true,
-                    RunSettings = LoadRunSettingsOrDefault(runSettings, targetFrameworkVersion)
-                });
```
JoeRobich, line 158: This should have been removed when I extracted out the GetTestRunningProcessStartInfo method

tests/OmniSharp.Tests/DotNetCliServiceFacts.cs
```diff
@@ -7,9 +8,20 @@ namespace OmniSharp.Tests
...
+        private string Release { get; }
+
         public DotNetCliServiceFacts(ITestOutputHelper output)
             : base(output)
         {
+            var version = SemanticVersion.Parse(DotNetVersion);
+            Major = version.Major;
+            Minor = version.Minor;
+            Patch = version.Patch;
+            Release = version.Release;
```
JoeRobich, line 24: This should make future SDK updates easier since the version string can be updated with a replace all.

PR #1888, opened 2020-08-14
Add flag so that all document changes are added together

JoeRobich:
Multi-cursor editing, refactor rename, and other operations are broken since the C# extension started sending changes instead of buffers to the OmniSharp. 

The reason is that changes in VSCode are reported with ranges that refer to the same version of the document being edited. The BufferManager, however, expects that changes should be applied sequentially and that the later ranges will compensate for previously applied changes.

This PR adds a flag to Request that indicates all changes should be applied together instead of sequentially. Since the flags default value would apply changes sequentially, this shouldn't break any existing clients.

### DIFF
--- a/src/OmniSharp.Abstractions/Models/Request.cs
+++ b/src/OmniSharp.Abstractions/Models/Request.cs
@@ -11,5 +11,6 @@ namespace OmniSharp.Models
         public int Column { get; set; }
         public string Buffer { get; set; }
         public IEnumerable<LinePositionSpanTextChange> Changes { get; set; }
+        public bool ApplyChangesTogether { get; set; }
     }
 }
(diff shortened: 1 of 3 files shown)


david-driscoll:
Interesting I wonder if this should just be the default behavior?

JoeRobich:
> Interesting I wonder if this should just be the default behavior?

Trying to test it locally and I think there is something I am missing. Gonna convert this to a draft in the meantime.

david-driscoll:
I wonder if order counts, in the past with things like this as long as the changes are ordered from the bottom up everything works correctly.

JoeRobich:
sending changes over bottom sounds like a good plan. I'll revisit this later

PR #1890, opened 2020-08-15
Updated lsp library to fix issue with signature help, etc

david-driscoll:
Fixes: #1887

PR #1892, opened 2020-08-17
Added System.Reflection.Emit.* to embedded mono script

david-driscoll:
Added `System.Reflection.Emit.*` assemblies to embedded mono.

Fixes test issue #1890


filipw:
it will require mono to be rebuilt and uploaded to blob storage

JoeRobich:
Happy to rebuild those packages. Anyone know why we don't copy the entire facades folder? is it just a size issue?

PR #1894, opened 2020-08-17
Fix lack of trailing italics

333fred:
When the section ended without a newline, we weren't correctly ending the italic section.


JoeRobich:
Thanks Fred!

PR #1895, opened 2020-08-18
Include version matched target files with minimal MSBuild

JoeRobich:
Instead of including the .target files from the Mono on the build machine. Overwrit those .target files with the versions from the MSBuild Runtime. Also, use the non-deprecated compiler toolset package.


JoeRobich:
@david-driscoll @mholo65 @filipw ptal

Issue #1482, opened 2019-05-08
Import completion for types


duki994:
Is there any possibility of supporting this in future?
https://github.com/dotnet/roslyn/pull/34808

It's still in preview and based on idea of IntelliSense Extender VS extension.
It will be out of preview in a few versions.

Roslyn itself is capable of providing unimported type information via TypeImportCompletionProvider implemented in above pull request. CompletionService should suffice and support this new CompletionProvider without any major changes. Maybe even without changes.

filipw:
Yes it's in the plans.

filipw:
I have built a prototype for this now and we'll have a look at integrating it soon. It will require some changes to the VS Code side of the extension too.

nickspoons:
Once the prototype is built, a brief description of that request->response->extra process would be very much appreciated, so we can implement it in other clients (OmniSharp-vim)

Gameghostify:
Any movement on this? Features like this is what makes roslyn/omnisharp so great and would be **amazing** to have 👀

filipw:
still work in progress 😄

fabiopicchi:
@filipw Can I help somehow? Is there a WIP branch?

JoshuaLight:
Any progress so far? Must have feature!

Gameghostify:
+1 this would be amazing to have

bazuka5801:
> Yes it's in the plans.

Any status?

bazuka5801:
![image](https://user-images.githubusercontent.com/11452353/84871464-3ede8500-b089-11ea-9d14-aafc22312271.png)
![image](https://user-images.githubusercontent.com/11452353/84871495-47cf5680-b089-11ea-989b-97446eae9705.png)

No suggestions? how to fix that?

phil-harmoniq:
@filipw Is this still in the works? This is a much needed feature imho.

PR #1896, opened 2020-08-18
Support completion of unimported types.

333fred:
This needs to call a few internal overloads for completion, instead of the public versions, in order to determine whether any completions can be expanded (ie, imported) and to get the full change for these items, including the import. The public APIs do not provide this information currently.

Todo:

* [x] More testing!
* [x] Make import completion optional. In large solutions it will almost certainly be expensive, we should provide an option to turn it off.

Preview of the change in vscode (watch the line numbers!):
import-completion

Fixes https://github.com/OmniSharp/omnisharp-roslyn/issues/1482

### DIFF
--- a/src/OmniSharp.Abstractions/Models/v1/Completion/CompletionItem.cs
+++ b/src/OmniSharp.Abstractions/Models/v1/Completion/CompletionItem.cs
@@ -85,7 +85,7 @@ namespace OmniSharp.Models.v1.Completion
         /// <summary>
         /// Index in the completions list that this completion occurred.
         /// </summary>
-        public int Data { get; set; }
+        public (int Index, bool IsExpanded) Data { get; set; }
 
         public override string ToString()
         {
--- a/src/OmniSharp.Abstractions/Models/v1/Completion/CompletionItem.cs
+++ b/src/OmniSharp.Abstractions/Models/v1/Completion/CompletionItem.cs
@@ -85,7 +85,7 @@ namespace OmniSharp.Models.v1.Completion
         /// <summary>
         /// Index in the completions list that this completion occurred.
         /// </summary>
-        public (int Index, bool IsExpanded) Data { get; set; }
+        public int Data { get; set; }
 
         public override string ToString()
         {
--- a/src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionService.cs
+++ b/src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionService.cs
@@ -116,14 +116,10 @@ namespace OmniSharp.Roslyn.CSharp.Services.Completion
                 return new CompletionResponse { Items = ImmutableArray<CompletionItem>.Empty };
             }
 
-            OptionSet completionsWithImport = await document.GetOptionsAsync();
-            completionsWithImport = completionsWithImport.WithChangedOption(CompletionItemExtensions.ShowItemsFromUnimportedNamespaces, LanguageNames.CSharp, true);
-
             var (completions, expandedItemsAvailable) = await completionService.GetCompletionsInternalAsync(
                 document,
                 position,
-                getCompletionTrigger(includeTriggerCharacter: false),
-                options: completionsWithImport);
+                getCompletionTrigger(includeTriggerCharacter: false));
             _logger.LogTrace("Found {0} completions for {1}:{2},{3}",
                              completions?.Items.IsDefaultOrEmpty != true ? 0 : completions.Items.Length,
                              request.FileName,
@@ -161,7 +157,8 @@ namespace OmniSharp.Roslyn.CSharp.Services.Completion
             // that completion provider is still creating the cache. We'll mark this completion list as not completed, and the
             // editor will ask again when the user types more. By then, hopefully the cache will have populated and we can mark
             // the completion as done.
-            bool isIncomplete = expandedItemsAvailable;
+            bool isIncomplete = expandedItemsAvailable &&
+                                _workspace.Options.GetOption(CompletionItemExtensions.ShowItemsFromUnimportedNamespaces, LanguageNames.CSharp) == true;
 
             for (int i = 0; i < completions.Items.Length; i++)
             {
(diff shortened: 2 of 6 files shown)


src/OmniSharp.Abstractions/Models/v1/Completion/CompletionItem.cs
```diff
@@ -85,7 +85,7 @@ public class CompletionItem
         /// <summary>
         /// Index in the completions list that this completion occurred.
         /// </summary>
-        public int Data { get; set; }
+        public (int Index, bool IsExpanded) Data { get; set; }
```
filipw, line 88: since we have a raw HTTP interface too, I think it would make more sense to capture it into a regular DTO

src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionService.cs
```diff
@@ -114,7 +116,14 @@ public async Task<CompletionResponse> Handle(CompletionRequest request)
                 return new CompletionResponse { Items = ImmutableArray<CompletionItem>.Empty };
             }
 
-            var completions = await completionService.GetCompletionsAsync(document, position, getCompletionTrigger(includeTriggerCharacter: false));
+            OptionSet completionsWithImport = await document.GetOptionsAsync();
+            completionsWithImport = completionsWithImport.WithChangedOption(CompletionItemExtensions.ShowItemsFromUnimportedNamespaces, LanguageNames.CSharp, true);
```
filipw, line 120: does this have to be changed per request? could we not set it as workspace option?

we already have infrastructure for binding `OmniSharpOptions` to workspace `OptionsSets` e.g. https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp.Roslyn.CSharp/Services/RenameWorkspaceOptionsProvider.cs This also takes care of watching for changes and updating the workspace if omnisharp.json changes. It seems to me like a way to go here

src/OmniSharp.Roslyn.CSharp/Services/Intellisense/CompletionItemExtensions.cs
```diff
@@ -22,21 +24,29 @@ internal static class CompletionItemExtensions
...
+        internal static readonly PerLanguageOption<bool?> ShowItemsFromUnimportedNamespaces = new PerLanguageOption<bool?>("CompletionOptions", "ShowItemsFromUnimportedNamespaces", defaultValue: null);
 
         static CompletionItemExtensions()
         {
             _symbolCompletionItemType = typeof(CompletionItem).GetTypeInfo().Assembly.GetType(SymbolCompletionItem);
             _getSymbolsAsync = _symbolCompletionItemType.GetMethod(GetSymbolsAsync, BindingFlags.Public | BindingFlags.Static);
 
             _getProviderName = typeof(CompletionItem).GetProperty(ProviderName, BindingFlags.NonPublic | BindingFlags.Instance);
+
+            _getCompletionsInternalAsync = typeof(CompletionService).GetMethod(nameof(GetCompletionsInternalAsync), BindingFlags.NonPublic | BindingFlags.Instance);
```
filipw, line 48: "GetCompletionsInternalAsync" oh joy 😅

src/OmniSharp.Abstractions/Models/v1/Completion/CompletionItem.cs
david-driscoll: Yeah I'm not sure how many users we have of the http interface at this point, but it is still supported at the moment.

333fred: I take it ValueTuple does something weird with the HTTP interface?

filipw: i don't think JSON.NET can serialize them https://github.com/JamesNK/Newtonsoft.Json/issues/1230

333fred: I can actually just remove this addition and leave it as it was, will do that.

src/OmniSharp.Roslyn.CSharp/Services/Intellisense/CompletionItemExtensions.cs
333fred: If I had another option... 😢

src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionService.cs
333fred: I think I've correctly done this. Let me know if I messed something up.

filipw:
it looks like the tests are a bit unstable for example this failed on Linux

```
    OmniSharp.Roslyn.CSharp.Tests.CompletionFacts.ImportCompletionResolvesOnSubsequentQueries(filename: "dummy.cs") [FAIL]
      System.OperationCanceledException : The operation was canceled.
      Stack Trace:
          at System.Threading.CancellationToken.ThrowOperationCanceledException () [0x00010] in <a17fa1457c5d44f2885ac746c1764ea5>:0 
          at System.Threading.CancellationToken.ThrowIfCancellationRequested () [0x00008] in <a17fa1457c5d44f2885ac746c1764ea5>:0 
          at OmniSharp.Roslyn.CSharp.Tests.CompletionFacts+<>c__DisplayClass9_0.<ImportCompletionResolvesOnSubsequentQueries>b__1 () [0x000d2] in <720a8e872b3940aca18dd68ab266aa89>:0 
          at OmniSharp.Roslyn.CSharp.Tests.CompletionFacts.ImportCompletionResolvesOnSubsequentQueries (System.String filename) [0x001ee] in <720a8e872b3940aca18dd68ab266aa89>:0 
```

but passed on a retry. Maybe some sort of a thread exhaustion, these agents are not too powerful

filipw (approved):
:shipit: :shipit: :shipit:

333fred:
I'll up the cancelation limit. 100 ms very stable locally, but I was a bit curious how they'd do on CI.

333fred:
> I'll up the cancelation limit. 100 ms is very stable locally, but I was a bit curious how they'd do on CI.

@filipw upped the timeout to a second/test.

filipw:
thanks this is good to go IMHO! 🍻

JoeRobich:
Thanks Fred!

PR #1902, opened 2020-08-19
Exclude Misc project from InternalsVisibleTo completion

filipw:
The misc project for loose .cs files should not show up in InternalsVisibleTo suggestion.

cc @333fred

### DIFF
--- a/src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionService.cs
+++ b/src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionService.cs
@@ -151,7 +151,7 @@ namespace OmniSharp.Roslyn.CSharp.Services.Completion
             }
 
             var triggerCharactersBuilder = ImmutableArray.CreateBuilder<char>(completions.Rules.DefaultCommitCharacters.Length);
-            var completionsBuilder = new List<CompletionItem>();
+            var completionsBuilder = ImmutableArray.CreateBuilder<CompletionItem>();
 
             // If we don't encounter any unimported types, and the completion context thinks that some would be available, then
             // that completion provider is still creating the cache. We'll mark this completion list as not completed, and the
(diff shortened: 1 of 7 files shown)


src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionService.cs
```diff
@@ -146,11 +147,13 @@ public async Task<CompletionResponse> Handle(CompletionRequest request)
             }
 
             var triggerCharactersBuilder = ImmutableArray.CreateBuilder<char>(completions.Rules.DefaultCommitCharacters.Length);
-            var completionsBuilder = ImmutableArray.CreateBuilder<CompletionItem>(completions.Items.Length);
+            var completionsBuilder = new List<CompletionItem>();
 
             for (int i = 0; i < completions.Items.Length; i++)
             {
                 var completion = completions.Items[i];
```
333fred, line 154: Consider doing this check in the switch that already checks the provider name.

filipw: ah yes

src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionService.cs
```diff
@@ -150,7 +151,7 @@ public async Task<CompletionResponse> Handle(CompletionRequest request)
             }
 
             var triggerCharactersBuilder = ImmutableArray.CreateBuilder<char>(completions.Rules.DefaultCommitCharacters.Length);
-            var completionsBuilder = ImmutableArray.CreateBuilder<CompletionItem>(completions.Items.Length);
+            var completionsBuilder = new List<CompletionItem>();
```
333fred, line 154: Consider leaving this as a builder, and instead calling `ToImmutable` on the builder below instead of `MoveToImmutable()` as the builder is no longer guaranteed to be correctly sized.

filipw: yes thanks it was `ToImmutable()`, just got messed up on the bad merge

Issue #1891, opened 2020-08-16
Cant build


GoldenHuaji:
This is the code

<pre><font color="#CC0000"><b>$ </b></font><font color="#4E9A06">./build.sh</font>
Feeds used:
  /home/goldenhuaji/.nuget/packages/
  https://api.nuget.org/v3/index.json
  https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json
  https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json
  https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet5/nuget/v3/index.json
  https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet5-transport/nuget/v3/index.json
  https://www.myget.org/F/omnisharp/api/v3/index.json

/home/goldenhuaji/.cache/vimfiles/repos/github.com/OmniSharp/omnisharp-vim/omnisharp-roslyn-server/tools/packages.config 中列出的所有程序包均已安装。
<span style="background-color:#2E3436"><font color="#FCE94F">/home/goldenhuaji/.cache/vimfiles/repos/github.com/OmniSharp/omnisharp-vim/omnisharp-roslyn-server/build.cake(48,20): warning CS0618: &apos;ITFBuildProvider.IsRunningOnTFS&apos; is obsolete: &apos;Please use ITFBuildProvider.IsRunningOnAzurePipelines instead.&apos;</font></span>
<span style="background-color:#CC0000"><font color="#EEEEEC">Error: </font></span><span style="background-color:#EF2929"><font color="#EEEEEC">One or more errors occurred. (GitVersion: Process returned an error (exit code 1).)</font></span>
	<span style="background-color:#EF2929"><font color="#EEEEEC">GitVersion: Process returned an error (exit code 1).</font></span>
</pre>

GoldenHuaji:
On  Ubuntu 20.04

filipw:
cc GitVersion experts @mholo65 @david-driscoll

david-driscoll:
I've not seen this issue myself on WSL (not saying it's not an issue).

I can look at updating our version, I'm not sure if we're fully up to date or not.

R0flcopt3r:
Also on Fedora 32

R0flcopt3r:
Also on Fedora 32.

```
> ./build.sh
[...]
[...]

PR #1912, opened 2020-08-24
updated gitversion to 5.3.7

filipw:
I can't reproduce https://github.com/OmniSharp/omnisharp-roslyn/issues/1891 but maybe it will help

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -31,7 +31,7 @@ variables:
   DotNetVersion: "3.1.401"
   CakeVersion: "0.32.1"
   NuGetVersion: "4.9.2"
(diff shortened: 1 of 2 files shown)


filipw:
aactually I will close this. Looks like GitVersion.CommandLine needs to be dropped and we should move to GitVersion.Tool. see https://github.com/cake-build/cake-action/issues/14#issuecomment-647545692

mholo65:
A quick solution would be to disable the GitVersion task in `build.cake` when running on non-windows (`.WithCriteria(() => ...)`). In CI we run GitVersion as an own job, and publish the JSON document from there so that it can be used in other jobs. Wasn't this how you set it up @david-driscoll?

PR #1914, opened 2020-08-26
Build Windows Arm64 package

JoeRobich:
This would make it possible to preview the language service experience on Windows Arm64 devices.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -42,7 +42,7 @@ Note: The arguments below should prefixed with a single hyphen on Windows (Power
   `-install-path Path`: Path used for the **Install** target.
     Defaults to `(%USERPROFILE%|$HOME)/.omnisharp`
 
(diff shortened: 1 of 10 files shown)


filipw (approved):
thanks! is there a way to test this? I think Azure Devops only has support for Linux ARM64

JoeRobich:
hmm, yeah, no automated tests on the Arm64 platform. We could perform manual validation prior to cutting a release, but certainly not on each master build. My line of thinking was that we wouldn't add this as a supported platform in the C# extension, but allow users to set their omnisharp.path to "latest" to opt-in to a preview experience. Another path for me to consider is just having the C# extension treat Win-Arm64 as Win-x86 and let emulation provide "an experience" =)

filipw:
I see, thanks. we can definitely include this as an experimental build

PR #1917, opened 2020-08-27
Use ExecutionPolicy Bypass when running powershell.exe

nohwnd:
Using Bypass policy will avoid errors when running build.cmd on system that does not have ExecutionPolicy set in powershell.exe, or when running build.ps1 from PowerShell Core (pwsh.exe).


filipw (approved):
thanks

PR #1936, opened 2020-09-11
Fix namespace icon

nohwnd:
Namespace was set to Text kind in the #1877, but it should be set to Module kind. 

Fixes https://github.com/OmniSharp/omnisharp-vscode/issues/4051


filipw (approved):
thanks!

PR #1937, opened 2020-09-11
Allow build.cmd to run from any directory

nohwnd:
Use paths relative to the location of build.cmd and build.ps1 when looking for dependencies, this allows the build to be run from any current directory.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -42,7 +42,7 @@ http://cakebuild.net
 Param(
     [parameter(position=0)]
     [string]$Target = "Default",
-    [string]$Script = "build.cake",
+    [string]$Script = "$PSScriptRoot/build.cake",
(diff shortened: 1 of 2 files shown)


build.ps1
```diff
@@ -85,10 +85,6 @@ function MD5HashFile([string] $filePath)
 
 Write-Host "Preparing to run build script..."
 
-if(!$PSScriptRoot){
-    $PSScriptRoot = Split-Path $MyInvocation.MyCommand.Path -Parent
-}
```
nohwnd, line 90: This and the usage on line 45 would break it for PowerShell 2 users, but I doubt there are any users trying to build this on Windows 7 without service pack 🙂

filipw (approved):
looks good, thanks

mholo65 (approved):

PR #1941, opened 2020-09-14
Fix completion on part of existing string

nohwnd:
When completion re-uses part of existing string the proposed change should re-use the part of the string in the original span and commit the change. 

Fix https://github.com/OmniSharp/omnisharp-vscode/issues/4063

### DIFF
--- a/src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionService.cs
+++ b/src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionService.cs
@@ -242,6 +242,19 @@ namespace OmniSharp.Roslyn.CSharp.Services.Completion
                                     break;
                                 }
 
+                                // If the span we are using is re-using part of the typed text we just need to grab the completion an prefix it
+                                // with the existing text. Such as Onenabled -> OnEnabled, this will re-use On of the typed text
+                                if (typedSpan.Start < change.TextChange.Span.Start && typedSpan.Start < change.TextChange.Span.End && typedSpan.End == change.TextChange.Span.End)
+                                {
+                                    var prefix = typedText.Substring(0, change.TextChange.Span.Start - typedSpan.Start);
+                                    
+                                    (insertText, insertTextFormat) = getAdjustedInsertTextWithPosition(change, position, newOffset: 0);
+
+                                    insertText = prefix + insertText;
+
+                                    break;
+                                }
+
                                 int additionalEditEndOffset;
                                 (additionalTextEdits, additionalEditEndOffset) = await GetAdditionalTextEdits(change, sourceText, (CSharpParseOptions)syntax!.Options, typedSpan, completion.DisplayText, providerName);
 
--- a/build.cmd
+++ b/build.cmd
@@ -1,2 +1,2 @@
 @echo off
-powershell.exe -NoProfile -ExecutionPolicy Bypass .\build.ps1 "%*"
+powershell.exe -NoProfile -ExecutionPolicy Bypass %~dp0build.ps1 "%*"
--- a/build.ps1
+++ b/build.ps1
@@ -42,7 +42,7 @@ http://cakebuild.net
 Param(
     [parameter(position=0)]
     [string]$Target = "Default",
-    [string]$Script = "build.cake",
+    [string]$Script = "$PSScriptRoot/build.cake",
     [ValidateSet("Release", "Debug")]
     [string]$Configuration = "Debug",
     [ValidateSet("Quiet", "Minimal", "Normal", "Verbose", "Diagnostic")]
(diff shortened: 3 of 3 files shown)


nohwnd:
the completion does not handle correctly the case when completion keeps a bit of the original text. It works with `one`, because the typed text is on the given position 34, 45 (line, col) the returned change would be `OnEnabled()\n ...` with position 34, 45. But with `One` the change is `Enabled()\n ...` on position 34,47. 

This will skip this condition and instead use the one the is more complicated, that allows overwriting the existing code, but that path does not support this case correctly. 

https://github.com/OmniSharp/omnisharp-roslyn/blob/0587b50d9baea974a76adbbe063a1572ae9382aa/src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionService.cs#L239-L243

I was able to naively fix it by just taking the start of the typedText and prepending it to the proposed change. But this seems like oversimplifying the problem.

filipw:
can you add some tests that reproduce this issue?

src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionService.cs
```diff
@@ -242,6 +242,19 @@ public async Task<CompletionResponse> Handle(CompletionRequest request)
...
 
+                                // If the span we are using is re-using part of the typed text we just need to grab the completion an prefix it
+                                // with the existing text. Such as Onenabled -> OnEnabled, this will re-use On of the typed text
+                                if (typedSpan.Start < change.TextChange.Span.Start && typedSpan.Start < change.TextChange.Span.End && typedSpan.End == change.TextChange.Span.End)
+                                {
+                                    var prefix = typedText.Substring(0, change.TextChange.Span.Start - typedSpan.Start);
+                                    
+                                    (insertText, insertTextFormat) = getAdjustedInsertTextWithPosition(change, position, newOffset: 0);
+
+                                    insertText = prefix + insertText;
```
filipw, line 253: what about a situation where insertText should correct the letter casing of the typed text?

nohwnd: Would that be triggered in this case? I don't think it will because that is covered by the if above this one, then then typedSpan is replaced as a whole. In this case the part of the span is already correct so the suggestion only contains the part that needs to be added. At least that is what I saw when trying `one<tab>` vs `One<tab>`.  In the first case `one` gets `OnEnabled() ...` suggestion, but `One` gets `Enabled() ...` suggestion. The span of the first suggestion is identical to the typedSpan, while the span of the second suggestion starts 2 characters after th typedSpan, because you already have `On` in the text.

nohwnd:
@filipw 

> can you add some tests that reproduce this issue?

Yeah, sure. I could not find where this is tested originally, because I was looking for CompletionService, but now I see there is CompletionFacts.

JoeRobich:
@333fred Maybe you have some thoughts on how we could handle this.

333fred:
I think the correct way to handle this is to use a TextEdit for the change, rather than use insert text. When I implemented this originally, I missed the part of the LSP spec that stated `InsertTextFormat` applied to both the text in `InsertText` and the text in a `TextEdit`. So, the best thing to do is to just have the main insertion be a `TextEdit` that specifies the exact span that should be replaced, including the current cursor location, rather than trying to fix up the insert text for different casing.

333fred:
@nohwnd I opened https://github.com/nohwnd/omnisharp-roslyn/pull/1 against your branch. I believe that it's a more robust implementation that will ensure that future errors like this are taken care of by explicitly providing the range of text to be replaced by the main edit.

src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionService.cs
```diff
@@ -233,14 +236,30 @@ public async Task<CompletionResponse> Handle(CompletionRequest request)
...
+                                if (change.TextChange.Span.Start > typedSpan.Start)
+                                {
+                                    // If the span we're using to key the replacement span within the original typed span
+                                    // span, we want to prepend the missing text from the original typed text to here. The
+                                    // reason is that some lsp clients, such as vscode, use the range from the text edit as
+                                    // the selector for what filter text to use. This can lead to odd scenarios where invoking
+                                    // completion and typing `EQ` will bring up the Equals override, but then dismissing and
+                                    // reinvoking completion will have a range that just replaces the Q. Vscode will then consider
+                                    // that capital Q to be the start of the filter text, and filter out the Equals overload
+                                    // leaving the user with no completion and no explanation
```
333fred, line 256: Technically, vscode does have an API to deal with this consistency. But it's not in the lsp spec, only in the native vscode completion handler api, so we can't use it here.

nohwnd:
@333fred Thanks, that's perfect.

Issue #1949, opened 2020-09-17
Embedded regex language provider support does not seem to escape regex constructs correctly


webczat:
Hi, currently using c# extension version 1.23.2 and testing embedded language providers, mainly regex.
If I type that somewhere in code `Regex.Match("test", "");` and try to trigger suggestions in the empty quotes of second argument, I see that the regex provider is triggered. Then, if I try to select some backslash suggestion like `\s`, what I get after pressing enter is just `"\s". However, because the string is non verbatim, it should be escaped as `"\\s"`.

In addition, if I do the same but trigger completion after first typing two backslashes and then select something like `\A`, I get `"\\\A"` instead of just `"\A"`.

I believe it does work in vs proper, so reporting it here.

CC: @333fred

JoeRobich:
> can you move this to omnisharp-roslyn?

@333fred TIL that I can =)

PR #1950, opened 2020-09-18
Correct regex completion

333fred:
Depending on whether the context is a verbatim string or not, the regex provider will escape the text, so we need to realize it ahead of time. Fixes https://github.com/OmniSharp/omnisharp-roslyn/issues/1949.

### DIFF
--- a/src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionService.cs
+++ b/src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionService.cs
@@ -174,6 +174,16 @@ namespace OmniSharp.Roslyn.CSharp.Services.Completion
                     string providerName = completion.GetProviderName();
                     switch (providerName)
                     {
+                        case CompletionItemExtensions.EmeddedLanguageCompletionProvider:
+                            // The Regex completion provider can change escapes based on whether
+                            // we're in a verbatim string or not
+                            {
+                                CompletionChange change = await completionService.GetChangeAsync(document, completion);
+                                Debug.Assert(typedSpan == change.TextChange.Span);
+                                insertText = change.TextChange.NewText!;
+                            }
+                            break;
+
                         case CompletionItemExtensions.InternalsVisibleToCompletionProvider:
                             // The IVT completer doesn't add extra things before the completion
                             // span, only assembly keys at the end if they exist.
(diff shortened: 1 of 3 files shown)


filipw (approved):
ai, good catch

333fred:
@filipw do you have any idea what could cause projects not be found on the Mac and Linux test legs? It seems that both the Mac and Linux MSBuild tests are seeing something like:

```
        [dbug]: OmniSharp.MSBuild.ProjectSystem
                Could not locate project for '/home/vsts/work/1/s/bin/Debug/OmniSharp.Roslyn.CSharp.Tests/net472/project.csproj'
        [dbug]: OmniSharp.MSBuild.ProjectSystem
                Could not locate project for 'dummy.csx'
        [dbug]: OmniSharp.Script.ScriptProjectSystem
                Could not locate project for 'dummy.csx'
        [trce]: OmniSharp.Roslyn.CSharp.Services.Completion.CompletionService
                Completions requested
        [trce]: OmniSharp.Roslyn.CSharp.Services.Completion.CompletionService
                Found 0 completions for dummy.csx:6,2
```

I'm especially confused because I'm doing nothing different in these tests than I'm doing for the other tests in the file, but these both are failing to locate projects.

filipw:
yes I saw this before here https://github.com/OmniSharp/omnisharp-roslyn/pull/1871/commits/56afd278dbead2d2bab11e26d6925764710deb27

at the time I just thought it was flaky and ended up removing it. Now it looks to me that CSX project system with project per CSX file somehow fails with the embedded language provider with Regex on *nix. I think it's OK to run these tests for *.cs only for now and we can investigate separately

333fred:
> I think it's OK to run these tests for *.cs only for now and we can investigate separately - at the end there is no regression here, it was just not being observed

Actually, these tests fail for .cs files as well.

filipw:
hmm yes you are right. I would imagine we could repro this on a simple console project on Mono - perhaps it's a Roslyn bug?
either way, there is no regression here so adding `[ConditionalFact(typeof(WindowsOnly))]` would restrict it to Windows only for the time being. it shouldn't hold up this PR IMHO

333fred:
Agreed, and one.

filipw:
thanks a lot

PR #1956, opened 2020-09-25
Correctly handle <ProjectReferences> that don't produce references

jasonmalinowski:
<ProjectReference> elements can have metadata that specify what type of item to create, or that they shouldn't create items at all. The existing code that was processing references wasn't checking for this metadata, so you could end up with project references between projects that shouldn't be there.

The approach to take here is to never look at the <ProjectReference> items directly, rather only look at the ReferencePath items that are produced, and if we see metadata that indicates it came from a project, convert it to a project reference at that point.

### DIFF
--- a/src/OmniSharp.MSBuild/ProjectFile/ProjectFileInfo.ProjectData.cs
+++ b/src/OmniSharp.MSBuild/ProjectFile/ProjectFileInfo.ProjectData.cs
@@ -240,8 +240,10 @@ namespace OmniSharp.MSBuild.ProjectFile
                     documentationFile, preprocessorSymbolNames, suppressedDiagnosticIds, warningsAsErrors, warningsNotAsErrors, signAssembly, assemblyOriginatorKeyFile, treatWarningsAsErrors, defaultNamespace, runAnalyzers, runAnalyzersDuringLiveAnalysis, ruleset: null);
             }
 
-            public static ProjectData Create(MSB.Execution.ProjectInstance projectInstance, MSB.Evaluation.Project project)
+            public static ProjectData Create(string projectFilePath, MSB.Execution.ProjectInstance projectInstance, MSB.Evaluation.Project project)
             {
+                var projectFolderPath = Path.GetDirectoryName(projectFilePath);
+
                 var guid = PropertyConverter.ToGuid(projectInstance.GetPropertyValue(PropertyNames.ProjectGuid));
                 var name = projectInstance.GetPropertyValue(PropertyNames.ProjectName);
                 var assemblyName = projectInstance.GetPropertyValue(PropertyNames.AssemblyName);
@@ -300,14 +302,16 @@ namespace OmniSharp.MSBuild.ProjectFile
                     // MSBuild logic is adding or removing properties too.
                     if (StringComparer.OrdinalIgnoreCase.Equals(referenceSourceTarget, ItemNames.ProjectReference))
                     {
-                        var sourceProjectFile = referencePathItem.GetMetadataValue(MetadataNames.MSBuildSourceProjectFile);
-                        if (sourceProjectFile.EndsWith(".csproj", StringComparison.OrdinalIgnoreCase))
+                        var projectReferenceOriginalItemSpec = referencePathItem.GetMetadataValue(MetadataNames.ProjectReferenceOriginalItemSpec);
+                        if (projectReferenceOriginalItemSpec.EndsWith(".csproj", StringComparison.OrdinalIgnoreCase))
                         {
-                            projectReferences.Add(sourceProjectFile);
+                            var projectReferenceFilePath = Path.GetFullPath(Path.Combine(projectFolderPath, projectReferenceOriginalItemSpec));
+
+                            projectReferences.Add(projectReferenceFilePath);
 
                             if (!string.IsNullOrEmpty(aliases))
                             {
-                                projectReferenceAliases[sourceProjectFile] = aliases;
+                                projectReferenceAliases[projectReferenceFilePath] = aliases;
                             }
 
                             continue;
--- a/src/OmniSharp.MSBuild/ProjectFile/ProjectFileInfo.ProjectData.cs
+++ b/src/OmniSharp.MSBuild/ProjectFile/ProjectFileInfo.ProjectData.cs
@@ -286,41 +286,30 @@ namespace OmniSharp.MSBuild.ProjectFile
 
                 var projectReferences = ImmutableArray.CreateBuilder<string>();
                 var projectReferenceAliases = ImmutableDictionary.CreateBuilder<string, string>();
-                var projectReferencesAdded = new HashSet<string>();
-                foreach (var projectReferenceItem in projectInstance.GetItems(ItemNames.ProjectReference))
-                {
-                    var fullPath = projectReferenceItem.GetMetadataValue(MetadataNames.FullPath);
-
-                    if (IsCSharpProject(fullPath) && projectReferencesAdded.Add(fullPath))
-                    {
-                        projectReferences.Add(fullPath);
-
-                        var aliases = projectReferenceItem.GetMetadataValue(MetadataNames.Aliases);
-                        if (!string.IsNullOrEmpty(aliases))
-                        {
-                            projectReferenceAliases[fullPath] = aliases;
-                        }
-                    }
-                }
 
                 var references = ImmutableArray.CreateBuilder<string>();
                 var referenceAliases = ImmutableDictionary.CreateBuilder<string, string>();
                 foreach (var referencePathItem in projectInstance.GetItems(ItemNames.ReferencePath))
                 {
                     var referenceSourceTarget = referencePathItem.GetMetadataValue(MetadataNames.ReferenceSourceTarget);
+                    var aliases = referencePathItem.GetMetadataValue(MetadataNames.Aliases);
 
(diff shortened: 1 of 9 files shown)


src/OmniSharp.MSBuild/ProjectFile/ProjectFileInfo.ProjectData.cs
```diff
@@ -286,41 +286,30 @@ public static ProjectData Create(MSB.Execution.ProjectInstance projectInstance,
...
-                        // If the reference was sourced from a project reference, we have two choices:
-                        //
-                        //   1. If the reference is a C# project reference, we shouldn't add it because it'll just duplicate
-                        //      the project reference.
-                        //   2. If the reference is *not* a C# project reference, we should keep this reference because the
-                        //      project reference was already removed.
-
-                        var originalItemSpec = referencePathItem.GetMetadataValue(MetadataNames.OriginalItemSpec);
-                        if (originalItemSpec.EndsWith(".csproj", StringComparison.OrdinalIgnoreCase))
+                        var sourceProjectFile = referencePathItem.GetMetadataValue(MetadataNames.MSBuildSourceProjectFile);
```
jasonmalinowski, line 303: Still confirming with the MSBuild team precise which property should be used here.

test-assets/test-projects/ProjectWithAnalyzersFromReference/ConsumingProject/ConsumingProject.csproj
```diff
@@ -0,0 +1,14 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+  </PropertyGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\Analyzer\Analyzer.csproj">
+      <OutputItemType>Analyzer</OutputItemType>
+      <ReferenceOutputAssembly>false</ReferenceOutputAssembly>
```
jasonmalinowski, line 10: This is a pattern we're seeing is more common with people writing source generators, since you often want to have the generator in your same solution as the project itself. It's not the _best_ supported because reloading the analyzer is often difficult if it changes on disk.

JoeRobich:
@jasonmalinowski If for any reason you need your test project to be built prior to your new unit test running then you can add the folder name to [this list](https://github.com/OmniSharp/omnisharp-roslyn/blob/master/build.json#L29)

jasonmalinowski:
@JoeRobich Shouldn't be necessary, and honestly if it _was_ that'd be a bug in it's own right.

filipw (approved):
this looks good to me, can we "undraft" it?

src/OmniSharp.MSBuild/ProjectFile/ProjectFileInfo.ProjectData.cs
```diff
@@ -286,41 +288,37 @@ public static ProjectData Create(MSB.Execution.ProjectInstance projectInstance,
...
-                        }
-                    }
-                }
 
                 var references = ImmutableArray.CreateBuilder<string>();
                 var referenceAliases = ImmutableDictionary.CreateBuilder<string, string>();
                 foreach (var referencePathItem in projectInstance.GetItems(ItemNames.ReferencePath))
                 {
                     var referenceSourceTarget = referencePathItem.GetMetadataValue(MetadataNames.ReferenceSourceTarget);
+                    var aliases = referencePathItem.GetMetadataValue(MetadataNames.Aliases);
```
filipw, line 297: I believe the old method `IsCSharpProject` can now be deleted

src/OmniSharp.MSBuild/ProjectFile/ProjectFileInfo.ProjectData.cs
rainersigwald: * `OriginalItemSpec`: available on all RAR outputs. The `Include` value of the RAR input that caused that output. (This is actually automatically set in the engine whenever populating an item from another item)
* `ProjectReferenceOriginalItemSpec`: available on RAR outputs that were resolved from `@(ProjectReference)` items. The `OriginalItemSpec` value of the `Reference` that `ResolveProjectReferences` produced from the `ProjectReference` item--so the `ProjectReference`'s `Include`.
* `OriginalProjectReferenceItemSpec: ¯\\(°_o)/¯
  > Copy OriginalItemSpec to OriginalProjectReferenceItemSpec, so that when ResolveAssemblyReferences
  > takes these items and resolves them to ReferencePath, we can still recover the _real_ OriginalItemSpec
  > for the unresolved reference items.
  
  Doesn't appear to be used anywhere. My advice: avoid.
* `MSBuildSourceProjectFile`: available on all items that are outputs of the `<MSBuild` task. Points to the project that created the item. For RAR outputs, inherited from the Reference generated by `<MSBuild` in `ResolveProjectReferences`


So the thing that most matches your use case is `ProjectReferenceOriginalItemSpec`. It has a weakness though: it's a relative path, not fully resolved. As long as you deal with that I'd go with it. `MSBuildSourceProjectFile` should always be an absolute path which is nice but it could get confused with custom reference-resolution schemes.

JoeRobich: @rainersigwald What we noticed is that the fully qualified `MSBuildSourceProjectFile` prepends the `/private/` folder name on MacOS for the shadowcopied unit tests which causes them to be different that the paths we qualifyw with `Path.GetFullPath`. I'll take your advice and update to use `ProjectReferenceOriginalItemSpec`.

src/OmniSharp.MSBuild/ProjectFile/ProjectFileInfo.ProjectData.cs
JoeRobich: Removed

jasonmalinowski: Actually it was probably necessary: if we have a reference coming from a VB project or a C++ project, we probably need to keep it as a metadata reference?

src/OmniSharp.MSBuild/ProjectFile/ProjectFileInfo.ProjectData.cs
```diff
@@ -286,41 +287,37 @@ public static ProjectData Create(MSB.Execution.ProjectInstance projectInstance,
...
-
-                        var originalItemSpec = referencePathItem.GetMetadataValue(MetadataNames.OriginalItemSpec);
-                        if (originalItemSpec.EndsWith(".csproj", StringComparison.OrdinalIgnoreCase))
+                        var projectReferenceOriginalItemSpec = referencePathItem.GetMetadataValue(MetadataNames.ProjectReferenceOriginalItemSpec);
+                        if (projectReferenceOriginalItemSpec.EndsWith(".csproj", StringComparison.OrdinalIgnoreCase))
                         {
+                            // Convert rooted project reference file paths ("/home/user/source/project/../other_project/project.csproj") to
+                            // full paths and unrooted project reference file paths ("./folder/project.csproj") to rooted full paths
+                            // by combining with the project file's path.
+                            var projectReferenceFilePath = Path.IsPathRooted(projectReferenceOriginalItemSpec)
```
jasonmalinowski, line 310: @JoeRobich Why the IsRooted check? Path.Combine() is documented as:

> However, if an argument other than the first contains a rooted path, any previous path components are ignored, and the returned string begins with that rooted path component.

JoeRobich: doh

src/OmniSharp.MSBuild/ProjectFile/ProjectFileInfo.ProjectData.cs
jasonmalinowski: Oh I see: there were two places this function checked if a project was a .csproj; I deleted the one using the function but the other one was never using it.

@filipw @JoeRobich any preference for having the .csproj check below call the function, versus just deleting it?

jasonmalinowski: @JoeRobich replied privately to keep it deleted.

filipw: I'd say pragmatically that when usage count goes to 1, the function is no longer needed 🙂

filipw:
sorry! it needs another rebase/merge due to a merge of #1964

jasonmalinowski:
@filipw Rebased. I swear you're the only repo I know that actually has that 'must be up to date' thing checked. :smile:

filipw (approved):
:shipit: thank you!

filipw:
also, I learnt something here and I can now drop `ProjectReferenceOriginalItemSpec` in some casual conversation around project structures

jasonmalinowski:
@filipw And note that generally, in Visual Studio for Windows we don't look at that property, and instead just wire up things based on output paths. So if project A is producing a DLL to an output path and project B is consuming that same (exact, full) path, we wire them up at that point. That may be something that needs to happen here longer term; this fix at least will help in unblocking us investigating other issues.

JoeRobich:
Thanks!

PR #1957, opened 2020-09-26
Property completion

333fred:
Based on https://github.com/OmniSharp/omnisharp-roslyn/pull/1941. In the process of playing around with the code, I discovered that I totally forgot about overriding properties when I implemented completion... 🤦‍♂️. This implements a more robust algorithm for finding the correct location in the changed text, and also handles property overrides correctly.


333fred:
The new part of this change is in commit 9. Commits 1-7 are based on #1941, and commit 8 is merging in master.

333fred:
Alright, I _really_ hope that nothing I did broke azdo package upload, because if it _is_ something I did then everything is a lie.

filipw (approved):
thanks a lot! I feel a bit embarrassed that I didn't notice the property overrides thing..

src/OmniSharp.Abstractions/Models/v1/Completion/CompletionItem.cs
```diff
@@ -58,13 +59,21 @@ public class CompletionItem
         /// A string that should be inserted into a document when selecting
         /// this completion.When null or empty the label is used.
         /// </summary>
+        [Obsolete("Use TextEdit instead", error: true)]
         public string? InsertText { get; set; }
```
filipw, line 63: I think you might as well break the contract and delete the property altogether. It's not populated anymore in the current implementation so it is not backwards compatible anymore anyway - therefore it would probably be more visible for the consumers that they need to change if it's not there at all

src/OmniSharp.Abstractions/Models/v1/Completion/CompletionItem.cs
```diff
@@ -58,13 +59,21 @@ public class CompletionItem
...
 
         /// <summary>
```
333fred, line 65:
```suggestion
```

333fred:
> thanks a lot! I feel a bit embarrassed that I didn't notice the property overrides thing..

Same...

filipw:
@333fred looks like this fails on linux

```
    OmniSharp.Cake.Tests.CompletionFacts.ShouldGetAdditionalTextEditsFromOverrideCompletion [FAIL]
      Assert.Equal() Failure
      Expected: String[] ["public override bool", "public override int", "public override void", "public override void", "public override string"]
      Actual:   SelectIListIterator<CompletionItem, String> ["Cake.Core.IO.FilePath assemblyInfoPath)\n{\n    re"..., "public override int", "public override void", "public override void", "public override string"]
      Stack Trace:
          at OmniSharp.Cake.Tests.CompletionFacts.ShouldGetAdditionalTextEditsFromOverrideCompletion () [0x002a5] in <895d19e3d19548c59ba9489d2daa2025>:0 
      Output:
```

not sure if this is transient or not, let's see if the retry helps.  Looks like Cake has some special host object property that is overrideable? @mholo65 not sure why it would affect Linux only  though

filipw:
hmm it passed now, but looking at the error above it looks like there is some race condition or other indeterministic state in the Cake tests. we can merge for now but we should keep an eye on this

mholo65:
@filipw thanks! I will dig into the Cake error and see if I find something obvious.

PR #1958, opened 2020-09-26
Apply bulk changes

333fred:
I rebased this on current master and added an additional setting to `ApplyChangesTogether` to default it to false when not present.

@JoeRobich can you describe the issues you were still seeing in vscode with this change applied? I tried reverting the revert of my vscode-side change and adding this flag (https://github.com/OmniSharp/omnisharp-vscode/pull/4088), and everything that I've tried since has seemed to work. Completion still triggers correctly, colorization appears to work, etc, but I'm not sure if I'm missing a test scenario you had? What I tried:
1. Wait for omnisharp to load
2. Rename a local used in multiple location/Use multiple cursors to type garbage in a bunch of locations
3. Verify that the buffer update had multiple updates in it, not one big update.
4. Tested colorization and intellisense after doing this

### DIFF
--- a/src/OmniSharp.Abstractions/Models/Request.cs
+++ b/src/OmniSharp.Abstractions/Models/Request.cs
@@ -11,5 +11,6 @@ namespace OmniSharp.Models
         public int Column { get; set; }
         public string Buffer { get; set; }
         public IEnumerable<LinePositionSpanTextChange> Changes { get; set; }
+        public bool ApplyChangesTogether { get; set; }
     }
 }
--- a/src/OmniSharp.Abstractions/Models/Request.cs
+++ b/src/OmniSharp.Abstractions/Models/Request.cs
@@ -11,6 +11,7 @@ namespace OmniSharp.Models
         public int Column { get; set; }
         public string Buffer { get; set; }
         public IEnumerable<LinePositionSpanTextChange> Changes { get; set; }
+        [JsonProperty(DefaultValueHandling = DefaultValueHandling.Populate)]
         public bool ApplyChangesTogether { get; set; }
     }
 }
--- a/src/OmniSharp.Roslyn/BufferManager.cs
+++ b/src/OmniSharp.Roslyn/BufferManager.cs
@@ -32,7 +32,7 @@ namespace OmniSharp.Roslyn
 
         public bool IsTransientDocument(DocumentId documentId)
         {
-            lock(_lock)
+            lock (_lock)
             {
                 return _transientDocumentIds.Contains(documentId);
             }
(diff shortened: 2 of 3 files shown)


JoeRobich:
> @JoeRobich can you describe the issues you were still seeing in vscode with this change applied?

1. Create a new console project `dotnet new console -o TestApp`
2. Open the TestApp folder in VS Code
3. Wait for OmniSharp to fully load the project
4. Run fix all in project for CS8019
image

Expected outcome:
All unused using directives are removed and the editor is in sync with OmniSharp

Actual outcome:
The editor and OmniSharp are out of sync . This is very noticable with CodeLens placement.

333fred:
Alright. So `dotnet new console` does not ship with any codefixes that need to be applied, but `dotnet new wpf` does the trick. As to the bug itself, it appears that what happens is we apply all code fixes, send the updates to vscode, and then vscode notices the updates and sends them back to us. So, for example, if 6 usings are deleted, we end up seeing the change twice. I need to dig into what the apply all codefixes action is doing, and why we're seeing double updates for that scenario.

333fred:
@savpek it appears that `RunFixAllCodeActionService` applies all changes to the workspace, rather than sending the requisite changes to the client, letting the client apply them, and then sending notification back to Omnisharp. Presumably, no one is both using the fix all service and sending any form of batched updates today, as this will break the existing form of batched updates in Omnisharp today. Do you have a sense of how hard it will be to break that dependency, such that RunFixAll does not apply the edits itself? I attempted to do a naive change of just taking the changed solution from each applyChangedOperation as the current solution, rather than applying the edit to the actual workspace, but it appears that this breaks the ability of the `FixAllProvider` to actually get the fix for the document being requested. I'm going to keep digging, but if you have any ideas I'd appreciate it :)

333fred:
Alright, I dug into this a bit more. The issue with the fix all provider occurs when attempting to fix all issues in the current document, ie no filter. I'm still not entirely _why_ that's an issue, because when the changes are applied to the workspace the document still changes and it should break in the same way, but it doesn't. I'm still digging into that part, but I'm also not sure if fix all issues in document is even used? I don't see a VSCode action to fix all items in the current file, and I'm unsure if it's been added to other omnisharp clients. A compromise could be to only support applying a fixall to a specific issue, rather than a fixall to all issues in the current document, and that should allow us to not apply the changes server side, remove the duplicate change notifications, and then enable batched changes for much better performance in large files.

333fred:
Alright, I've been able to decouple the fix all service from the workspace, submitted a PR for that here: https://github.com/OmniSharp/omnisharp-roslyn/pull/1962. I've then tested this change in combination with that, and fix all works correctly with it applied, and buffer updates are incremental.

333fred:
https://github.com/OmniSharp/omnisharp-roslyn/pull/1962 decoupled the dependency here, so I think this is safe to merge now.

333fred:
I reverted the linquify change because I noticed it was actually capturing state that I missed on my first pass, which is not great for a hot loop like that.

333fred:
@JoeRobich @filipw I believe this is ready for review.

filipw (approved):
LGTM, thanks a lot!

JoeRobich (approved):

Issue #1960, opened 2020-09-27
Fix All always does the entire project, regardless of input


333fred:
https://github.com/OmniSharp/omnisharp-roslyn/blob/ed3d4173580283de22fb92341f2bfaa9c7923f5b/src/OmniSharp.Roslyn.CSharp/Services/Refactoring/RunFixAllCodeActionService.cs#L81

This ignores the input fix all scope, and always runs fixers over the entire project, regardless of whether just one document is requested.

Example of the behavior:
1. Create a new WPF app with `dotnet new wpf`.
2. Open in vscode
3. Open Main.xaml.cs
4. Run "Fix all in document" on unused usings.

Expected:

App.xaml.cs is not touched.

Actual:

Unused usings are removed from App.xaml.cs as well.

PR #1962, opened 2020-09-28
Decouple FixAll from the workspace

333fred:
Today, the RunFixAllInCodeActionService is tightly coupled to the current workspace, applying its changes to the workspace as it processes the request and counting on the updates from doing so to invalidate diagnostic caches. While this does work, it's has 2 major downsides:

1. It's a race condition waiting to happen, where the propagation of the changed documents invalidating the diagnostic provider's caches races with the next fix all operation attempting to get the diagnostics it needs to fix.
2. It has bad interations with the buffer updater from clients. The service returns the needed changes to clients, who then apply the changes. The client will then see that changes have been applied, and send them back to the server. The server's workspace has already been updated with these changes, so it ends up double applying the changes, leaving the server's workspace in a bad state with garbage in many files.

This commit breaks the dependency of the FixAll service on the current workspace. We now have the fixes occur on top of the previous fix's changed solution. At the end, if the client requested the changes be applied on the workspace side, then we apply the changes to the workspace at that point. In order for this work, I had to modify the diagnostic workers a bit to add an option to force a document to be analyzed, skipping the cache entirely. This is necessary because the cache is only invalidated on document add/remove/change in the workspace itself. The documents we work on linearly in the fix all provider do not get added/changed/removed in the workspace, so we need to analyze them again, right then and there. To support this, the AnalyzerWorkQueue is now based on documents, not document ids, though the caches in the workers are still based on ids.

In the process of doing this PR, I also fixed https://github.com/OmniSharp/omnisharp-roslyn/issues/1960.


333fred:
@filipw I'm not sure if any of these test failures could be related to this change. Can you advise?

filipw:
no, all three separate failures are "known" to be flaky already... let's see the retry

src/OmniSharp.Roslyn.CSharp/Workers/Diagnostics/CSharpDiagnosticWorkerWithAnalyzers.cs
```diff
@@ -83,28 +83,54 @@ public async Task<ImmutableArray<DocumentDiagnostics>> GetDiagnostics(ImmutableA
...
+
+                return resultsBuilder.ToImmutable();
+            }
+
+            var documentIds = documents.SelectAsArray(d => d.Id);
+
+            return await GetDiagnosticsByDocumentIds(documents, waitForDocuments: true);
+        }
+
+        private async Task<ImmutableArray<DocumentDiagnostics>> GetDiagnosticsByDocumentIds(ImmutableArray<Document> documents, bool waitForDocuments)
```
JoeRobich, line 112: How about `GetDiagnosticsForDocuments`?

src/OmniSharp.Roslyn.CSharp/Workers/Diagnostics/CSharpDiagnosticWorkerWithAnalyzers.cs
```diff
@@ -301,24 +341,24 @@ private void EmitDiagnostics(DocumentDiagnostics results)
...
         {
-            var allDocumentsIds = _workspace.CurrentSolution.Projects.SelectMany(x => x.DocumentIds).ToImmutableArray();
-            return await GetDiagnosticsByDocumentIds(allDocumentsIds, waitForDocuments: false);
+            var allDocuments = _workspace.CurrentSolution.Projects.SelectMany(x => x.Documents).ToImmutableArray();
+            return await GetDiagnosticsByDocumentIds(allDocuments, waitForDocuments: false);
         }
 
         public ImmutableArray<DocumentId> QueueDocumentsForDiagnostics(ImmutableArray<ProjectId> projectIds)
         {
             var documentIds = projectIds
```
JoeRobich, line 357: `documents`?

JoeRobich (approved):

src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/BaseCodeActionService.cs
```diff
@@ -16,22 +17,25 @@
...
 using OmniSharp.Models.V2.CodeActions;
+using OmniSharp.Roslyn.CSharp.Helpers;
 using OmniSharp.Roslyn.CSharp.Services.Diagnostics;
 using OmniSharp.Roslyn.CSharp.Workers.Diagnostics;
 using OmniSharp.Roslyn.Utilities;
 using OmniSharp.Services;
 using OmniSharp.Utilities;
 using FixAllScope = OmniSharp.Abstractions.Models.V1.FixAll.FixAllScope;
 
+#nullable enable
```
david-driscoll, line 28: How dare you nullable! (j/k!)

src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/BaseCodeActionService.cs
```diff
@@ -211,33 +215,47 @@ private IEnumerable<AvailableCodeAction> ConvertToAvailableCodeAction(IEnumerabl
...
-        private async Task<ImmutableArray<DocumentDiagnostics>> GetCorrectDiagnosticsInScope(FixAllScope scope, string fileName)
+        protected async Task<ImmutableArray<DocumentDiagnostics>> GetDiagnosticsAsync(FixAllScope scope, Document document)
         {
             switch (scope)
             {
                 case FixAllScope.Solution:
-                    var documentsInSolution = Workspace.CurrentSolution.Projects.SelectMany(x => x.Documents).Select(x => x.FilePath).ToImmutableArray();
-                    return await diagnostics.GetDiagnostics(documentsInSolution);
+                    var documentsInSolution = document.Project.Solution.Projects.SelectMany(p => p.Documents).ToImmutableArray();
+                    return await _diagnostics.GetDiagnostics(documentsInSolution, skipCache: true);
```
david-driscoll, line 251: what are the advantages / disadvantages of using `skipCache` here?

src/OmniSharp.Roslyn.CSharp/Workers/Diagnostics/AnalyzerWorkQueue.cs
```diff
@@ -44,7 +44,7 @@ public AnalyzerWorkQueue(ILoggerFactory loggerFactory, int timeoutForPendingWork
             _maximumDelayWhenWaitingForResults = timeoutForPendingWorkMs;
         }
 
-        public void PutWork(IReadOnlyCollection<DocumentId> documentIds, AnalyzerWorkType workType)
+        public void PutWork(IReadOnlyCollection<Document> documents, AnalyzerWorkType workType)
```
david-driscoll, line 47: would this mean we would have stale documents in the queue potentially?

src/OmniSharp.Roslyn.CSharp/Workers/Diagnostics/CSharpDiagnosticWorkerWithAnalyzers.cs
```diff
@@ -152,7 +178,7 @@ private void EventIfBackgroundWork(AnalyzerWorkType workType, string projectPath
                 _forwarder.ProjectAnalyzedInBackground(projectPath, status);
         }
 
-        private void QueueForAnalysis(ImmutableArray<DocumentId> documentIds, AnalyzerWorkType workType)
+        private void QueueForAnalysis(ImmutableArray<Document> documentIds, AnalyzerWorkType workType)
```
david-driscoll, line 181: `documentIds` -> `documents`

src/OmniSharp.Roslyn.CSharp/Workers/Diagnostics/CSharpDiagnosticWorkerWithAnalyzers.cs
```diff
@@ -177,7 +203,7 @@ private void OnWorkspaceChanged(object sender, WorkspaceChangeEventArgs changeEv
                 case WorkspaceChangeKind.ProjectChanged:
                 case WorkspaceChangeKind.ProjectReloaded:
                     _logger.LogDebug($"Project {changeEvent.ProjectId} updated, reanalyzing its diagnostics.");
-                    var projectDocumentIds = _workspace.CurrentSolution.GetProject(changeEvent.ProjectId).Documents.Select(x => x.Id).ToImmutableArray();
+                    var projectDocumentIds = _workspace.CurrentSolution.GetProject(changeEvent.ProjectId).Documents.ToImmutableArray();
```
david-driscoll, line 206: `projectDocumentIds` -> `projectDocuments`

src/OmniSharp.Roslyn.CSharp/Workers/Diagnostics/AnalyzerWorkQueue.cs
```diff
@@ -56,21 +56,21 @@ public void PutWork(IReadOnlyCollection<DocumentId> documentIds, AnalyzerWorkTyp
                 if (queue.WorkPendingToken == null)
                     queue.WorkPendingToken = new CancellationTokenSource();
 
-                queue.WorkWaitingToExecute = queue.WorkWaitingToExecute.Union(documentIds);
+                queue.WorkWaitingToExecute = queue.WorkWaitingToExecute.Union(documents);
```
david-driscoll, line 59: If an old version of the document exists in the queue, any idea what this will do?

1) Leave both documents (old and new)?
2) Leave the old document (union will check equality)?
3) Remove the old document and use the new document?

Logically I don't think 3 will ever happen, but I'm not sure about the equality behavior for 1 or 2.

david-driscoll (changes requested):

src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/BaseCodeActionService.cs
david-driscoll: reread the pr, I'm slow ignore me here, lol.

src/OmniSharp.Roslyn.CSharp/Workers/Diagnostics/AnalyzerWorkQueue.cs
333fred: `Document`s are not equal across `Solution`s, so both would be in the queue. This is fine on its own, but I think it points out an issue that the caching layer shouldn't operate on the DocumentId's, but rather on the `Document`s themselves. The solution is to use a CWT, like Roslyn's analyzer driver, from `Document` to the diagnostics. I'll go ahead and do that.

src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/BaseCodeActionService.cs
333fred: Since I'm moving the caching layer to use a CWT, I can remove the `skipCache` parameter entirely now.

src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/BaseCodeActionService.cs
333fred: I know, we put too much time into it to use!

src/OmniSharp.Roslyn.CSharp/Workers/Diagnostics/AnalyzerWorkQueue.cs
333fred: Yep. That's really fine: Documents are immutable, so there's no threading concern. It could mean that we do a bit of extra work, but analysis on a single document or even group of documents isn't _that_ expensive, particularly since we don't have a UI thread dependency here.

333fred:
@david-driscoll I believe I've addressed your feedback.

src/OmniSharp.Roslyn.CSharp/Services/Refactoring/RunFixAllCodeActionService.cs
```diff
@@ -44,76 +46,140 @@ public class RunFixAllCodeActionService : BaseCodeActionService<RunFixAllRequest
...
+            // Currently, there are no roslyn changes that will result in multiple ApplyChangesOperations
+            Debug.Assert(operations.OfType<ApplyChangesOperation>().Count() < 2);
+            if (operations.OfType<ApplyChangesOperation>().FirstOrDefault() is ApplyChangesOperation applyChangesOperation)
+            {
+                _logger.LogTrace("Found apply changes operation for {0}.", diagnosticId);
+                return applyChangesOperation.ChangedSolution.GetDocument(originalDoc.Id);
+            }
+            else
+            {
+                _logger.LogTrace("No apply changes operation for {0}.", diagnosticId);
```
filipw, line 180: there will be some cases when the code action did not produce `ApplyChangesOperation` (e.g. modifying files that are not part of the solution)

currently in the `RunCodeActionService` we'd try to apply those directly (https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/RunCodeActionService.cs#L73-L74). I think it would be consistent to have same behavior here

333fred: I agree, but I think that's outside the scope of this PR. I'd be happy to file an issue for following up on this but I'd prefer this change not have any more changes then necessary to decouple the service from the workspace.

filipw (approved):
:shipit:

JoeRobich:
@david-driscoll Would you like to give this PR your blessing? =)

david-driscoll (approved):
:shipit:

PR #1964, opened 2020-09-28
added binding redirects for Microsoft.CodeAnalysis.Features and Microsoft.CodeAnalysis.CSharp.Features

filipw:
This is a best guess attempt to improve the behavior mentioned in https://github.com/OmniSharp/omnisharp-vscode/issues/4090

The error is vague but looks like somehow there are different versions of `CompletionList` loaded simultaneously. 
At the moment there are no binding redirects for its parent assembly `Microsoft.CodeAnalysis.Features` so it wouldn't hurt to add that.

cc @333fred


mholo65 (approved):
LGTM

PR #1972, opened 2020-10-05
When an analyzer fails to load, log an error

jasonmalinowski:
Right now we just silently fail which isn't a good debugging experience. Ideally these would be shown in the error list but this is good enough when trying to track things down.


filipw (approved):
thanks, makes sense. there is a somewhat related thing to it that for code actions that come from predefined assemblies (roslyn features. roslynator etc) we currently ignore those with generics (in case you'd wonder why something should light up but it doesn't)

https://github.com/OmniSharp/omnisharp-roslyn/blob/62b3b52d01251fdc0564a600010936e677f24a2e/src/OmniSharp.Roslyn/Services/AbstractCodeActionProvider.cs#L31

JoeRobich (approved):

PR #1974, opened 2020-10-06
Bump Roslyn to 3.8.0-4.20503.2

JoeRobich:
This closely matches what will ship in VS 16.8 Preview 4


filipw (approved):
LGTM. do you know if there is a public date for 5.0 RC2 set already?

JoeRobich:
> do you know if there is a public date for 5.0 RC2 set already?

I don't know if we've announced a date, but it is soon.

PR #1979, opened 2020-10-10
Quote config_file in Mono ENV Options

mholo65:
This should fix an issue when there's a space in the path where OmniSharp is installed.

### DIFF
--- a/mono-packaging/run
+++ b/mono-packaging/run
@@ -19,7 +19,7 @@ if [ "$1" = "--no-omnisharp" ]; then
 fi
 
 export MONO_CFG_DIR=${etc_dir}
-export MONO_ENV_OPTIONS="--assembly-loader=strict --config ${config_file}"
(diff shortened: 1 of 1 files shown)


mholo65:
See https://twitter.com/Nick_Craver/status/1315008491812270085?s=19

filipw (approved):

NickCraver (approved):
:+1: I have verified this locally - solved my issue. Thank you!

henryju:
Hello, I'm affected by the same bug. It seems that even latest release (v1.37.10) still doesn't contain the fix. How is it possible?
You can download https://github.com/OmniSharp/omnisharp-roslyn/releases/download/v1.37.10/omnisharp-linux-x64.zip and open the `run` file, and the content is:
`export MONO_ENV_OPTIONS="--assembly-loader=strict --config ${config_file}"`

PR #1981, opened 2020-10-12
Update included Mono to 6.12.0

JoeRobich:
I saw 6.12.0 was listed for download as stable.

Also, noticed that the 6.12.0 Mono ships with a 16.8 MSBuild meaning we can revert the meaning of `omnisharp.useGlobalMono` when we take the next release. Verified that the 6.12 Mono's MSBuild handles loading the dotnet/sdk sdk.sln which is firmly .NET 5. =)
```patch
        Located 2 MSBuild instance(s)
!            1: Mono 16.8.0 - "/Library/Frameworks/Mono.framework/Versions/6.12.0/lib/mono/msbuild/Current/bin"
            2: StandAlone 16.8.0 - "/Users/joeyrobichaud/.vscode-insiders/extensions/ms-dotnettools.csharp-1.23.3/.omnisharp/1.37.1/omnisharp/.msbuild/Current/Bin"
```

### DIFF
--- a/.pipelines/init.yml
+++ b/.pipelines/init.yml
@@ -44,7 +44,7 @@ steps:
       arguments: install tools/packages.config -ExcludeVersion -OutputDirectory tools
   - ${{ if ne(parameters.MonoVersion, '') }}:
     - script: |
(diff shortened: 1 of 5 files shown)


filipw:
that's great news! because last time I checked out 6.12 preview it still had msbuild 16.6

filipw (approved):

PR #1986, opened 2020-10-16
Add async completion support

333fred:
The recent completion rewrite makes completion faster for a large number of cases, but it has a couple of issues:

* Any completion provider that needs to make edits beyond the few we've special cased doesn't work correctly. Roslyn is looking to add more of these, such as https://github.com/dotnet/roslyn/pull/47511.
* Completion providers that we do special case can be _slow_. Override and partial method completion in big types is painful, taking over a minute to show up sometimes.

To resolve this, I've added support for async edits by adding a new completion end point: /completion/afterInsert. It takes the item that was inserted, the file, and the new cursor position. It relies on the InsertText that was inserted still allowing for resolving the completion in the same way as before, so I had to force a few providers that don't behave well here to be eagerly resolved, regardless. I also kept import completion using the standard /completion/resolve step to resolve extra edits, as they don't need to modify the current cursor location at all.


333fred:
@filipw @JoeRobich @david-driscoll I've opened this as a draft because I'm not 100% sure we actually want to take this. It'll require implementation work in every client as LSP does not support this. LSP can use the `command` field in a completionitem to trigger a custom command, but every editor client extension will have to know how to handle that. It certainly is _much_ faster for many things though.

Pictures speak 1000 words, and animated gifs even more so.  
Sync completion (the experience today):
sync-completion
Async completion (the experience with this PR):
async-completion

david-driscoll:
Completion is getting better resolve support in 3.16.
https://microsoft.github.io/language-server-protocol/specifications/specification-3-16/#textDocument_completion

333fred:
@david-driscoll better, but not good enough for our purposes. Unless they're planning on changing it, vscode only supports additionalTextEdits and detal/documentation fields for lazy resolve, not the main textEdit.

333fred:
Because, to be clear, I do not like this code. It's a race condition waiting to happen. But the vscode lsp implementation does report `textEdit` as a resolvable property: https://github.com/microsoft/vscode-languageserver-node/blob/248560e74c9a715be59e62fdbdd554b87ddf8dba/client/src/common/client.ts#L1506-L1508.

david-driscoll:
command should be sufficient, however I don't know if there is a good logical flag for if an client supports it or not.

david-driscoll:
vscode: supports it [via native commands](https://github.com/microsoft/vscode-languageserver-node/blob/248560e74c9a715be59e62fdbdd554b87ddf8dba/client/src/common/client.ts#L2478)
emacs: lsp-mode supports it [here](https://github.com/emacs-lsp/lsp-mode/blob/7c6ace2d9e4ca4e7230c5989ae47991c7db61c7f/lsp-completion.el#L540-L541)
eclipse: lsp4e supports it [here](https://git.eclipse.org/r/plugins/gitiles/lsp4e/lsp4e/+/refs/heads/master/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/completion/LSIncompleteCompletionProposal.java#558)
vim: I can't tell if vim-lsp supports it, my assumption is it does not.

cc @nickspoons has vim knowledge

nickspoons:
There are many LSP clients for vim:

 - [prabirshrestha/vim-lsp](https://github.com/prabirshrestha/vim-lsp)
 - [neoclide/coc.nvim](https://github.com/neoclide/coc.nvim)
 - [autozimu/LanguageClient-neovim](https://github.com/autozimu/LanguageClient-neovim)
 - [dense-analysis/ale](https://github.com/dense-analysis/ale)
 - upcoming native LSP in neovim
 - various others I haven't looked at

I flicked through a few of them and can't find any reference to this `command` parameter in any. But I don't know these code bases well at all.

As for OmniSharp-vim, we're not an LSP client so we can adapt 😁

filipw:
I think since this is based on a request flag, there is absolutely no problem in adding this and thus giving more advanced clients an option to use a superior approach for extra edits

333fred:
> I think since this is based on a request flag, there is absolutely no problem in adding this and thus giving more advanced (more O# protocol aware) clients an option to use a superior approach for extra edits

Note, I think I can (and probably should) make this a startup flag for omnisharp instead, so LSP clients that support the `command` parameter could map it appropriately.

ffMathy:
I love this.

ffMathy:
In fact, I think this is a must if Omnisharp and VSCode needs any chance to be feasible instead of Visual Studio.

ffMathy:
@333fred what are the next steps for this? Can it be marked as a real PR instead of a draft?

I think this is super important.

333fred:
@ffMathy I'm really not happy with this code and think it's very likely to result in a subpar experience, which is why I haven't moved forward with it yet.

JoeRobich:
@333fred I hate how the gulp command swallows errors. You can get the error locally by running `tsc -p .`.

```console
~/Source/omnisharp-vscode [auto-doc-gen +0 ~2 -0 !]> tsc -p .
test/integrationTests/documentationCommentAutoFormatting.integration.test.ts:6:1 - error TS6133: 'skip' is declared but its value is never read.

6 import { skip } from 'rxjs/operators';
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


Found 1 error.
```

ffMathy:
What can be done to help though? The current experience is insanely slow (as seen in the first GIF).

333fred:
> What can be done to help though? The current experience is insanely slow (as seen in the first GIF).

Frankly, I'm not certain. There's an inherent race condition with this PR and we really can't do anything to solve it.

ffMathy:
Hmm, but how does the full Visual Studio do it then? It manages to be much faster in terms of completions and suggestions.

I think it's crucial to get to the same speed for larger projects. Is this still the plan abd/or vision?

ffMathy:
Similarly one could ask, how does TypeScript do it? It also remains super fast for very large projects.

Could we get inspired by it?

333fred:
> Hmm, but how does the full Visual Studio do it then?

VS has native support for this operation. VS Code does not have support for this type of asynchronous completion, and thus this remains a hack around functionality that VS Code does not have.

> Similarly one could ask, how does TypeScript do it? It also remains super fast for very large projects.

Typescript doesn't do this type of operation, so they don't have to be fast for it :)

ffMathy:
Thank you for answering my questions so far.

Is there an issue in the VS Code repo for supporting this type of async operation?

333fred:
I've talked with @jrieken about it before, I'm not sure if there is any specific issue open about making `TextEdit` asynchronously-resolvable.

ffMathy:
I made an issue (#112527) for it now.

jrieken:
> I made an issue microsoft/vscode#112527 for it now.

I have closed the issue as out-of-scope and this is why: accepting a completion is a synchronous operation and no further blocking or waiting can happen. You have to assume that users are typing and that typing has accepted a completion. Needing to asynchronously resolve the text edit would mean that we (a) block typing or (b) employ some kind of OT strategy which inserts the text edit when resolving. Option (a) is not possible because you don't block in JS and (b) is complex to implement and yields in a subpar experience, e.g when I accept a completion I wanna see it now, not after a delay.

333fred:
> (b)... yields in a subpar experience, e.g when I accept a completion I wanna see it now, not after a delay.

FWIW, and I've brought this up before (though not in talking with you Johannes) I fundamentally disagree with this assertion. The experience _today_ is subpar because bringing up the completion list takes 10+ seconds in these complex cases due to the nature of how we need to calculate the changes up front for every single completion in the list. I'd much rather have a 100 ms delay after insertion, rather than a 10+ second delay on bringing up the list.

ffMathy:
@333fred indeed. The current experience is simply unusable. Not just for the average user - for everyone.

Every C# developer I know (which is a lot), wants to use VS Code, but they always back out because Omnisharp is slow. This is across all kinds of solutions that are just a little bigger than the average Hello World example (multiple projects, dependencies, etc).

The current project I have is only 7000 lines of code. Visual Studio can produce suggestions instantly. In VS Code, it takes me 10+ seconds(!!!) for the completions to arrive. But completions is not the only thing. Also errors and warnings pop up very late.

@jrieken I agree blocking is probably not the option. That's why many people love VS Code in the first place in favor of Visual Studio for every other language than C#. However, whatever solution we end up with, it can't possibly be worse than the current state of things, and in that, I agree with @333fred.

ffMathy:
I think it's as simple as asking "does anyone want 10+ seconds of wait time for sguiggly error lines and/or warning lines to disappear, and the same wait time for waiting for completions and code-fixes?"

The answer would clearly be no. What we have now is essentially Notepad with syntax highlighting, because of that wait time.

And don't get me wrong. Omnisharp and VS Code are great. Just slow. The feature set is amazing and an excellent piece of work. I just think it's a waste when it doesn't get the opportunity to be used on a larger scale.

It's almost like an API that presents an amazing feature, but doesn't have documentation for it.

jrieken:
This seems to be turning into a random rant. I am out of here

ffMathy:
@jrieken it is not a rant. I am providing observations and arguments for discussion.

What is it that makes you think it is a rant, and not a civil discussion between engineers?

What part of my arguments sound offensive? I'm really sorry if that's your impression of anything I have said. I literally just praised VS Code too 🤷‍♂️

ffMathy:
And whether or not you see it as a rant, it's still an important issue to discuss. I feel backing out of it would be backing out of any form of responsibility over the project.

jrieken:
> But completions is not the only thing. Also errors and warnings pop up after the same duration.

This is where it turned into a rant. Diagnostics use a push model, it's up to omnisharp to push diagnostics into VS Code. If that takes 10+ sec than that's your problem. Apart from the fact that diagnostics has nothing to with this discussion

tverboon:
@ffMathy I think you are being too negative.

>  What we have now is essentially Notepad with syntax highlighting, because of that wait time.
> And don't get me wrong. Omnisharp and VS Code are great. Just slow.

I have been working with VS Code and Omnisharp for 4 years. Some operations are slow like this one, but it's not slow overall. I have seen bugs and improvements and regressions over this period. That said, I will never boot Windows again and work with Visual Studio. I really don't like that environment and I am much more productive using VS Code and Omnisharp.

@jrieken has been focussing on performance for VS Code for a very long time. Just saying it's slow and unusable, is a bit of an exaggeration and honestly not really fair. VS Code feels much snappier in a lot of use cases.

ffMathy:
> > But completions is not the only thing. Also errors and warnings pop up after the same duration.
> 
> This is where it turned into a rant. Diagnostics use a push model, it's up to omnisharp to push diagnostics into VS Code. If that takes 10+ sec than that's your problem. Apart from the fact that diagnostics has nothing to with this discussion

Well @jrieken forgive my ignorance then, but I don't know how VS Code's internal code works. I am experiencing that wait time, but I did not know that it was "my own problem" - how could I? I thought they were connected because of similar durations.

Thank you for the insight! :pray:

ffMathy:
> @ffMathy I think you are being too negative.
> 
> > What we have now is essentially Notepad with syntax highlighting, because of that wait time.
> > And don't get me wrong. Omnisharp and VS Code are great. Just slow.
> 
> I have been working with VS Code and Omnisharp for 4 years. Some operations are slow like this one, but it's not slow overall. I have seen bugs and improvements and regressions over this period. That said, I will never boot Windows again and work with Visual Studio. I really don't like that environment and I am much more productive using VS Code and Omnisharp.
> 
> @jrieken has been focussing on performance for VS Code for a very long time. Just saying it's slow and unusable, is a bit of an exaggeration and honestly not really fair. VS Code feels much snappier in a lot of use cases.

@tverboon ah yes. By saying "Omnisharp is slow" I meant completions alone, because that was what we were talking about, and because completions is such an essential part of any form of IDE. That is indeed generalizing too much, and I am sorry about that. Omnisharp is VERY responsive about anything else than completions (and warnings/errors for me only apparently) :pray: ANd that's why I (and many other developers) want to move to VS Code in the first place.

As a software developer myself, I understand how it can be a big blow to receive negative feedback about a subject which has been greatly improved over the past years. That's also why I mentioned that I love VS Code, and Omnisharp, and its feature set.

However, objectively from a user's perspective, "faster" is not enough. It (completion suggestions) needs to be "fast enough", and it currently isn't. And that's what is *stopping* many developers (who I have talked to) from making the jump away from VS into VS Code.

I hope that clears things up. It was honestly never my intent to rant about anything, although I understand how it may have been perceived as negative ❤️

jrieken:
>  By saying "Omnisharp is slow" I meant completions alone, because that was what we were talking about, and because completions is such an essential part of any form of IDE.

Which raises the question why are completions show. Sure, it's easy to blame the API but then I say that TS and Java, which use the same API, are fast

ffMathy:
> > By saying "Omnisharp is slow" I meant completions alone, because that was what we were talking about, and because completions is such an essential part of any form of IDE.
> 
> Which raises the question why are completions show. Sure, it's easy to blame the API but then I say that TS and Java, which use the same API, are fast

Yes, indeed. I was not trying to play a blame game here. I was simply trying to provide more insight.

I must have failed at that, and I am truly sorry.

However, I don't think anyone are really pointing fingers here. Yes, the experience is slow. That's all *I* know, because I'm just the "user". I don't know where the "blame" is, and I don't really care. 

I originally just wanted to show the support for @333fred's PR, which I think solves one of the biggest issues of VSCode/Omnisharp/Roslyn (or where ever the "problem" is) right now.

And from what I can gather, @333fred isn't blaming anyone either. From what I understand with my limited experience within this repo, he argues that the he isn't proud of the current proposed solution, and that to solve it, VS Code needs a new set of features in its API. There is no implicit "blame" here.

Sure, it's a valid argument to say that Java and TS don't have the same problem (which I actually also made further above in the comments), but like @333fred told me, those are also different ecosystems with different requirements.

Either way, we have a problem here in the C# context. That may not be VS Code's "problem", and perhaps not even Omnisharp or Roslyn's problem, but it is certainly a problem for the users either way, despite whatever KPI is present or who is to "blame" 🤷‍♂️

What I do hear you saying @jrieken is that it's not worth the effort though? On behalf of myself and many other C# developers who are looking for faster completion times, that's a bummer. But it's definitely an okay response. I don't know the kind of effort that's required, so I respect that decision.

jrieken:
> What I do hear you saying @jrieken is that it's not worth the effort though?

No, what I say is that we considered async text edits for completions and discarded the idea as not feasible. Now it's on the extensions to improve. Tho, having said that we DO have additional text edits and a best effort approach to apply them async, after making the main edit. See this [doc](https://github.com/microsoft/vscode/blob/f211a2bbe9685e0468000bfe34e87390c7eced22/src%2Fvs%2Fvscode.d.ts#L4081)

333fred:
> Tho, having said that we DO have additional text edits and a best effort approach to apply them async, after making the main edit.

Yes, this is how import completion works in omnisharp today (and in TS, I believe). The issue with override completion is that the expensive part is calculating the minimum text necessary for the argument types and what the auto-inserted base call should look like. These edits intersect the current cursor location, and indeed where the cursor should finally end up is based on this calculation. This means that additionalTextEdits is not suitable, as we cannot control the cursor location or intersect the cursor location.

There's other types of completion that become impossible or very hard to do well without asynchronicity. I know that rust-analyzer has brought this up in the past as well, where they want to have a `match` completion that rewrites the current expression and returns a snippet, but they, like us, don't want to do the expensive part of that calculation unless the user actually wants the completion. Roslyn is also starting to add these types of completions, and in order to keep general completions responsive in O#, I'm going to have to block them. Either that, or try to work around vscode's limitations and provide a significantly worse experience than vscode could do natively, as I fundamentally cannot resolve this race condition in an extension.

ffMathy:
@jrieken does @333fred's above elaboration make the idea more "feasible"? :pray:

src/OmniSharp.Abstractions/Models/v1/Completion/CompletionItem.cs
```diff
@@ -87,7 +87,12 @@ public class CompletionItem
         /// <summary>
         /// Index in the completions list that this completion occurred.
         /// </summary>
-        public int Data { get; set; }
+        public (long CacheId, int Index) Data { get; set; }
```
filipw, line 90: does this work well serialized over the wire? I could imagine some client implementations might have troubles

src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionListBuilder_Sync.cs
```diff
@@ -0,0 +1,248 @@
...
+                        }
+                    }
+                    else
+                    {
+                        filterText = labelText == completion.FilterText ? null : completion.FilterText;
+                    }
+                }
+            }
+
+            static (string?, InsertTextFormat) getPossiblySnippitizedInsertText(TextChange change, int? adjustedNewPosition)
```
filipw, line 223: I am not sure this is a real word, but if it were, it would likely be `snippetized` 🙂

src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionListCache.cs
```diff
@@ -0,0 +1,97 @@
...
+    /// <summary>
+    /// Caches completion lists in between calls to CompletionHandler and
+    /// CompletionResolveHandler. Used to avoid unnecessary recomputation.
+    /// </summary>
+    internal class CompletionListCache
+    {
+        /// <summary>
+        /// Maximum number of completion lists allowed in cache. Must be >= 1.
+        /// </summary>
+        private const int MaxCacheSize = 2;
```
filipw, line 22: we used to have a single statically cached item, I am not sure I can follow why is there now two? 
is this "just in case"?

src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionListBuilder_Async.cs
```diff
@@ -0,0 +1,114 @@
...
+            CSharpCompletionList completions,
+            TextSpan typedSpan,
+            bool expectingImportedItems, bool isSuggestionMode)
+        {
+            var completionsBuilder = new List<CompletionItem>(completions.Items.Length);
+            var seenUnimportedCompletions = false;
+            var commitCharacterRuleCache = new Dictionary<ImmutableArray<CharacterSetModificationRule>, IReadOnlyList<char>>();
+            var commitCharacterRuleBuilder = new HashSet<char>();
+            var isOverrideOrPartialCompletion = completions.Items.Length > 0
+                && completions.Items[0].GetProviderName() is CompletionItemExtensions.OverrideCompletionProvider or CompletionItemExtensions.PartialMethodCompletionProvider;
```
filipw, line 36: is it enough to just check index 0 here?

src/OmniSharp.Abstractions/IsExternalInit.cs
```diff
@@ -0,0 +1,4 @@
+﻿namespace System.Runtime.CompilerServices
+{
+    internal sealed class IsExternalInit { }
```
filipw, line 3: I assume this is so that records compile? maybe it should also have `[EditorBrowsable(EditorBrowsableState.Never)]`

src/OmniSharp.Shared/Options/RoslynExtensionsOptions.cs
```diff
@@ -10,6 +10,7 @@ public class RoslynExtensionsOptions : OmniSharpExtensionsOptions
         public bool EnableDecompilationSupport { get; set; }
         public bool EnableAnalyzersSupport { get; set; }
         public bool EnableImportCompletion { get; set; }
+        public bool EnableAsyncCompletion { get; set; }
```
filipw, line 13: feature flag 👍

src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionListBuilder_Async.cs
333fred: Yes. If this is a partial or override completion session, these will be the only types of completions shown.

src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionListCache.cs
333fred: We now have an after insert step, and the user could technically have triggered a completion session in between. I want to make sure that we handle it somewhat gracefully.

src/OmniSharp.Abstractions/Models/v1/Completion/CompletionItem.cs
333fred: I would expect it should be just fine, it should be serialized as `{ 'Item1': 1, 'Item2': 2 }`.

src/OmniSharp.Abstractions/IsExternalInit.cs
333fred: Yeah, it's for records. I can certainly add that attribute.

src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionService.cs
```diff
@@ -498,16 +207,60 @@ public async Task<CompletionResolveResponse> Handle(CompletionResolveRequest req
...
+            var lastCompletionItem = completions.Items[index];
+            if (lastCompletionItem.DisplayTextPrefix + lastCompletionItem.DisplayText + lastCompletionItem.DisplayTextSuffix != request.Item.Label)
+            {
+                _logger.LogError("Inconsistent completion data. Requested data on {0}, but found completion item {1}", request.Item.Label, lastCompletionItem.DisplayText);
+                return new CompletionAfterInsertResponse();
+            }
+
+            if (lastCompletionItem.GetProviderName() is not (CompletionItemExtensions.OverrideCompletionProvider or
+                                                             CompletionItemExtensions.PartialMethodCompletionProvider)
+                                                        and var name)
```
filipw, line 240: oh I didn't know you could do something like this 🧐

src/OmniSharp.Abstractions/Models/v1/Completion/CompletionItem.cs
filipw: I see. one other thing that jumped to my mind is that it is technically a breaking change on the contract.

It would be better to pass the CacheId as a separate property, this way it would be an additive change and it wouldn't affect the existing completion service implementations

333fred: I don't know if I agree. This data is only used by the server, not by clients: the vscode implementation just treats this as an `any` type: https://github.com/OmniSharp/omnisharp-vscode/blob/master/src/omnisharp/protocol.ts#L541. Ideally, it would be a totally opaque blob, but that isn't how json works.

filipw: yes I understand that in the C# extension it is treated this way (and that it was intended to flow back to the server as sort of a "state"), but it is part of a contract after all. we don't follow all the implementations of Omnisharp (we are not even aware of most!) but we try not to have breaking contract changes

filipw: actually since this is new anyway I am pretty sure *you* are the only user of it so I guess it doesn't matter 😀

src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionService.cs
333fred: Composition!

333fred:
@filipw any more comments?

filipw (approved):
for me this is good to go - thanks!

333fred:
@filipw what's the next step?

filipw:
it failed the last test run, I've enabled auto merge now

ffMathy:
@333fred congrats on getting this out, and thank you for making it happen! :heart_eyes:

When can this be enabled, and how?

333fred:
> @333fred congrats on getting this out, and thank you for making it happen! :heart_eyes:
> 
> When can this be enabled, and how?

When the vscode side is merged, and using the setting.

PR #1998, opened 2020-11-02
Upgraded lsp to 0.18.3

david-driscoll:
No breaking changes (other than the analyzer that is getting chained from NerdBank).


filipw (approved):
looks good

PR #2002, opened 2020-11-04
fix lsp mode logging

yatli:
I was looking forward to a working lsp mode but when I come back to check this, even the logging functionality is toast...
`¯\_(ツ)_/¯`

### DIFF
--- a/src/OmniSharp.LanguageServerProtocol/LanguageServerHost.cs
+++ b/src/OmniSharp.LanguageServerProtocol/LanguageServerHost.cs
@@ -56,6 +56,8 @@ namespace OmniSharp.LanguageServerProtocol
             _options = new LanguageServerOptions()
                 .WithInput(input)
                 .WithOutput(output)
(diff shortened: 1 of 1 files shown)


src/OmniSharp.LanguageServerProtocol/LanguageServerHost.cs
```diff
@@ -44,25 +44,24 @@ public class LanguageServerHost : IDisposable
...
             Stream output,
             CommandLineApplication application,
-            CancellationTokenSource cancellationTokenSource)
+            CancellationTokenSource cancellationTokenSource,
+            Action<ILoggingBuilder> configureLogging = null)
         {
             _options = new LanguageServerOptions()
                 .WithInput(input)
                 .WithOutput(output)
-                .ConfigureLogging(x => x
```
yatli, line 57: hmm maybe this is still needed.

yatli: done

PR #2008, opened 2020-11-08
Response file can now used enviroment variables in the path + more error handling. 

ByronMayne:
For custom users scripting I don't want users to have to know where the response file is so I have it defined as an environment variable. This change adds support for that. I also went ahead and added some more error handling because if the .rsp file had invalid members that errors where just thrown into the void. 

```csharp
{
  "script": {
     "rspFilePath": "%PLUGIN_DIRECTORY%\\custom.rsp"
  }
}
```

I was also looking into trying to unit test the parsing of the .rsp file but it's quite difficult at the moment. I was looking at adding checks to make sure we are reading lib paths and source paths however with the current way things are laid out we don't have any accesses to the members without doing dirty reflection checking in tests which is not ideal.


filipw (approved):
this looks good, thank you

PR #2009, opened 2020-11-08
Added github actions runners

david-driscoll:


build.cake
```diff
@@ -478,22 +481,37 @@ Task("PrepareTestAssets")
...
+        try {
+            DotNetCoreBuild(folder, new DotNetCoreBuildSettings()
+            {
+                ToolPath = env.DotNetCommand,
+                WorkingDirectory = folder,
+                Verbosity = DotNetCoreVerbosity.Minimal
+            });
+        } catch {
+            // ExternalAlias has issues once in a while, try building again to get it working.
+            if (project == "ExternAlias") {
```
filipw, line 500: ha ha 🙈

.config/dotnet-tools.json
```diff
@@ -0,0 +1,12 @@
+{
+  "version": 1,
+  "isRoot": true,
+  "tools": {
```
filipw, line 4: could  you update git-version  to the dotnet tool variant too?  the current one we use doesn't work on  Unix

build.cake
```diff
@@ -478,22 +481,37 @@ Task("PrepareTestAssets")
...
+                    WorkingDirectory = folder,
+                    Verbosity = DotNetCoreVerbosity.Minimal
+                });
+            }
+        }
     });
 
 Task("PrepareTestAssets:RestoreOnlyTestAssets")
     .IsDependeeOf("PrepareTestAssets")
+    .WithCriteria(testProjects.Any(z => nonCakeTestProjects.Any(x => x == z)))
```
filipw, line 514: what does this do?

david-driscoll: This was part of a matrix text, but basically if you're running the cake test project it only needs the cake assets, so it makes the run a little faster.  

It could be removed (but doesn't cause any harm).

filipw:
ooops I now created conflicts on this one by merging #2010  sorry 😢

david-driscoll:
you're fired!  I'll get to back to this, this week.

filipw:
@david-driscoll is there any hold up? can we get rid of ADO? 😄

david-driscoll:
Fixed conflicts, next step will be to update the release process to no longer require ado, however that might take some help on the microsoft side as I'll need the secret for uploading to azure storage.

david-driscoll:
I'm going to push this through, and make a follow up.

PR #2023, opened 2020-11-23
Add documentation comment creation to the FormatAfterKeystrokeService

333fred:
IDocumentationCommentSnippetService was recently moved down to the Features layer, which means omnisharp can now take advantage of it and generate documentation comment snippets on typing. There are a couple of caveats:

1. LSP's "/onTypingFormat" endpoint has no support for setting the cursor location after format, which means that the cursor ends up after the inserted block. This is fine for newlines inside doc comments, but awkward for the initial expansion. However, unless we want to add special support with a custom endpoint, we can't get around this (the custom endpoint is how Roslyn does it).
2. The service itself is still internal, so I had to add a few new reflection points.

Server-side of https://github.com/OmniSharp/omnisharp-vscode/issues/8. It also needs a vscode change, but it's a one-line change I'll submit later after we get this merged and I can write some integration tests.

### DIFF
--- a/src/OmniSharp.Roslyn.CSharp/DocumentationComments/DocumentationCommentSnippetService.cs
+++ b/src/OmniSharp.Roslyn.CSharp/DocumentationComments/DocumentationCommentSnippetService.cs
@@ -1,0 +1,110 @@
+﻿#nullable enable
+
+using System;
+using System.Reflection;
+using System.Threading;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Completion;
+using Microsoft.CodeAnalysis.Host;
+using Microsoft.CodeAnalysis.Options;
+using Microsoft.CodeAnalysis.Text;
+
+namespace OmniSharp.Roslyn.DocumentationComments
+{
+    /// <summary>
+    /// Proxy service for Microsoft.CodeAnalysis.DocumentationComments.IDocumentationCommentSnippetService.
+    /// Implementation was based on the service as of this commit: 2834b74995bb66a7cb19cb09069c17812819afdc
+    /// See: https://github.com/dotnet/roslyn/blob/2834b74995bb66a7cb19cb09069c17812819afdc/src/Features/Core/Portable/DocumentationComments/IDocumentationCommentSnippetService.cs
+    /// </summary>
+    public struct DocumentationCommentSnippetService
+    {
+        /// <summary>
+        /// IDocumentationCommentService HostLanguageServices.GetRequiredService<IDocumentationCommentService>()
+        /// </summary>
+        private static MethodInfo s_getRequiredService;
+        /// <summary>
+        /// DocumentationCommentSnippet IDocumentationCommentService.GetDocumentationCommentSnippetOnCharacterTyped(SyntaxTree, SourceText, int, DocumentOptionSet, CancellationToken)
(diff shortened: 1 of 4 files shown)


src/OmniSharp.Roslyn.CSharp/DocumentationComments/DocumentationCommentSnippetService.cs
```diff
@@ -0,0 +1,110 @@
...
+        private static PropertyInfo s_spanToReplace;
+        /// <summary>
+        /// string DocumentationCommentSnippet.SnippetText
+        /// </summary>
+        private static PropertyInfo s_snippetText;
+
+        static DocumentationCommentSnippetService()
+        {
+            var iDocumentationCommentSnippetServiceType = typeof(CompletionItem).Assembly.GetType("Microsoft.CodeAnalysis.DocumentationComments.IDocumentationCommentSnippetService");
+            s_getDocumentationCommentSnippetOnCharacterTyped = iDocumentationCommentSnippetServiceType.GetMethod(nameof(GetDocumentationCommentSnippetOnCharacterTyped));
```
filipw, line 45: I saw that there is also `GetDocumentationCommentSnippetOnCommandInvoke`, I think it is used in VS when you invoke "insert comment" command from insert snippet feature. we may also want to expose it in the future

filipw (approved):
thanks a lot!

filipw:
I'd like to add that the current popular solution to do this in VS Code is to use `k--kato.docomment` extension.
There have been several times already that we had to deal with an issue report in OmniSharp (unresponsive etc) that turned out to be caused by this extension.

JoeRobich (approved):

lonix1:
It seems that this feature requires us to set `"editor.formatOnType": true`.

That is unfortunate as many people (me included) don't like it - it ruins one's coding workflow, and besides there are many formatting tools that handle that sort of thing.

Is it possible for this feature to work without that setting?

(BTW: the `k--kato.docomment` extension doesn't need that setting in order to expand `///`.)

333fred:
> It seems that this feature requires us to set `"editor.formatOnType": true`.
> 
> That is unfortunate as many people (me included) don't like it - it ruins one's coding workflow, and besides there are many formatting tools that handle that sort of thing.
> 
> Is it possible for this feature to work without that setting?
> 
> (BTW: the `k--kato.docomment` extension doesn't need that setting in order to expand `///`.)

No, it is not. Not and be compatible with LSP. We need a new LSP feature to make it more generally available.

filipw:
also, `k--kato.docomment` runs the check "should I expand comment" on every text change in the editor which is very aggressive

lonix1:
Thank you for explaining. I didn't realize that plugin was that aggressive!

> We need a new LSP feature

In vscode or omnisharp? And is there already an issue tracking this, or do you want me to open one?

333fred:
> In vscode or omnisharp?

Neither, in LSP. And I imagine there's an open proposal somewhere for an LSP endpoint that could accomplish this, but I haven't really looked.

lonix1:
@333fred Well then it's obviously beyond my understanding, so I wouldn't even know how to write the issue. :smile:

I hope you'll find some time to investigate. Until then thanks for this PR, it's a decent approach for now.

PR #2030, opened 2020-11-27
Gracefully handle GitVersion failure, and set a default version.

mholo65:


scripts/common.cake
```diff
@@ -293,7 +293,22 @@ public class BuildEnvironment
         }
         else
         {
-            return context.GitVersion();
+            try
+            {
+                return context.GitVersion();
+            }
+            catch
```
filipw, line 300: Can we have this catch block only kick in when not running on CI?

mholo65: Sure, but currently when on CI, we run GitVersion as separate job, and get GitVersion stuff as env vars. So CI will never run GitVersion in this context.

filipw: ah right

filipw (approved):

Issue #1995, opened 2020-11-01
Interop to omnisharp protocol via LSP is not working properly


razzmatazz:
This is probably more of a question/note to @david-driscoll . I am trying to bring back unit test functionality to emacs/lsp-mode of which one attempt was #1965 , but then I realized you have already aded interop functionality added with 26e08ceece10e89c06f82b40bc16c2fbbb148701 (Upgrade to the latest lsp library). So I am trying to use that.

But then when trying to invoke, e.g. `o#/checkalivestatus` and I get this response. The same (or similar error) is raised for other handlers as well. Is this something to do with the server implemention (a bug) or rather an incorrect invocation of the handler? Do you have some code (in vscode or elsewhere) that uses this interop that I could refer to?

The error raised on invocation follows:

```
lsp-request: Internal Error - System.InvalidOperationException: Error constructing handler for request of type MediatR.IRequestHandler‘2[OmniSharp.Extensions.JsonRpc.DelegatingRequest‘1[Newtonsoft.Json.Linq.JToken],MediatR.Unit]. Register your handlers with the container. See the samples in GitHub for examples. ---> DryIoc.ContainerException: code: Error.NoConversionOperatorFoundWhenInterpretingTheConvertExpression;
message: There is no explicit or implicit conversion operator found when interpreting OmniSharp.Extensions.JsonRpc.DelegatingJsonRequestHandler to MediatR.IRequestHandler<OmniSharp.Extensions.JsonRpc.DelegatingRequest<Newtonsoft.Json.Linq.JToken>, MediatR.Unit> in expression: Convert(value(OmniSharp.Extensions.JsonRpc.DelegatingJsonRequestHandler), IRequestHandler‘2)
  at DryIoc.Throw.It (System.Int32 error, System.Object arg0, System.Object arg1, System.Object arg2, System.Object arg3) [0x00012] in <fba9e575570149efaee3484292339424>:0
  at DryIoc.Converter.ConvertWithOperator (System.Object source, System.Type targetType, FastExpressionCompiler.LightExpression.Expression expr) [0x0003a] in <fba9e575570149efaee3484292339424>:0
[...]

PR #2044, opened 2020-12-16
Updated to the latest beta of lsp.  Spec version 3.16

david-driscoll:
The latest version of the spec just dropped, this beta has all the latest spec features, plus a few minor breaking changes, documented with the release [here](https://github.com/OmniSharp/csharp-language-server-protocol/releases/tag/v0.19.0-beta.4)


david-driscoll:
NOTE: I'll probably remove this from beta in about 2 weeks, everything should be good, unless there is feedback that folks are unhappy with the minor breaking changes.

build/Settings.props
```diff
@@ -2,7 +2,7 @@
 <Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
   <PropertyGroup>
-    <LangVersion>8.0</LangVersion>
+    <LangVersion>9.0</LangVersion>
```
david-driscoll, line 5: @filipw @JoeRobich anyone have an issue with this change?

JoeRobich: I don't have a problem with it.

david-driscoll: good because the latest library has lots of records... lots of with statements 🤣

filipw: one small issue with this is that some features of C# 9 e.g. positional records or `init` are not available without .NET 5.0, so it is easy to fall into the trap of using something that won't compile 😄

filipw (approved):
looks good to me

build/Settings.props
david-driscoll: records have been working fine for me, the only gotcha is much like NRT you have to provide an interop type.  
https://github.com/OmniSharp/csharp-language-server-protocol/blob/81a38cca80a4e35ec198de75dbab5a5571aae82a/src/JsonRpc/Nullable/Records.cs

razzmatazz:
I was trying to figure out the cause for #1995 with the latest changes checked out from this PR but still failed -- the issue contains updated stack trace.

Sorry for trying to shamelessly plug in another issue, but hopefully this is looks to be related/easy to fix?

This would allow me to implement other features which are now available only via native omnisharp protocol, like assembly inspection/metadata/test runners, etc in `lsp-mode` on emacs and in other editors too.

razzmatazz:
> Sorry for trying to shamelessly plug in another issue, but hopefully this is looks to be related/easy to fix?
> 
> I was trying to figure out the cause for #1995 with the latest changes checked out from this PR but still failed -- the issue contains updated stack trace.
> 
> This would allow me to implement other features which are now available only via native omnisharp protocol, like assembly inspection/metadata/test runners, etc in `lsp-mode` on emacs and in other editors too.

I have created a PR to OS.E.LSP that fixes my issue:
 - https://github.com/OmniSharp/csharp-language-server-protocol/pull/488

So I guess my previous comment with a shameless plug can be ignored here.

JoeRobich (approved):

razzmatazz:
I can confirm this branch works works (for me, at least).

With [this patch in O.E.L](https://github.com/OmniSharp/csharp-language-server-protocol/commit/32a135922dcf794158588fdacd66213b102ea5c4) it enables me to use the legacy endpoints via LSP to enable test-runner, metadata browsing and other functionality in emacs, at least. This will also allow us to use the `codeaction/resolve` method from LSP 3.16 and drop the custom `omnisharp/executeCodeAction` method as per #2068.

Would it be feasible to merge this PR before the next release? This unlocks quite a few things LSP-wise.

src/OmniSharp.LanguageServerProtocol/Handlers/OmniSharpCodeActionHandler.cs
```diff
@@ -101,14 +95,14 @@ public override async Task<CommandOrCodeActionContainer> Handle(CodeActionParams
...
                         Command = Command.Create("omnisharp/executeCodeAction")
-                            .WithTitle(ca.Name)
                             .WithArguments(new CommandData()
                             {
                                 Uri = request.TextDocument.Uri,
                                 Identifier = ca.Identifier,
                                 Name = ca.Name,
                                 Range = request.Range,
                             })
+                            with { Title = ca.Name }
```
333fred, line 105: We can't also with the arguments?

src/OmniSharp.LanguageServerProtocol/Handlers/OmniSharpCompletionHandler.cs
```diff
@@ -11,7 +11,7 @@
 
 namespace OmniSharp.LanguageServerProtocol.Handlers
 {
-    class OmniSharpCompletionHandler : CompletionHandler
+    class OmniSharpCompletionHandler : CompletionHandlerBase
```
333fred, line 14: Man, this whole handler needs to be ported to the new one, doesn't it.

build/Settings.props
333fred: I've tried to use `or` patterns several times now, please give them to me :)

src/OmniSharp.LanguageServerProtocol/Handlers/OmniSharpCodeActionHandler.cs
david-driscoll: as it stands yes and no..., the arguments is a [JArray](https://github.com/OmniSharp/csharp-language-server-protocol/blob/e0c9fc66da91ab389874f34ff9a5b1b18d932be6/src/Protocol/Models/Command.cs#L27-L28) so you'd have to do `with { Arguments = JArray.FromObject(...) }`

333fred (approved):

src/OmniSharp.LanguageServerProtocol/Handlers/OmniSharpCompletionHandler.cs
filipw: yes

filipw (approved):
LGTM

src/OmniSharp.LanguageServerProtocol/Handlers/OmniSharpCompletionHandler.cs
david-driscoll: thankfully the changed merged in cleanly.

Issue #2048, opened 2020-12-22
Records break syntax highlighting


333fred:
```
System.Collections.Generic.KeyNotFoundException: The given key 'record name' was not present in the dictionary.\
  at System.Collections.Generic.Dictionary`2[TKey,TValue].get_Item (TKey key) [0x0001e] in <9f0df102fe6e4cfea29d2e46f585d8a5>:0 \
  at OmniSharp.Roslyn.CSharp.Services.SemanticHighlight.SemanticHighlightService.CreateSemanticSpan (System.Collections.Generic.IEnumerable`1[T] results, Microsoft.CodeAnalysis.Text.TextLineCollection lines) [0x000db] in <88f2819c12ee464e94d6700e849310cf>:0 \
  at OmniSharp.Roslyn.CSharp.Services.SemanticHighlight.SemanticHighlightService+<>c.<Handle>b__1_2 (System.Linq.IGrouping`2[TKey,TElement] grouping) [0x0000c] in <88f2819c12ee464e94d6700e849310cf>:0 \
  at System.Linq.Enumerable+SelectEnumerableIterator`2[TSource,TResult].ToArray () [0x0001d] in <9ba54d07696a449db4a4279fc05aa435>:0 \
  at System.Linq.Enumerable.ToArray[TSource] (System.Collections.Generic.IEnumerable`1[T] source) [0x0001f] in <9ba54d07696a449db4a4279fc05aa435>:0 \
  at OmniSharp.Roslyn.CSharp.Services.SemanticHighlight.SemanticHighlightService.Handle (OmniSharp.Models.SemanticHighlight.SemanticHighlightRequest request) [0x002fa] in <88f2819c12ee464e94d6700e849310cf>:0 \
  at OmniSharp.Endpoint.EndpointHandler`2[TRequest,TResponse].GetFirstNotEmptyResponseFromHandlers (OmniSharp.Endpoint.Exports.ExportHandler`2[TRequest,TResponse][] handlers, TRequest request) [0x00099] in <954ed0cf974c4f038a543c1ed1dfaa98>:0 \
  at OmniSharp.Endpoint.EndpointHandler`2[TRequest,TResponse].HandleRequestForLanguage (System.String language, TRequest request, OmniSharp.Protocol.RequestPacket packet) [0x00163] in <954ed0cf974c4f038a543c1ed1dfaa98>:0 \
[...]

PR #2049, opened 2020-12-22
Handle records in syntax highlighting

333fred:
Fixes https://github.com/OmniSharp/omnisharp-roslyn/issues/2048.

### DIFF
--- a/src/OmniSharp.Roslyn.CSharp/Services/SemanticHighlight/SemanticHighlightService.cs
+++ b/src/OmniSharp.Roslyn.CSharp/Services/SemanticHighlight/SemanticHighlightService.cs
@@ -107,6 +107,7 @@ namespace OmniSharp.Roslyn.CSharp.Services.SemanticHighlight
                 [ClassificationTypeNames.VerbatimStringLiteral] = SemanticHighlightClassification.VerbatimStringLiteral,
                 [ClassificationTypeNames.StringEscapeCharacter] = SemanticHighlightClassification.StringEscapeCharacter,
                 [ClassificationTypeNames.ClassName] = SemanticHighlightClassification.ClassName,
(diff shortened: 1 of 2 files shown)


333fred:
@JoeRobich for review.

src/OmniSharp.Roslyn.CSharp/Services/SemanticHighlight/SemanticHighlightService.cs
```diff
@@ -107,6 +107,7 @@ class ClassifiedResult
                 [ClassificationTypeNames.VerbatimStringLiteral] = SemanticHighlightClassification.VerbatimStringLiteral,
                 [ClassificationTypeNames.StringEscapeCharacter] = SemanticHighlightClassification.StringEscapeCharacter,
                 [ClassificationTypeNames.ClassName] = SemanticHighlightClassification.ClassName,
+                [ClassificationTypeNames.RecordName] = SemanticHighlightClassification.ClassName,
```
333fred, line 110: The other option would be to map a separate classification for record names and have the clients deal with it. Which is fine, just need to know what to do.

filipw: no, this is fine IMHO

filipw (approved):

src/OmniSharp.Roslyn.CSharp/Services/SemanticHighlight/SemanticHighlightService.cs
333fred: Well, the only concern is that we'll need to do something different when struct records come out, but that's a concern for next year :).

PR #2058, opened 2020-12-31
fix exception with null sdk version

corngood:
I've been getting errors like:

`{"Event":"log","Body":{"LogLevel":"ERROR","Name":"OmniSharp.MSBuild.ProjectLoadListener","Message":"Unexpected exception got thrown from project load listener: System.NullReferenceException: Object reference not set to an instance of an object\n  at OmniSharp.MSBuild.ProjectLoadListener.GetSdkVersion (OmniSharp.MSBuild.Notification.ProjectLoadedEventArgs args) [0x0000b] in <7e6cb753aca94549b693616e0e7d1016>:0 \n  at OmniSharp.MSBuild.ProjectLoadListener.ProjectLoaded (OmniSharp.MSBuild.Notification.ProjectLoadedEventArgs args) [0x0001a] in <7e6cb753aca94549b693616e0e7d1016>:0 "},"Seq":137,"Type":"event"}`

This happens when `dotnet` isn't available in PATH.  Omnisharp runs `dotnet --info` and if it fails, `SdkVersion` ends up null.

My change is a bit hacky.  Perhaps it would be better to propagate the null through to `ProjectConfigurationMessage.SdkVersion`?  That seems to be the only place it's used, but it might be a breaking protocol change.

### DIFF
--- a/src/OmniSharp.MSBuild/ProjectLoadListener.cs
+++ b/src/OmniSharp.MSBuild/ProjectLoadListener.cs
@@ -87,7 +87,7 @@ namespace OmniSharp.MSBuild
 
         private static HashedString GetSdkVersion(ProjectLoadedEventArgs args)
         {
-            return _tfmAndFileHashingAlgorithm.HashInput(args.SdkVersion.ToString());
+            return _tfmAndFileHashingAlgorithm.HashInput($"{args.SdkVersion}");
         }
 
         private static HashedString GetSessionId(ProjectLoadedEventArgs args)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/OmniSharp/omnisharp-roslyn?pullRequest=2058) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/OmniSharp/omnisharp-roslyn?pullRequest=2058) before we can accept your contribution.<br/><br/>:x: corngood [sign now](https://cla.dotnetfoundation.org/OmniSharp/omnisharp-roslyn?pullRequest=2058)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/OmniSharp/omnisharp-roslyn?pullRequest=2058) it.</sub>

src/OmniSharp.MSBuild/ProjectLoadListener.cs
```diff
@@ -87,7 +87,7 @@ private static HashedString GetProjectId(ProjectLoadedEventArgs args)
 
         private static HashedString GetSdkVersion(ProjectLoadedEventArgs args)
         {
-            return _tfmAndFileHashingAlgorithm.HashInput(args.SdkVersion.ToString());
+            return _tfmAndFileHashingAlgorithm.HashInput($"{args.SdkVersion}");
```
filipw, line 90: we could just return a `null` I think
however, a better question is if we should allow to start without `dotnet` on the path at all.
There would be other things that will fail without it e.g. auto restore or test adapter

PR #2061, opened 2021-01-07
use "variable" kind for parameter completion

filipw:
We used to do it this way until the switch to CompletionService. This PR restores the old mapping of parameter to variable.

Fixes https://github.com/OmniSharp/omnisharp-roslyn/issues/2060
Fixes https://github.com/OmniSharp/omnisharp-vscode/issues/4314

### DIFF
--- a/src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionService.cs
+++ b/src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionService.cs
@@ -63,7 +63,7 @@ namespace OmniSharp.Roslyn.CSharp.Services.Completion
             { WellKnownTags.Method, CompletionItemKind.Method },
             { WellKnownTags.Module, CompletionItemKind.Module },
             { WellKnownTags.Operator, CompletionItemKind.Operator },
-            { WellKnownTags.Parameter, CompletionItemKind.Value },
+            { WellKnownTags.Parameter, CompletionItemKind.Variable },
             { WellKnownTags.Property, CompletionItemKind.Property },
             { WellKnownTags.RangeVariable, CompletionItemKind.Variable },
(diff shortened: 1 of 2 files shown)


filipw:
cc @333fred

333fred:
Interesting. This is copied directly from roslyn, https://github.com/dotnet/roslyn/blob/master/src/Features/LanguageServer/Protocol/Extensions/ProtocolConversions.cs#L57. @dibarbet, is Roslyn intentionally using value for parameters there, or is it an oversight?

filipw:
yes I saw this too and got confused. but then tested in VS directly and saw there that locals and parameters use same glyph. is it remapped elsewhere?

333fred:
> yes I saw this too and got confused. but then tested in VS directly and saw there that locals and parameters use same glyph. is it remapped elsewhere?

That map is only used by the LSP, so you won't see it unless you're using codespaces.

mholo65 (approved):

dibarbet:
> @dibarbet, is Roslyn intentionally using value for parameters there, or is it an oversight?

good question, it very well could be an oversight, or VS just uses the same icon for both, @allisonchou do you happen to remember?

JoeRobich (approved):

allisonchou:
>good question, it very well could be an oversight, or VS just uses the same icon for both, @allisonchou do you happen to remember?

I'm not sure. 😦 It may have been an oversight. Looking at the file history, it looks like this mapping has been here since LSP was first tested out in Roslyn.

333fred (approved):

333fred:
Filed https://github.com/dotnet/roslyn/issues/50324 on Roslyn to investigate whether we're using the wrong value as well.

filipw:
thanks!

PR #2066, opened 2021-01-20
changed the default RunFixAllRequest timeout to 10 seconds

filipw:
This is now equal to the go to metadata timeout which is also 10 seconds. Three seconds is too short especially for solution wide fixes.

That said, it would be good for editors to have their own setting.


JoeRobich (approved):

tverboon:
Not sure if commenting here is the right way to do this, but is there a reason not increase this timeout after 30 sec? 10 seconds is an improvement, but even 10 seconds are passing quick sometimes with a medium sized project. I know, eventually 30 seconds won't be enough at some point.

My guess is that users are more willing to wait for this FixAll operation, than go to metadata. My alternative is doing it by hand and that will take a lot longer than 30 secs 😄

filipw:
The idea was that this is only a reasonable default, and that each editor would expose a configurable value on its own (which VS Code extension does not do at the moment).

Issue #2029, opened 2020-11-26
Catch "The reference assemblies (...) were not found." project loader exception and provide better information


filipw:
Currently when ` The reference assemblies for .NETFramework,Version=v4.7.1 were not found.` is thrown (typically when trying to open desktop .NET project when running on built-in Mono), it is accompanied by a link to download the reference assemblies for Windows only (the default MSBuild error).

A lot of users, especially Unity users, are overwhelmed by this and do not know how to proceed.

We should catch that exception and provide more relevant entry in the log e.g. detecting if user is on *nix, describing they should install global Mono and so on. This could allow users to self-diagnose and fix the problem.

PR #2069, opened 2021-01-22
use an OmniSharp specific message for MSB3644

filipw:
fixes https://github.com/OmniSharp/omnisharp-roslyn/issues/2029 

Instead of printing the default MSBuild message for MSB3644 which is Windows specific and has been very confusing to the users:

> The reference assemblies for {0} were not found. To resolve this, install the Developer Pack (SDK/Targeting Pack) for this framework version or retarget your application. You can download .NET Framework Developer Packs at https://aka.ms/msbuild/developerpacks

we now change it into a more relevant OmniSharp message on Unix:

> This project targets .NET version that requires reference assemblies that do not ship with OmniSharp out of the box (e.g. .NET Framework). The most common solution is to make sure Mono is installed on your machine (https://mono-project.com/download/) and that OmniSharp is started with that Mono installation (e.g. 'omnisharp.useGlobalMono':'always' in C# Extension for VS Code).


JoeRobich (approved):
Thanks for this change

PR #2072, opened 2021-01-25
LSP and Completion tweaks

333fred:
This is a couple of small tweaks to resolve LSP and completion behavior:

1. We don't resolve partial and override completion on ` ` after an `override`, which is a behavior gap from VS.
2. Making the add using code action `SourceOrganizeImports` triggers special behavior and filtering in vscode, which is undesirable. Instead, just categorize it as a quickfix.


333fred:
Incorporated into https://github.com/OmniSharp/omnisharp-roslyn/pull/2044.

PR #2074, opened 2021-01-26
Move LSP Completion to the new CompletionService

333fred:
This is a bit overdue, but as vscode will hopefully be moving to the LSP shortly, we can look at starting to convert its providers over to the LSP. As the new service was designed to be a mirror of LSP, this simplifies the implementation of the handler: just convert the request to the internal one, pass through, and convert the internal result back to LSP's version. As part of this, I ported all of the completion tests in the regular service to the LSP, with only a few small changes around host setup and where names/nesting differed a bit from the LSP.


filipw (approved):
very nice, thank you

333fred:
Hmm, getting cancellations I wasn't getting locally. Will have a look tonight.

filipw:
as discussed on Slack - this is good to go

PR #2078, opened 2021-01-31
Adds SyntaxTree Visualizer Endpoints

333fred:
This adds a new set of endpoints designed for implementing a syntax tree visualizer. There is an accompanying set of commits using these endpoints in vscode to display syntax information for the currently-active file. This is a stateful API that requires calls in the correct order. A session can be started by querying one of 2 endpoints:

1. SyntaxTreeRequest with a null Parent.
2. SyntaxNodeAtRangeRequest with a file path that does not match the current path and text version.

The rest of the tree can be filled in by repeated requests to SyntaxTreeRequest with the parent filled in, in order to get the child nodes at that level.

It is also useful to build the tree in reverse order, starting from a nested node and working up. This enabled a combination of SyntaxNodeAtRangeRequest followed by repeated SyntaxNodeParentRequests to build up the spine of the tree, which allows clients to figure out what node to reveal in the tree when a document range is selected.

Finally, SyntaxNodeInfoRequest returns detailed information about a particular tree item. It returns the node type (and the symbol kind of that type), the SyntaxKind, the semantic classification of the node, all public properties of the node, and if the node is a SyntaxNode, the results of calling GetTypeInfo, GetConversion, GetSymbolInfo, and GetDeclaredSymbol on that node, enabling a detail view of the node.


333fred:
Note: I still need to write some tests, which is why this is draft for the moment.

filipw (commented):
have you thought about building this as a plugin (like Razor does)?
it sounds like a perfect candidate for a plugin, because it is a specialized type of a feature. At the same time, we need better plugin ecosystem and we need to have some good examples of plugins 🙂

333fred:
> have you thought about building this as a plugin (like Razor does)?
> it sounds like a perfect candidate for a plugin, because it is a specialized type of a feature. At the same time, we need better plugin ecosystem and we need to have some good examples of plugins 🙂

I'm definitely not sure how to go about doing that. If you have examples, I'd happily break this out.

filipw:
Razor is one example https://github.com/dotnet/aspnetcore-tooling/tree/main/src/Razor/src/Microsoft.AspNetCore.Razor.OmniSharpPlugin  
There is another one here https://github.com/dmgonch/fastcodenav

I am not aware of any other, and to be honest, I am not sure if it is even possible to have a plugin at the moment that contributes endpoints (instead of an entire project system for example). It would be good to have a tangible use case like this one though to further explore it and make sure we can support it, because that was the intention all along 😀

333fred:
Alright, I think it would be possible with some commands exposed from the vscode plugin. Here's what I'm thinking:

* `omnisharp.installPlugin` - takes a string dll path and a version number. Installs the dll into the correct folder and returns a structure of `installedVersion, needsRestart`. `needsRestart` will be true if the plugin was newly restarted and the plugin extension should prompt the user to restart the omnisharp server.
* `omnisharp.executePluginCommand` - takes a command id and an `any...` of args, and returns a `Thenable<T> | undefined`, sends that command to the server, and returns the result (if any).

@filipw thoughts?

NTaylorMullen:
If O# enabled plugins to provide [workspace commands](https://microsoft.github.io/language-server-protocol/specification#workspace_executeCommand) it wouldn't require exposing a new endpoint but instead would just require O# to route to specific command holders. I doubt it'd be that bad to implement 😄

333fred:
> If O# enabled plugins to provide [workspace commands](https://microsoft.github.io/language-server-protocol/specification#workspace_executeCommand) it wouldn't require exposing a new endpoint but instead would just require O# to route to specific command holders. I doubt it'd be that bad to implement smile

Only if we have the LSP mode enabled though, right? https://github.com/OmniSharp/omnisharp-vscode/pull/3817 isn't merged yet.

NTaylorMullen:
Eh, endpoints are just endpoints. I don't recall how it all fits together but I'd be surprised if that was 100% required

333fred:
> 
> 
> Eh, endpoints are just endpoints. I don't recall how it all fits together but I'd be surprised if that was 100% required

I would be. There's no way to send the request to the language server currently.

ryzngard:
@filipw can we move forward with this in some way? I really like this idea :) So much so I started writing my own extension to do this until I was mad aware of this PR. I'd like if this wasn't blocked on capabilities of omnisharp extensions because of endpoints.

333fred:
I have pulled this into a separate extension on the vscode marketplace, https://marketplace.visualstudio.com/items?itemName=333fred.net-compiler-developer-sdk, which supports the new roslyn-based LSP host as well as devkit.

PR #2088, opened 2021-02-09
do not download mono assets if they already exist (unless running full build)

filipw:
This speeds up local development loop. The storage account is slow and on for me it speeds up the quick build by roughly 30s.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -132,10 +132,23 @@ Task("ValidateMono")
     ValidateMonoVersion(buildPlan);
 });
 
+Task("CleanUpMonoAssets")
+    .WithCriteria(() => !Platform.Current.IsWindows)
+    .Does(() =>
+{
+    DirectoryHelper.Delete(env.Folders.Mono, recursive: true);
+});
+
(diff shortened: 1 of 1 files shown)


bjorkstromm (approved):
LGTM

build.cake
```diff
@@ -934,6 +947,7 @@ Task("Install")
 /// </summary>
 Task("All")
     .IsDependentOn("Cleanup")
+    .IsDependentOn("CleanUpMonoAssets")
```
bjorkstromm, line 950: `Default` is dependent on `All` and is dependent on `CleanUpMonoAssets`. Will this not result in Mono being installed every time when running default target on a non-Windows OS?

filipw: this is already the case today, and in some way it makes sense since it's a full rebuild then (as stated in the PR title).
The PR however addresses the quick build only, which also currently reinstalls the assets every time

bjorkstromm: Ah, yes! Makes sense.

JoeRobich (approved):

Issue #2090, opened 2021-02-14
VB Support


Youssef1313:
Having VB support is a great addition for OmniSharp IMO. It can be added with relatively little cost. I'll open a PR with sample work for BlockStructure endpoint

PR #2091, opened 2021-02-14
Support BlockStructure endpoint for VB

Youssef1313:
See https://github.com/OmniSharp/omnisharp-roslyn/issues/2090

Let me know if you want me to proceed with more work supporting VB.

If incomplete work shouldn't go to master, you may consider creating a feature branch that I get my PRs into instead of master, and merge it to master after the whole work is completed.

VB would be a really good addition.

### DIFF
--- a/OmniSharp.sln
+++ b/OmniSharp.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.27004.2005
+# Visual Studio Version 16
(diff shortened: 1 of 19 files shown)


marvin90m (approved):

StevenRotelli:
> See #2090
> 
> Let me know if you want me to proceed with more work supporting VB.
> 
> If incomplete work shouldn't go to master, you may consider creating a feature branch that I get my PRs into instead of master, and merge it to master after the whole work is completed.
> 
> VB would be a really good addition.
I'd like to see this happen, anyway I could be of use?

PR #2097, opened 2021-02-26
Improve custom error messages for MSB3644

filipw:
https://github.com/OmniSharp/omnisharp-roslyn/pull/2069 introduced custom error message for MSB3644 (avoid the default `The reference assemblies for {0} were not found. To resolve this, install the Developer Pack (SDK/Targeting Pack)...`)

This works well, however, .NET 5.0 and 6.0 are special cases, because the old MSBuilds assume these are ".NETFramework,Version=v5.0" and ".NETFramework,Version=v6.0" and our solution suggestion (to use global Mono instead of embedded one) is invalid. In fact, in those cases we need to suggest the opposite.

This PR extends https://github.com/OmniSharp/omnisharp-roslyn/pull/2069 to provide special casing for .NET 5.0 and 6.0, which hopefully will make troubleshooting easier for both users and us.


JoeRobich (approved):
Thanks!

PR #2100, opened 2021-03-01
Removed Azure DevOps test runs

david-driscoll:
This is where we see the most failures is with the tests on ADO.

The release pipeline will still run through ADO for the time being (on master and tags) but this should speed up builds a bit I hope.


filipw (approved):
thanks! that's great

PR #2103, opened 2021-03-02
Update included build tools to closely match NET 6 Preview 1 SDK

JoeRobich:


filipw:
very nice could you also please upgrade the bundled MSBuild to 16.9 as well? It seems to work with .NET 6.0 currently, but it's still 16.8

JoeRobich:
> could you also please upgrade the bundled MSBuild to 16.9 as well? It seems to work with .NET 6.0 currently, but it's still 16.8

This does update the bundled MSBuild. 
https://github.com/OmniSharp/omnisharp-roslyn/blob/5cb88c36b360555b1dc03af4b265be397a65c808/tools/packages.config#L5-L9
The version in the Packages.props isn't deployed and is just used to compile against.

filipw:
ah you are right. We used to ship those as separate zips, but I forgot [we stopped doing that](https://github.com/OmniSharp/omnisharp-roslyn/pull/1697#issuecomment-582137266)

filipw (approved):
I think when this is merged, it's a good time to trigger a new release, it's been a while

JoeRobich:
> it's a good time to trigger a new release, it's been a while

I want to follow up this afternoon with a PR to update Roslyn to either our 3.9.0 package or possible 3.10 preview1.

PR #2107, opened 2021-03-04
added binding redirects for System.Threading.Tasks.Dataflow

filipw:


david-driscoll (approved):
I hate binding redirects.

JoeRobich (approved):
Thanks Filip

PR #1952, opened 2020-09-19
Support solution filters (*.slnf)

sharwell:
Addresses https://github.com/dotnet/roslyn/pull/47531#pullrequestreview-485152426

I'm not sure how to test this.

PR #2121, opened 2021-03-25
Support Solution filter (.slnf)

JoeRobich:
Based on https://github.com/OmniSharp/omnisharp-roslyn/pull/1952 with changes for a passing CI.


filipw (approved):
LGTM. you will need to tweak the auto start in the extension too https://github.com/OmniSharp/omnisharp-vscode/blob/3117b4e9e053b4928ec979f360f93b39ab697c11/src/omnisharp/server.ts#L505

JoeRobich:
> you will need to tweak the auto start in the extension too

Yeah, I think the changes in https://github.com/OmniSharp/omnisharp-vscode/pull/4076 should do it.

AffluentOwl:
Any links to documentation on how to use this / what the intended workflow is? What if my project has hundreds of csproj files, is supposed to have no sln files, but a few spurious ones popped up that I want to filter out from omnisharp?

I'd like to avoid having to add and manage a sln file or slnf file for hundreds of projects I don't own, and instead just filter out all the bad sln files which are corrupting omnisharp.

Issue #2123, opened 2021-03-26
ArgumentOutOfRangeException from `completion/resolve` on the 1st line of top level program


filipw:
This reproduces for me with an empty `.cs` file + a regular project file, when trying to use the top level statements.

As I start typing `Console` I see the following in the logs (using OmniSharp 1.37.6):

```
[warn]: OmniSharp.Stdio.Host
        ************ Request ************
{
  "Type": "request",
  "Seq": 581,
  "Command": "/completion/resolve",
  "Arguments": {
    "Item": {
      "Label": "Console",
      "Kind": 7,
      "Tags": null,
      "Detail": "System",
      "Documentation": null,
      "Preselect": false,
      "SortText": "1Console System",
      "FilterText": "Console",
      "InsertTextFormat": 1,
      "TextEdit": {
        "NewText": "Console",
        "StartLine": 0,
        "StartColumn": 0,
        "EndLine": 0,
        "EndColumn": 1
      },
      "CommitCharacters": [
        " ",
        "{",
        "}",
        "[",
        "]",
        "(",
        ")",
        ".",
        ",",
        ":",
        ";",
        "+",
        "-",
        "*",
        "/",
        "%",
        "&",
        "|",
        "^",
        "!",
        "~",
        "=",
        "<",
        ">",
        "?",
        "@",
        "#",
        "'",
        "\"",
        "\\"
      ],
      "AdditionalTextEdits": null,
      "Data": 455
    }
  }
}
[fail]: OmniSharp.Stdio.Host
        ************  Response ************ 
{
  "Request_seq": 581,
  "Command": "/completion/resolve",
  "Running": true,
  "Success": false,
[...]

PR #2126, opened 2021-04-02
Rework completion resolution

333fred:
Recently, Roslyn undeprecated the CompletionChange.TextChanges API. This API gives changes in a significantly simpler manner for us to deal with, as it splits up things like imports and the actual main change element, so we can remove a whole bunch of code dedicated to finding the changed elements and mapping original source locations to the new document. The new pattern is much simpler: except for import completion, we always resolve the change up front. For most providers, this is an extremely quick call, as most providers just return the label, and for the ones that don't we can't do much about it anyway. We can then loop through the individual changes, putting changes that are not touching the current cursor location into AdditionalTextChanges. This shrinks the completion payload pretty significantly for many scenarios, and gets rid of a bunch of special handling around it. The only remaining special handling is adjusting the filter texts and snippitizing completions that want to move the cursor. I've also aligned the behavior of the Preselect flag with Roslyn's completion handler, and removed additional filtering of completion items so that they can be filtered by the client instead.

Fixes https://github.com/OmniSharp/omnisharp-roslyn/issues/2123 as well.

### DIFF
--- a/OmniSharp.sln
+++ b/OmniSharp.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.27004.2005
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.31129.286
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{2C348365-A9D8-459E-9276-56FC46AAEE31}"
 EndProject
@@ -67,9 +67,11 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "OmniSharp.Stdio.Driver", "s
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "OmniSharp.Script.Tests", "tests\OmniSharp.Script.Tests\OmniSharp.Script.Tests.csproj", "{9E4BA68C-7F4B-429A-A0C7-8CE7D41D610F}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "OmniSharp.Shared", "src\OmniSharp.Shared\OmniSharp.Shared.csproj", "{9571E3FE-E742-44AC-9E1F-64156815B8E1}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "OmniSharp.Shared", "src\OmniSharp.Shared\OmniSharp.Shared.csproj", "{9571E3FE-E742-44AC-9E1F-64156815B8E1}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "OmniSharp.Lsp.Tests", "tests\OmniSharp.Lsp.Tests\OmniSharp.Lsp.Tests.csproj", "{D67AA10B-8DB6-408D-A4C5-0B1DDCF5B3CC}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "OmniSharp.Lsp.Tests", "tests\OmniSharp.Lsp.Tests\OmniSharp.Lsp.Tests.csproj", "{D67AA10B-8DB6-408D-A4C5-0B1DDCF5B3CC}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "OmniSharp.Benchmarks", "src\OmniSharp.Benchmarks\OmniSharp.Benchmarks.csproj", "{6F5B209E-8DD3-4E90-A1F3-D85E7AF56588}"
 EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
@@ -369,6 +371,18 @@ Global
 		{D67AA10B-8DB6-408D-A4C5-0B1DDCF5B3CC}.Release|x64.Build.0 = Release|Any CPU
 		{D67AA10B-8DB6-408D-A4C5-0B1DDCF5B3CC}.Release|x86.ActiveCfg = Release|Any CPU
 		{D67AA10B-8DB6-408D-A4C5-0B1DDCF5B3CC}.Release|x86.Build.0 = Release|Any CPU
+		{6F5B209E-8DD3-4E90-A1F3-D85E7AF56588}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{6F5B209E-8DD3-4E90-A1F3-D85E7AF56588}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{6F5B209E-8DD3-4E90-A1F3-D85E7AF56588}.Debug|x64.ActiveCfg = Debug|Any CPU
+		{6F5B209E-8DD3-4E90-A1F3-D85E7AF56588}.Debug|x64.Build.0 = Debug|Any CPU
+		{6F5B209E-8DD3-4E90-A1F3-D85E7AF56588}.Debug|x86.ActiveCfg = Debug|Any CPU
+		{6F5B209E-8DD3-4E90-A1F3-D85E7AF56588}.Debug|x86.Build.0 = Debug|Any CPU
+		{6F5B209E-8DD3-4E90-A1F3-D85E7AF56588}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{6F5B209E-8DD3-4E90-A1F3-D85E7AF56588}.Release|Any CPU.Build.0 = Release|Any CPU
+		{6F5B209E-8DD3-4E90-A1F3-D85E7AF56588}.Release|x64.ActiveCfg = Release|Any CPU
+		{6F5B209E-8DD3-4E90-A1F3-D85E7AF56588}.Release|x64.Build.0 = Release|Any CPU
+		{6F5B209E-8DD3-4E90-A1F3-D85E7AF56588}.Release|x86.ActiveCfg = Release|Any CPU
+		{6F5B209E-8DD3-4E90-A1F3-D85E7AF56588}.Release|x86.Build.0 = Release|Any CPU
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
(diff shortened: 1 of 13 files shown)


src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionService.cs
```diff
@@ -509,86 +455,17 @@ public async Task<CompletionResolveResponse> Handle(CompletionResolveRequest req
             };
         }
 
-        private (IReadOnlyList<LinePositionSpanTextChange>? edits, int endOffset) GetAdditionalTextEdits(
```
333fred, line 512: Man am I happy to delete this method :).

tests/OmniSharp.Roslyn.CSharp.Tests/CompletionFacts.cs
```diff
@@ -586,15 +575,10 @@ public MyClass2()
...
-                switch (c.Label)
-                {
-                    case "text:":
-                        Assert.True(c.Preselect);
-                        break;
-                    default:
-                        Assert.False(c.Preselect);
-                        break;
-                }
+                if (c.Label == "ToString")
```
333fred, line 578: I honestly can't tell you why Roslyn wants to preselect this, but it does. The behavior in VSCode is that `ToString` is filtered out because of the prefix mismatch.

tests/OmniSharp.Roslyn.CSharp.Tests/CompletionFacts.cs
```diff
@@ -270,12 +270,34 @@ public static void Test(this object o)
 
             Assert.Single(resolved.Item.AdditionalTextEdits);
             var additionalEdit = resolved.Item.AdditionalTextEdits[0];
-            Assert.Equal(NormalizeNewlines("using N2;\n\nnamespace N1\r\n{\r\n    public class C1\r\n    {\r\n        public void M(object o)\r\n        {\r\n            o"),
+            Assert.Equal(NormalizeNewlines("using N2;\n\n"),
```
333fred, line 273: @bjorkstromm @NTaylorMullen these change will, I think, make import completion easier to support from Cake/Razor, as we're not sending the whole document anymore. Just the import changes.

tests/OmniSharp.Roslyn.CSharp.Tests/CompletionFacts.cs
```diff
@@ -270,12 +270,34 @@ public static void Test(this object o)
...
-            Assert.Equal(13, additionalEdit.EndColumn);
+            Assert.Equal(0, additionalEdit.EndLine);
+            Assert.Equal(0, additionalEdit.EndColumn);
+            VerifySortOrders(completions.Items);
+        }
+
+        [Theory]
+        [InlineData("dummy.cs")]
+        [InlineData("dummy.csx")]
+        public async Task ImportCompletion_OnLine0(string filename)
```
filipw, line 285: thanks for fixing this!

src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionService.cs
```diff
@@ -173,167 +171,134 @@ public async Task<CompletionResponse> Handle(CompletionRequest request)
...
+                    case CompletionItemExtensions.InternalsVisibleToCompletionProvider:
+                        // if the completion is for the hidden Misc files project, skip it
+                        if (completion.DisplayText == Configuration.OmniSharpMiscProjectName) continue;
+                        goto default;
+
+                    default:
+                        {
+                            // Except for import completion, we just resolve the change up front in the sync version. It's only expensive
+                            // for override completion, but there's not a heck of a lot we can do about that for the sync scenario
+                            var change = await completionService.GetChangeAsync(document, completion);
```
filipw, line 201: does it now make sense to update the client to not call resolve on the server for each item except in specific cases?

filipw:
looks very nice, thank you

src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionService.cs
333fred: No, because resolve will still fill in documentation for all items.

david-driscoll:
`Timeout of 2400000ms hit` 😱

src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionService.cs
david-driscoll: Any benefit (or problems!) with using `Task.WhenAll` to group all the async operations.  Not that we were doing this before anyway.

david-driscoll (approved):
Not sure what's causing the issues with GitHub Actions, but that's concerning at least because Azure Devops was just fine.

:shipit:  !

333fred:
I'll try to look into the timeouts sometime this weekend.

david-driscoll:
Oh wait, azure devops doesn't run tests anymore, so never mind there may be something there.  I'll try and run them on my machines later as well.

333fred:
> I'll try to look into the timeouts sometime this weekend.

So, I have no idea why the tests failing is causing the github action runner to take so long (perhaps all the output?), but the main issue is that I forgot I cloned all the tests for the lsp handler when I updated LSP to use the new service, and they're of course now all failing.

src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionService.cs
333fred: I'd need to get a real benchmark setup to determine this. A quick test didn't show any _obvious_ impact either way.

333fred: I've created a benchmark project using benchmarkdotnet and included in this PR. Some statistics from running it with and without `Task.WhenAll`:


|                    Method |     Mean |     Error |    StdDev |
|-------------------------- |---------:|----------:|----------:|
| ImportCompletionListAsyncOriginal | 3.715 ms | 0.0203 ms | 0.0190 ms |
| ImportCompletionListAsyncWhenAll | 4.232 ms | 0.0200 ms | 0.0178 ms |

|                  Method | NumOverrides |      Mean |    Error |   StdDev |    Median |
|------------------------ |------------- |----------:|---------:|---------:|----------:|
| OverrideCompletionAsyncOriginal |           10 |  11.54 ms | 0.051 ms | 0.045 ms |  11.53 ms |
| OverrideCompletionAsyncOriginal |          100 |  34.25 ms | 0.681 ms | 1.852 ms |  35.29 ms |
| OverrideCompletionAsyncOriginal |          250 |  65.92 ms | 1.312 ms | 3.194 ms |  67.20 ms |
| OverrideCompletionAsyncOriginal |          500 | 111.92 ms | 0.566 ms | 0.529 ms | 111.89 ms |
| OverrideCompletionAsyncWhenAll |           10 |  6.769 ms | 0.1011 ms | 0.0945 ms |
| OverrideCompletionAsyncWhenAll |          100 | 16.147 ms | 0.3139 ms | 0.2936 ms |
| OverrideCompletionAsyncWhenAll |          250 | 28.739 ms | 0.4544 ms | 0.3547 ms |
| OverrideCompletionAsyncWhenAll |          500 | 73.618 ms | 1.4516 ms | 1.7827 ms |
```

File: src/OmniSharp.LanguageServerProtocol/Handlers/OmniSharpCompletionHandler.cs [hunk 6]
```diff
@@ -112,7 +112,7 @@ private static void VerifyEnumsInSync(Type enum1, Type enum2)
             Debug.Assert(lspValues.Length == modelValues.Length);
             for (int i = 0; i < lspValues.Length; i++)
             {
-                Debug.Assert((int?)lspValues.GetValue(i) == (int?)modelValues.GetValue(i));
+                Debug.Assert((int)lspValues.GetValue(i) == (int)modelValues.GetValue(i));
```
filipw, line 115: I changed this to `int?` because I noticed on the .NET 5.0 branch that it reports `CS8605 "Unboxing possibly null value"` there

OmniSharp.sln
```diff
@@ -67,9 +67,11 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "OmniSharp.Stdio.Driver", "s
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "OmniSharp.Script.Tests", "tests\OmniSharp.Script.Tests\OmniSharp.Script.Tests.csproj", "{9E4BA68C-7F4B-429A-A0C7-8CE7D41D610F}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "OmniSharp.Shared", "src\OmniSharp.Shared\OmniSharp.Shared.csproj", "{9571E3FE-E742-44AC-9E1F-64156815B8E1}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "OmniSharp.Shared", "src\OmniSharp.Shared\OmniSharp.Shared.csproj", "{9571E3FE-E742-44AC-9E1F-64156815B8E1}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "OmniSharp.Lsp.Tests", "tests\OmniSharp.Lsp.Tests\OmniSharp.Lsp.Tests.csproj", "{D67AA10B-8DB6-408D-A4C5-0B1DDCF5B3CC}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "OmniSharp.Lsp.Tests", "tests\OmniSharp.Lsp.Tests\OmniSharp.Lsp.Tests.csproj", "{D67AA10B-8DB6-408D-A4C5-0B1DDCF5B3CC}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "OmniSharp.Benchmarks", "src\OmniSharp.Benchmarks\OmniSharp.Benchmarks.csproj", "{6F5B209E-8DD3-4E90-A1F3-D85E7AF56588}"
```
filipw, line 74: thank you for adding this... It has been long long long overdue ✨

src/OmniSharp.LanguageServerProtocol/Handlers/OmniSharpCompletionHandler.cs
333fred: Well, if it's int? the assert fails because these are ints, not nullable ints :)

filipw: oh 😅

PR #2134, opened 2021-04-09
Update build tools to match .NET 6 Preview 3 SDK.

JoeRobich:
Reworked the packages.config with details on how it can be updated.


JoeRobich:
FYI @nohwnd

JoeRobich:
@david-driscoll Getting failures in the LSP tests about being unable to load Newtonsoft.Json 11.0. Not sure if there is a good place to put a binding redirect or if we need to update the language server protocol libraries.

filipw (approved):
nice thanks

PR #2140, opened 2021-04-24
Do not crash on startup when configuration is invalid

filipw:
At the moment when bootstrapping configuration, if anything invalid is found (e.g. JSON structure is bad) we have an unhandled exception and failure to start.

Since the config is really optional I think it is better to swallow the exception, log an error and start with default config. This has been source of several issues already (got closed after the user fixed the config on their side).

since when the config is bootstrapped, the DI container is not yet initialized, logging of the possible error is deferred until the logger is available.


bjorkstromm (approved):
LGTM!

Would it be possible (useful) to also add a test for the configuration builder?

JoeRobich (approved):

filipw:
> Would it be possible (useful) to also add a test for the configuration builder?

I added some tests

PR #2149, opened 2021-05-05
Only subscribe to AppDomain.AssemblyResolve once

jasonmalinowski:
This is a port of https://github.com/dotnet/roslyn/pull/53166. There's a bigger question here about how to reuse this code better, but this caused a bunch of confusion while debugging an issue.


JoeRobich (approved):

filipw (approved):
👀

PR #2150, opened 2021-05-07
Remove private reflection from OmniSharp

333fred:
This updates OmniSharp to consume the new ExternalAccess assemblies that were added to Roslyn, providing a more structured form of internals access that will give us compile-time diagnostics and avoid the overhead of reflection-based calls. As a part of this, I'm updating our Roslyn dependency to 3.11, and I'm removing the deprecated AutoComplete handler in favor of the modern completion handlers.


333fred:
This is in draft and tests will fail until https://github.com/dotnet/roslyn/pull/53253 is merged and we can get a version with the bugfix in it, as the pickmembers service is currently failing.

333fred:
@filipw @JoeRobich for review.

filipw:
> @filipw @JoeRobich for review.

I have a feeling it would be good to do one last release with everything we have so far before merging this

src/OmniSharp.Roslyn.CSharp/Services/Intellisense/IntellisenseService.cs
```diff
@@ -1,213 +0,0 @@
...
-using OmniSharp.Mef;
-using OmniSharp.Models.AutoComplete;
-using OmniSharp.Options;
-using OmniSharp.Roslyn.CSharp.Services.Documentation;
-
-namespace OmniSharp.Roslyn.CSharp.Services.Intellisense
-{
-    [Obsolete("Please use CompletionService.")]
-    [OmniSharpHandler(OmniSharpEndpoints.AutoComplete, LanguageNames.CSharp)]
-    public class IntellisenseService : IRequestHandler<AutoCompleteRequest, IEnumerable<AutoCompleteResponse>>
```
filipw, line 20: we cannot remove this, not all clients have switched to the completion endpoint (e.g. vim)

src/OmniSharp.Roslyn/WorkspaceServices/ExtractClassWorkspaceService.cs
```diff
@@ -1,86 +1,33 @@
...
 using System.Threading.Tasks;
 using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.ExternalAccess.OmniSharp.ExtractClass;
 
 namespace OmniSharp
 {
-    public class ExtractClassWorkspaceService : DispatchProxy
+    [Shared]
+    [Export(typeof(IOmniSharpExtractClassOptionsService))]
+    internal class ExtractClassWorkspaceService : IOmniSharpExtractClassOptionsService
```
filipw, line 12: all of these look dramatically better 😅

filipw:
there is also a bunch of reflection calls here https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp.Script/ScriptProjectProvider.cs#L127-L140 looks like this is not part of the current change

333fred:
> there is also a bunch of reflection calls here https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp.Script/ScriptProjectProvider.cs#L127-L140 looks like this is not part of the current change

Unfortunately, I didn't see these when I was creating the external access library, and didn't include them in it. We can look at them as a follow up.

src/OmniSharp.Roslyn.CSharp/Services/Intellisense/IntellisenseService.cs
333fred: This makes me very sad :(.

333fred: I'm especially sad because I assumed I could and didn't expose the necessary things this service needed in external access.

333fred:
> Unfortunately, I didn't see these when I was creating the external access library, and didn't include them in it. We can look at them as a follow up.

@filipw the things being used by `ScriptProjectProvider` are compiler APIs, not IDE APIs. ExternalAccess cannot provide IVT to them as it doesn't have IVT to them itself. Those are going to have to stay as is.

filipw:
> @filipw the things being used by `ScriptProjectProvider` are compiler APIs, not IDE APIs. ExternalAccess cannot provide IVT to them as it doesn't have IVT to them itself. Those are going to have to stay as is.

ah I see, makes sense! 

given that we reset binder flags there and you could use it for all kinds of mad things (e.g ignore accessibility checks), I can imagine it's probably not one that will ever be exposed to anyone anyway 😅

src/OmniSharp.Roslyn.CSharp/Services/Intellisense/IntellisenseService.cs
filipw: thanks a lot for bringing this back. We'll engage with folks to try to get them to move, it's just not there yet

filipw (approved):
thanks! looks good to me, pending a 'current' release before the merge https://github.com/OmniSharp/omnisharp-roslyn/pull/2150#issuecomment-839520018

src/OmniSharp.Roslyn.CSharp/Services/Intellisense/IntellisenseService.cs
nickspoons: What is the minimum requirement to use the new endpoint instead of the old one in OmniSharp-vim? Can we make a couple of small changes and continue as we are, ignoring the extra code changes for now?

Someone has had a look at switching but we haven't had time to get into it properly. But if we can make a small change for the sake of keeping the server codebase lean, that is doable.

333fred: > Can we make a couple of small changes and continue as we are, ignoring the extra code changes for now?

I expect so. The new endpoint is modeled on the `completion` LSP endpoint. The one concern would be override completion, as I assume that you won't really be able to consume the async completion prototype I've made.

333fred:
> given that we reset binder flags there and you could use it for all kinds of mad things (e.g ignore accessibility checks), I can imagine it's probably not one that will ever be exposed to anyone anyway 😅

Yeah, the compiler layer gives IVT to no-one, and that's not going to be changing at any point... ever :).

JoeRobich:
> thanks! looks good to me, pending a 'current' release before the merge

Release created - https://github.com/OmniSharp/omnisharp-roslyn/releases/tag/v1.37.9

JoeRobich (approved):

PR #2153, opened 2021-05-14
Merge master to demo/net50

JoeRobich:
- Bumping the Packages.props NuGet version higher than 5.2.0 breaks compatibility with the .NET Core 2.1 SDK. If user have the SDK version set to 2.1.x by a global.json, they will need to move to update it to a more recent SDK. .NET Core 2.1 is end of support August 21,  2021 (see [Support Policy](https://dotnet.microsoft.com/platform/support/policy/dotnet-core)).
- Add NET 5.0 Tests CI without Cake tests
- Use Microsoft.Build.Locator for finding VS, DevConsole, and SDK instance of MSBuild
- For .NET 5.0 O# load MSBuild from SDK


filipw (approved):
fantastic, thank you

PR #2158, opened 2021-05-23
Add support for InlineValues

333fred:
Adds support for calculating expressions for language clients to have debuggers show values inline. As part of this, I've added a few test helper utilities from roslyn to make test diffing much better for things that use AssertEx instead of Xunit's Assert, as it adds diffing and full text display.

InlineValues suffers from a small limitation in generic contexts: static fields and properties cannot be determined. We need to give an expression for the debugger to evaluate, and since we don't know the current generic type parameters we can't know how to do this.

API shape is based on https://code.visualstudio.com/api/references/vscode-api#InlineValuesProvider.


333fred:
In action:

InlineValues

333fred:
Currently, I return references on every line they occur. It might be a better design to only return the first instance of the reference that occurs inside the current viewport, or perhaps try to find the closest instance to the current line. The main concern with doing so would be that it's much harder to cache such results.

src/OmniSharp.Roslyn.CSharp/Services/InlineValues/InlineValuesService.cs
```diff
@@ -0,0 +1,235 @@
...
+            {
+                // Couldn't find the context, just return early.
+                _logger.LogWarning("Could not find context for {0}:{1}", document.FilePath, request.Context.StoppedLocation);
+                return new InlineValuesResponse();
+            }
+            else if (containingMember is TypeDeclarationSyntax or EventDeclarationSyntax)
+            {
+                _logger.LogInformation("Request was for type or event {0}:{1}", document.FilePath, request.Context.StoppedLocation);
+                return new InlineValuesResponse();
+
```
JoeRobich, line 74: nit: empty line before closing brace
```suggestion
```

JoeRobich (approved):

333fred:
@JoeRobich did you have any opinions on the design question around how many things to return?

JoeRobich:
> @JoeRobich did you have any opinions on the design question around how many things to return?

@333fred other implementations provide the values for each occurrence which seems like a fine place to start from.

333fred:
Closing this out as stale and not going to be implemented.

PR #2168, opened 2021-05-30
Add V2 version of GotoDefinitionService

333fred:
The V1 GotoDefinitionService can only return one definition for a symbol. This is particularly annoying for partial types, which have multiple locations, and the location the service returns is pretty much never the one I want. The V2 service is very similar, but it uses V2 location models and returns a list of definitions, not a single one.


tests/OmniSharp.Roslyn.CSharp.Tests/GoToDefinitionV2Facts.cs
```diff
@@ -0,0 +1,636 @@
...
+}
+");
+
+            var testFile2 = new TestFile("baz.cs", @"
+partial class {|def:Class|}
+{
+}");
+
+            await TestGoToSourceAsync(testFile1, testFile2);
+        }
```
333fred, line 213: These are the only new tests in this PR. Everything else is a carbon copy of the V1 tests, with the same results.

333fred:
@filipw @david-driscoll could you take a look?

src/OmniSharp.Abstractions/Models/v2/Range.cs
```diff
@@ -29,30 +26,5 @@ public bool Contains(int line, int column)
...
-
-        public override bool Equals(object obj)
-            => Equals(obj as Range);
-
-        public bool Equals(Range other)
-            => other != null
-                && EqualityComparer<Point>.Default.Equals(Start, other.Start)
-                && EqualityComparer<Point>.Default.Equals(End, other.End);
-
-        public override int GetHashCode()
```
filipw, line 41: ✨

src/OmniSharp.Roslyn.CSharp/Services/Navigation/GoToDefinitionHelpers.cs
```diff
@@ -0,0 +1,51 @@
...
+        internal static async Task<FileLinePositionSpan?> GetMetadataMappedSpan(
+            Document document,
+            ISymbol symbol,
+            ExternalSourceServiceFactory externalSourceServiceFactory,
+            IExternalSourceService externalSourceService,
+            OmniSharpOptions options,
+            int timeout)
+        {
+
+            var cancellationToken = externalSourceServiceFactory.CreateCancellationToken(options, timeout);
```
filipw, line 39: in the past the cancellation token was only used for the metadata stuff, but since you are refactoring into a cleaner v2 version anyway, I think it is time we start doing this properly and create the cancellation token based on the request timeout at the entry point to the handler and flow it through all the async APIs used. Would you mind changing this?

tests/OmniSharp.Roslyn.CSharp.Tests/GoToDefinitionV2Facts.cs
filipw: would it be possible to make one test class inherit from the other then?

filipw (commented):
Thank, you this is a very good idea - I left some small comments only

src/OmniSharp.Roslyn.CSharp/Services/Navigation/GoToDefinitionHelpers.cs
```diff
@@ -0,0 +1,51 @@
...
+using System.Threading.Tasks;
+
+namespace OmniSharp.Roslyn.CSharp.Services.Navigation
+{
+    internal static class GoToDefinitionHelpers
+    {
+        internal static async Task<ISymbol?> GetDefinitionSymbol(Document document, int line, int column)
+        {
+            var sourceText = await document.GetTextAsync();
+            var position = sourceText.GetPositionFromLineAndOffset(line, column);
```
filipw, line 16: I think we can use this opportunity to clean up a bit - looks like we have two competing extension methods that do the same thing https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp.Roslyn/Extensions/SourceTextExtensions.cs#L10 and https://github.com/OmniSharp/omnisharp-roslyn/blob/aa1018cc281bc8fe2ead8e237b560d15bae1f646/src/OmniSharp.Roslyn/Extensions/TextExtensions.cs#L26-L33

I think we can get rid of one, or am I wrong on something?

tests/OmniSharp.Roslyn.CSharp.Tests/GoToDefinitionV2Facts.cs
333fred: Done.

333fred:
@filipw addressed your feedback, thanks.

filipw (approved):
thanks!

src/OmniSharp.Cake/Extensions/ResponseExtensions.cs
```diff
@@ -260,7 +260,7 @@ private static async Task<CodeElement> TranslateAsync(this CodeElement element,
                     continue;
                 }
 
-                builder.AddChild(childElement);
+                builder.AddChild(translatedElement);
```
333fred, line 263: This only ever worked because mutation.

filipw: 🙈

Issue #1934, opened 2020-09-10
What plans exist regarding C# source generator support?


alexrp:
Hello,

I saw [this blog post](https://devblogs.microsoft.com/dotnet/new-c-source-generator-samples/) on the .NET blog recently, which indicates that source generators are far enough along that they're actually usable in practice. So, in one of my projects, I ported a T4 template to a C# source generator to try it out. While this worked fine during build, it turns out that OmniSharp doesn't currently support them well: The generated types do not show up in IntelliSense, so lots of errors appear throughout the solution.

In [an earlier post](https://devblogs.microsoft.com/dotnet/introducing-c-source-generators/), it says:

> Note: you will currently need to restart Visual Studio to see IntelliSense and get rid of errors with the early tooling experience

Restarting OmniSharp, or even VS Code as a whole, does not appear to help.

So, I just wanted to check if there are any plans regarding support for C# source generators in OmniSharp? (Assuming that this is a lack of support, and not just a bug.)

jasonmalinowski:
So from the Roslyn perspective, we expected that the core support added to the Roslyn Workspace APIs was going to mean support in Omnisharp lit up automatically, but it seems it's not always working reliably. There's a couple of bugs we're currently investigating and working on fixing. We are trying to use source generators in the Roslyn project itself, and since we have some developers using VS Code we are working to make sure we don't break ourselves!

333fred:
To update this, basic scenarios (completion, hover, semantic, colorization) seem to work. However, things like go-to-def create ugly errors that need to be resolved somehow.

alexrp:
With `omnisharp.path` set to `latest`, I get this in the OmniSharp Log window:

```
        Failure while loading the analyzer reference 'libflare-gen': Could not load file or assembly 'System.Runtime, Version=5.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The system cannot find the file specified.
```

`dotnet build` on the command line works fine.

The generator project looks like this:

```
<Project Sdk="Microsoft.NET.Sdk">
    <PropertyGroup>
        <AssemblyName>libflare-gen</AssemblyName>
        <RootNamespace>Flare.Generator</RootNamespace>
        <TargetFramework>net5.0</TargetFramework>
    </PropertyGroup>

    <ItemGroup>
        <PackageReference Include="Microsoft.CodeAnalysis" Version="3.8.0-3.final" />
[...]

PR #2170, opened 2021-06-01
Add support for GoToDefinition on source-generated files

333fred:
This adds a new response element to gotodefinition responses: SourceGeneratedFileInfo. This is similar to MetadataSource, except that unlike MetadataSource it's not tracked on a type/project basis, but rather as a document/project basis. Retrieving info about a source generated file can be done through the SourceGeneratedFileService endpoints:

* SourceGeneratedFileInfo - Gets the file content of a source generated file.
* UpdateSourceGeneratedFileInfo - Gets the updated content of a source generated file, if it has changed since the last time information was returned.
* SourceGeneratedFileClosed - Sent to the server to inform it that the editor has closed the generated file and it can stop tracking Document version info for that file.

Currently, the only way to get the info needed to open a source-generated file is via the gotodefinition endpoint. We'll want to add info to find-usages as well, but that's a job for another day. Contributes to https://github.com/OmniSharp/omnisharp-roslyn/issues/1934.

### DIFF
--- a/src/OmniSharp.Abstractions/Models/v1/GotoDefinition/GotoDefinitionResponse.cs
+++ b/src/OmniSharp.Abstractions/Models/v1/GotoDefinition/GotoDefinitionResponse.cs
@@ -1,16 +1,20 @@
+#nullable enable
+
 using Newtonsoft.Json;
 using OmniSharp.Models.Metadata;
+using OmniSharp.Models.v1.SourceGeneratedFile;
 
 namespace OmniSharp.Models.GotoDefinition
 {
     public class GotoDefinitionResponse : ICanBeEmptyResponse
     {
-        public string FileName { get; set; }
+        public string? FileName { get; set; }
         [JsonConverter(typeof(ZeroBasedIndexConverter))]
         public int Line { get; set; }
         [JsonConverter(typeof(ZeroBasedIndexConverter))]
         public int Column { get; set; }
-        public MetadataSource MetadataSource { get; set; }
+        public MetadataSource? MetadataSource { get; set; }
+        public SourceGeneratedFileInfo? SourceGeneratedInfo { get; set; }
         public bool IsEmpty => string.IsNullOrWhiteSpace(FileName) && MetadataSource == null;
     }
 }
--- a/src/OmniSharp.Abstractions/Models/v1/GotoDefinition/GotoDefinitionResponse.cs
+++ b/src/OmniSharp.Abstractions/Models/v1/GotoDefinition/GotoDefinitionResponse.cs
@@ -15,6 +15,6 @@ namespace OmniSharp.Models.GotoDefinition
         public int Column { get; set; }
         public MetadataSource? MetadataSource { get; set; }
         public SourceGeneratedFileInfo? SourceGeneratedInfo { get; set; }
-        public bool IsEmpty => string.IsNullOrWhiteSpace(FileName) && MetadataSource == null;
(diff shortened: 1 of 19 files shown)


333fred:
@filipw another go-to-def PR for you to review 😊.

src/OmniSharp.Abstractions/Models/v1/GotoDefinition/GotoDefinitionResponse.cs
```diff
@@ -1,16 +1,20 @@
...
-        public string FileName { get; set; }
+        public string? FileName { get; set; }
         [JsonConverter(typeof(ZeroBasedIndexConverter))]
         public int Line { get; set; }
         [JsonConverter(typeof(ZeroBasedIndexConverter))]
         public int Column { get; set; }
-        public MetadataSource MetadataSource { get; set; }
+        public MetadataSource? MetadataSource { get; set; }
+        public SourceGeneratedFileInfo? SourceGeneratedInfo { get; set; }
         public bool IsEmpty => string.IsNullOrWhiteSpace(FileName) && MetadataSource == null;
```
filipw, line 18: you need to include `SourceGeneratedInfo` into this empty check too

src/OmniSharp.Roslyn.CSharp/Services/Navigation/SourceGeneratedFileService.cs
```diff
@@ -0,0 +1,105 @@
...
+            }
+
+            return new SourceGeneratedFileResponse
+            {
+                Source = text.ToString(),
+                SourceName = document.FilePath
+            };
+        }
+
+        public async Task<UpdateSourceGeneratedFileResponse> Handle(UpdateSourceGeneratedFileRequest request)
```
filipw, line 62: when would the client call the update point?

src/OmniSharp.Roslyn.CSharp/Services/Navigation/SourceGeneratedFileService.cs
```diff
@@ -0,0 +1,105 @@
...
+            if (document == null)
+            {
+                lock (_lock)
+                {
+                    _ = _lastSentVerisons.Remove(documentId);
+                }
+                return new UpdateSourceGeneratedFileResponse() { UpdateType = UpdateType.Deleted };
+            }
+
+            var docVersion = await document.GetTextVersionAsync();
```
filipw, line 75: I would say the version calculation should be done inside the lock as well?

filipw:
whoa! thank you!

src/OmniSharp.Roslyn.CSharp/Services/Navigation/SourceGeneratedFileService.cs
333fred: Can't, it's an async call.

src/OmniSharp.Roslyn.CSharp/Services/Navigation/SourceGeneratedFileService.cs
333fred: The way I've done it for vscode, when the buffer is shown.

src/OmniSharp.Roslyn.CSharp/Services/Navigation/SourceGeneratedFileService.cs
333fred: Also, as `Document`s are immutable, it doesn't matter. It'll always give the same version back.

filipw (approved):
LGTM!

Issue #2171, opened 2021-06-03
Omnisharp failed to load with InternalLoggerExeception error


dhp279:
## Issue Description ##
Omnisharp crashes when opening some project.

## Expected Behavior ##
Omnisharp starts successfully
## Actual Behavior ##
Omnisharp unable to load project

### OmniSharp log ###
<details>Starting OmniSharp server at 6/3/2021, 3:49:52 PM
    Target: c:\DsMainDev\testsrc\ntdbms\SEIPipeline\PipelineInfrastructure.sln

OmniSharp server started.
    Path: c:\Users\dapha.REDMOND.000\.vscode\extensions\ms-dotnettools.csharp-1.23.12\.omnisharp\1.37.11-beta.23\OmniSharp.exe
    PID: 7648

[info]: OmniSharp.Stdio.Host
        Starting OmniSharp on Windows 6.2.9200.0 (x64)
[info]: OmniSharp.Services.DotNetCliService
        DotNetPath set to dotnet
[info]: OmniSharp.MSBuild.Discovery.MSBuildLocator
        Located 4 MSBuild instance(s)
            1: Visual Studio Build Tools 2017 15.9.28307.1500 - "C:\Program Files (x86)\Microsoft Visual Studio\2017\BuildTools\MSBuild\15.0\Bin"
            2: Visual Studio Enterprise 2017 15.9.28307.1500 - "C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\MSBuild\15.0\Bin"
            3: Visual Studio Enterprise 2019 16.10.31321.278 - "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Current\Bin"
            4: StandAlone 16.11.0 - "c:\Users\dapha.REDMOND.000\.vscode\extensions\ms-dotnettools.csharp-1.23.12\.omnisharp\1.37.11-beta.23\.msbuild\Current\Bin"
[info]: OmniSharp.MSBuild.Discovery.MSBuildLocator
        Registered MSBuild instance: Visual Studio Enterprise 2019 16.10.31321.278 - "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Current\Bin"
[info]: OmniSharp.WorkspaceInitializer
        Invoking Workspace Options Provider: OmniSharp.Roslyn.CSharp.Services.CSharpFormattingWorkspaceOptionsProvider, Order: 0
[...]

PR #2172, opened 2021-06-04
Defend against null value in BuildErrorEventArgs

JoeRobich:
Resolves https://github.com/OmniSharp/omnisharp-roslyn/issues/2171

Looking at the deserialize code for BuildErrorEventArgs and it appears that `null` is a valid value for the `Code` property - https://github.com/dotnet/msbuild/blob/206d7ae3e6187e2d214041bf79719600ec197c56/src/Framework/BuildErrorEventArgs.cs#L300


filipw (approved):
that is new to me 🧐

PR #2175, opened 2021-06-18
Update build tools to match NET 6 Preview 5

JoeRobich:
also bumps other packages


filipw (approved):
Roslyn 4! should we make a release before updating?

JoeRobich:
> should we make a release before updating?

I've created a release from the commit prior to this merging.

PR #2181, opened 2021-06-19
Use the Microsoft.Build.Locator package for discovery

JoeRobich:
Thought this might be a good change to backport from the vnext branch.


filipw (approved):
This looks like a terrific idea to me 👏
thank you for porting this. one thing that occurred to me is that it may change the meaning of `"useglobalmono": "always"`. 

Perhaps we could consider to add a flag to be able to force the Mono instance provider with highest priority

JoeRobich:
> one thing that occurred to me is that it may change the meaning of `"useglobalmono": "always"`.

@filipw hmm.. This should only replace the VS and DevConsole providers. The new instance provider shouldn't return instances on non-Windows platforms. Should I change the check to not return instances on Mono platforms?

filipw:
ugh sorry I missed the `(!PlatformHelper.IsWindows)` check that returns no instances 🤦‍♂️

PR #2196, opened 2021-07-26
Update included build tools to match .NET 6 Preview 7 SDK

JoeRobich:
@captainsafia This change will require updating the Roslyn version you are using in aspnetcore-tooling.


filipw (approved):
thanks!

captainsafia (approved):
I think that should be fine. It should be fine to upgrade the Roslyn version in aspnetcore-tooling without updating aspnetcore.

JoeRobich:
@333fred Can you review this commit https://github.com/OmniSharp/omnisharp-roslyn/pull/2196/commits/f4a665441867e9d5faafd4db7a828caa71983263

333fred (approved):
LGTM. I do think we'll want tests and to verify what the behavior is, and we'll need to see whether we should keep this provider enabled for regular completion. It's possible it'll slow down standard completion too much.

Issue #2201, opened 2021-07-29
Error notification when LocationPaths is missing .dll file(s)


marcospgp:
Is there any way we can get an error when one of the `LocationPaths` does not contain a `.dll`? Otherwise it's relatively easy for analyzers to silently stop working.

![image](https://user-images.githubusercontent.com/3464445/127536998-129295d8-ce03-48c0-8f79-a70b76cbd285.png)

bjorkstromm:
Looking at the code and currently no.

I'd say an error might probably be too intrusive, but maybe a warning or at least a debug log could be added.

Relevant code is around here:

https://github.com/OmniSharp/omnisharp-roslyn/blob/aa1018cc281bc8fe2ead8e237b560d15bae1f646/src/OmniSharp.Roslyn.CSharp/Services/ExternalFeaturesHostServicesProvider.cs#L17-L32

https://github.com/OmniSharp/omnisharp-roslyn/blob/aa1018cc281bc8fe2ead8e237b560d15bae1f646/src/OmniSharp.Shared/Options/RoslynExtensionsOptions.cs#L21-L40

https://github.com/OmniSharp/omnisharp-roslyn/blob/aa1018cc281bc8fe2ead8e237b560d15bae1f646/src/OmniSharp.Host/Services/AssemblyLoader.cs#L38-L62

Thoughts, @filipw @david-driscoll?

PR #2202, opened 2021-07-30
log a warning when external features path has no assemblies

filipw:
Fixes https://github.com/OmniSharp/omnisharp-roslyn/issues/2201


bjorkstromm (approved):
LGTM! I'd use `Count > 0` instead of `Any()`,  but in this case it doesn't really matter. `Any()` reads better.

PR #2217, opened 2021-08-30
Update versions to match .NET SDK 6 RC1

nohwnd:

PR #2218, opened 2021-08-30
Update versions to match .NET SDK 6 RC1 (fix test)

JoeRobich:
Attempt to fix test failure in https://github.com/OmniSharp/omnisharp-roslyn/pull/2217

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -256,7 +256,6 @@ Task("CreateMSBuildFolder")
         "NuGet.Common",
         "NuGet.Configuration",
         "NuGet.DependencyResolver.Core",
-        "NuGet.Frameworks",
         "NuGet.LibraryModel",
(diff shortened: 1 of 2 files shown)


JoeRobich:
@filipw @nohwnd PTAL

filipw (approved):
looks legit thanks

filipw:
FWIW this would be my fix too, though I wouldn't be able to explain why 😅

JoeRobich:
Basically I have been pulling most of the NuGet libraries from the Microsoft.NET.Sdk package since they would all be version matched and that save a lot of repetition in the packages.config. In this case the NuGet packages in the Sdk package don't include this API change. We can revert my change once we move to a newer Sdk package that includes the change.

PR #2034, opened 2020-12-07
Update NuGet to version 5.10.0

grendello:
Fixes: https://github.com/OmniSharp/omnisharp-roslyn/issues/2027

Fixes the following exception when using OmniSharp on Linux (with Emacs,
but possibly breaks with other LSP clients as well):

    ERROR: OmniSharp.MSBuild.ProjectLoader, The "ResolvePackageAssets" task failed unexpectedly.
    System.MissingMethodException: Method not found: NuGet.ProjectModel.LockFileTarget NuGet.ProjectModel.LockFile.GetTarget(string,string)
      at Microsoft.NET.Build.Tasks.ResolvePackageAssets+CacheReader.CreateReaderFromDisk (Microsoft.NET.Build.Tasks.ResolvePackageAssets task, System.Byte[] settingsHash)
      at Microsoft.NET.Build.Tasks.ResolvePackageAssets+CacheReader..ctor (Microsoft.NET.Build.Tasks.ResolvePackageAssets task)
      at Microsoft.NET.Build.Tasks.ResolvePackageAssets.ReadItemGroups ()
      at Microsoft.NET.Build.Tasks.ResolvePackageAssets.ExecuteCore ()
      at Microsoft.NET.Build.Tasks.TaskBase.Execute ()
      at Microsoft.Build.BackEnd.TaskExecutionHost.Microsoft.Build.BackEnd.ITaskExecutionHost.Execute ()
      at Microsoft.Build.BackEnd.TaskBuilder.ExecuteInstantiatedTask (Microsoft.Build.BackEnd.ITaskExecutionHost taskExecutionHost, Microsoft.Build.BackEnd.Logging.TaskLoggingContext taskLoggingContext, Microsoft.Build.BackEnd.TaskHost taskHost, Microsoft.Build.BackEnd.ItemBucket bucket, Microsoft.Build.BackEnd.TaskExecutionMode howToExecuteTask)

Currently used NuGet 5.2.0 is too old, update to 5.10.0 makes the issue
disappear.

PR #2219, opened 2021-08-30
Remove .NET Core 2.1

JoeRobich:
With .NET Core 2.1 SDK out of support (see https://dotnet.microsoft.com/platform/support/policy/dotnet-core) we can remove tests and update our targeted framework.

This would also allow us to merge https://github.com/OmniSharp/omnisharp-roslyn/pull/2034/.


filipw:
that's correct I was meaning to remove it too - thanks!

filipw (approved):

PR #2223, opened 2021-08-31
Update the vnext branch

JoeRobich:


filipw:
"Failure while loading the analyzer reference 'StyleCop.Analyzers': Assembly with same name is already loaded"

does SDK now ship stylecop by default?

JoeRobich:
> does SDK now ship stylecop by default?
@filipw It does not. However, I just realized I'm targeting the wrong branch here.

JoeRobich:
We needed as similar set of changes to analyzer assembly loading as Roslyn made in https://github.com/dotnet/roslyn/pull/55098

src/OmniSharp.Host/Services/DefaultAnalyzerAssemblyLoader.Desktop.cs
```diff
@@ -0,0 +1,53 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+#if !NETCOREAPP
```
filipw, line 5: I think maybe we will just drop desktop builds from the branch altogether?

filipw (approved):
thanks for keeping this dream alive

filipw:
> We needed as similar set of changes to analyzer assembly loading as Roslyn made in [dotnet/roslyn#55098](https://github.com/dotnet/roslyn/pull/55098)

this is a very clever solution indeed, nice

PR #2236, opened 2021-09-17
Reuse Roslyn's analyzer assembly loader

jasonmalinowski:
Roslyn has made some improvements to better handle the case where analyzers or generators have a mix of assembly versions, such as in microsoft/CsWin32#218. The Roslyn changes were made in dotnet/roslyn#56432, and this consumes it.


filipw (commented):
thanks a lot for keeping this up to date

src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionOptionsProvider.cs
```diff
@@ -15,9 +15,9 @@ public class CompletionOptionsProvider : IWorkspaceOptionsProvider
         public int Order => 0;
 
         public OptionSet Process(OptionSet currentOptionSet, OmniSharpOptions omniSharpOptions, IOmniSharpEnvironment omnisharpEnvironment)
-            => currentOptionSet.WithChangedOption(
+            => currentOptionSet/*.WithChangedOption(
```
jasonmalinowski, line 18: @tmat: do we need to make a new API for Omnisharp so they can call into the completion service and specify the import completion flag? I guess this worked before by taking the value into an OptionSet here, and that was passed into  OmniSharpCompletionService.GetCompletionsAsync in CompletionService.cs; we didn't expose any way to pass your new struct through, so I'm guessing we need to revise that?

src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionService.cs
```diff
@@ -243,6 +259,11 @@ public async Task<CompletionAfterInsertResponse> Handle(CompletionAfterInsertReq
             }
 
             var completionService = CSharpCompletionService.GetService(document);
+            if (completionService is null)
```
jasonmalinowski, line 262: More is null annotated now, so this had to be added.

src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionOptionsProvider.cs
tmat: Looking now.

jasonmalinowski: This has now all been fixed up in #2306.

src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionService.cs
jasonmalinowski: This all happened in #2306 now.

src/OmniSharp.Roslyn/Utilities/ShadowCopyAnalyzerAssemblyLoader.cs
```diff
@@ -0,0 +1,10 @@
+﻿using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.ExternalAccess.OmniSharp.Analyzers;
+
+namespace OmniSharp.Roslyn.Utilities
+{
+    public static class ShadowCopyAnalyzerAssemblyLoader
+    {
+        public static readonly IAnalyzerAssemblyLoader Instance = OmnisharpAnalyzerAssemblyLoaderFactory.CreateShadowCopyAnalyzerAssemblyLoader();
```
jasonmalinowski, line 8: This is here because the Omnisharp.Host project doesn't have internals access to Microsoft.CodeAnalysis.ExternalAccess.Omnisharp; I don't know if that was an accidental oversight or intentional layering, but following it for now.

jasonmalinowski:
This is now ready for review now that we sorted everything else out.

src/OmniSharp.Roslyn/Utilities/ShadowCopyAnalyzerAssemblyLoader.cs
JoeRobich: I'm sure it just wasn't necessary for the initial round of reflection removal. This seems fine for now.

JoeRobich (approved):

PR #2239, opened 2021-10-01
Update versions from 6.0.1xx

nohwnd:


tools/packages.config
```diff
@@ -9,44 +9,44 @@
...
         The version for these packages can be found in the dotnet/installer eng/Versions.props file in
         the branch that corresponds with the SDK Toolset being modeled.
-        ex. https://github.com/dotnet/installer/blob/release/6.0.1xx-rc1/eng/Versions.props
+        ex. https://github.com/dotnet/installer/blob/release/6.0.1xx/eng/Versions.props
     -->
 
     <!-- These packages should match the "MicrosoftNETSdkPackageVersion" property-->
-    <package id="Microsoft.NET.Sdk" version="6.0.100-rc.1.21427.66" />
-    <package id="Microsoft.DotNet.MSBuildSdkResolver" version="6.0.100-rc.1.21427.66" />
+    <package id="Microsoft.NET.Sdk" version="6.0.100-rtm.21480.28" />
```
filipw, line 16: are these the ones envisioned for RTM already?!

filipw (approved):

tools/packages.config
nohwnd: @filipw Sorry, I don't know how to answer that. I only copied the versions from the release/6.0.1xx branch.

JoeRobich: I think ideally we would have used the packages from https://github.com/dotnet/sdk/tree/release/6.0.1xx-rc2 since RC2 is about to release. Once we have a new build we can validate against RC2 and see if this is going to be a problem or not.

filipw: right - that was exactly my question here too - I thought that perhaps RC2 was not happening anymore if the RTM versions are there now? I am not privvy to all these plans 😀

PR #2249, opened 2021-10-18
Fix typo ("Initalize")

CyberShadow:


filipw:
thanks - unfortunately this typo is in the public interface which is published to nuget and it is a breaking change.
However IMHO this is still acceptable change to make, as it should have never been like this in the first place, it's a clear mistake.

filipw:
@NTaylorMullen do you guys implement this interface in Razor tooling? would this be a breaking change if we rename this due to a typo?

PR #2262, opened 2021-11-01
Update versions to match dotnet SDK 6.0.1xx

nohwnd:


build/Packages.props
```diff
@@ -7,7 +7,7 @@
         <MicrosoftTestPackageVersion>16.9.4</MicrosoftTestPackageVersion>
         <MSBuildPackageVersion>16.10.0</MSBuildPackageVersion>
         <NuGetPackageVersion>5.10.0</NuGetPackageVersion>
-        <RoslynPackageVersion>4.0.0-5.21480.15</RoslynPackageVersion>
+        <RoslynPackageVersion>4.0.0-6.21526.21</RoslynPackageVersion>
```
JoeRobich, line 10: Can we update the version of Roslyn used by the language server to match? https://github.com/OmniSharp/omnisharp-roslyn/blob/master/build/Packages.props#L10

JoeRobich (approved):

build/Packages.props
nohwnd: What do you mean? You are linking to this file, but in main. I took the version from here https://github.com/dotnet/sdk/blob/release/6.0.1xx/eng/Versions.props#L124 but there is also final version on nuget https://www.nuget.org/packages/Microsoft.CodeAnalysis/4.0.0-6.final.

JoeRobich: haha. I am sorry. I totally missed this was exactly what I was asking for.

filipw (approved):
thanks!

FooqX:
so how to make it work with .net 6?? in vsc

PR #2263, opened 2021-11-02
Update changelog for 1.37.17

nohwnd:


CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
 # Changelog
 All changes to the project will be documented in this file.
 
+## [1.37.17]
```
nohwnd, line 4: In some issue we agreed to not write "no released yet" or the date, because we don't know when it will actually release, and the not released yet shows up in VSCode extension readme if we release with it in place.

filipw (approved):

CHANGELOG.md
filipw: this applies to the VS Code changelog, not Omnisharp server release notes - those are independent

nohwnd: Right, my bad.

PR #2267, opened 2021-11-08
Check diagnostic analyzer's configured severity before running it

JoeRobich:
This change will filter the analyzers which are run by whether they can produce a meaningful diagnostic. If a diagnostic has been configured in the .editorconfig with a severity of `none`, then it should be filtered out.

Resolves https://github.com/OmniSharp/omnisharp-vscode/issues/4718

### DIFF
--- a/src/OmniSharp.Roslyn.CSharp/Workers/Diagnostics/AnalyzerOptionExtensions.cs
+++ b/src/OmniSharp.Roslyn.CSharp/Workers/Diagnostics/AnalyzerOptionExtensions.cs
@@ -1,0 +1,201 @@
+using System;
+using System.Linq;
+using System.Threading;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace OmniSharp.Roslyn.CSharp.Workers.Diagnostics
+{
+    internal static class AnalyzerOptionsExtensions
+    {
+        internal const string DotnetDiagnosticPrefix = "dotnet_diagnostic";
+        internal const string DotnetAnalyzerDiagnosticPrefix = "dotnet_analyzer_diagnostic";
+        internal const string CategoryPrefix = "category";
+        internal const string SeveritySuffix = "severity";
+
+        internal const string DotnetAnalyzerDiagnosticSeverityKey = DotnetAnalyzerDiagnosticPrefix + "." + SeveritySuffix;
+
+        internal static string GetDiagnosticIdBasedDotnetAnalyzerDiagnosticSeverityKey(string diagnosticId)
+            => $"{DotnetDiagnosticPrefix}.{diagnosticId}.{SeveritySuffix}";
+        internal static string GetCategoryBasedDotnetAnalyzerDiagnosticSeverityKey(string category)
+            => $"{DotnetAnalyzerDiagnosticPrefix}.{CategoryPrefix}-{category}.{SeveritySuffix}";
+
+        /// <summary>
+        /// Tries to get configured severity for the given <paramref name="descriptor"/>
+        /// for the given <paramref name="tree"/> from analyzer config options, i.e.
+        ///     'dotnet_diagnostic.%descriptor.Id%.severity = %severity%',
+        ///     'dotnet_analyzer_diagnostic.category-%RuleCategory%.severity = %severity%',
+        ///         or
+        ///     'dotnet_analyzer_diagnostic.severity = %severity%'
+        /// </summary>
+        public static bool TryGetSeverityFromConfiguration(
+            this AnalyzerOptions analyzerOptions,
+            SyntaxTree tree,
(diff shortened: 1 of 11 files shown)


src/OmniSharp.Roslyn.CSharp/Workers/Diagnostics/DiagnosticAnalyzerExtensions.cs
```diff
@@ -0,0 +1,180 @@
...
+
+namespace OmniSharp.Roslyn.CSharp.Workers.Diagnostics
+{
+    public static class DiagnosticAnalyzerExtensions
+    {
+        private static MethodInfo TryGetMappedOptionsMethod { get; }
+
+        static DiagnosticAnalyzerExtensions()
+        {
+            TryGetMappedOptionsMethod = Assembly.Load(new AssemblyName("Microsoft.CodeAnalysis.Features"))
```
filipw, line 18: is the reflection here necessary? 
if yes, then for consistency it would make sense to use https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp.Shared/Utilities/ReflectionExtensions.cs

src/OmniSharp.Roslyn.CSharp/Workers/Diagnostics/AnalyzerOptionExtensions.cs
```diff
@@ -0,0 +1,201 @@
...
+                return true;
+            }
+
+            // Analyzer bulk configuration does not apply to:
+            //  1. Disabled by default diagnostics
+            //  2. Compiler diagnostics
+            //  3. Non-configurable diagnostics
+            if (analyzerOptions == null ||
+                !descriptor.IsEnabledByDefault ||
+                descriptor.CustomTags.Any(tag => tag == WellKnownDiagnosticTags.Compiler || tag == WellKnownDiagnosticTags.NotConfigurable))
```
filipw, line 57: I saw somewhere that there is also `WellKnownDiagnosticTags.Unnecessary`. do you know what is the purpose it? does it need to be include here too?

src/OmniSharp.Roslyn.CSharp/Workers/Diagnostics/CSharpDiagnosticWorkerWithAnalyzers.cs
```diff
@@ -255,7 +303,47 @@ private async Task AnalyzeProject(Solution solution, IGrouping<ProjectId, Docume
...
+                var diagnostics = ImmutableArray<Diagnostic>.Empty;
+
+                // Only basic syntax check is available if file is miscellanous like orphan .cs file.
+                // Those projects are on hard coded virtual project
+                if (project.Name != $"{Configuration.OmniSharpMiscProjectName}.csproj" && analyzers.Any())
+                {
+                    var compilationWithAnalyzers = compilation.WithAnalyzers(analyzers, new CompilationWithAnalyzersOptions(
+                        workspaceAnalyzerOptions,
+                        onAnalyzerException: OnAnalyzerException,
+                        concurrentAnalysis: true,
```
filipw, line 323: we always had it set to false. I am a bit worried that setting it to true will cause even more people to complain about the Mono process CPU usage. do you see any performance improvements when this is set to true?

src/OmniSharp.Roslyn.CSharp/Workers/Diagnostics/AnalyzerOptionExtensions.cs
JoeRobich: Unnecessary is a tag that indicates the generated diagnostics will be spans of unnecessary code and they will be tagged so that they are faded out. They shouldn't be included in this check.

src/OmniSharp.Roslyn.CSharp/Workers/Diagnostics/DiagnosticAnalyzerExtensions.cs
JoeRobich: I can work on adding an accessor to the ExternalAccess layer to remove this reflection down the road.

src/OmniSharp.Roslyn.CSharp/Workers/Diagnostics/CSharpDiagnosticWorkerWithAnalyzers.cs
JoeRobich: Let me dig into this. My assumption is that running them serially would prolong the load on the CPU, but I don't want to regress.

JoeRobich: Going to revert this for now. Would like to look again at performance some time soon.

filipw (approved):
LGTM

PR #2279, opened 2021-11-25
Move OmniSharp vNext to run on .NET 6

JoeRobich:


filipw (approved):
thanks, looks good!

filipw:
I think might be worth considering to yank the .NET Framework targets completely from the vnext branch as well and just target .NET 6.0 there

PR #2285, opened 2021-12-02
run analyzers on multiple threads if allowed to

### LINKED ISSUE #2241: Analyzing projects is extremely slow compared to MSBuild or Visual Studio
I am working on a .NET Framework-project where a full build using all .NET 5 SDK analyzers takes about 40 seconds, Visual Studio can analyze the project in a similar amount of time.

OmniSharp on the other hand can analyze for hours on end and use many gigabytes of RAM. In fact, it never gets done because as I work things change so one CPU core is running hot all day long. I just had to kill the OmniSharp process because it was using 21 GB of RAM.

I have noticed this performance difference in other projects too, even in pure .NET Core projects. This has been the status quo for years.

So there seems to be a fundamental difference in how analysis is done in OmniSharp vs regular Roslyn since there is at least an order of magnitude or more difference in speed.

Would it be possible to involve Microsoft to get this properly fixed once and for all?

TomasEkeli:
fixes #2241

## background
running analyzers is is currently  sequential and can take a long time on big projects. this change can help by running
in parallel, particularly if the machine has many cores.

## configuration
introduces a new option the user can set in `omnisharp.json`: `RoslynExtensionsOptions.DiagnosticWorkersThreadCount`.
- controls how many threads the diagnostic workers will use for analysis. this allows users to control how much of their available compute they want to give the analysers access to. 
- any value up (exclusive) the number of cores on their computer would be good. 
- setting this to 1 will make the analysis run in one worker (i.e. sequentially, like before this patch)
- not setting this value will default to using 75% of the available cores, or 1 if that is all that is available

## observability
the analysers report their status as a percentage done of the total number of documents to analyse. this is done instead of listing the currently analysed project, and gives the user a better experience. thanks to @darosenberg for this, and help throughout!

in my experience this decreases the wall-time of analysers by a good amount at the expense of running the users' computer hotter. exposing it as a configuration allows the users to make their own trade-off.

## documentation
the documentation for `omnisharp.json` in the [wiki](https://github.com/OmniSharp/omnisharp-roslyn/wiki/Configuration-Options#roslyn-extensions-options) should be updated once this is merged.


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/OmniSharp/omnisharp-roslyn?pullRequest=2285) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/OmniSharp/omnisharp-roslyn?pullRequest=2285) before we can accept your contribution.<br/><br/>:x: TomasEkeli [sign now](https://cla.dotnetfoundation.org/OmniSharp/omnisharp-roslyn?pullRequest=2285)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/OmniSharp/omnisharp-roslyn?pullRequest=2285) it.</sub>

src/OmniSharp.Shared/Options/RoslynExtensionsOptions.cs
```diff
@@ -12,6 +12,9 @@ public class RoslynExtensionsOptions : OmniSharpExtensionsOptions
         public bool EnableImportCompletion { get; set; }
         public bool EnableAsyncCompletion { get; set; }
         public int DocumentAnalysisTimeoutMs { get; set; } = 10 * 1000;
+        public int ThreadsToUseForAnalyzers { get; set; } = Environment.ProcessorCount == 1
+            ? 1
+            : Environment.ProcessorCount / 2;
```
DavidZidar, line 17: The expression can be simplified as `Math.Max(1, Environment.ProcessorCount / 2)`.

The option name is a bit complicated and does not seem to follow the naming pattern of the other options. How about `AnalyzerThreadCount` instead?

So here's my suggestion
```suggestion
        public int AnalyzerThreadCount { get; set; } = Math.Max(1, Environment.ProcessorCount / 2);
```

DavidZidar (commented):
Seems like a simple solution, I think it might work. When I did my quick hack however I did get some timeout errors, have you noticed anything like that? I probably got timeouts because I was going full blast with no "throttler" though.

Also, there's a similar foreach loop here, do you think it might need the same treatment?
https://github.com/OmniSharp/omnisharp-roslyn/blob/6a5ccc447b1fbe0fb4a3579a6e1dbdfcde427daf/src/OmniSharp.Roslyn.CSharp/Workers/Diagnostics/CSharpDiagnosticWorker.cs#L149-L157

TomasEkeli:
> Seems like a simple solution, I think it might work. When I did my quick hack however I did get some timeout errors, have you noticed anything like that? I probably got timeouts because I was going full blast with no "throttler" though.
> 
> Also, there's a similar foreach loop here, do you think it might need the same treatment?
> 
> https://github.com/OmniSharp/omnisharp-roslyn/blob/6a5ccc447b1fbe0fb4a3579a6e1dbdfcde427daf/src/OmniSharp.Roslyn.CSharp/Workers/Diagnostics/CSharpDiagnosticWorker.cs#L149-L157

nice catch, I'll look into it

DavidZidar:
I just pulled your branch and tried it out and it is indeed much faster, but I'm still getting the timeout errors I mentioned and tested this on a small/medium sized project on a 16 core Ryzen 5950X.

They look like this in the OmniSharp log (I truncated the list of analyzers because it is very long):

```
[fail]: OmniSharp.Roslyn.CSharp.Services.Diagnostics.CSharpDiagnosticWorkerWithAnalyzers
        Analysis of document SomeController.cs failed or cancelled by timeout: The operation was canceled., analysers: Microsoft.CodeAnalysis.UseSystemHashCode.UseSystemHashCodeDiagnosticAnalyzer, Microsoft.CodeAnalysis.UseExplicitTupleName.UseExplicitTupleNameDiagnosticAnalyzer, Microsoft.CodeAnalysis.MakeFieldReadonly.MakeFieldReadonlyDiagnosticAnalyzer, Microsoft.CodeAnalysis.Formatting.FormattingDiagnosticAnalyzer, ...
```

The timeout seems to be this one which is only 10 seconds, I'm thinking the default value would have to be increased when running things in parallel. Maybe 30 seconds?

https://github.com/OmniSharp/omnisharp-roslyn/blob/6a5ccc447b1fbe0fb4a3579a6e1dbdfcde427daf/src/OmniSharp.Shared/Options/RoslynExtensionsOptions.cs#L14

TomasEkeli:
ah, I've had that at 30 seconds for a long time, that's probably why I wasn't seeing the timeouts.

is there a problem if we just increase the default to that?

TomasEkeli:
i've set it to 30s default.

DaRosenberg:
I opened an alternative PR #2312 with some further improvements upon this work.

src/OmniSharp.Roslyn.CSharp/Workers/Diagnostics/CSharpDiagnosticWorkerWithAnalyzers.cs
```diff
@@ -215,8 +243,9 @@ public async Task<IEnumerable<Diagnostic>> AnalyzeProjectsAsync(Project project,
                 .Concat(project.AnalyzerReferences.SelectMany(x => x.GetAnalyzers(project.Language)))
                 .ToImmutableArray();
 
-            var compilation = await project.GetCompilationAsync();
+            var compilation = await project.GetCompilationAsync(cancellationToken);
             var workspaceAnalyzerOptions = (AnalyzerOptions)_workspaceAnalyzerOptionsConstructor.Invoke(new object[] { project.AnalyzerOptions, project.Solution });
+            var analyzerTasks = new List<Task>();
```
DaRosenberg, line 248: Can be removed?

TomasEkeli:
can we get a review on this @filipw or someone with write access? just picked filip to mention as he's the top contributor and probably know who would be right.

JoeRobich:
@TomasEkeli Can you update the PR description to describe the changes the code is currently making. I didn't see a new RoslynOption for number of analyzer threads for instance.

There are certainly multiple places we could add parallelism to the Analyzer runner. In `AnalyzeDocument()` we could set `concurrentAnalysis` to `true` when getting the CompilationWithAnalyzers. However, with that approach, we don't have any knobs to control how many resources it will consume. I think the approach in this PR is OK, since it gives O# the ability to decide how to use resources.

TomasEkeli:
> @TomasEkeli Can you update the PR description to describe the changes the code is currently making. I didn't see a new RoslynOption for number of analyzer threads for instance.
> 
> There are certainly multiple places we could add parallelism to the Analyzer runner. In `AnalyzeDocument()` we could set `concurrentAnalysis` to `true` when getting the CompilationWithAnalyzers. However, with that approach, we don't have any knobs to control how many resources it will consume. I think the approach in this PR is OK, since it gives O# the ability to decide how to use resources.

i've updated the pull-request description - it was a little outdated. sorry about that, and i hope it's clearer now.

DaRosenberg:
@JoeRobich We did consider setting `concurrentAnalysis` to `true` as well, but decided not to because we found this mention:

https://www.csharpcodi.com/csharp-examples/Microsoft.CodeAnalysis.Diagnostics.EngineV2.DiagnosticIncrementalAnalyzer.CompilationManager.GetAnalyzerExceptionFilter(Project)/

```
// in IDE, we always set concurrentAnalysis == false otherwise, we can get into thread starvation due to
// async being used with syncronous blocking concurrency.
```

Not sure if that's applicable at all in the OmniSharp context given that it's in a separate process from the IDE, but decided not the risk it nonetheless.

JoeRobich (approved):
Thanks!

JoeRobich:
CC: @filipw, in case you have any concerns

filipw:
I think this should be *OK* because it is gated. Also, normally we hide such features behind feature flags, but since the entire analysis is behind a feature flag already, it is fine. I am a bit worried that it can lead to state corruption on the client side but we can give it a try.

Thanks a lot for your work.

TomasEkeli:
i'v updated the wiki

filipw:
> I am a bit worried that it can lead to state corruption on the client side but we can give it a try.

Unfortunately, looking at the feedback in https://github.com/OmniSharp/omnisharp-vscode/issues/5017 from users using the latest build from master, it seems like there is indeed a regression where diagnostics retrieval crashes with a null reference.

DaRosenberg:
@filipw I can take a look later tonight. What's the correct process here, open another PR against `master` with the fix?

filipw:
thanks, yes that is correct. Every merge to master is published as beta prerelease and users who have `"omnisharp.path": "latest"` set in their VS Code receive that build then on next VS Code restart. This usually allows catching regressions early 🙂

DavidZidar:
@filipw I'm fairly sure I found the issue and opened a PR with a fix #2333

DaRosenberg:
@filipw Yep, that's the same improvement I had previously made in the `CSharpDiagnosticWorkerWithAnalyzers` when I started helping out on this PR. I wasn't aware that the `CSharpDiagnosticWorker` had been changed also, otherwise I would have done the same there..

Anyway, kudos for tracking it down and fixing! 👍🏻

DavidZidar:
@DaRosenberg Ah, and a good improvement it was. :) I reviewed the code at the time but failed to consider concurrency. And thanks for the initial investigation on how to reproduce the issue, it really helped!

PR #2291, opened 2021-12-03
Merge vNext into master

333fred:


JoeRobich (approved):

filipw:
This is the summary of what this change brings:

1. There is now a new .NET 6.0 build of OmniSharp published as build artifact. It gets published for the following RIDs
 - win7-x64
 - win7-x86
 - win10-arm64
 - osx-x64
 - osx-arm64
 - linux-x64

The previous .NET Framework targets for Windows and Mono based variants for MacOS and Linux (including arm64) are still all there and OmniSharp is cross compiled for them.

2. The .NET 6.0 build uses .NET SDK MSBuild only on *nix. On Windows it queries for VS instances as long as they come with SDK 6. In either case **.NET SDK 6.0.100 must be installed.**
3. When new .NET 6 SDK comes out, OmniSharp will not be able to handle it until it is upgraded to that version (this is caused by having runtime dependency on NuGet assemblies, which are typically revved up in SDK releases)
4. Both HTTP and STDIO servers have been upgraded to .NET 6.0.
5. The 6.0 build is a drop-in replacement for the old server. For example in VS Code it can be enabled by setting `omnisharp.path` to point to the `OmniSharp` executable from the given distro. For example:

```
{
    "omnisharp.useGlobalMono": "never",
    "omnisharp.path": "/Users/filipw/omnisharp-roslyn/artifacts/publish/OmniSharp.Stdio.Driver/osx-arm64/net6.0/OmniSharp"
}
```
6. On *nix .NET Framework is not supported. For that the global Mono variant is still needed as in the past.

filipw (approved):
:shipit:

PR #2300, opened 2021-12-08
Do not include msbuild folder in net6 packages

JoeRobich:
This is OK since the net6 builds will use the MSBuild included in the SDK.


filipw (approved):
thanks!

PR #2301, opened 2021-12-08
Don't build self-contained packages since we require the .NET SDK be installed.

JoeRobich:


filipw (approved):
thanks. are they much smaller now?

JoeRobich:
> thanks. are they much smaller now?

I wish I had checked at each step of my changes from this evening. The unzipped size is ~110MB for the ReadyToRun, non-self-contained, no-msbuild omnisharp. Compared to ~165MB, the size of 1.37.18-beta.97-net6.0. I think it is likely the savings from not publishing a self-contained build were used by enabling ReadyToRun.

PR #2302, opened 2021-12-08
Require a known good version of .NET SDK be installed.

JoeRobich:
Since we fail when loading SDKs with mismatched NuGet.Framework, we can limit our SDK discovery to a known good SDK version.

### DIFF
--- a/src/OmniSharp.Host/HostHelpers.cs
+++ b/src/OmniSharp.Host/HostHelpers.cs
@@ -1,5 +1,6 @@
 using System;
 using Microsoft.Extensions.Logging;
+using OmniSharp.MSBuild.Discovery;
 using OmniSharp.Roslyn;
 using OmniSharp.Utilities;
 
@@ -27,6 +28,11 @@ namespace OmniSharp
                 }
 
                 return action();
+            }
+            catch (MSBuildNotFoundException mnfe)
+            {
+                Console.Error.WriteLine(mnfe.Message);
+                return 0xbad;
             }
             catch (Exception e)
             {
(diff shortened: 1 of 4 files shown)


src/OmniSharp.Host/HostHelpers.cs
```diff
@@ -28,6 +29,11 @@ public static int Start(Func<int> action)
 
                 return action();
             }
+            catch (MSBuildNotFoundException mnfe)
+            {
+                Console.Error.WriteLine(mnfe.Message);
+                return 0xbad;
```
JoeRobich, line 35: Can make this a unique exit code.

src/OmniSharp.Host/MSBuild/Discovery/Extensions.cs
```diff
@@ -72,7 +72,7 @@ public static void RegisterDefaultInstance(this IMSBuildLocator msbuildLocator,
             }
             else
             {
-                logger.LogError("Could not locate MSBuild instance to register with OmniSharp");
+                throw new MSBuildNotFoundException("Could not locate MSBuild instance to register with OmniSharp.");
```
JoeRobich, line 75: Further loading isn't useful and clutters the output.

src/OmniSharp.Host/MSBuild/Discovery/Providers/MicrosoftBuildLocatorInstanceProvider.cs
```diff
@@ -22,9 +22,17 @@ public override ImmutableArray<MSBuildInstance> GetInstances()
         {
 
 #if NETCOREAPP
+            const string DotNetSdkVersion = "6.0.100";
```
JoeRobich, line 25: Should be easy to find and replace this version string.

filipw (approved):
yes I've suggested this yesterday  in a comment somewhere - thanks!

PR #2304, opened 2021-12-08
Allow for alternate versions of documents to be semantically highlighted

JoeRobich:
To fix diff highlighting in the VSCode C# extension, we need to pass along alternate versions of the document we are highlighting. Request has a `Buffer` property but populating that will invoke the updateBufferHandler when the request is handled (as alluded to in https://github.com/OmniSharp/omnisharp-vscode/pull/4915#discussion_r760604439).
https://github.com/OmniSharp/omnisharp-roslyn/blob/52f9ed3efbd2aa85afd3feb1433462474ad30fcd/src/OmniSharp.Host/Endpoint/EndpointHandler.cs#L109-L116

This PR adds an optional `VersionedText` property to the highlight request which can carry the alternate version of the document contents. The highlight service will then swap the document text for the duration of the request when it is populated.


filipw (approved):
thanks! 👏

PR #2306, opened 2021-12-09
Pass Completion, Rename and Block Structure options directly instead of updating the Workspace

tmat:
This approach will also apply to the remaining IWorkspaceOptionsProviders once necessary refactorings are complete in Roslyn.

- [x] Update Roslyn to the latest build that contains necessary changes to External Access APIs: https://github.com/dotnet/roslyn/pull/58355
- [x] Requires fix in https://github.com/dotnet/roslyn/pull/58381


tmat:
@JoeRobich @333fred PTAL

tmat:
FYI @jasonmalinowski

333fred (approved):

src/OmniSharp.Roslyn.CSharp/Services/Refactoring/RenameService.cs
```diff
@@ -41,9 +44,14 @@ public async Task<RenameResponse> Handle(RenameRequest request)
...
                 {
+                    var options = new OmniSharpRenameOptions(
+                        RenameOverloads: _omniSharpOptions.RenameOptions.RenameOverloads,
+                        RenameInStrings: _omniSharpOptions.RenameOptions.RenameInStrings,
+                        RenameInComments: _omniSharpOptions.RenameOptions.RenameInComments);
+
                     try
                     {
-                        solution = await Renamer.RenameSymbolAsync(solution, symbol, request.RenameTo, _workspace.Options);
+                        solution = await OmniSharpRenamer.RenameSymbolAsync(solution, symbol, request.RenameTo, options, nonConflictSymbols: null, CancellationToken.None);
```
jasonmalinowski, line 54: This seems particularly unfortunate that an API that was entirely public is now having to get external access versions for it -- I assume we have a plan to re-make these public in some form?

tmat: Possibly at some point...

333fred:
@tmat looks like the version you updated to doesn't have the updated types?

JoeRobich (approved):

333fred (approved):
LGTM, assuming tests are passing

jasonmalinowski:
Ah, this is now blocked on https://github.com/dotnet/roslyn/pull/58940

tmat:
@333fred @JoeRobich Any idea what might be wrong with the CI build? Seems like a loader exception. Perhaps some dependency of the new Roslyn build needs binding redirects or something?

tmat:
@jasonmalinowski Is O# not using VSMEF?

jasonmalinowski:
We could bump Omnisharp to use VS MEF, but given this is a regression we need to fix _anyways_ I'd say let's just wait for the fix.

jasonmalinowski:
The fix on the Roslyn side has been merged, so @tmat we just need to bump the Roslyn version in this PR (...again...).

build/Packages.props
```diff
@@ -7,7 +7,7 @@
         <MicrosoftTestPackageVersion>17.0.0</MicrosoftTestPackageVersion>
         <MSBuildPackageVersion>17.0.0</MSBuildPackageVersion>
         <NuGetPackageVersion>6.0.0</NuGetPackageVersion>
-        <RoslynPackageVersion>4.1.0-3.22069.8</RoslynPackageVersion>
+        <RoslynPackageVersion>4.2.0-1.22074.8</RoslynPackageVersion
```
JoeRobich, line 10:
```suggestion
        <RoslynPackageVersion>4.2.0-1.22074.8</RoslynPackageVersion>
```

PR #2308, opened 2021-12-14
Allow net6 build of O# to load newer .NET SDKs

JoeRobich:
By not shipping NuGet libraries ourselves in our net6 packages, we will load the NuGet libraries shipped with the SDK registered by the MSBuildLocator. I did need to add a SemanticVersion implementation in order to remove our use of NuGet.Versioning.

To test you can install 6.0.2xx version of the .NET SDK from [dotnet/installers](https://github.com/dotnet/installer).

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -919,51 +919,51 @@ Task("PublishNet6Builds")
         {
             if (!Platform.Current.IsWindows)
             {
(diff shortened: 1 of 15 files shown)


filipw:
This is a spectacular solution and I love how simple it turned out to be!
Do you want to include it in this release or in the next one?

filipw (approved):

JoeRobich:
This release would be great.

PR #2312, opened 2021-12-19
Adds parallel execution of background Roslyn analyzers

### LINKED ISSUE #2241: Analyzing projects is extremely slow compared to MSBuild or Visual Studio

DaRosenberg:
Fixes #2241 

This PR builds upon the work started by @TomasEkeli in #2285 and adds some further improvements.

The original PR:

- uses one semaphore for foreground worker and another for background worker, which could potentially use twice the configured threads when both workers are active
- parallelizes only at the *project* level. This can yield a nice perf boost when there are many projects to analyze, but will be less effective when there are fewer projects (but potentially containing many files)
- keeps the same diagnostics events (i.e. project start/stop) which, as visualized in a client like VS Code with just one status bar message at a time, has no real meaning anymore since multiple projects are now actually analyzed simultaneously

This PR addresses the above points by:

- Parallelizing both at the project level (with no throttling) **and** at the document level (with throttling) to ensure the intended concurrency in both scenarios (many projects vs. few projects) while still respecting the global configured limit
- Using a member-level semaphore for the throttling mechanism
- Adding a new `BackgroundDiagnosticStatus` event (and associated model types) in addition to the existing `ProjectDiagnosticStatus` event. The new event is modeled around total and remaining number of documents currently being analyzed, instead of the notion of current project. An update event is emitted every 24 documents (results in reasonable frequency on most machines, could be made time-based as an improvement) allowing clients to display meaningful progress. Code for handling the new status event can eventually be added to `omnisharp-vscode` and other clients.
- Retaining the old status events for backward compat with clients, "shimmed" - displays as `Analyzing (23%)`

This PR also parallelizes slightly more aggressively by defaulting to 75% of available processors (rather than 50%).


DaRosenberg:
The improvements in this PR have been merged into the original PR #2285 - closing this one.

PR #2314, opened 2021-12-21
Add MSBuild project to solution and apply the change to Roslyn workspace as a unit

dmgonch:
I often hit 'Failed to add project to workspace: <project-file>' error in OmniSharp log when opening one of internal projects. I only see this issue on Linux (Ubuntu 20.04) and not on Windows. I always have omnisharp.enableMsBuildLoadProjectsOnDemand set to "true". The problem occurs often but not all the time.

In both success and failure cases I see the following sequence of events (shortened to highlight only important parts):

```
[info]: OmniSharp.MSBuild.ProjectManager
	Queue project update for '/home/<path_to_project>/ProjectName.csproj'
[info]: OmniSharp.MSBuild.ProjectManager
	Loading project: /home/<path_to_project>/ProjectName.csproj
[dbug]: OmniSharp.Stdio.Host
        ************ Request ************
{
  "Type": "request",
  "Seq": 7,
  "Command": "/filesChanged",
  "Arguments": [
    {
      "FileName": "/home/<path_to_project>/obj/Debug/net6.0/linux-x64/ProjectName.AssemblyInfo.cs",
      "changeType": "Change"
    }
  ]
}
[dbug]: OmniSharp.Roslyn.BufferManager
        Adding transient file for /home/<path_to_project>/obj/Debug/net6.0/linux-x64/ProjectName.AssemblyInfo.cs
[info]: OmniSharp.MSBuild.ProjectManager
        Successfully loaded project file '/home/<path_to_project>/ProjectName.csproj'.
[info]: OmniSharp.MSBuild.ProjectManager
        Adding project '/home/<path_to_project>/ProjectName.csproj'
```

In the case of the failure though the sequence is followed by:
	
```
[fail]: OmniSharp.MSBuild.ProjectManager
	Failed to add project to workspace: '/home/<path_to_project>/ProjectName.csproj'
...
[fail]: OmniSharp.MSBuild.ProjectManager
	Could not locate project in workspace: /home/<path_to_project>/ProjectName.csproj
```

The documentation for "bool Workspace.TryApplyChanges(Solution newSolution)" states: “… The specified solution must be one that originated from this workspace. If it is not, or the workspace has been updated since the solution was obtained from the workspace, then this method returns false …"

ProjectManager.AddProject clearly passes the solution that is originated from the same workspace. Therefore, the workspace somehow gets changed b/w adding the project to the solution and applying the changes to the workspace. The theory is that the file change notification received for ProjectName.AssemblyInfo.cs causes the workspace change in ProjectManager.OnDirectoryFileChanged. Thus, the lock added around these two places. My testing shows that the change does address the issue I'm seeing.

### DIFF
--- a/src/OmniSharp.MSBuild/ProjectManager.cs
+++ b/src/OmniSharp.MSBuild/ProjectManager.cs
@@ -60,6 +60,7 @@ namespace OmniSharp.MSBuild
         private readonly ConcurrentDictionary<string, int/*unused*/> _projectsRequestedOnDemand;
         private readonly ProjectLoader _projectLoader;
         private readonly OmniSharpWorkspace _workspace;
+        private readonly object _workspaceGate = new();
         private readonly ImmutableArray<IMSBuildEventSink> _eventSinks;
         private const int LoopDelay = 100; // milliseconds
         private readonly BufferBlock<ProjectToUpdate> _queue;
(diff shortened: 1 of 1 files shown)


dmgonch:
@filipw, @JoeRobich, appreciate if you could take a look.

JoeRobich (approved):
Thanks!

PR #2319, opened 2021-12-31
Handle completions with trailing whitespace on previous lines

333fred:
LSP filters expect that the range of the main text edit is going to start on the same line as the cursor. However, when trailing whitespace preceeds the override marker, Roslyn will return a single text edit for the whole range that trims that whitespace. This confuses the vscode intellisense engine, which then unhelpfully displays no completions with no user feedback that this filtering is happening. To avoid this, we now detect when this case occurs, and split up the change into non-overlapping edits from before the current line. Fixes https://github.com/OmniSharp/omnisharp-vscode/issues/4600.

### DIFF
--- a/src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionListBuilder_Sync.cs
+++ b/src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionListBuilder_Sync.cs
@@ -159,7 +159,7 @@ namespace OmniSharp.Roslyn.CSharp.Services.Completion
             {
                 if (!textChange.Span.IntersectsWith(position))
                 {
-                    HandleNonIntersectingEdit(sourceText, ref additionalTextEdits, ref adjustedNewPosition, textChange);
+                    handleNonInsertsectingEdit(sourceText, ref additionalTextEdits, ref adjustedNewPosition, textChange);
                 }
                 else
                 {
@@ -204,7 +204,7 @@ namespace OmniSharp.Roslyn.CSharp.Services.Completion
 
                         // Now that we've found the cuttoff, we can build our two subchanges
                         var prefixChange = new TextChange(new TextSpan(changeSpanStart, length: lineStartPosition - changeSpanStart), textChange.NewText!.Substring(0, cutoffPosition));
-                        HandleNonIntersectingEdit(sourceText, ref additionalTextEdits, ref adjustedNewPosition, prefixChange);
+                        handleNonInsertsectingEdit(sourceText, ref additionalTextEdits, ref adjustedNewPosition, prefixChange);
                         updatedChange = new TextChange(new TextSpan(lineStartPosition, length: textChange.Span.End - lineStartPosition), textChange.NewText.Substring(cutoffPosition));
                     }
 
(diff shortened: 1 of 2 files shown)


333fred:
https://github.com/microsoft/vscode/issues/41022 would have been very useful for helping debug this issue.

tests/OmniSharp.Roslyn.CSharp.Tests/CompletionFacts.cs
```diff
@@ -2220,6 +2220,39 @@ public static void Main()
             }
         }
 
+        [Theory]
+        [InlineData("dummy.cs")]
+        [InlineData("dummy.csx")]
+        public async Task TestOverrideWithTrailingWhitespacePrior(string filename)
+        {
+            const string input = @"
+namespace OmniSharp.Cake
```
filipw, line 2229: would you mind changing this to a generic Foo or something of that sort like the rest of the tests? someone in the future will surely spend time wondering what is Cake specific about this (in fact, that was my first automatic reaction too)

src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionListBuilder_Sync.cs
```diff
@@ -148,38 +149,67 @@ internal static partial class CompletionListBuilder
...
+                        (adjustedNewPosition.Value <= changeSpanStart + textChange.NewText!.Length) &&
+                        (adjustedNewPosition.Value >= changeSpanStart));
+
+                    // Filtering needs a range that is a _single_ line. Consider a case like this (whitespace documented with escapes):
+                    //
+                    // 1: class C
+                    // 2: {\t\r\n
+                    // 3:    override $$
+                    //
+                    // Roslyn will see the trailing \t on line 2 and remove it when creating the _main_ text change. However, that will
```
filipw, line 179: very interesting

src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionListBuilder_Sync.cs
```diff
@@ -236,6 +266,24 @@ internal static partial class CompletionListBuilder
 
                 return ($"{beforeText}$0{afterText}", InsertTextFormat.Snippet);
             }
+
+            static void HandleNonIntersectingEdit(SourceText sourceText, ref List<LinePositionSpanTextChange>? additionalTextEdits, ref int? adjustedNewPosition, TextChange textChange)
```
filipw, line 270: I don't think you need `ref` on the list?

filipw: another thing that jumps out to me, is that it doesn't seem like we are consistent here. In one place we have `getPossiblySnippetizedInsertText` which returns a tuple and doesn't modify references, the locals get reassigned based on what was returned. 
and the new one `HandleNonIntersectingEdit` returns void and operates on references. I think it would be good to align either way. 

also, one is pascal cased the other is camel cased

333fred: I'll rename the function, but I don't think that any other changes are necessary. The `ref` is necessary because the edits list gets initialized if it's not already initialized. The local function is also different from `getPossiblySnippetizedInsertText`, and I think that's fine. One of them is a static input/output function, the other is a handler that has side effects. I don't think it's worth it try and make this function an input/output: that would make a good chunk of the code sharing between the two locations not possible.

filipw: my physical brain prefers to think without side effects `f : x -> y` but OK 😄

filipw (approved):

PR #2343, opened 2022-02-11
Eliminate more instances of IWorkspaceOptionsProvider

tmat:
All options are now passed explicitly to the APIs that need them, except for default values for editorconfig-based formatting options. These will be eliminated later on.

Requires Roslyn changes: https://github.com/dotnet/roslyn/pull/59469


tmat:
@JoeRobich @[ryzngard](https://github.com/ryzngard)

src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionService.cs
```diff
@@ -119,7 +119,7 @@ CompletionListBuilder.PartialMethodCompletionProvider or
             // that completion provider is still creating the cache. We'll mark this completion list as not completed, and the
             // editor will ask again when the user types more. By then, hopefully the cache will have populated and we can mark
             // the completion as done.
-            bool expectingImportedItems = expandedItemsAvailable && _omniSharpOptions.RoslynExtensionsOptions.EnableImportCompletion;
+            bool expectingImportedItems = options.ShowItemsFromUnimportedNamespaces;
```
333fred, line 122: This change isn't correct, and is the source of the test failures. We need to only mark this as true when the cache is complete and returning full info, which is what the `expandedItemsAvailable` variable was marking. Seems like with @genlu's import change, this may need to be reworked.

genlu: I will add a flag for partial results from import completion providers

genlu: Fix PR: https://github.com/dotnet/roslyn/pull/59958
Once this change is merged, O# need to set the ForceExpandedCompletionIndexCreation option to true before calling CompletionService to make test behavior deterministic

JoeRobich:
@333fred @tmat I updated with a build of Roslyn that includes Gen's change. Feedback welcome on https://github.com/OmniSharp/omnisharp-roslyn/pull/2343/commits/43ca31876ec9266f83854b29d4a14245f0e5f3bc

tests/OmniSharp.Roslyn.CSharp.Tests/CompletionFacts.cs
```diff
@@ -175,7 +175,7 @@ public Class1()
             {
                 while (completions.IsIncomplete)
                 {
-                    completions = await FindCompletionsAsync(filename, input, host);
+                    completions = await FindCompletionsAsync(filename, input, host, forceExpandedCompletionIndexCreation: false);
```
333fred, line 178: We could likely avoid this entire loop if we just set the option to true.

JoeRobich: I ended up doing that for `FindCompletionsWithImportedAsync()`. Will let this test remain our test that the index will eventually be created.

JoeRobich (approved):

src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionService.cs
333fred: This condition is still incorrect. We need to know if we got partial results back from the `GetCompletionsAsync`, or `IsIncomplete` will not be correct.

genlu: I was just discussing this with @JoeRobich offline. You are right that the `IsIncomplete` flag will not be correct, but you couldn't decide if the result is complete with the removed "expandedItemsAvailable" flag either. That flag only means import completion would trigger in current context, it doesn't mean the result the provider returns is complete or any results would be returned at all. 

The behavior for unimported types was "all or nothing" before (depends on cache state), which might gives you an impression that the existence of unimported type is an indication of complete results, but it's just an implementation detail and was changed in [this PR](https://github.com/dotnet/roslyn/pull/59125) to return partial results, which is consistent with how unimported extension methods works.

If you'd like to support `IsIncomplete` flag, maybe take a look at the approach used in [this PR](https://github.com/dotnet/roslyn/pull/59045)

genlu (approved):
Completion part LGTM

src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionService.cs
333fred: Ok. I'll try to look at that at a later date, we don't need to block this PR on it. At worst, the first few completion sessions will be inconsistent, but following sessions will be fine.

333fred (approved):

PR #2347, opened 2022-02-17
Update versions from 6.0.200

nohwnd:


nohwnd:
Saw it build locally. Will add release notes here tomorrow.

filipw (approved):

nohwnd:
@filipw please give the change log a quick look. I removed one item that looked very internal (update deps for cake), rest looks like stuff that people would be interested to know.

filipw (approved):
thanks!

PR #2357, opened 2022-03-16
Support inlay hints

333fred:
I've based the O# api on the proposed inlay hint LSP api, https://github.com/microsoft/vscode-languageserver-node/blob/main/protocol/src/common/protocol.inlayHint.ts, which is pretty much stable now and 3.17 should be out shortly with this shape. I've also done a couple of refactors while touching code:

* Update to C# 10.
* Add some additional assert utils from roslyn, to make test results more readable and actionable.
* Add an interpolated string handler for logging, so that interpolated strings can be used in logging without a perf penalty when they're uneeded.

### DIFF
--- a/src/OmniSharp.Roslyn.CSharp/Services/InlayHints/InlayHintService.cs
+++ b/src/OmniSharp.Roslyn.CSharp/Services/InlayHints/InlayHintService.cs
@@ -42,6 +42,7 @@ internal class InlayHintService :
         _omniSharpOptions = omniSharpOptions;
         _cache = new(_logger);
     }
+
     public async Task<InlayHintResponse> Handle(InlayHintRequest request)
     {
         var document = _workspace.GetDocument(request.Location.FileName);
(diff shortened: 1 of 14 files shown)


src/OmniSharp.Roslyn.CSharp/Services/InlayHints/InlayHintService.cs
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    public async Task<InlayHint> Handle(InlayHintResolveRequest request)
+    {
+        if (!_cache.TryGetFromCache(request.Hint, out var roslynHint, out var document))
+        {
+            return request.Hint;
+        }
+
+        var descriptionTags = await roslynHint.GetDescrptionAsync(document, CancellationToken.None);
```
JoeRobich, line 97: Opened https://github.com/dotnet/roslyn/pull/60217 for this typo

JoeRobich (approved):
Looks good to me

filipw (approved):
very cool

333fred:
I need to adjust one test to account for platform differences (`int` is a `readonly struct` on some platforms, and that affects the ToString() behavior).

PR #2359, opened 2022-03-17
Delete System.Configuration.ConfigurationManager from deployed packages

333fred:
The SDK ships a newer version of System.Configuration.ConfigurationManager than we do, and it's causing load issues. This is a quick patch to fix https://github.com/OmniSharp/omnisharp-vscode/issues/5113 for now, but we'll want to remove all dlls that the SDK ships to avoid this long-term.


JoeRobich (approved):

filipw (approved):
thanks this will fix the prerelease so let's merge asap

Issue #2339, opened 2022-02-02
Remove included MSBuild tools and Mono


JoeRobich:
**Background:**

The included Mono and MSBuild tools were added to support easy development of C# using .NET Core. We have frequently updated these build tools to match the current shipping SDK tooling. 

On Windows we did not need to ship a copy of Mono because we could rely on the .NET Framework being installed. For MacOS and Linux we shipped a minimal Mono that would support running MSBuild and OmniSharp. With these tools, the .NET SDK installed, and the dotnet CLI discoverable from the environment PATH, developers would have good support when developing against the newer SDK-style projects. 

There was some support for Full Framework projects with this configuration, but that was considered a bonus. Our most common Full Framework projects (Unity) were unsupported with this configuration because the .NET 4.7.1 targeting pack was not included.

**Proposal:**

The included MSBuild and Mono should be removed. 

We now ship a .NET 6 build of OmniSharp which runs against the build tools in the .NET SDK. The .NET 6 build, with an install of the .NET 6 SDK and the dotnet CLI discoverable from the environment PATH, covers all the scenarios that the included MSBuild Tools and Mono were intended to cover. Users working with Full Framework projects typically install the VS MSBuild Tools or a full Mono to get support for additional project types. 

Everyone would benefit from reduced download size and OmniSharp would benefit from the reduced maintenance burden.

PR #2360, opened 2022-03-17
Remove MSBuild and Mono from release packages

JoeRobich:
Closes https://github.com/OmniSharp/omnisharp-roslyn/issues/2339


JoeRobich:
Getting mixed signals from the Mono test failures.

```
                Located 1 MSBuild instance(s)
                            1: Mono 16.10.1 - "/Library/Frameworks/Mono.framework/Versions/6.12.0/lib/mono/msbuild/Current/bin"
        [dbug]: OmniSharp.CompositionHostBuilder
                MSBuild instance Mono 16.10.1 scored at 2
                Registered MSBuild instance: Mono 16.10.1 - "/Library/Frameworks/Mono.framework/Versions/6.12.0/lib/mono/msbuild/Current/bin"
                Loaded Microsoft.Build from '/Library/Frameworks/Mono.framework/Versions/6.12.0/lib/mono/msbuild/Current/bin/Microsoft.Build.dll'
                Loaded Microsoft.Build.Framework from '/Library/Frameworks/Mono.framework/Versions/6.12.0/lib/mono/msbuild/Current/bin/Microsoft.Build.Framework.dll'
                Loaded Microsoft.Build.Tasks.Core from '/Library/Frameworks/Mono.framework/Versions/6.12.0/lib/mono/msbuild/Current/bin/Microsoft.Build.Tasks.Core.dll'
                Loaded Microsoft.Build.Utilities.Core from '/Library/Frameworks/Mono.framework/Versions/6.12.0/lib/mono/msbuild/Current/bin/Microsoft.Build.Utilities.Core.dll'
```

but then the exception is:

```
        System.TypeLoadException: Could not load file or assembly 'Microsoft.Build.Utilities.Core, Version=15.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies.
          at OmniSharp.MSBuild.ProjectFile.ProjectFileInfo.Load (System.String filePath, OmniSharp.MSBuild.ProjectIdInfo projectIdInfo, OmniSharp.MSBuild.ProjectLoader loader, System.Guid sessionId, OmniSharp.Services.DotNetInfo dotNetInfo) [0x00023] in <9e48cc6cdb05429387c2464a8132425b>:0 
          at OmniSharp.MSBuild.ProjectManager+<>c__DisplayClass32_0.<LoadProject>b__0 () [0x00000] in <9e48cc6cdb05429387c2464a8132425b>:0 
          at (wrapper delegate-invoke) System.Func`1[System.ValueTuple`3[OmniSharp.MSBuild.ProjectFile.ProjectFileInfo,System.Collections.Immutable.ImmutableArray`1[OmniSharp.MSBuild.Logging.MSBuildDiagnostic],OmniSharp.MSBuild.Notification.ProjectLoadedEventArgs]].invoke_TResult()
          at OmniSharp.MSBuild.ProjectManager.LoadOrReloadProject (System.String projectFilePath, System.Func`1[TResult] loader) [0x0001b] in <9e48cc6cdb05429387c2464a8132425b>:0 
```

JoeRobich:
I believe I will need to learn more about this comment https://github.com/OmniSharp/omnisharp-roslyn/pull/2360/files#diff-c6eace5a70f0c2d920dfa97e988d4cad3d0131dd2b029bb2646c96222e866cf8L710-L711

filipw:
At the moment the "pending" checks are appearing here because they are pinned by name in the branch protection rules, and since the names now changed due to version bumps, they appear as not-satisfied.

I do not believe there is a way to selectively apply different rule to this PR than it is to the other ones, so what I will suggest is that once we are ready to merge this, we merge using admin permissions without requirements met, and then adjust the branch protection rules once it's merged, as then all subsequent PRs will be built according to new task names.

src/OmniSharp.Host/MSBuild/Discovery/MSBuildLocator.cs
```diff
@@ -34,33 +34,15 @@ protected override void DisposeCore(bool disposing)
 
         public static MSBuildLocator CreateDefault(ILoggerFactory loggerFactory, IAssemblyLoader assemblyLoader, IConfiguration msbuildConfiguration)
         {
-            var useBundledOnly = msbuildConfiguration.GetValue<bool>("UseBundledOnly");
```
filipw, line 37: we could still look at this value and if we detect it to be `true`, we print a warning into the log saying that this is not supported anymore?

the way the configuration works, it does not even need to be part of the typed model anymore for us to be able to extract it manually here (like it is already done)

filipw:
I think this looks great and removes a ton of complexities. I would just wait for us to successfully resolve the Nuget key issue and publish 1.38.2 before this is merged

src/OmniSharp.Host/MSBuild/Discovery/MSBuildLocator.cs
```diff
@@ -34,33 +34,15 @@ protected override void DisposeCore(bool disposing)
...
-            if (useBundledOnly)
-            {
-                var logger = loggerFactory.CreateLogger<MSBuildLocator>();
-                logger.LogInformation("Because 'UseBundledOnly' is enabled in the configuration, OmniSharp will only use the bundled MSBuild.");
-                return CreateStandAlone(loggerFactory, assemblyLoader);
-            }
-
             return new MSBuildLocator(loggerFactory, assemblyLoader,
```
JoeRobich, line 37:
```suggestion
            var useBundledOnly = msbuildConfiguration.GetValue<bool>("UseBundledOnly");
            if (useBundledOnly)
            {
                logger.LogWarning("The MSBuild option 'UseBundledOnly' is no longer supported. Please update your OmniSharp configuration files.");
            }

            return new MSBuildLocator(loggerFactory, assemblyLoader,
```

src/OmniSharp.Host/MSBuild/Discovery/MSBuildLocator.cs
JoeRobich: Good idea. Updated with a warning.

filipw (approved):
looks great!

PR #2373, opened 2022-03-30
provide constructor accepting hostServices

stg609:
aggregator.CreateHostServices is a time-consuming job, which can be handled outside of the omnisharp.


filipw (approved):
thanks

PR #2377, opened 2022-04-04
Support .NET 7 SDKs

JoeRobich:
Adds new SdkOptions allowing users to specify a SDK path, a SDK version, and the ability to include prerelease SDKs during MSBuild locating.

In order to discover .NET 7.0 preview SDKs our hosts need to run on .NET 7.0 runtime
- Tests now build against net7.0 so that the test runner will run them on the 7.0 runtime.
- O# hosts still target net6.0 but now rollforward in a way that includes .NET 7.0 preview runtimes.


tests/OmniSharp.Roslyn.CSharp.Tests/FindSymbolsFacts.cs
```diff
@@ -161,7 +161,7 @@ public async Task Returns_struct_kind(string filename)
 
             var usages = await FindSymbolsAsync(code, filename);
             var symbols = usages.QuickFixes.Cast<SymbolLocation>().Select(q => q.Kind);
-            Assert.Equal("Structure", symbols.First());
+            Assert.Contains(symbols.First(), new[] { "Struct", "Structure" });
```
JoeRobich, line 164: In the TypeKind enum both `Struct` and `Structure` have the same value. A runtime change must have effected the order they are returned. Either is valid since they both represent the value `10`.

filipw (approved):
thanks!

PR #2376, opened 2022-04-03
Fixing #2372 (Using sumary instead of xmldoc on signature requests)

AnonymusRaccoon:
This pr fixes #2372.

It uses the xml summary instead of the whole xmldoc if a summary exists, if it does not, it falls back to the whole documentation.

image

PR #2392, opened 2022-04-27
Use human readable doc in lsp's signature help

### LINKED ISSUE #2372: Signature Help Documentation is the xml and not human-readable
The [signature help](https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp.Roslyn.CSharp/Services/Signatures/SignatureHelpService.cs#L171) uses the xml data and not the summary tag for the documentation of the signature. The specification wants plain text or markdown instead. 

Plain text is already used for parameters documentation, it should be the same everywhere. This issue was already raised during the implementation of Structured [here](https://github.com/OmniSharp/omnisharp-roslyn/pull/1085/files#r161122095), but it was not modified (I don't know why).

I confirmed that [VS Code](https://github.com/OmniSharp/omnisharp-vscode/blob/d1785942e51f3afee1d8e62b406ee46b35168c63/src/features/signatureHelpProvider.ts#L32) is not using the documentation field so changing it would not break their implementation, but I don't know about other editors, so should I PR or not?

For now on editors that don't do a custom thing for Roslyn, this is shown instead:
image

AnonymusRaccoon:
Fixes #2372

Continuation of #2376, but for the lsp mode only. Following what @JoeRobich said on the issue.

### DIFF
--- a/src/OmniSharp.LanguageServerProtocol/Handlers/OmniSharpSignatureHelpHandler.cs
+++ b/src/OmniSharp.LanguageServerProtocol/Handlers/OmniSharpSignatureHelpHandler.cs
@@ -49,7 +49,7 @@ namespace OmniSharp.LanguageServerProtocol.Handlers
 
             var containerSignatures = omnisharpResponse.Signatures.Select(x => new SignatureInformation
             {
-                Documentation = x.Documentation,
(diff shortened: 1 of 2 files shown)


JoeRobich (approved):

JoeRobich:
Thanks!

Issue #2366, opened 2022-03-19
No `linux-musl-$(Platform)` release


ayakael:
Tests on https://github.com/redhat-developer/dotnet-regular-tests pull Omnisharp release from here for testing. In working on https://github.com/redhat-developer/dotnet-regular-tests/pull/192 to adapt the testing suite to Alpine, it fails due to no build of for `linux-musl`

PR #2395, opened 2022-05-04
added support for linux-musl-x64 and linux-musl-arm64

filipw:
For lightweight distributions using [musl](https://wiki.musl-libc.org/projects-using-musl.html) like Alpine Linux.
Fixes https://github.com/OmniSharp/omnisharp-roslyn/issues/2366

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -42,7 +42,7 @@ Note: The arguments below should prefixed with a single hyphen on Windows (Power
   `-install-path Path`: Path used for the **Install** target.
     Defaults to `(%USERPROFILE%|$HOME)/.omnisharp`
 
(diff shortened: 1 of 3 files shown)


bjorkstromm (approved):

JoeRobich:
Looks like [VS Code supports](https://code.visualstudio.com/api/working-with-extensions/publishing-extension#platformspecific-extensions) extensions targeting alpine-x64 and alpine-arm64. Should we be proactive and also target linux-musl-arm64?

filipw:
That makes sense. I did realize this is supported, because it was missing from the [RID catalog docs](https://docs.microsoft.com/en-us/dotnet/core/rid-catalog) but it is indeed visible [here](https://docs.microsoft.com/en-us/dotnet/core/rid-catalog).

I added it now.

JoeRobich (approved):

Issue #2421, opened 2022-07-25
Semantic Highlighting Failures


333fred:
Open the dotnet/razor-compiler repo, and open a file in the main compiler project, like `src/Microsoft.AspNetCore.Razor.Language/src/Syntax/GreenNode.cs`. You'll see the following stack:

```
System.InvalidOperationException: Sequence contains more than one element
   at System.Linq.ThrowHelper.ThrowMoreThanOneElementException()
   at System.Linq.Enumerable.TryGetSingle[TSource](IEnumerable`1 source, Boolean& found)
   at OmniSharp.Roslyn.CSharp.Services.SemanticHighlight.SemanticHighlightService.CreateSemanticSpan(IEnumerable`1 results, TextLineCollection lines) in /Users/runner/work/1/s/src/OmniSharp.Roslyn.CSharp/Services/SemanticHighlight/SemanticHighlightService.cs:line 85
   at OmniSharp.Roslyn.CSharp.Services.SemanticHighlight.SemanticHighlightService.<>c.<Handle>b__3_2(IGrouping`2 grouping) in /Users/runner/work/1/s/src/OmniSharp.Roslyn.CSharp/Services/SemanticHighlight/SemanticHighlightService.cs:line 75
   at System.Linq.Enumerable.SelectEnumerableIterator`2.ToArray()
   at OmniSharp.Roslyn.CSharp.Services.SemanticHighlight.SemanticHighlightService.Handle(SemanticHighlightRequest request) in /Users/runner/work/1/s/src/OmniSharp.Roslyn.CSharp/Services/SemanticHighlight/SemanticHighlightService.cs:line 78
   at OmniSharp.Endpoint.EndpointHandler`2.GetFirstNotEmptyResponseFromHandlers(ExportHandler`2[] handlers, TRequest request) in /Users/runner/work/1/s/src/OmniSharp.Host/Endpoint/EndpointHandler.cs:line 217
   at OmniSharp.Endpoint.EndpointHandler`2.HandleRequestForLanguage(String language, TRequest request, RequestPacket packet) in /Users/runner/work/1/s/src/OmniSharp.Host/Endpoint/EndpointHandler.cs:line 233
   at OmniSharp.Endpoint.EndpointHandler`2.Process(RequestPacket packet, LanguageModel model, JToken requestObject) in /Users/runner/work/1/s/src/OmniSharp.Host/Endpoint/EndpointHandler.cs:line 142
[...]

Issue #2050, opened 2020-12-23
SemanticHighlighing Exception in dotnet/roslyn/...ArrayBuilder.cs


333fred:
Repro steps:

1. Open Compilers.sln from dotnet/roslyn
2. Open `src/Dependences/PooledObjects/ArrayBuilder.cs`

Expected: Semantic highlighting

Actual:
```
System.InvalidOperationException: Sequence contains more than one element\
  at System.Linq.Enumerable.Single[TSource] (System.Collections.Generic.IEnumerable`1[T] source) [0x00070] in <9ba54d07696a449db4a4279fc05aa435>:0 \
  at OmniSharp.Roslyn.CSharp.Services.SemanticHighlight.SemanticHighlightService.CreateSemanticSpan (System.Collections.Generic.IEnumerable`1[T] results, Microsoft.CodeAnalysis.Text.TextLineCollection lines) [0x00058] in <88f2819c12ee464e94d6700e849310cf>:0 \
  at OmniSharp.Roslyn.CSharp.Services.SemanticHighlight.SemanticHighlightService+<>c.<Handle>b__1_2 (System.Linq.IGrouping`2[TKey,TElement] grouping) [0x0000c] in <88f2819c12ee464e94d6700e849310cf>:0 \
  at System.Linq.Enumerable+SelectEnumerableIterator`2[TSource,TResult].ToArray () [0x0001d] in <9ba54d07696a449db4a4279fc05aa435>:0 \
  at System.Linq.Enumerable.ToArray[TSource] (System.Collections.Generic.IEnumerable`1[T] source) [0x0001f] in <9ba54d07696a449db4a4279fc05aa435>:0 \
  at OmniSharp.Roslyn.CSharp.Services.SemanticHighlight.SemanticHighlightService.Handle (OmniSharp.Models.SemanticHighlight.SemanticHighlightRequest request) [0x002fa] in <88f2819c12ee464e94d6700e849310cf>:0 \
  at OmniSharp.Endpoint.EndpointHandler`2[TRequest,TResponse].GetFirstNotEmptyResponseFromHandlers (OmniSharp.Endpoint.Exports.ExportHandler`2[TRequest,TResponse][] handlers, TRequest request) [0x00099] in <954ed0cf974c4f038a543c1ed1dfaa98>:0 \
[...]

PR #2424, opened 2022-07-27
Only get first document's highlights

333fred:
For files that are included in multiple projects (such as via `Compile Include` directives), we currently attempt to retrieve semantic highlights for every document. However, this means that highlight spans will have overlapping results from every project, and the service will crash. Further, we have no way of actually selecting what project to display the file as, so retrieving any but the first document doesn't serve the user. Therefore, we remove the loop, and just look through the first document found.

Fixes https://github.com/OmniSharp/omnisharp-roslyn/issues/2421.
Fixes https://github.com/OmniSharp/omnisharp-roslyn/issues/2050.

### DIFF
--- a/src/OmniSharp.Roslyn.CSharp/Services/SemanticHighlight/SemanticHighlightService.cs
+++ b/src/OmniSharp.Roslyn.CSharp/Services/SemanticHighlight/SemanticHighlightService.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Generic;
 using System.Composition;
 using System.Linq;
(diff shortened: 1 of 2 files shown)


333fred:
Note: I would highly recommend reviewing without whitespace changes, the diff is significantly smaller.

filipw (approved):

filipw:
thank you!

333fred:
I'm going to force merge this. @filipw @JoeRobich, we need to update our mac leg, as 10.15 is deprecated, and they are intentionally starting to fail jobs. https://github.com/actions/virtual-environments/issues/5583

JoeRobich (commented):
👍

PR #2431, opened 2022-08-12
Provide SourceGeneratedFileInfo for workspace symbols requests

333fred:
Updates the workspace symbols requests to provide a SourceGeneratedFileInfo when the symbol is from a source generated file, so aware editors can retrieve information about the file and display it correctly, rather than showing no information.

### DIFF
--- a/src/OmniSharp.Roslyn/Extensions/SolutionExtensions.cs
+++ b/src/OmniSharp.Roslyn/Extensions/SolutionExtensions.cs
@@ -1,10 +1,14 @@
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.FindSymbols;
 using OmniSharp.Models;
+using OmniSharp.Models.v1.SourceGeneratedFile;
+
+#nullable enable
 
 namespace OmniSharp.Extensions
 {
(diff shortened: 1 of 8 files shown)


333fred:
@JoeRobich @filipw for review.

src/OmniSharp.Roslyn/Extensions/SolutionExtensions.cs
```diff
@@ -86,7 +90,25 @@ private static QuickFix ConvertSymbol(Solution solution, ISymbol symbol, Locatio
...
                 EndLine = lineSpan.EndLinePosition.Line,
                 EndColumn = lineSpan.EndLinePosition.Character,
-                Projects = projects
+                Projects = projects,
+                ContainingSymbolName = symbol.ContainingSymbol?.Name ?? "",
+                GeneratedFileInfo = GetSourceGeneratedFileInfo(solution, location),
+            };
+        }
+
+        internal static SourceGeneratedFileInfo? GetSourceGeneratedFileInfo(Solution solution, Location location)
```
333fred, line 99: Copy/paste from `GoToDefinitionHelpers`, and modified to take a `Solution` instead of an `OmniSharpWorkspace`.

filipw: I believe this was intended to be the home for extension methods, so maybe it would be good to make this one too for consistency

filipw (approved):
this all makes sense to me. would it possible to add a test case too?

333fred:
@filipw any other comments?

filipw (approved):
thanks!

PR #2434, opened 2022-08-13
Return generated file info for find references

333fred:
Allows find references to provide information about references in source-generated files.


333fred:
@JoeRobich @filipw for review.

JoeRobich (approved):

PR #2436, opened 2022-08-14
Host dependency cleanup

333fred:
Today, we delete a few dlls in the cake packaging scripts, to remove DLLs that should come from the runtime we're loaded with, rather than the dlls we built with. This is fine for packaging, but is annoying when doing local tests by just rebuilding the relevant driver, because you have to remember to go into the output directly and delete them every time. I addressed this by adding explicit packagereferences to the dlls in question and setting them to not keep any assets.


Directory.Build.targets
```diff
@@ -1,6 +1,20 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
+  <ItemGroup Condition="'$(SkipHostDlls)' == 'true'">
```
333fred, line 4: Still need to only do this if we're building for .NET 6.

src/OmniSharp.Http.Driver/OmniSharp.Http.Driver.csproj
```diff
@@ -11,6 +11,7 @@
         <!-- Always run on the latest runtime installed. -->
         <RuntimeFrameworkVersion>6.0.0-preview.7.21317.1</RuntimeFrameworkVersion>
         <RollForward>LatestMajor</RollForward>
+        <SkipHostDlls>true</SkipHostDlls>
```
JoeRobich, line 14: We weren't removing these libraries from the Full Framework builds. Should this be conditioned by TFM?

filipw:
This looks good to me, can we move it out of draft @333fred ?

filipw (approved):

Issue #2442, opened 2022-08-30
Method not found: 'System.ReadOnlySpan`1<Char> Microsoft.IO.Path.GetFileName(System.ReadOnlySpan`1<Char>)'


mwaaas:
Failing to load project on vscode with the following error 

System.MissingMethodException: Method not found: 'System.ReadOnlySpan`1<Char> Microsoft.IO.Path.GetFileName(System.ReadOnlySpan`1<Char>)'.
   at Microsoft.Build.Shared.FileMatcher.IsFileNameMatch(String path, String pattern)
   at System.Linq.Enumerable.WhereArrayIterator`1.MoveNext()
   at System.Linq.Buffer`1..ctor(IEnumerable`1 source)
   at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
   at Microsoft.Build.Shared.FileMatcher.GetFilesForStep(RecursiveStepResult stepResult, RecursionState recursionState, String projectDirectory, Boolean stripProjectDirectory)
   at Microsoft.Build.Shared.FileMatcher.GetFilesRecursive(ConcurrentStack`1 listOfFiles, RecursionState recursionState, String projectDirectory, Boolean stripProjectDirectory, IList`1 searchesToExclude, Dictionary`2 searchesToExcludeInSubdirs, TaskOptions taskOptions)
   at Microsoft.Build.Shared.FileMatcher.GetFilesImplementation(String projectDirectoryUnescaped, String filespecUnescaped, List`1 excludeSpecsUnescaped)
   at Microsoft.Build.Shared.FileMatcher.<>c__DisplayClass67_0.<GetFiles>b__1(String _)
   at System.Collections.Concurrent.ConcurrentDictionary`2.GetOrAdd(TKey key, Func`2 valueFactory)
   at Microsoft.Build.Shared.FileMatcher.GetFiles(String projectDirectoryUnescaped, String filespecUnescaped, List`1 excludeSpecsUnescaped)
[...]

PR #2443, opened 2022-08-30
Add explicit System.Memory dependency to Hosts

JoeRobich:
Resolves https://github.com/OmniSharp/omnisharp-vscode/issues/5352

Similar to https://github.com/dotnet/roslyn/pull/63545. O# needs to have the newer version of System.Memory available when running against newer releases of VS Build Tools or .NET SDKs.

Also updates SDKs to a version that would have issues loading.


nickspoons:
From the error logs it looks like this PR will also fix https://github.com/OmniSharp/omnisharp-roslyn/issues/2442 and https://github.com/OmniSharp/omnisharp-vim/issues/803

src/OmniSharp.Http.Driver/app.config
```diff
@@ -62,6 +62,11 @@
                 <assemblyIdentity name="System.Composition.TypedParts" publicKeyToken="b03f5f7f11d50a3a" culture="neutral"/>
                 <bindingRedirect oldVersion="0.0.0.0-6.0.0.0" newVersion="6.0.0.0"/>
             </dependentAssembly>
+
+            <dependentAssembly>
+                <assemblyIdentity name="System.Memory" publicKeyToken="cc7b13ffcd2ddd51" culture="neutral" />
+                <bindingRedirect oldVersion="0.0.0.0-4.0.1.2" newVersion="4.0.1.2" />
```
333fred, line 68: Why doesn't this version match up with the dependency version?

JoeRobich: Because the assembly version doesn't match the nuget package version.

333fred: Might be worth a comment.

333fred (approved):

PR #2460, opened 2022-10-13
Don't remap line mappings in Razor files

allisonchou:
**Problem:** O# sets the start/end characters of a location to 0 if the location is within a line mapping directive. This is problematic in Razor, where we're attempting to implement document highlighting for the first time and need exact start/end characters.

**Solution**: Use the original line and character locations (edit: only in Razor files). This also means we need to return the non-mapped file path.

I confirmed features that go through this call path (document highlighting, find all references, go-to-impl, etc.) still work correctly with this change in both C# and Razor files. Most features still work since many [do their own re-mapping](https://github.com/OmniSharp/omnisharp-vscode/blob/1f2d7de27a9e89d8a7303104a710a918eb699827/src/features/referenceProvider.ts#L32) behind the scenes. As a result of this change, document highlighting in Razor now works too. 🙂

### DIFF
--- a/src/OmniSharp.Roslyn.CSharp/Helpers/LocationExtensions.cs
+++ b/src/OmniSharp.Roslyn.CSharp/Helpers/LocationExtensions.cs
@@ -16,9 +16,7 @@ namespace OmniSharp.Helpers
             if (!location.IsInSource)
                 throw new Exception("Location is not in the source tree");
 
-            var lineSpan = Path.GetExtension(location.SourceTree.FilePath).Equals(".cake", StringComparison.OrdinalIgnoreCase)
-                ? location.GetLineSpan()
-                : location.GetMappedLineSpan();
+            var lineSpan = location.GetLineSpan();
 
             var documents = workspace.GetDocuments(lineSpan.Path);
             var sourceText = GetSourceText(location, documents, lineSpan.HasMappedPath);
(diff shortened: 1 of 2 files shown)


filipw (commented):
I am not convinced this is the best thing to do, and it is demonstrated by the test you modified. Line mappings can be injected at any point in the code, for any arbitrary reason and I think the language service should respect that, instead of making a deliberate decision to ignore them.

Now, since we already exclude it from Cake, how about we exclude that from Razor files too, instead of making a global change to ignore those mappings? That would then solve your use case without affecting the 'normal' C# code.

allisonchou:
Thanks for the suggestion, Filip - that makes sense. I took your feedback and excluded Razor files. I also added tests specifically for the Razor scenario. Could you take another look?

allisonchou:
Also, it doesn't have to be addressed in this PR, but it seems that VS and VS Code have different behaviors for line mappings in *.cs files. VS doesn't respect line mappings while VS Code does - for example, when running FindAllRefs in VS Code with the following:

``` C#
public class Foo
{
    public void bar() { }
}
public class FooConsumer
{
    public FooConsumer()
    {
#line 100
        new Foo().bar();
#line default
    }
}
```

VS Code result:
image

VS result:
image

Not sure if we want to make these consistent eventually, just food for thought.

filipw (approved):
thanks!

PR #2463, opened 2022-10-25
Add missing LSP Handlers

JoeRobich:
Discovered as part of the work to move the C# extension over to using the O# LSP.

Other missing handlers:
  InlayHintsHandler - Required csharp-language-server-protocol to update with the 3.17 spec. I believe this is currently in a feature branch.

### DIFF
--- a/src/OmniSharp.LanguageServerProtocol/Handlers/OmniSharpDocumentHighlightHandler.cs
+++ b/src/OmniSharp.LanguageServerProtocol/Handlers/OmniSharpDocumentHighlightHandler.cs
@@ -1,0 +1,69 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using OmniSharp.Extensions.JsonRpc;
+using OmniSharp.Extensions.LanguageServer.Protocol.Client.Capabilities;
+using OmniSharp.Extensions.LanguageServer.Protocol.Document;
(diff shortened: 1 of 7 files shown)


src/OmniSharp.LanguageServerProtocol/Handlers/OmniSharpDocumentHighlightHandler.cs
```diff
@@ -0,0 +1,69 @@
...
+            var omnisharpRequest = new FindUsagesRequest
+            {
+                FileName = Helpers.FromUri(request.TextDocument.Uri),
+                Column = Convert.ToInt32(request.Position.Character),
+                Line = Convert.ToInt32(request.Position.Line),
+                OnlyThisFile = true,
+                ExcludeDefinition = false
+            };
+
+            var omnisharpResponse = await _findUsagesHandler.Handle(omnisharpRequest);
```
dibarbet, line 47: interesting - findusages returns a quick fix response that actually contains document highlights?  Some strange naming

dibarbet (approved):
one place of weird naming but otherwise seems OK

src/OmniSharp.LanguageServerProtocol/Handlers/OmniSharpDocumentHighlightHandler.cs
filipw: a lot of history behind it 😀

filipw (approved):
LGTM

PR #2503, opened 2023-02-01
Use new version of workflow that allows multiple installations

dibarbet:
attempt to fix failing CI


.github/workflows/build.yml
```diff
@@ -2,7 +2,7 @@
 name: 'Build'
 env:
   DotNetVersion: "7.0.100-rc.1.22431.12"
-  DotNetVersion2: "3.1.422"
+  DotNetVersion2: "6.0.202"
```
dibarbet, line 5: can't use a higher version since 6.0.300 requires msbuild 17+, but that doesn't appear to be available in mono
https://www.mono-project.com/docs/about-mono/versioning/#mono-source-versioning

filipw (approved):
looks.. green :)

PR #2529, opened 2023-05-04
Remove OmniSharpSymbolRenamedCodeActionOperationFactoryWorkspaceService

Cosifne:
Roslyn use 
`ISymbolRenamedCodeActionOperationFactoryWorkspaceService` to do extra work when rename.
And O# provides the implementation in the extern layer `OmniSharpSymbolRenamedCodeActionOperationFactoryWorkspaceService`

However, this looks like wrong, because O# is just create an unnecessary rename action. 
This would be cleaned by this change https://github.com/dotnet/roslyn/pull/68095


JoeRobich:
@Cosifne I believe we will need to remove https://github.com/dotnet/roslyn/blob/main/src/Tools/ExternalAccess/OmniSharp/CodeRefactorings/WorkspaceServices/IOmniSharpSymbolRenamedCodeActionOperationFactoryWorkspaceService.cs#L9 from the EA library before we can merge this change. Otherwise our MEF composition doesn't work.

Cosifne:
@JoeRobich 
I have submitted this https://github.com/dotnet/roslyn/pull/68097

PR #2536, opened 2023-05-31
Update to the latest SDKs and Package versions

JoeRobich:
- Add validation against the .NET 8 preview 4 SDK
- Move to latest release .NET 7 SDK version
- Move to the latest .NET 6 SDK version still support by Mono MSBuild 6.0.203
- Update Roslyn to 4.7.0-2.23225.15
- Move test projects from netcoreapp3.1 to net6.0 and from netstandard 1.3 to netstandard2.0


JoeRobich:
Ugh, Mono ships with a MSBuild that incompatible with the newer SDKs. Will have to revisit this later. Perhaps stay on the older 6.0.2xx SDK for test projects and skip the net7/net8 tests on Mono.

JoeRobich:
Downgraded .NET 6 SDK back to a 6.0.2xx release. Long term we should keep separate projects for testing OmniSharp.MSBuild against Mono and SDK. Not sure we are exercising framework style project loading under Mono at all.

dibarbet (approved):

PR #2542, opened 2023-06-16
Respond to vscode breaking change in 1.79.2 that triggers completion inside words

dibarbet:
Fixes https://github.com/dotnet/vscode-csharp/issues/5801

VSCode made a change in completion somewhere between 1.78.2 and 1.79.2 where they now trigger completion automatically inside words.  This was confirmed as an intentional change here - https://github.com/microsoft/vscode/issues/185286

However, this has caused lots of bug reports on us because O# will replace the entire word (including after the cursor) if a completion is accepted, instead of just inserting the selected completion item before the cursor. 

This does the same change as we did to fix this in Roslyn LSP - we limit the range to the cursor position.  See https://github.com/dotnet/roslyn/pull/68624

Now looks like 
completion_1_79_2

### DIFF
--- a/tests/OmniSharp.Roslyn.CSharp.Tests/CompletionFacts.cs
+++ b/tests/OmniSharp.Roslyn.CSharp.Tests/CompletionFacts.cs
@@ -2255,16 +2255,16 @@ namespace N
         [InlineData("dummy.csx")]
         public async Task ReplacesUpUntilCursorInMiddleOfWord(string filename)
         {
-            const string input =
-                @"pub$$class}";
+            const string input = @"public class C1 {}
+pub$$class";
 
             var completions = await FindCompletionsAsync(filename, input, SharedOmniSharpTestHost);
             Assert.All(completions.Items, (completion) =>
             {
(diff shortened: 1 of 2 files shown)


tests/OmniSharp.Roslyn.CSharp.Tests/CompletionFacts.cs
```diff
@@ -2250,6 +2250,24 @@ public C()
             }
         }
 
+        [Theory]
+        [InlineData("dummy.cs")]
+        [InlineData("dummy.csx")]
+        public async Task ReplacesUpUntilCursorInMiddleOfWord(string filename)
+        {
+            const string input =
+                @"pub$$class}";
```
genlu, line 2259: Based on the bug I created recently, we might want to have a multi-line code file and have the word not start at the beginning of the file for a more generalized test XD

genlu (approved):

JoeRobich (approved):
Looks good. I saw Gen's change and was hoping someone would port it.

Issue #2483, opened 2022-12-12
Semantic tokens do not conform to the LSP specification


JManch:
Semantic highlighting was recently merged into Neovim. Unfortunately the token types provided by Omnisharp don't conform to the [LSP semantic tokens specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_semanticTokens) so Neovim cannot support it (neovim/neovim#21391).

To demonstrate the issue, here's a comparison of Lua Language Server's tokens to Omnisharp's tokens. Omnisharp's tokens are fundamentally wrong as well as having illegal formatting.

<details>
  <summary>Lua Language Server (correct tokens)</summary>

  ``` lua
{
   full = true,
   legend = {
      tokenModifiers = {
         "declaration",
         "definition",
         "readonly",
         "static",
         "deprecated",
         "abstract",
         "async",
         "modification",
         "documentation",
         "defaultLibrary",
         "global",
      },
      tokenTypes = {
         "namespace",
         "type",
         "class",
         "enum",
         "interface",
         "struct",
         "typeParameter",
         "parameter",
         "variable",
         "property",
         "enumMember",
         "event",
         "function",
         "method",
         "macro",
         "keyword",
         "modifier",
         "comment",
         "string",
         "number",
         "regexp",
         "operator",
         "decorator",
      },
   },
   range = true,
}
  ```

</details>

<details>
  <summary>Omnisharp-roslyn (incorrect Tokens)</summary>

  ``` lua
{
   full = vim.empty_dict(),
   legend = {
      tokenModifiers = { "static symbol" },
      tokenTypes = {
         "comment",
         "excluded code",
         "identifier",
         "keyword",
         "keyword - control",
         "number",
         "operator",
         "operator - overloaded",
         "preprocessor keyword",
         "string",
         "whitespace",
         "text",
         "static symbol",
         "preprocessor text",
         "punctuation",
         "string - verbatim",
         "string - escape character",
         "class name",
         "delegate name",
         "enum name",
         "interface name",
         "module name",
         "struct name",
         "type parameter name",
         "field name",
         "enum member name",
         "constant name",
         "local name",
         "parameter name",
         "method name",
         "extension method name",
         "property name",
         "event name",
         "namespace name",
         "label name",
         "xml doc comment - attribute name",
         "xml doc comment - attribute quotes",
         "xml doc comment - attribute value",
         "xml doc comment - cdata section",
[...]

PR #2548, opened 2023-07-01
Use core LSP TokenTypes where possible and validate token names

JoeRobich:
Closes https://github.com/OmniSharp/omnisharp-roslyn/issues/2483

- Adds LSP SemanticTokensHandler tests.

### DIFF
--- a/src/OmniSharp.LanguageServerProtocol/Handlers/OmniSharpSemanticTokensHandler.cs
+++ b/src/OmniSharp.LanguageServerProtocol/Handlers/OmniSharpSemanticTokensHandler.cs
@@ -8,12 +8,11 @@ using OmniSharp.Extensions.LanguageServer.Protocol.Client.Capabilities;
 using OmniSharp.Extensions.LanguageServer.Protocol.Document;
 using OmniSharp.Extensions.LanguageServer.Protocol.Models;
 using OmniSharp.Models.SemanticHighlight;
-using OmniSharp.Roslyn.CSharp.Services.SemanticHighlight;
 using static OmniSharp.LanguageServerProtocol.Helpers;
 
 namespace OmniSharp.LanguageServerProtocol.Handlers
 {
-    class OmniSharpSemanticTokensHandler : SemanticTokensHandlerBase
+    internal class OmniSharpSemanticTokensHandler : SemanticTokensHandlerBase
     {
         public static IEnumerable<IJsonRpcHandler> Enumerate(RequestHandlers handlers)
         {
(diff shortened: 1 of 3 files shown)


src/OmniSharp.LanguageServerProtocol/Handlers/OmniSharpSemanticTokensHandler.cs
```diff
@@ -22,34 +21,49 @@ public static IEnumerable<IJsonRpcHandler> Enumerate(RequestHandlers handlers)
...
-                        return dictionary;
-                    })
-                    .ToImmutableDictionary();
+            = System.Enum.GetValues(typeof(SemanticHighlightClassification))
+                .Cast<SemanticHighlightClassification>()
+                .ToImmutableDictionary(value => value,
+                    // Use Core LSP token types where possible
+                    value => _coreTokenMap.ContainsKey(value)
+                        ? _coreTokenMap[value]
+                        : new SemanticTokenType(MakeLSPCompatibleString(value.ToString())));
```
JoeRobich, line 61: Previously we had been using the Roslyn ClassificationTypeName which contained spaces. This changes us to use LSP TokenTypes when possible, otherwise SemanticHighlightClassification EnumMember names fixed up to begin with a lowercase letter.

src/OmniSharp.LanguageServerProtocol/Handlers/OmniSharpSemanticTokensHandler.cs
```diff
@@ -22,34 +21,49 @@ public static IEnumerable<IJsonRpcHandler> Enumerate(RequestHandlers handlers)
...
+                [SemanticHighlightClassification.EventName] = SemanticTokenType.Event,
+                [SemanticHighlightClassification.PreprocessorKeyword] = SemanticTokenType.Macro,
+                [SemanticHighlightClassification.LabelName] = SemanticTokenType.Label,
+            }.ToImmutableDictionary();
+
         private readonly Mef.IRequestHandler<SemanticHighlightRequest, SemanticHighlightResponse> _definitionHandler;
         private readonly DocumentSelector _documentSelector;
 
+        private static string MakeLSPCompatibleString(string str)
+            => char.ToLower(str[0]) + str.Substring(1);
```
333fred, line 52:
```suggestion
            => char.ToLower(str[0]) + str[1..];
```

333fred: I will also say that I spent a minute trying to figure out if there was a way to do this in one simple line without doing a substring at all, using spans, but I would have to make a local out of `char.ToLower(str[0])` to get that to work.

333fred (approved):

src/OmniSharp.LanguageServerProtocol/Handlers/OmniSharpSemanticTokensHandler.cs
JoeRobich: I would have to pull in Range to make your suggestion work for net472. Maybe we can clean this all up one day.

JustALawnGnome7:
Very exciting! Any idea when we should expect this change to be included in a future OmniSharp release, or which release version will include the change?

333fred:
The next version will include it, but we don't have a regular release schedule. It's usually whenever there's something requiring a bugfix release.

SvanRaaij:
Isn't "making sure the software works out of the box again for nvim users" a reason for a bugfix release?

333fred:
That seems reasonable. @dibarbet, do you think we do a release?

jsomedon:
Please do release, I have been constantly seeing my neovim reporting Invalid character in group name, and I am sure many users been noticing this too.

333fred:
Thanks to @JoeRobich, 1.39.8 with this fix was released just now: https://github.com/OmniSharp/omnisharp-roslyn/releases/tag/v1.39.8

PR #2549, opened 2023-07-06
Move from Packages.props to NuGet Central Package Management

JoeRobich:
- Enables central package management
- Enables transitive version pinning
  - Removes a lot of System.Memory references
- Enables static restore graph
- Adds package source mapping
- Removes ReferenceAssemblies references which are added when necessary by SDK


NuGet.Config
```diff
@@ -1,5 +1,8 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
+    <config>
+        <add key="globalPackagesFolder" value=".nuget" />
```
333fred, line 4: Why do we want to override this?

JoeRobich: Good catch. The tool I used to generate the package mapping required it. Removed.

333fred (approved):

PR #2567, opened 2023-09-25
Update Roslyn to version 4.8.0-3.23475.6

JoeRobich:


JoeRobich:
From the CI errors, it seems like source generated documents aren't being returned when requested, completion is using incorrect end of line sequences in empty documents, and completion changed how they return text edits.

src/OmniSharp.Roslyn.CSharp/Services/Navigation/SourceGeneratedFileService.cs
```diff
@@ -20,6 +22,9 @@ public class SourceGeneratedFileService :
         IRequestHandler<UpdateSourceGeneratedFileRequest, UpdateSourceGeneratedFileResponse>,
         IRequestHandler<SourceGeneratedFileClosedRequest, SourceGeneratedFileClosedResponse>
     {
+        private static MethodInfo? CreateDocumentIdFromSerialized => typeof(DocumentId)
```
333fred, line 25: Should we update the EA for this? Would prefer that over readding reflection.

JoeRobich: @333fred Updated to remove reflection and use EA.

333fred (approved):
Make sure to update the title when merging.

PR #2569, opened 2023-10-04
Do not include commit characters if the typed span is empty

JoeRobich:
This is the equivalent of how Roslyn treats this situation in https://github.com/dotnet/roslyn/commit/73ed40522c213eaf97ef19813adb731e1287eef5


src/OmniSharp.Roslyn.CSharp/Services/Completion/CompletionListBuilder_Sync.cs
```diff
@@ -110,7 +110,15 @@ internal static partial class CompletionListBuilder
...
 
-                var commitCharacters = BuildCommitCharacters(completion.Rules.CommitCharacterRules, isSuggestionMode, commitCharacterRuleCache, commitCharacterRuleBuilder);
+                var treatAsASuggestion = isSuggestionMode ||
+                    (
+                        // The user hasn't actually typed anything and completion provider does
+                        // not request the item be hard-selected.
+                        completion.Rules.MatchPriority != MatchPriority.Preselect &&
+                        typedSpan.Length == 0 &&
+                        completion.Rules.SelectionBehavior != CompletionItemSelectionBehavior.HardSelection
+                    );
```
333fred, line 120: Could we extract this as a shared helper, rather than duplicating twice?

JoeRobich: @333fred extracted!

333fred (approved):

PR #2592, opened 2023-12-21
Disallow concurrent RestoreAsync on same working directories.

MarkCiliaVincenti:


Directory.Packages.props
```diff
@@ -14,6 +14,8 @@
     </PropertyGroup>
 
     <ItemGroup>
+        <PackageVersion Include="AsyncKeyedLock" Version="6.2.4" />
```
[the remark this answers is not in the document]
MarkCiliaVincenti: Currently the code allows concurrent calls to `RestoreAsync` for the same `workingDirectory` value. I believe this could lead to undesired results so it is safer to stop concurrent processing for the same working directory, which is what this PR achieves.

src/OmniSharp.Host/Services/DotNetCliService.cs
```diff
@@ -98,10 +99,6 @@ public Task RestoreAsync(string workingDirectory, string arguments = null, Actio
                     }
                     finally
                     {
-                        _semaphore.Release();
-
-                        _locks.TryRemove(workingDirectory, out _);
```
JoeRobich, line 103: Would it not be simpler to just remove this line and force all restores to the same WorkingDirectory to share the same lock object?

MarkCiliaVincenti: Sharing the same lock object would mean keeping a dictionary of semaphoreslim. But then you need to clean up when you're ready otherwise the semaphoreslim remains in memory. That's where this library comes in handy.

JoeRobich: So, just so I am sure, you are trying to make restore operations run in parallel as long as they are in different working directories?

MarkCiliaVincenti: Yes, the way I understand it is that they currently already run in parallel, irrespective of working directory.

JoeRobich: Stepping back and looking at the restore code. I think ideally, we wouldn't be passing in a working directory. Instead, we should be passing in the project path. Reason being that a directory could contain a mix of project and solution files which will cause the dotnet restore command to fail. Of course, then we would want to ensure we only start one Restore operation at a time for the project path.

MarkCiliaVincenti: @JoeRobich so basically you'd approve this PR but would want improvements? I don't want to spend more time if you then have an issue with the 3rd party library.

JoeRobich: Yeah, sorry, wish we could have discussed this in an issue first. I think the effort would be better spent getting the restore code to do the right thing. Thanks for bringing this to my attention.

MarkCiliaVincenti: @JoeRobich I'm sorry but your answer is a bit ambiguous. Would you like me to make changes to my PR to pass on the projectPath instead of the workingDirectory?

JoeRobich: @MarkCiliaVincenti I don't want to take on a new dependency for this fix. I think the best course of action would be to use a ConcurrentDictionary (or something similar) to track restore tasks running for a particular path and avoid locking entirely. If you want to work on a solution like that, I would be happy to take it.

MarkCiliaVincenti: @JoeRobich if you don't want to take the library I suggest you go for AsyncDuplicateLock from https://stackoverflow.com/a/31194647/9945524

This will not be as performant as the AsyncKeyedLock library and also has much more memory allocations. Moreover you're stuck with code you'll have to maintain in future.

MarkCiliaVincenti: @JoeRobich shall we close this PR?

PR #2610, opened 2024-04-11
Remove NuGet.Packaging.Core, as it's an assembly that only contains forwarders

nkolev92:
Remove NuGet.Packaging.Core since it only contains type forwarders. 

image


JoeRobich (approved):
Thanks @nkolev92! I will work on getting this change in.

Issue #2507, opened 2023-02-21
[LSP]  Feature request: Progress support


Anakael:
It would be nice to have support of lsp progress:
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#progress

Example of usage: https://github.com/j-hui/fidget.nvim

If you interested in that I want to try implement it by myself :)

gabrieldechichi:
+1 I would like that

GustavEikaas:
+1

joseph280996:
+ 1

Chibur:
that would be great. omnisharp does the the full solution indexing and linting at the beginning so it would be great to know when it is already done :)

Anakael:
I've started to work on it.
![Image](https://github.com/user-attachments/assets/a52df24f-04c5-4876-a0a2-c8d2bf662a41)

PR #2649, opened 2025-01-20
feat: lsp-progress

Anakael:
Resolves https://github.com/OmniSharp/omnisharp-roslyn/issues/2507

Example:
image

### DIFF
--- a/src/OmniSharp.Abstractions/Eventing/IEventEmitterExtensions.cs
+++ b/src/OmniSharp.Abstractions/Eventing/IEventEmitterExtensions.cs
@@ -45,6 +45,11 @@ namespace OmniSharp.Eventing
                 });
         }
 
+        public static void ProjectLoadingStarted(this IEventEmitter emitter, string projectPath) =>
+            emitter.Emit(
+                    EventTypes.ProjectLoadingStarted,
+                    projectPath);
+
         public static void ProjectInformation(this IEventEmitter emitter,
                                               HashedString projectId,
                                               HashedString sessionId,
@@ -55,7 +60,8 @@ namespace OmniSharp.Eventing
                                               IEnumerable<HashedString> references,
                                               IEnumerable<HashedString> fileExtensions,
                                               IEnumerable<int> fileCounts,
-                                              bool sdkStyleProject)
+                                              bool sdkStyleProject,
+                                              string projectFilePath)
         {
             var projectConfiguration = new ProjectConfigurationMessage()
             {
@@ -68,7 +74,8 @@ namespace OmniSharp.Eventing
                 References = references.Select(hashed => hashed.Value),
                 FileExtensions = fileExtensions.Select(hashed => hashed.Value),
                 FileCounts = fileCounts,
-                SdkStyleProject = sdkStyleProject
+                SdkStyleProject = sdkStyleProject,
+                ProjectFilePath = projectFilePath
             };
 
             emitter.Emit(
--- a/src/OmniSharp.Abstractions/Models/Events/EventTypes.cs
+++ b/src/OmniSharp.Abstractions/Models/Events/EventTypes.cs
@@ -2,6 +2,7 @@ namespace OmniSharp.Models.Events
 {
     public static class EventTypes
     {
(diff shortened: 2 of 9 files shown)


src/OmniSharp.LanguageServerProtocol/Eventing/LanguageServerEventEmitter.cs
```diff
@@ -46,6 +50,14 @@ public void Emit(string kind, object args)
                         }
                     }
                     break;
+                case EventTypes.ProjectLoadingStarted:
+                    string projectPath = (string)args;
+                    IWorkDoneObserver projectObserver = _server.WorkDoneManager
+                        .Create(new WorkDoneProgressBegin { Title = $"Loading {projectPath}" })
+                        .GetAwaiter()
```
Anakael, line 57: Can we change signature to async here?

src/OmniSharp.Abstractions/Eventing/IEventEmitterExtensions.cs
```diff
@@ -68,7 +74,8 @@ public static void ProjectInformation(this IEventEmitter emitter,
                 References = references.Select(hashed => hashed.Value),
                 FileExtensions = fileExtensions.Select(hashed => hashed.Value),
                 FileCounts = fileCounts,
-                SdkStyleProject = sdkStyleProject
+                SdkStyleProject = sdkStyleProject,
+                ProjectFilePath = projectFilePath
```
JoeRobich, line 78: The ProjectConfigurationMessage is used for some rudimentary telemetry. Passing ProjectFilePath may include some PII. You would likely want to add a new Message type for ProjectLoadFinished.

src/OmniSharp.LanguageServerProtocol/LanguageServerHost.cs
```diff
@@ -137,6 +139,19 @@ public async Task Start()
...
+                    .GetExports<IProjectSystem>()
+                    .Select(ps => ps.WaitForIdleAsync())
+                    .ToArray());
+
+            _workDoneObserver?.OnNext(new WorkDoneProgressReport
+            {
+                Message = "Language Server ready",
+                Percentage = 100,
+            });
+            _workDoneObserver?.OnCompleted();
```
JoeRobich, line 153: I don't think we want to wait for project loading to complete here. You could extract it to a method and just fire and forget.

src/OmniSharp.LanguageServerProtocol/Eventing/LanguageServerEventEmitter.cs
JoeRobich: Sure!

src/OmniSharp.LanguageServerProtocol/LanguageServerHost.cs
JoeRobich: Actually, why not move this back to the Initialized method?

Anakael: Because in order to send lsp-progress messages to client server must be initialized.
Otherwise there are warnings that server it not initialized and messages about initialization are got all at once after server is initilized.

src/OmniSharp.LanguageServerProtocol/Eventing/LanguageServerEventEmitter.cs
Anakael: Done. But diff got big.

src/OmniSharp.Abstractions/Models/Events/EventTypes.cs
```diff
@@ -2,6 +2,8 @@ namespace OmniSharp.Models.Events
 {
     public static class EventTypes
     {
+        public const string ProjectLoadingStarted = nameof(ProjectLoadingStarted);
```
JoeRobich, line 5: Does this event mean that we have started loading all projects or that we have started loading a single project? If it means a single project then I would recommend `ProjectLoadStarted`. Same with the Finished event.

Anakael: Single one.
Done.

PR #2670, opened 2025-04-30
Add `--locale` commandline option

### LINKED ISSUE #2661: IntelliSense defaults to system language
Coming here from https://github.com/dotnet/vscode-csharp/issues/2513
Not sure if this is the right place or if there is an open issue for this already.

I'm using the C# extension by muhammad-sammy for VSCodium, which uses omnisharp.
When hovering over a warning or error, IntelliSense will localize it into the language chosen for the operating system. According to the linked issue this is a long existing bug, which is hugely problematic for international programmers, as looking up error messages in languages other than english will often yield no results.
Ideally it should default to the IDE's language setting or have a custom one instead.

The issue was closed with the comment "O# does not set the CurrentUICulture so it defaults to your system locale. If this is still a problem, please open an issue at https://github.com/OmniSharp/omnisharp-roslyn/issues".

A workaround for this issue is to delete some .dll files in this folder, but this has to be repeated with every update.

Image

JoeRobich:
The option is optional and defaults to the system locale. Extensions will need to be updated to pass this along when starting O#.

resolves #2661

### DIFF
--- a/src/OmniSharp.Host/CommandLineApplication.cs
+++ b/src/OmniSharp.Host/CommandLineApplication.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Threading;
@@ -21,6 +22,7 @@ namespace OmniSharp
         private readonly CommandOption _logLevel;
         private readonly CommandOption _applicationRoot;
         private readonly CommandOption _debug;
+        private readonly CommandOption _locale;
 
         public CommandLineApplication()
         {
@@ -37,6 +39,7 @@ namespace OmniSharp
             _zeroBasedIndices = Application.Option("-z | --zero-based-indices", "Use zero based indices in request/responses (defaults to 'false').", CommandOptionType.NoValue);
             _plugin = Application.Option("-pl | --plugin", "Plugin name(s).", CommandOptionType.MultipleValue);
             _debug = Application.Option("-d | --debug", "Wait for debugger to attach", CommandOptionType.NoValue);
+            _locale = Application.Option("--locale", "Locale to use for localization (defaults to system locale).", CommandOptionType.SingleValue);
         }
 
         public int Execute(string[] args)
@@ -70,6 +73,7 @@ namespace OmniSharp
         {
             Application.OnExecuteAsync((_) =>
             {
+                SetLocale();
                 DebugAttach();
                 return func();
             });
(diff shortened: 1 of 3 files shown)


src/OmniSharp.Host/CommandLineApplication.cs
```diff
@@ -117,5 +123,19 @@ private void DebugAttach()
...
+
+        private void SetLocale()
+        {
+            if (!string.IsNullOrEmpty(Locale))
+            {
+                try
+                {
+                    CultureInfo.CurrentUICulture = CultureInfo.GetCultureInfo(Locale);
+                }
+                catch (CultureNotFoundException) { }
```
333fred, line 137: Can we log a warning or something?

tests/OmniSharp.Tests/CommandLineApplicationFacts.cs
```diff
@@ -24,5 +26,33 @@ public void AllowsEqualsSignInSolutionPath(string arg1, string arg2, string arg3
...
+            var app = new CommandLineApplication();
+            app.OnExecute(() => { runtimeLocale = CultureInfo.CurrentUICulture.Name; return 0; });
+            app.Execute(["--locale", locale]);
+
+            Assert.Equal(locale, runtimeLocale);
+        }
+
+        [Fact]
+        [UseCulture("de-DE", "de-DE")]
+        public void NotPassingLocaleUsesSystemLocale()
```
333fred, line 46: Let's add a test for an invalid locale.

src/OmniSharp.Host/CommandLineApplication.cs
JoeRobich: This is before we have a logger. Could use `Console.WriteLine`, but unsure how that might affect the STDIO host.

tests/OmniSharp.Tests/CommandLineApplicationFacts.cs
JoeRobich: Done!

src/OmniSharp.Host/CommandLineApplication.cs
333fred: Could we use `Console.Error.WriteLine`? Presumably the stdio hosts aren't looking for response output from that?

JoeRobich: Good call.

JoeRobich: Updated

333fred (approved):

slarticodefast:
Thank you so much for fixing this.
This is such an important QOL feature for international programmers!

Issue #2375, opened 2022-04-03
DocumentVersion shouldn't be reset to 0 when handling "textDocument/didSave"


Swassie:
I'm using the LSP client Eglot in emacs together with OmniSharp, and it mostly works well. One problem I have encountered is triggered by the following sequence:

1. Edit a file. (Client increments the DocumentVersion).
2. Save the file. (Server resets the version to 0 internally).
3. Request and execute a code action for the file, e.g. add missing 'using ...'
4. Server requests "workspace/applyEdit" for the file, with version 0.
5. Client fails to apply the edit since the version doesn't match.

I think that OmniSharp shouldn't reset the version in this case. I have looked at a few of the other major LSP servers and I couldn't find anyone that did this. Also I did try removing the line src/OmniSharp.LanguageServerProtocol/Handlers/OmniSharpTextDocumentSyncHandler.cs:158 and it solves the problem.

Thanks for your work!

jameschensmith:
I'm running into this issue as well, and it's very annoying to have to restart the LSP any time I want to do code actions (renames don't require a restart, though, which is nice 😅 ). I've dug into this a bit and would like to see if I can help bring traction to this issue.

- [In the LSP specification for `textDocument/didSave`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_didSave), there is no mention that the server should reset the version to 0.
- When looking at the [blame of the line mentioned by @Swassie](https://github.com/OmniSharp/omnisharp-roslyn/blame/ed467d0ad2d877b837380a849f856dcd210d69f7/src/OmniSharp.LanguageServerProtocol/Handlers/OmniSharpTextDocumentSyncHandler.cs#L158), it appears that this change was made almost 3 years ago by @razzmatazz in [this commit](https://github.com/OmniSharp/omnisharp-roslyn/commit/96600de10ba16caed0379b1439d1cf5ff382e468).
- @razzmatazz, is [working on his own C# language server](https://github.com/razzmatazz/csharp-language-server). In his [implementation of `textDocument/didSave`](https://github.com/razzmatazz/csharp-language-server/blob/1ef7d166f3834d673cfdd68222c944bf60c199ad/src/CSharpLanguageServer/Server.fs#L433-L460), there doesn't appear to be any document version resetting.

Based on this analysis, @razzmatazz, would you be able to help with any of the following:
1. Provide some history / context as to why a version reset was added to `textDocument/didSave`.
[...]

PR #2695, opened 2025-11-06
fix: document version reset on textDocument/didSave

S222em:
Fixes https://github.com/OmniSharp/omnisharp-roslyn/issues/2375

This PR fixes an issue where the document version is reset to `0` on `textDocument/didSave`.

It is not documented in the LSP that the version should reset, see: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_didSave.

This caused a document version mismatch in editors when applying code actions (for example see https://github.com/zed-industries/zed/issues/31014).


S222em:
@microsoft-github-policy-service agree

JoeRobich (approved):
@S222em Thanks!

